Issue #379, opened 2014-02-04
Sorting is off for filenames with (2) numbers in parentheses


jancborchardt:
As you know we generate »filename (2)« for duplicates. Currently sorting goes like this:
- filename (2)
- filename (3)
- filename

Which makes it pretty confusing since the original document is all the way on the bottom. Instead it should go like this:
- filename
- filename (2)
- filename (3)

Even worse, when you have numbers in parentheses which exceed 10, it sorts like this:
- filename (18)
- filename (19)
- filename (2)
- filename (20)
- filename (21)

PR #624, opened 2014-09-14
Added sorting by name, date or size ascending/descending

tobiasKaminsky:
Hello,
I have added the possibility to sort files/folders by name, date or size.
The option is stored temporary, so it is obeyed if the directory is changed.

Filename sorting fixes also #379 
Sorting by size: Folder are sorted by name, as there is no information about the number of files containing the directory.

Sorting is, like dolphin or windows explorer, separated between folder and files.

TODO: Translate to other languages.


tobiasKaminsky:
@davivel I have created a PR to use the folder size of the server: #1655

Issue #649, opened 2014-10-09
feature request: auto-rename for Photo-upload


marlboro2k:
I think it would be a good improvement if automaticly uploaded photos get renamed like: yyyy_mm_dd_time or something like this.

davivel:
After carefully considering all the points, this will not be done. Reasons are:
 - it does not add an important value with respect to sorting by date;
 - it adds complexity to the mapping between local and remote names;

What we really need is fixing the setting of modification time in server side to gran that dates match the moment the picture was taken, and not the moment is was uploaded. That's already somewhere in the repo.

Thanks everybody.

PR #672, opened 2014-10-22
Instant upload: rename files on upload

Kernald:
Hi,
So, this is a draft implementation for #649. It's not fully tested yet, as it still have at least an issue (about which I have a question, hence this pull request.

I'm using `[Images|Video].Media.DATE_TAKEN` to allow the user to rename the file depending on the date the photo or video was taken. It's working fine when the device is online at the time the media is taken, but not when the device is offline and the media uploaded later.

In this specific case, `InstantUploadBroadcastReceiver.handleConnectivityAction` manages the uploading of the file which has not been sent yet. However, at that time, we only have the (local) file path. The mime type was needed here, but instead of using the media store to retrieve it, a simple map lookup with the file extension is made. Is there any reason to this, excepted the fact it's simpler for the mime type only? I'll need to use the media store here to be able to fetch the date the media was taken, I think. I don't really see any other way.

Thanks

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -20,6 +20,9 @@
     <string name="prefs_pincode_summary">Protect your client</string>
     <string name="prefs_instant_upload">Instant picture uploads</string>
     <string name="prefs_instant_upload_summary">Instantly upload pictures taken by camera</string>
+    <string name="prefs_instant_upload_rename_files">Rename uploaded files</string>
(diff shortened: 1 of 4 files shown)


davivel:
Thanks for the effort, but this won't be in. 

I really hope this doesn't discourage you. We need to improve and fasten our decision take and contributions handling processes. Working on it.

My apologies.

PR #707, opened 2014-11-05
Text file preview

Stoyicker:
This PR adds support for previewing text files.

~~Depends on #661.~~


jancborchardt:
@davivel @masensio @rperezb @MTRichards this has been hanging around for quite some time now. Can we please get this worked in?

jancborchardt:
Btw @Stoyicker please join our IRC channel #owncloud-android! :)

Stoyicker:
@jancborchardt I don't think I can be any active on IRC, but drop me an e-mail or comment here if there is something that should be addressed before merging, or the merge itself, and I will find a time to do it.

jancborchardt:
Awaiting your comment @davivel @masensio @rperezb @MTRichards

MTRichards:
Please bring this up on the next call so we can look at it. Right now, 
the customer and oC 8.1 bug backlog (a bunch of new bugs just identified 
with File Locking on 8.1) will take precedent.

.Matt

On 6/22/2015 5:35 PM, Jan-Christoph Borchardt wrote:
>
> Awaiting your comment @davivel <https://github.com/davivel> @masensio 
> <https://github.com/masensio> @rperezb <https://github.com/rperezb> 
> @MTRichards <https://github.com/MTRichards>
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/owncloud/android/pull/707#issuecomment-114270547>.
>

-- 
Matt Richards
VP Product
ownCloud, Inc

978-760-3294 (m)
www.owncloud.com

davivel:
Thanks for commenting, @MTRichards

build.gradle
```diff
@@ -63,8 +63,8 @@ android {
 
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_7
-        targetCompatibility JavaVersion.VERSION_1_7
+        sourceCompatibility JavaVersion.VERSION_1_5
```
davivel: This won't be valid. We will move soon to Material Design. That requieres Java 7, according to Google documentation.

res/layout/text_file_preview.xml
```diff
@@ -0,0 +1,31 @@
...
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_height="match_parent"
+    android:layout_width="match_parent"
+    android:fillViewport="true">
+
+    <TextView
+        android:id="@+id/text_preview"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="8dp"
```
davivel: I'd rather include both layout_marginStart and layout_marginLeft (and same for End and Right) for increasing compatibility. *Start and *End where introduced in API level 17. We support from API level 14.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        OCFile file = getFile();
+
+        Bundle args = getArguments();
+
+        if (file == null)
+            file = args.getParcelable(FileDisplayActivity.EXTRA_FILE);
```
davivel: Would prefer that { } are used even in one-liners.

build.gradle
Stoyicker: Are you sure about this? Material Design is just a set of design guidelines, not limited to Android, nor even Java, and anyway, Java 7 is almost fully compatible with all Android versions: http://stackoverflow.com/questions/20480090/does-android-support-jdk-6-or-7

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+        Log_OC.e(TAG, "onPause");
+        super.onPause();
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        Log_OC.e(TAG, "onResume");
+
+        loadAndShowTextPreview();
```
davivel: Please, don't reload the text in ```onResume()```. Do it in ```onCreateView()``` or ```onActivityCreated()```; ```onPause()``` and ```onResume()``` may be called when other windows partially hide the fragment (such as the menu showing apps to open the file outside of ownCloud) even though it stays visible.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+    private void loadAndShowTextPreview() {
+        new TextLoadAsyncTask().execute(getFile().getStoragePath(), mTextPreview);
+    }
+
+    /**
+     * Reads the file to preview and shows its contents. Too critical to be anonymous.
+     */
+    private class TextLoadAsyncTask extends AsyncTask<Object, Void, StringWriter> {
+        private final String DIALOG_WAIT_TAG = "DIALOG_WAIT";
+        private TextView mTextView;
```
davivel: ```WeakReference<TextView>``` would be better for the reference to the view field in the async task; would also need to asign its value in a new contstructor.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+        protected void onPreExecute() {
+            showLoadingDialog();
+        }
+
+        @Override
+        protected StringWriter doInBackground(java.lang.Object... params) {
+            if (params.length != 2)
+                throw new IllegalArgumentException("The parameters to " + TextLoadAsyncTask.class.getName() + " must be (1) the file location and (2) the text view to update");
+            final String location = (String) params[0];
+            mTextView = (TextView) params[1];
```
davivel: Still would prefer to set the TextView reference in a constructor; the semantics of AsyncTask grant that fields set in constructor or ```onPreExecute``` are 'done' before ```doInBrackground``` is started.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+                sc = new Scanner(inputStream);
+                while (sc.hasNextLine()) {
+                    bufferedWriter.append(sc.nextLine());
+                    if (sc.hasNextLine()) bufferedWriter.append("\n");
+                }
+                bufferedWriter.close();
+                IOException exc = sc.ioException();
+                if (exc != null) throw exc;
+            } catch (IOException e) {
+                finish();
```
davivel: Please, log ```e``` through ```Log_OC.e(,,...)```, including the exception object as last parameter to dump the stack trace.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+            } finally {
+                if (inputStream != null) {
+                    try {
+                        inputStream.close();
+                    } catch (IOException e) {
+                        finish();
```
davivel: Please, log e through Log_OC.e(,,...), including the exception object as last parameter to dump the stack trace.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -0,0 +1,391 @@
...
+                if (sc != null) {
+                    sc.close();
+                }
+            }
+            return source;
+        }
+
+        @Override
+        protected void onPostExecute(final StringWriter stringWriter) {
+            mTextView.setText(new String(stringWriter.getBuffer()));
```
davivel: After changing the type of ```mTextView``` to ```WeakReference<TextView>``` , here will be needed to check that ```mTextView.get()``` is not null, then do ```setText()``` on the returned ```TextView```.

davivel:
Generally speaking, it's OK. Most of my comments are minor. Please, observe specially the comment about the ```WeakReference```. Consider also add a field to save the AsyncTask in progress, and cancel it in ```onStop```. That will help to avoid bugs in rotations.

Needs rebasing, but we can take care of it when it's finally scheduled for QA.

Stoyicker:
I believe I have done all the changes except the Java version and the location of the text load request. The reason for the former is both my inline response and, mostly, the fact that it is more of a whole build configuration change that I believe should go into some other PR, along other fixes that this configuration should get, like for example the plugin version. As for moving the request of the text load to onCreateView(), I believe that the reason why I put it in onResume was because, if you preview a text file and then send the Oc app to background, edit the file in another app and come back to Oc, the changes would show, but it is possible that this would also work with having the call in onCreateView(), and in this case I understand and agree that it is preferred. However, our "organization"'s test environment is long gone and therefore I cannot feasibly test this. Will you be able to take care of this?

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
LukeOwncloud: @Stoyicker wrote:
> I put it in onResume was because, if you preview a text file and then send the Oc app to background, edit the file in another app and come back to Oc, the changes [*should*] show.

A valid point, but an editor covers the whole activity. So ``onStart()`` would be the best place: https://developer.android.com/training/basics/activity-lifecycle/starting.html#lifecycle-states

Stoyicker: I'm not that sure. Navigating away from the activity triggers a pause, not a stop, and therefore neither onRestart() nor onStart() would be called. See the first response to [this SO question.](http://stackoverflow.com/questions/3325767/onstop-not-getting-called-when-i-press-the-back-button)
But again, as I said in my comment below, I cannot test it so this is just what I believe should theoretically happen.

LukeOwncloud: > Navigating away from the activity triggers a pause

 According to http://developer.android.com/reference/android/app/Activity.html#ActivityLifecycle you are right.
According to my previous link it trigger PAUSE + STOP.

Oh, I love inconsistent (or misinterpret-able) documentation... not.

Stoyicker: I have verified the lifecycle by myself and I have not managed to create a UC in which onResume() is called without onStart(), so I have performed the move.

davivel:
@Stoyicker, thank you very much for the changes. 

We'll take care of testing the specific case you comment about the test load; in any case, you moved it to ```onStart```, what is pretty better than ```onResume``` IMHO.

About JDK 7, I can't find the reference I remembered. The most near I get is in http://developer.android.com/tools/sdk/eclipse-adt.html ; it's specific of Eclipse-based environments.

We'll check during the validation if there is any real problem with it or not.

masensio:
This PR has a lot of conflicts because the branch is outdated. It should be necessary to updated with the last version of the develop branch.
@Stoyicker, no worries I've created the branch text_file_preview_pr_707_with_develop for testing.
@purigarcia the branch text_file_preview_pr_707_with_developis ready to be test. :)

purigarcia:
There is a bug when you rotate the device
STEPS TO REPRODUCE IT
1. Download a txt file
2. once it is downloaded, change the orientation of the device

CURRENT BEHAVIOR
         -- Loading is shown and it is there forever. You need to kill the app to keep on using it

- EXPECTED BEHAVIOR
         -- Loading is not there forever.

- DEVICE
         -- It has been checked in Nexus 5 v.5.0.1 and  Samsung Galaxy s3 v.4.1.2

@Stoyicker could you take a look? thank you

Stoyicker:
I believe to have the problem located, but these days I'm remarkably busy, plus I still need to setup a new test environment, so although it may take it around a month I'm estimating, I can definitively fix it.

Stoyicker:
@purigarcia can't replicate on real Nexus 4, nor Nexus 5, Galaxy S3 and Galaxy S5 emulators (everything I've tried so far). Can you share the exact file you are able to trigger the issue with?

tobiasKaminsky:
I could replicate it with a nexus 5 emulator:
download file, after that it shows text file content, after that ctrl + f12, modal "wait a moment" appears.

Text file is 0.txt with only content "laptop".

purigarcia:
@Stoyicker , I update the steps indicating the you should be seen the preview of the file.
I reproduce it with a txt created from the web with a word, for instance

Stoyicker:
Are you testing this branch or text_file_preview_pr_707_with develop?

purigarcia:
I am testing text_file_preview_pr_707_with develop
Because it is up to date.

purigarcia:
Maybe the bug is affected with new code, but we should fix it anyway.
Thank for the effort

Stoyicker:
It has to be that. My attempts have been with us4_view_text_files and the bug is not there.

tobiasKaminsky:
I have fixed the bug.
The main problem was that the "wait a moment" dialog was added twice, instead of just being shown the second time.
And as there were existing two dialogs, but only one was removed, the other still blocked the UI.

tobiasKaminsky:
I have pushed the fix to text_file_preview_pr_707_with develop

tobiasKaminsky:
I think this one can be closed and therefore I open a PR for the new branch.
If I am wrong, please revert.

purigarcia:
Great, @tobiasKaminsky !
I think pushing the fix to branch ext_file_preview_pr_707_with develop is correct.
But I am not sure if this pr should be closed. @davivel ?

davivel:
It's OK. History is kept, and a single PR for a single feature.

Let's follow the work in #1095 .

Thanks to everybody.

purigarcia:
I find another issue

***rtf files are not been shown correctly***
- STEPS TO REPRODUCE IT
1. select to download a .rtf

- CURRENT BEHAVIOR
     -- It is seen with a lot of strange characters

- EXPECTED BEHAVIOR
      -- Try to showing it correctly and if it is not possible, not preview it

- DEVICE
       -- It has been checked in Nexus 5 v.5.0.1

@Stoyicker ,@tobiasKaminsky  could any of you take a look of it? Thanks

Stoyicker:
An RTF file is a proprietary format by Microsoft that describes a file that contains information which is not plain text. I don't know if it is possible to properly read the file (the results would probably be something similar to what you get when you open a .docx file with LibreOffice), but even if it is, the effort and the likely overhead added to the app just for this is huge, and makes the feature completely infeasible in my opinion.

The fact that the file attempts to be previewed is just because [the interpretation of its type depends on its mime type](https://github.com/owncloud/android/blob/be5dbcbb2e5afef17d47b4cbbf9e0eb08de69da4/src/com/owncloud/android/datamodel/OCFile.java#L554-557), and this is defined by ownCloud, not in the PR.

Note how the same problem occurs with non-text files. For example, take [this PSD file](https://wiki.teamfortress.com/w/images/8/8f/New_Tool_Demo_Thumbnail.PSD) and upload it to your server, then open the app and, when it's synced, download it. You'll see how it tries to show the preview, because of the recognized mime type, but is unable to do so because the format is not "understandable" and shows an error instead.

In conclusion, this is a "problem" of the policy on how you identify file types, not preview text files itself. Furthermore, and since there is always the option of opening the file with an external app (Word or Photoshop for example in the described cases) I would not even identify it as a problem. Mime types are a more reliable way to identify a file type than extension parsing as extensions can easily be edited.

tobiasKaminsky:
Yeah.
I would also disable it for RTF and show preview just for plain txt files.
Should we?

Stoyicker:
I would say that's inconsistent in practice. In the case that you want to reliably identify RTF files you would probably need to look for certain bytes that show that structure; otherwise you could be facing, for example, a plain text file that got its extension manually changed to .rtf, and although since it is a plain text file it should be previewable, the .rtf extension would filter it out. And, even if the filtering is applied, what happens then with .doc, .docx, .odt, etc? I say let's not touch it.

tobiasKaminsky:
According to http://www.sitepoint.com/web-foundations/mime-types-complete-list/
the mimetypes should be different:
txt -> text/plain
rtf -> text/richtext

So this should work.

Of course there is still the function mimetypeByName which should be dropped for the reason you mentioned.

Stoyicker:
The point is it is precisely because of that it doesn't work. [The filtering method implemented in OCFile](https://github.com/owncloud/android/blob/be5dbcbb2e5afef17d47b4cbbf9e0eb08de69da4/src/com/owncloud/android/datamodel/OCFile.java#L554-557) only cares about the file being text, and not whether it is or not plain, which causes the issue here.
Even if we drop getMimeTypeByName for this functionality and decide to make a different implementation for it, we would have to consider all 99 text mimetypes, and the criteria to identify supported ones would probably end up being rather floffy (for example, HTML files can be displayed as source, but does the user expect them to be interpreted instead? It's the same problem that with RTFs). For this I believe the current approach serves as a good-enough solution. Let's also not forget that this mimetype identification was already being used before this PR, so it is not like sticking to it is bad as long as it continues to be good enough for the purposes that it was being used before, whichever they were.

tobiasKaminsky:
Hmm. Maybe I am not understanding you correctly.
My idea was to use the text preview only for mimetype "text/plain".
What are you suggesting? Keep it the way it is currently and displaying even rtfs as "corrupted" preview?

Stoyicker:
In summary, I strongly want to keep the "corrupted" preview. I'll use an example to try to better explain why. Think of your suggestion: let's imagine that we decide that this functionality only applies to purely plain text files. This is already arguable by the fact that a user downloading, let's say, an HTML (text/html) file, probably expects to be able to see the contents, but instead the functionality would not support it. But even if we decide to go through all 99 text/*something* types and filter in those that are properly readable even if not typed as plain, like text/html, this is inconsistent with the current behaviour of the app.
Take a look at the image previewing procedure: it should be pretty similar (actually the text preview is designed in strict accordance with it). Any file that is considered an image by the app - like, for example, PSD files - will be attempted to preview. However, trying to preview a PSD will result in an "improper" preview. Therefore, the same should apply to other types of previews.
Making a proper fix for this is, in my opinion, an application-level issue, that is independent of this PR; this is, filtering only concrete, fully-qualified mimetypes in all UCs where a file type needs to be identified, this is, by rewriting the is*() methods in OCFile. If this is not seen as an independent problem and we decide to fix it here instead, the behaviour will not beg consistent with other previews.

tobiasKaminsky:
Ok. thanks for your very detailed explanation.
To compare it with a "normal" computer: it is like we open all text files with notepad/kwrite/etc.
A normal user should also know that this is not meaningful in every situation.

This is fine for me.

davivel:
I'm sorry, @Stoyicker, but I disagree.

I love consistency in code, but it cannot be more important than providing the user with a right functionality. I cannot imagine a situation where rendering the non-printable characters of an RTF file is a nice result for a regular user. 

I agree that trying to render RTF correctly is too hard with our resources and current knowledge. Too much for this PR, that brings a very valuable functionality and doesn't need to increase its scope due to problems find in QA. Due to this, the best decision to take is *not rendering the file*. The user will be able to open it with other apps installed in the device, as for any other file that the OC app is not able to preview (that are a lot).

> The filtering method implemented in OCFile only cares about the file being text

Sorry, but that's not the filtering method, it's a (helper) method about a property of the file. The filtering method is 

https://github.com/owncloud/android/pull/1095/files#diff-d2ee4c2ecc1b4aee74165cb73d21bbcdR391

, that may be as complex as necessary.

Stoyicker:
Sure, will get into it later.

Issue #465, opened 2014-04-14
Instant upload only when charging


fduraibi:
I would like if there is an option for the instant upload to only upload when the phone is charging. Similar to what e.g MEGA is doing (see photo)

This saves battery when you don't have charger around. also stops it from syncing when connected to slow wifis at public places and coffee shops.

![screenshot_2014-04-14-12-42-53](https://cloud.githubusercontent.com/assets/432460/2698045/5ffed370-c3f5-11e3-974c-02cc30ebac35.png)

_______________

EDIT (July 2021):

An interesting feature to include in the app, taking advantage of the current WorkManager implementation.

Related #3315 

ACs:

- New check in Camera upload photos/videos called `Upload pictures/videos only when charging`
- If it is checked, all new stuff will be enqueued till device is charging. Proper message to show in the uploads view (`Waiting for charging` or similar) depending¿?
- If it is unchecked, nothing different will happen. Stuff will be uploaded depending other conditions but not this one
- Pictures and videos will behave separately
- It will be visible only if `Only wifi` is enabled. Otherwise, it will be hidden
- If camera upload section is switched off, check becomes disabled (same as wifi only)

(open to discussion)

davivel:
@KurtLopes, sorry, no estimation for this.

@tobiasKaminsky, about the appropiate set of sync options, it's something that needs to be discussed in a wide scope, probably in its own issue.

rperezb:
@jancborchardt I'd like to have your input on this one.

MTRichards:
@jancborchardt wanted to see your view on the settings complexity vs. feature.

tobiasKaminsky:
We are discussing it in the PR #745

davigonz:
Changing the label from Instant uploads to Camera uploads since instant uploads feature doesn't longer exist...

PR #745, opened 2014-11-22
Upload when charging

tobiasKaminsky:
See #465


jancborchardt:
@tobiasKaminsky can you post a screenshot of how this looks like so I can design-review it?

tobiasKaminsky:
upload_1
upload_2

Currently there is only one "upload when charging", but two "upload on wifi"...
I do not know what my intention was...

masensio:
Hi @tobiasKaminsky!

I'm reviewing this PR. The branch is outdated, but don't worry. I will updated it. 
I'll cross my fingers when resolve conflicts

res/xml/preferences.xml
```diff
@@ -18,15 +18,29 @@
...
 	<PreferenceCategory android:title="@string/prefs_category_security">
-	    <android.preference.CheckBoxPreference android:title="@string/prefs_passcode" android:key="set_pincode" />
+	    <android.preference.CheckBoxPreference
+			android:title="@string/prefs_passcode"
+			android:key="set_pincode" />
 	</PreferenceCategory>
 
-    <PreferenceCategory android:title="@string/prefs_category_instant_uploading" android:key="instant_uploading_category">
-         <com.owncloud.android.ui.CheckBoxPreferenceWithLongTitle android:key="instant_uploading"
+    <PreferenceCategory android:title="@string/prefs_category_instant_uploading"
```
masensio: Maybe, it is necessary to review the order in the preferences.xml

tobiasKaminsky: @jancborchardt can review the order?

tobiasKaminsky:
@jancborchardt Please have a look here

res/xml/preferences.xml
jancborchardt: Which order is this actually?

jancborchardt:
Looks good! Just a few details:
* can you change the text to »Upload folder«?
* can you change »WiFi« to »wifi«?
* I assume the folder, wifi and charging options are greyed out when the main instant upload option is not enabled?

tobiasKaminsky:
ref 1:"Upload folder"? We do not support this currently, or?
This is all about the option, that the upload only starts if the phone is charging.

2: I will change
3: As far as I remember, they are not visible, if the main option is disabled.

jancborchardt:
@tobiasKaminsky right, if we call it »Upload folder« it might get confusing. »Upload path« is better, but with »path« in lowercase.

tobiasKaminsky:
instantupload1

Should "when charging" be independent for picture and video as "on wifi"?

tobiasKaminsky:
Updated text:
instantupload2

@jancborchardt Remaining question is, if "when charging" should be an independent setting for picture and video? This could be a good idea as videos are very big and can drain the battery very quick...

jancborchardt:
@tobiasKaminsky yeah, I would make the setting independent. Right now on the very top it’s kind of confusing anyway.

tobiasKaminsky:
instantupload3

jancborchardt:
Yeah, looks good!

PR #773, opened 2014-12-02
Trigger media scan

davivel:


fluffy88:
This is great :)
I'm running version 1.6.2 and this issue was annoying me so I came here to report/fix the problem.
This merge seems to have resolved the problem for renames, moves, downloads and deletes.

However I don't think uploads (or instant uploads) trigger the media scan.
I've created a pull request with a fix for the same, #839
It may seem weird that uploads need a media scan, but I guess until #819 is implemented all uploaded files will be stored on the users device in their ownCloud directory.

PR #777, opened 2014-12-03
Copy files

Stoyicker:
Adds support for file copy.

Depends on owncloud/android-library#47.


.gitmodules
```diff
@@ -1,5 +1,4 @@
-
 [submodule "owncloud-android-library"]
 	path = owncloud-android-library
-	url = git://github.com/owncloud/android-library.git
-	branch = develop
+	url = https://github.com/owncloudChalmers/android-library.git
```
davivel: Wrong URL here; I guess it cannot be fixed until the PR in the library is merged...

davivel:
Please, @Stoyicker , disable the automatic formatting. This is going to take a while to review :(

build.gradle
```diff
@@ -31,6 +31,12 @@ dependencies {
 android {
     compileSdkVersion 19
     buildToolsVersion "20.0.0"
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_5
```
davivel: Any concrete reason for this change?

.gitmodules
Stoyicker: I'll change it to the proper one when merging. As mentioned in #661, we had to change this to be able to build because the Gradle support was not yet merged into your 'android-library' and we therefore could not build our 'android' using Gradle, but now that it is merged the problem is gone.

build.gradle
Stoyicker: It is not a "change". If I don't misremember, the amount of warnings that were popped when using 1.7 is annoyingly huge, and most of them are due to complaints of the type "you could do this that is possible since JDK 1.X", where X>5. Obviously we were not going to review the whole codebase to update everything to "Java 7-acceptable" code, because the automated formatting issues you speak about (and that I understand, just didn't really think of it) would be piece of cake when compared to reviewing such a huge, huge set of changes. By setting these options, the IDE stops highlighting everything so you can look at the code without hurting your eyes, and we researched and made sure that there are no compatibility problems at all as long as the specified version is not beyond 1.7.

Stoyicker:
@davivel Both PRs are now ready to go. I must say that since we no longer have our test server I haven't been able to test it out as much as desired, so some confirmatory testing once the library PR is merged (which obviously is the reason why this one fails in Travis right now) probably won't hurt.

build.gradle
davivel: :+1:

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -726,10 +725,60 @@ public void moveLocalFile(OCFile file, String targetPath, String targetParentPat
...
+                if (!targetFolder.exists()) {
+                    targetFolder.mkdirs();
+                }
+                copied = copyFile(localFile, targetFile);
+            }
+            Log_OC.d(TAG, "Local file COPIED : " + copied);
+        }
+    }
+
+    private static boolean copyFile(File src, File target) {
```
davivel: There is no need this is```static```

src/com/owncloud/android/operations/CopyFileOperation.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Constructor
+     *
+     * @param context          Context of the application.
+     * @param srcPath          Remote path of the {@link OCFile} to move.
+     * @param targetParentPath Path to the folder where the file will be copied into.
+     * @param account          OwnCloud account containing both the file and the target folder
+     */
+    public CopyFileOperation(Context context, String srcPath, String targetParentPath, Account account) {
+        mContext = context;
```
davivel: This is an undesired dependancy we should have get rid of long time ago. We'll take care.

src/com/owncloud/android/datamodel/FileDataStorageManager.java
Stoyicker: Agree, fixed.

fxshpac:
Hope this feature gets into next release soon. This is one of the features I have been waiting for to complete my use scenario. Keep it up :)

davivel:
So I hope. Right now is our top priority in terms of community PRs. 

We will fix the conflicts ourselves when the time comes. Crossing fingers.

masensio:
This PR has a lot of conflicts because the branch is outdated. It should be necessary to updated with the last version of the develop branch.
@Stoyicker, no worries I've created the branch copy_operation_pr_47_with_develop for testing the copy feature.
@purigarcia the branch copy_operation_pr_47_with_develop is ready to be test. :)

purigarcia:
it is not possible to copy a file from the detail view, gallery or video preview.
The option is shown but do nothing
- STEPS TO REPRODUCE IT
1. download an image
2. select to copy a file 

- CURRENT BEHAVIOR
       -- Nothing happens

- EXPECTED BEHAVIOR
      -- File is copied

- DEVICE
      -- It happens using Nexus 9 v.5.0.2 and htc v.4.2.2

cc @Stoyicker @davivel @masensio

purigarcia:
There is another bug

- STEPS TO REPRODUCE IT
1. turn off the wifi in the device
2. select to copy a folder

- CURRENT BEHAVIOR
      -- Message says : "Unable to copy. Check wether the file exists"

- EXPECTED BEHAVIOR
      -- Message should says:  "The operation couldn't be completed, server is unavailable" or  "An error occurred while trying to copy this file or folder" 
         in the majority of the actions it is said the first message.

- DEVICE
         -- nexus 9 v.5

cc @Stoyicker @davivel @masensio

purigarcia:
Another bug, because copy option should not be available when the file/folder is uploading or downloading

- STEPS TO REPRODUCE IT
1. select to download a big file
2. select to show the context menu

- CURRENT BEHAVIOR
        -- the option of copy is shown

- EXPECTED BEHAVIOR
         -- the option of copy is not shown

- DEVICE
          -- nexus 9 v.5

cc @Stoyicker @davivel @masensio

Stoyicker:
These issues are reproduced in the original branch. I'll address them soon.

Stoyicker:
I've just pushed the fixes. The "Copy" option is no longer shown when a file is downloading, the correct message is reported upon impossibility to reach the server and the "Copy" option is not available in preview fragments (I know that this is not the expected behavior you mentioned, but it is the same as the "Move" one, so this seems more consistent IMO).

Note that if #1095 is merged, a block similar to [this](https://github.com/owncloud/android/blob/be5dbcbb2e5afef17d47b4cbbf9e0eb08de69da4/src/com/owncloud/android/ui/preview/PreviewTextFragment.java#L266-271) but corresponding to R.id.action_copy will have to be added to onPrepareOptionsMenu in order to prevent the "Copy" option from appearing in the text preview fragment.

davivel:
@Stoyicker , 404s are working fine for me.

Closing in favor of #1102 , after rebasing. Let's follow the validation there.

Thanks.

PR #798, opened 2014-12-11
on downgrading upload db, delete and re-create

LukeOwncloud:
This fixes a crash after downgrading the upload db (that is, when coming back from reliable_uploads to develop branch) and then uploading a file.
Without this patch upload succeeds but oc app crashes.

@davivel  This will make live easier when inspecting reliable_uploads branch. Please grant it priority. It is only 3 lines of code.


davivel:
This is OK to QA @purigarcia

davivel:
Validated. :+1:

Issue #680, opened 2014-10-25
Slide-show does not obey sorting order


tobiasKaminsky:
When changing sorting order to "newest -> oldest" the slide show does not obey this.
Instead it shows the next file according to "sorting by name" files when swiping left/right.

jabarros:
It is merged and will be available in next release.
Thanks @tobiasKaminsky!

PR #820, opened 2014-12-26
Obey sort order fixmerge

tobiasKaminsky:
According to #680


davivel:
Validated and there it goes.

Thanks, @tobiasKaminsky

PR #822, opened 2014-12-27
PIN: Replaced AlertDialog with Toast, request PIN 10s after leaving app

### LINKED ISSUE #76: "App Pin Removed" message doesn't necessarily get you back to the settings.
After having entered the last digit of the pin, a dialog appears telling me that the pin has been removed. I can tap on blank space around the dialog to close it without geting back to the settings view.

### LINKED ISSUE #257: PIN not requested on startup
I have configured the ownCloud Android app on my phone with a PIN. Nevertheless, opening the application never actually requests entering the PIN.

App version: 1.4.4
Android: 4.1.2

### LINKED ISSUE #667: Display PIN prompt when uploading
With the PIN prompt enabled, when the user uploads they're not prompted to input their PIN.

This is a security risk and should be labeled as such.

tobiasKaminsky:
- fix #76 
- fix #257 
- fix #667

### DIFF
--- a/res/xml/preferences.xml
+++ b/res/xml/preferences.xml
@@ -27,8 +27,7 @@
 	        android:title="@string/prefs_select_oc_account"
 	        android:summary="@string/prefs_summary_select_oc_account"
 	        / -->
-	    <com.owncloud.android.ui.CheckBoxPreferenceWithLongTitle android:title="@string/prefs_pincode" android:key="set_pincode" 
-                        android:summary="@string/prefs_pincode_summary"/>
+	    <android.preference.CheckBoxPreference android:title="@string/prefs_pincode" android:key="set_pincode" />
 	</PreferenceCategory>
 
     <PreferenceCategory android:title="@string/prefs_category_instant_uploading">
--- a/src/com/owncloud/android/authentication/PinCheck.java
+++ b/src/com/owncloud/android/authentication/PinCheck.java
@@ -1,0 +1,28 @@
+package com.owncloud.android.authentication;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+
+import com.owncloud.android.MainApp;
+import com.owncloud.android.ui.activity.PinCodeActivity;
+
+public class PinCheck extends Activity {
+
+    private static Long timestamp = 0l;
+    
+    public static void setUnlockTimestamp() {
+        timestamp = System.currentTimeMillis();
+    }
+
+    public static boolean checkIfPinEntry(){
+        if ((System.currentTimeMillis() - timestamp) > 10000){
+            SharedPreferences appPrefs = PreferenceManager.getDefaultSharedPreferences(MainApp.getAppContext());
+            if (appPrefs.getBoolean("set_pincode", false)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/src/com/owncloud/android/authentication/PinCheck.java
+++ b/src/com/owncloud/android/authentication/PinCheck.java
@@ -10,14 +10,15 @@ public class PinCheck extends Activity {
 
     private static Long timestamp = 0l;
     private static Long lastStart = 0l;
+    private static int PINCODE_TIMEOUT = 10000;
     
     public static void setUnlockTimestamp() {
         timestamp = System.currentTimeMillis();
     }
 
     public static boolean checkIfPinEntry(){
-        if ((System.currentTimeMillis() - timestamp) > 10000 &&
-            (System.currentTimeMillis() - lastStart) > 10000){
+        if ((System.currentTimeMillis() - timestamp) > PINCODE_TIMEOUT &&
+            (System.currentTimeMillis() - lastStart) > PINCODE_TIMEOUT){
             SharedPreferences appPrefs = PreferenceManager.getDefaultSharedPreferences(MainApp.getAppContext());
             if (appPrefs.getBoolean("set_pincode", false)) {
                 lastStart = System.currentTimeMillis();
--- a/src/com/owncloud/android/authentication/PinCheck.java
+++ b/src/com/owncloud/android/authentication/PinCheck.java
@@ -1,25 +1,26 @@
 package com.owncloud.android.authentication;
 
 import android.app.Activity;
-import android.content.Intent;
 import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 
 import com.owncloud.android.MainApp;
-import com.owncloud.android.ui.activity.PinCodeActivity;
 
 public class PinCheck extends Activity {
 
     private static Long timestamp = 0l;
+    private static Long lastStart = 0l;
     
     public static void setUnlockTimestamp() {
         timestamp = System.currentTimeMillis();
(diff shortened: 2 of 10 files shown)


jancborchardt:
@rperezb @davivel this is reviewed from my side. Can you look into it?

res/xml/preferences.xml
```diff
@@ -27,8 +27,7 @@
 	        android:title="@string/prefs_select_oc_account"
 	        android:summary="@string/prefs_summary_select_oc_account"
 	        / -->
-	    <com.owncloud.android.ui.CheckBoxPreferenceWithLongTitle android:title="@string/prefs_pincode" android:key="set_pincode" 
-                        android:summary="@string/prefs_pincode_summary"/>
+	    <android.preference.CheckBoxPreference android:title="@string/prefs_pincode" android:key="set_pincode" />
```
jancborchardt: There’s really nothing needed to add. We shouldn’t add additional information just for the sake of it.

jabarros: Cool! Thanks @jancborchardt

src/com/owncloud/android/authentication/PinCheck.java
```diff
@@ -0,0 +1,29 @@
...
+    private static Long timestamp = 0l;
+    private static Long lastStart = 0l;
+    
+    public static void setUnlockTimestamp() {
+        timestamp = System.currentTimeMillis();
+    }
+
+    public static boolean checkIfPinEntry(){
+        if ((System.currentTimeMillis() - timestamp) > 10000 &&
+            (System.currentTimeMillis() - lastStart) > 10000){
```
tobiasKaminsky: Done

src/com/owncloud/android/authentication/PinCheck.java
```diff
@@ -0,0 +1,30 @@
+package com.owncloud.android.authentication;
+
+import android.app.Activity;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
+
+import com.owncloud.android.MainApp;
+
+public class PinCheck extends Activity {
```
davivel: This should not be an Activity; it's never started, it only holds a couple of static methods... it should be a Utils class, or would be possible just moving the methods to the PinCodeActivity and allow accessing them publicly, out the of its regular life cycle.

src/com/owncloud/android/authentication/PinCheck.java
```diff
@@ -0,0 +1,30 @@
...
+    private static int PINCODE_TIMEOUT = 10000;
+    
+    public static void setUnlockTimestamp() {
+        timestamp = System.currentTimeMillis();
+    }
+
+    public static boolean checkIfPinEntry(){
+        if ((System.currentTimeMillis() - timestamp) > PINCODE_TIMEOUT &&
+            (System.currentTimeMillis() - lastStart) > PINCODE_TIMEOUT){
+            SharedPreferences appPrefs = PreferenceManager.getDefaultSharedPreferences(MainApp.getAppContext());
```
davivel: Whenever possible is better not using the Application Context; in this case, a parameter Context ctx in the method would do the trick; the callers are always Activities, they only need to pass 'this' as the value for the parameter.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -802,6 +804,12 @@ protected void onResume() {
         super.onResume();
         Log_OC.e(TAG, "onResume() start");
         
+        if (PinCheck.checkIfPinEntry()){
+            Intent i = new Intent(MainApp.getAppContext(), PinCodeActivity.class);
```
davivel: Since this is duplicated in onCreate(...) and in onResume(), the code to check the PIN code is being called twice when the Activity is initialized. Now it's working because the timeout for the PIN code is 10 seconds, and it's impossible so much time goes from onCreate() to onResume() execution; but, if there were no timeout, then we would ask the PIN code twice to the user. It is risky, thinking about possible future changes.

Besides, when the user comes back from a modal dialog, onResume() is also called. I didn't test it yet, but I bet that if we click the 'upload' icon, wait for 10 seconds doing nothing, and the cancel, we will be requested for the PIN. All of it without exiting the app.

Probably the best place to do the check is onStart()  (onCreate() is not good enough); it is called every time the activity is becoming visible, but not after modal dialogs partially hiding the app disappear.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -847,7 +855,7 @@ protected void onPause() {
             mDownloadFinishReceiver = null;
         }
         
-        
+        PinCheck.setUnlockTimestamp();
```
davivel: This is correctly paired with onResume(), but should move to onStop() when the check is moved to onStart().

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -308,6 +309,8 @@ public boolean onPreferenceClick(Preference preference) {
     @Override
     protected void onPause() {
         super.onPause();
+        
+        PinCheck.setUnlockTimestamp();
```
davivel: This is not enough to protect the app when the user press HOME while showing the Settings activity.

src/com/owncloud/android/ui/activity/PinCodeActivity.java
```diff
@@ -553,32 +530,15 @@ public void onClick(DialogInterface dialog, int which) {
...
-            @Override
-            public void onClick(DialogInterface dialog, int which) {
-                finish();
-                return; 
-            }
-            
-        });
-        aDialog.show(); 
+        
+        Toast.makeText(this, cseq, Toast.LENGTH_LONG).show();
```
davivel: :+1:

src/com/owncloud/android/ui/preview/PreviewImageActivity.java
```diff
@@ -103,8 +105,10 @@ protected void onCreate(Bundle savedInstanceState) {
         actionBar.hide();
         
         // PIN CODE request
-        if (getIntent().getExtras() != null && savedInstanceState == null && fromNotification()) {
-            requestPinCode();
+        if (PinCheck.checkIfPinEntry()){
```
davivel: We should refactor all this, so that FileActivity does the work with the PIN code, and not every of it's children. That would not cover all the activities, but would save some repeated code.

davivel:
Not ready to go, but since this became a top priority for this sprint, I'd go on from here.

Thanks, @tobiasKaminsky . If you'd like to continue working on it directly, ping me here or in IRC.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
davivel: For the record: onStart() / onStop() will not do it; when you go back from an activity to the previous one, onStart() in the older is called before onStop() in the exiting one; the time that will compared in onStart() is with its own stop before starting the newest.

tobiasKaminsky:
@davivel Currently I am very busy.
But I hope, you can continue with this.

davivel:
@tobiasKaminsky sure; thanks a lot.

PR #824, opened 2014-12-29
Account creation fix

### LINKED ISSUE #603: Create account page shows "unknown error" when adding URL with trailing slash
When I enter the owncloud URL with a trailing slash:
https://192.168.0.3/owncloud/
I get an "unknown error" message when hitting the enter button.

The Apache2 access log was showing the following entries:
192.168.0.246 - - [01/Sep/2014:20:35:21 +0000] "GET /owncloud//status.php HTTP/1.1" 200 773 "-" "Android-ownCloud"
192.168.0.246 - - [01/Sep/2014:20:35:21 +0000] "HEAD /owncloud//remote.php/webdav/ HTTP/1.1" 403 681 "-" "Android-ownCloud"

You could see that the HEAD request gets a 403 Forbidden response and the context path was showing a double slash "owncloud//status.php".

After removing the trailing slash the error message went away. The Android App should strip the trailing slash so that the user does not get confused by this kind of error.

tobiasKaminsky:
fixes #603

### DIFF
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -733,7 +733,6 @@ SsoWebViewClientListener, OnSslUntrustedCertListener {
     private void checkOcServer() {
         // Remove multiple "/" at the end
         String uri = mHostUrlInput.getText().toString().trim().replaceFirst("/*$", "/");
-        Log_OC.d("Auth", "uri: " + uri);
         mServerIsValid = false;
         mServerIsChecked = false;
         mOkButton.setEnabled(false);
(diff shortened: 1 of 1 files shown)


jabarros:
This is a bug fix fast and clean.

davivel:
As told in #603, our tests show this was previously fixed. 

Besides, the fix is not strict enough. This allows the user to input a URL finishing with any length of slashes, and take it as a right URL, *but it is not*. In a general sense, trying to "guess" what was the user trying to input instead of showing an error when she inputs somethign wrong is a dangerous approach. IMHO, trimming blanks is the limit.

The error of two slashes was not due to a problem in the input field, but to a wrong concatenation of the URL set by the user and the path to the WebDAV entry point. This was already fixed, and should be working correcly from version 1.6.1 (or 1.6.0, not sure) of the OC app. 

Of course, maybe the fix was not complete and there are still situations where it happens. But according to our tests, the context described in #603 is working fine. If any other issue related to double slash appears, we will review again.

In any case, @tobiasKaminsky , as always, thanks a lot for your help.

PR #844, opened 2015-01-19
Trigger Media Scan on parent folder when deleting a file or a folder fro...

jabarros:
...m locally


src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -1498,4 +1499,23 @@ public void triggerMediaScan(String path) {
         MainApp.getAppContext().sendBroadcast(intent);
     }
 
+    public void deleteFileInMediaScan(String path) {
+
+        String mimetypeString = FileStorageUtils.getMimeTypeFromName(path);
+        if (mimetypeString.startsWith("image/")) {
+            // Images
+            getContentResolver().delete(MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
```
davivel: Problem here: getContentResolver() returns a NULL value when in context of SyncAdapter. It's necessary to check that the returned value is not NULL and, otherwise, call instead getContentProviderClient(). Have a look to other methods in this class calling getContentResolver().

masensio: Done.

davivel:
:+1:

PR #858, opened 2015-01-21
Fixed. Footer message should be showed in the list view mode. Also added...

jabarros:
... in grid view mode.


davivel:
:+1:

jospoortvliet:
Do you guys not delete branches, @davivel ? Is that intentional?

PR #868, opened 2015-01-30
Fixed start of SAML dialog

davivel:


davivel:
@purigarcia , ready to test.

davivel:
Not on develop.

PR #872, opened 2015-02-04
Update strings.xml

wanam:
Fixed some french translations.


masensio:
Hi @wanam,

thanks for your contribution.
For updates in translations we are using Transifex instead of the code. So we'd appreciate if you update this through Transifex https://www.transifex.com/projects/p/owncloud/.

Otherwise we don't accept pull request into android:master. So please, for future contributions you must do the pull request on android:develop.
Also, we need you sign and send the contributor agreement at http://owncloud.org/about/contributor-agreement/ to @karlitschek .

Thanks.

PR #918, opened 2015-03-10
Share password support

masensio:


davivel:
:+1:

if12b100:
hey nice work! =) I was working on this too. 
1 more thing: I think it should be possible to share with password even if the "enforce password" option on the server isn't checked.

PR #920, opened 2015-03-17
Fix crash in android 2: gridview

masensio:


davivel:
This may go for the moment, but need to refactor as soon as we get rid of Android 2.x support definitely.

:+1:

PR #924, opened 2015-03-20
Add user agent support

masensio:


src/com/owncloud/android/MainApp.java
```diff
@@ -117,4 +120,24 @@ public static String getLogName() {
...
 
+    // user agent
+    public static String getUserAgent() {
+        String appString = getAppContext().getResources().getString(R.string.user_agent);
+        String packageName = getAppContext().getPackageName();
+        PackageInfo pInfo = null;
+        try {
+            pInfo = getAppContext().getPackageManager().getPackageInfo(packageName, 0);
+        } catch (PackageManager.NameNotFoundException e) {
+
```
davivel: Should log a warning here

src/com/owncloud/android/MainApp.java
```diff
@@ -117,4 +120,24 @@ public static String getLogName() {
...
+        }
+        String version = "";
+        if (pInfo != null) {
+            version = "/" + pInfo.versionName;
```
davivel: This should go inside the previous try ... catch block to avoid a NullPointerException in ```pInfo.versionname```

davivel:
We need to review your autoformatting options; should apply only to lines modified, not to whole files.

masensio:
Changes done. :smile:

davivel:
:+1:

PR #943, opened 2015-04-12
Implemented soft reference in memory cache for thumbnails

Spacefish:
Drastically improves scrolling performance in folders with a lot of images.

PR #944, opened 2015-04-12
implemented soft reference in memory cache for thumbnails

Spacefish:
this is pr #943 but against develop instead of master. I also some functionality to make sure that the cache is consistent, when using put, remove, clear on the class.


nedimhx:
When I open folder where are pics every time it load me first 2 thumbnails and for other 5-6 need about 30 sec to wait until they are loaded, but when I "pull to refresh" every 1-2 sec thumbnails are loading faster.

Is it possible to compres thumbnails  to reduce quality which would speed up thumbnails previews ?

tobiasKaminsky:
@nedimhx This has absolutely nothing to do with this pull request. 
The thumbnails have currently the png format, which is not ideal for images.
I hope that /core will change this soon.

Regarding your first problem: please open a new issue and describe your problem with the template.

Spacefish:
I have seen this problem as well. The picture box is not repainted after the image is set. But this has really nothing to do with my pull request. The problem already existed before.

Spacefish:
Any updates on this? Will this be merged?

masensio:
Hi @Spacefish,

Thanks for your contribution.

We are fighting with the thumbnails problem.
But before merging your code we need you sign and send the contributor agreement at http://owncloud.org/about/contributor-agreement/ to @karlitschek .

Also we need to validate this PR before merging it.

Thanks in advance

Spacefish:
@masensio  thanks for your answer, i have already signed the contributor agreement, i have commit rights  for most of the repositories here, but android is not one of them.

masensio:
@Spacefish, thanks for your answer. Sorry I've looked for your name in Android-devs team and I haven't  found it. 
After your answer, I've seen your name in core-devs team. :smile:

Spacefish:
Any updates on this? The commit is still compatible with the latest source in develop.. Can we merge it? I have signed the developers agreement, but i have no rights in the android repository here :(

tobiasKaminsky:
There is a small typo in the task: thumbnail is missing a n

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -344,6 +321,48 @@ public View getView(int position, View convertView, ViewGroup parent) {
         return view;
     }
 
+    private class ThumbailLoaderTask extends AsyncTask<Object, Void, Bitmap> {
```
masensio: ThumbnailLoaderTask instead of ThumbailLoaderTask ;)

masensio:
@Spacefish could you update this PR with the last version of develop?

As soon as this branch is updated, this PR will be ready to be tested.

davivel:
Sorry, I need more info.

What's the advantage of using ```SoftReference``` in this case?

Thanks in advance.

davivel:
Reviewing better this time.

So, the point is not the ```SoftReference``` as much as there is a new cache level in memory, besides the disk cache. Makes sense.

I'll add a couple of comments.

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -46,6 +50,8 @@
             
     private static final String TAG = DiskLruImageCache.class.getSimpleName();
 
+    private Map<String, SoftReference<Bitmap>> mBitmapCache = Collections.synchronizedMap(new HashMap<String, SoftReference<Bitmap>>());
```
davivel: Please, wrap lines at 120 characters for better readability in Github

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -76,6 +82,18 @@ public void put( String key, Bitmap data ) {
         DiskLruCache.Editor editor = null;
         String validKey = convertToValidKey(key);
         try {
+            // put into in memmory cache
+            if(mBitmapCache.containsKey(key)) {
```
davivel: Is this really needed? Is not the value replaced calling directly ```mBitmapCache.put(...)```?

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -120,10 +151,12 @@ public Bitmap getBitmap( String key ) {
             }
             final InputStream in = snapshot.getInputStream( 0 );
             if ( in != null ) {
-                final BufferedInputStream buffIn = 
+                final BufferedInputStream buffIn =
                 new BufferedInputStream( in, IO_BUFFER_SIZE );
-                bitmap = BitmapFactory.decodeStream( buffIn );              
-            }   
+                bitmap = BitmapFactory.decodeStream( buffIn );
+                buffIn.close();
```
davivel: Nice. Even better, both close() methods could be moved into the ```finally``` block below, with a NULL check condition, so that the are done even in case of Exceptions.

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
davivel: Refactoring the original block in a reusable item is really needed, great point. 

At first I didn't understand the new flow, but in a second thought, seems that makes sense.

Anyway, is not there an easier way to write it? Maybe merging both ```ThumbnailLoaderTask``` and ```ThumbnailGenerationTask``` in a single ```AsyncTask```? Just wondering, didn't think it in depth.

davivel:
@Spacefish , I made some comments.

Could you also rebase your branch on ```owncloud:master``` and then make a new PR onto it? We changed the PR workflow since the time you created this PR. I'm really sorry, I know it may be disturbing, specially since you already changed the target branch before.

PR #951, opened 2015-04-17
Loging different servers same pattern

masensio:


src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -168,6 +168,11 @@ public boolean fileExists(String path) {
         return ret;
     }
 
+    public Vector<OCFile> getDownloadedFilesByAccount(String accountName) {
```
davivel: This method is not used; remove it please.

davivel:
:+1: , waiting for Travis to merge

PR #957, opened 2015-04-22
Fix crash invalid server url

masensio:


davivel:
:+1: ; but would prefer a more general solution

PR #967, opened 2015-04-30
Fix upload fails sending another app

masensio:


src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
```diff
@@ -0,0 +1,138 @@
+/**
+ *   ownCloud Android client application
+ *
+ *   Copyright (C) 2015 ownCloud Inc.
```
davivel: @author should go here

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
```diff
@@ -0,0 +1,138 @@
...
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package com.owncloud.android.utils;
+
+import android.app.Activity;
+import android.content.ContentResolver;
+import android.content.Context;
```
davivel: This import is not used.

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
```diff
@@ -0,0 +1,138 @@
...
+    private String mAccountName;
+    private ContentResolver mContentResolver;
+    private int mIndex;
+
+    public int getIndex(){
+        return mIndex;
+    }
+
+    public CopyTmpFileAsyncTask(Activity activity) {
+        mContentResolver = ((FileActivity) activity).getContentResolver();
```
davivel: So much casting is not a good sign; I would change the constructor to receive one single parameter of type OnCopyTmpFileTaskListener. I would pass both the account name and the content resolver as additional parameters to the execute call, and receive them in doInBackground().

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
```diff
@@ -0,0 +1,138 @@
...
+                }
+
+                if (fullTempPath != null) {
+                    File f = new File(fullTempPath);
+                    f.delete();
+                }
+                result =  null;
+            }
+        } else {
+            Log_OC.e(TAG, "Error in parameters number");
```
davivel: This should be more "assertive"; throw an IllegalArgumentException to break the app. This would be a programming error, not an error condition in runtime.

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
```diff
@@ -0,0 +1,138 @@
...
+            listener.OnCopyTmpFileTaskListener(result, mIndex);
+        }
+    }
+
+    /*
+     * Interface to retrieve data from recognition task
+     */
+    public interface OnCopyTmpFileTaskListener{
+
+        void OnCopyTmpFileTaskListener(String result, int index);
```
davivel: The first letter should not be uppercase in a method; besides, would be better a name not exactly the same as the class, remarking what event just happened; for instance, ```onTmpFileCopied(...)```

src/com/owncloud/android/ui/activity/Uploader.java
```diff
@@ -29,6 +29,7 @@
 import java.util.Stack;
 import java.util.Vector;
 
+
```
davivel: Add yourself as @author in the starting comment

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
masensio: I've added @author masensio here.

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
masensio: Imports updated :)

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
masensio: +1.
I've changed the parameters for doInBackground. Now, execute method needs:
     - Uri: uri of file
     - String: path for saving the file into the app
     - int: index of upload
     - String: accountName
     - ContentResolver: content resolver

src/com/owncloud/android/ui/activity/Uploader.java
masensio: Done

src/com/owncloud/android/utils/CopyTmpFileAsyncTask.java
masensio: +1

davivel:
:+1:

PR #969, opened 2015-05-03
Convert AsyncTask to IntentService

yulin2:
Hey owncloud-android developers, I'm doing research on Android async programming, particularly on 
AsyncTask and IntentService. AsyncTask can lead to memory leak and losing task result when GUI is recreated during task running (such as orientation change). [This article](http://bon-app-etit.blogspot.com/2013/04/the-dark-side-of-asynctask.html) describe the problems very well. 

We discussed with some Android experts and they agree with this issue, and claim that AsyncTask can be considered only for short tasks (less than 1 second). However, using IntentService (or AsyncTaskLoader) can avoid such problems since their lifecycles are independent from `Activity`.

For example, in `LogHistoryActivity`, even if you use `WeakReference` for a text view in the AsyncTask, 
you still hold a strong reference to the activity itself, which can lead to the problems described above. (I also opened a related pr before #892)

I refactored three AsyncTasks in `owncloud-android` to IntentService, with the help of a refactoring tool we developed. Do you think using IntentService is better in these three activities, and merge this pr?

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -151,6 +151,7 @@
         <service android:name=".files.services.FileUploader" />
         <service android:name=".media.MediaService" />
         <service android:name=".services.LoadingLogService" android:exported="false"/>
+        <service android:name=".services.MoveFilesService" android:exported="false"/>
         
         <activity android:name=".ui.activity.PinCodeActivity" />
         <activity android:name=".ui.activity.ConflictsResolveActivity"/>
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -152,6 +152,7 @@
         <service android:name=".media.MediaService" />
         <service android:name=".services.LoadingLogService" android:exported="false"/>
         <service android:name=".services.MoveFilesService" android:exported="false"/>
+        <service android:name=".services.CheckAvailableSpaceService" android:exported="false"/>
         
         <activity android:name=".ui.activity.PinCodeActivity" />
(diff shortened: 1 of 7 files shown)


yulin2:
I sent this pr serval days ago. Any comments/suggestions on the changes?

masensio:
Hi @yulin2,

Thanks for your contribution.

Is it your first contribution in the OwnCloud Project? I've been searching your nickname in the developers list and I have not found it. If you haven't signed the Contributor Agreement, please send it ( https://owncloud.org/contribute/agreement/) to @karlitschek. We can't merged your PR without this agreement.

On the other hand, could you update your branch with the last version of the code? It is better to do the code review easily.

A question, why did you select IntentService instead of AsyncTaskLoader as in your PR https://github.com/owncloud/android/pull/892?

Thanks again

davivel:
Need to think a bit about this. Not sure that replacing ```AsyncTasks``` with ```IntentService``` as a rule is fine for any case. 

At first sight, I see a possible problem with returning results through ```LocalBroadcastManager```. If I recall correctly, these broadcast messages cannot be made sticky, so could be lost by the listener ```Activity``` during rotations.

Besides, we'd like reduce the number of ```Services``` in the app. Right now we have too many for operations that could be merged in a single one. This seems to go in the other way. But this should not be a decision criteria for itself.

davigonz:
Need to check if this still relevant in the context of new architecture #2351

PR #970, opened 2015-05-05
PIN code no more: Pass code protects all the activities in the app

davivel:


src/com/owncloud/android/ui/preview/PreviewImageActivity.java
```diff
@@ -42,6 +40,7 @@
 import com.ortiz.touch.ExtendedViewPager;
 import com.owncloud.android.R;
 import com.owncloud.android.authentication.AccountUtils;
+import com.owncloud.android.authentication.PassCodeManager;
```
masensio: This import is not used

src/com/owncloud/android/ui/preview/PreviewImageActivity.java
```diff
@@ -90,6 +88,8 @@
     
     private View mFullScreenAnchorView;
     
+    private Boolean mUnlocked = false;
```
masensio: mUnlocked is not used

PR #975, opened 2015-05-11
Fix refresh large default storage

masensio:
Adapt WebDav parser for receiving exponential values in QUOTAS: Update Reference


davivel:
:+1:

Issue #714, opened 2014-11-08
Image slideshow supported by a LRU-cache of reduced-size images


tobiasKaminsky:
I want to make a "new" suggestion, how to handle images and the storage of them.
This is a summarized version of my ideas in #588 and #429.
I will update this "user story" according to our discussion in this thread.

IDEA:
When the user opens a file he will get a reduced version of the image. This will be adapted to his screen (and 75% compress rate). Therefore we have no visual quality loss, but have to transfer a much smaller file (3,1MB --> 83,8Kb).
This file will be cached and re-used. Caching system is already implemented. The cache should be adjustable in the settings (default to 100Mb).
If the user wants to use the file (share, open with another app) he will get prompted:
- use reduced size (e.g. 1024x768 [shows real image dimension])
- Download & use original file
  Also there is a new option on long-press for images and in options in the slideshow: Download original file.

EXPLANATION:
A cache is used to have a specific amount of storage our app will use for "background operations". If the user manually downloads a file he wants to have it and cannot complain about the big file. But in all other cases the user should not get the idea that owncloud is flooding his storage.

Prompting the user to use reduced/full size is also very useful as Whatsapp/Facebook... always shrink the images by themself, even if one uploads a full sized image. Also for sending an image via email it is useful to have this option. 

Offering the option to download the full sized image is mandatory as it is the current behaviour and there might be (a few) user who want to do image processing or see a very small detail of the image and therefore have to zoom in.

I think this is a good compromise regarding storage size and data traffic.

Later on, as @davivel suggested, we can try to implement different user profiles.

Mention all user that have posted on #588 #429:
@davivel @jancborchardt @Nowaker @NetAction @enoch85 @stefan-niedermann @LukeOwncloud  @LukasReschke

tobiasKaminsky:
An update: This is working for me right now.
Problem remaining is to share a cached file and https://github.com/owncloud/core/issues/16250

But I am getting closer to it :)

tobiasKaminsky:
I cannot create a PR since https://github.com/owncloud/core/issues/16250 is still not solved.
@davivel can you vote for this to get this implemented soon?

davivel:
I'm afraid that won't be so easy. Is there a PR for this? Is the code in a branch in this repo or in your fork?

tobiasKaminsky:
@davivel 
A PR is done in #1044.
The branch is in this repo.

Looking forward to have this feature included :)

DeepDiver1975:
[...]

PR #982, opened 2015-05-18
resized image is loaded instead of full sized image

tobiasKaminsky:
Implemented as discussed in https://github.com/owncloud/android/issues/714


src/com/owncloud/android/ui/adapter/DiskLruImageCacheFileProvider.java
```diff
@@ -0,0 +1,121 @@
+/**
+ *   ownCloud Android client application
+ *
+ *   Copyright (C) 2015 Tobias Kaminsky
```
masensio: Please don't add your Copyright in the license. You can put your name as @author

tobiasKaminsky:
Closing at is in my local repo and #1044 is the same.

PR #1012, opened 2015-06-13
App Install Location

SweetLion:
The internal memory of my device is quite small, this PR should allow users to easily move the app to the microSD.

"To allow the system to install your application on the external storage, modify your manifest file to include the android:installLocation attribute in the element, with a value of either "preferExternal" or "auto"." (source: https://developer.android.com/guide/topics/data/install-location.html)

Thank you very much for your awesome app :)

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -18,6 +18,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
  -->
 <manifest package="com.owncloud.android"
+    android:installLocation="auto"
     android:versionCode="10700100"
     android:versionName="1.7.1" xmlns:android="http://schemas.android.com/apk/res/android">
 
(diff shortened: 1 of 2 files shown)


masensio:
Hi @SweetLion,

Thanks for your contribution.

I think it is interesting to can move the app to the sdcard.
But before merging your code we need you sign and send the contributor agreement at http://owncloud.org/about/contributor-agreement/ to @karlitschek .

Also we need to validate this PR before merging it. Please, update your code with our last changes in develop branch, we need the last version of the code for the validation process. 

Thanks in advance

SweetLion:
Hi @masensio and @karlitschek, thank you for your reply and interest in this topic. My "code" is just one additional element in the Android manifest. Is it possible that someone of the ownCloud team pushes this change to the project? I do not want to sign the contributor agreement.

karlitschek:
@SweetLion @masensio. Great this works for me. And yes. No contributor agreement is needed for this change :-)

masensio:
Ok, I've created another branch from develop including the change in the manifest.  Also, I've created another Pull Request with this change (https://github.com/owncloud/android/pull/1017).
So, I'll close this pull request.

@SweetLion  thanks for your contribution in any case :smile:

PR #1014, opened 2015-06-15
Forbidden characters from server

masensio:


davivel:
This is fine, but depends on owncloud-library#71 , and I need to check that a bit more...

davivel:
:+1:

PR #1017, opened 2015-06-18
App install location.

masensio:
Add android:installLocation = "auto" into manifest.

This pull request contains the changes from https://github.com/owncloud/android/pull/1012
Thanks to @SweetLion

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -18,6 +18,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
  -->
 <manifest package="com.owncloud.android"
+    android:installLocation="auto"
     android:versionCode="10700100"
     android:versionName="1.7.1" xmlns:android="http://schemas.android.com/apk/res/android">
 


masensio:
This change has more implications than apparent at first, so we need to study if we introduce it in the app or not. 

Now, I'll close the PR and if in the future we consider that we need this change, we will create again.

tobiasKaminsky:
@masensio For me it would be interesting what the implications are.
thank you

davivel:
@tobiasKaminsky , some excerpts from http://developer.android.com/guide/topics/data/install-location.html : 

> Warning: When the user enables USB mass storage to share files with a computer or unmounts the SD card via the system settings, the external storage is unmounted from the device and all applications running on the external storage are immediately killed.

> You should not allow your application to be installed on the external storage if it uses any of the following features, due to the cited consequences when the external storage is unmounted:
> (...)
> Your running Service will be killed and will not be restarted when external storage is remounted.
> Your accounts created with AccountManager will disappear until external storage is remounted.
> Your AbstractThreadedSyncAdapter and all its sync functionality will not work until external storage is remounted.

tobiasKaminsky:
@davivel Thank you for the explanation!
Sounds indeed very complicated. 

I think that this is not needed as the app itself is very small.
Changing the default storage path would be the right way. (But this could cause problems as well if the directory is unmounted).

davivel:
Yes. And we'll have to gracefully deal with the errors that could occur.

You're welcome!

PR #1028, opened 2015-06-23
return false if remove failed

LukeOwncloud:
if OCFile is null --> cannot delete ==> return false.

### DIFF
--- a/src/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/src/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -467,7 +467,7 @@ public class FileDataStorageManager {
     
 
     public boolean removeFile(OCFile file, boolean removeDBData, boolean removeLocalCopy) {
-        boolean success = true;
+        boolean success = false;
         if (file != null) {
             if (file.isFolder()) {
                 success = removeFolder(file, removeDBData, removeLocalCopy);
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -467,7 +467,7 @@ public void saveFolder(
     
 
     public boolean removeFile(OCFile file, boolean removeDBData, boolean removeLocalCopy) {
-        boolean success = true;
+        boolean success = false;
```
davivel: I'm sorry to say this change breaks the condition in line 497 for input parameters ```file != null && !file.isFolder()``` ```removeDBData == false```. 

I'd say the easiest way to achieve what you want is with adding an ```else``` block in line 509.

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -512,7 +512,7 @@ public boolean removeFile(OCFile file, boolean removeDBData, boolean removeLocal
     
 
     public boolean removeFolder(OCFile folder, boolean removeDBData, boolean removeLocalContent) {
-        boolean success = true;
+        boolean success = false;
```
davivel: Similarly, this breaks the condition in line 520 for input parameters ```folder != null && folder.isFolder() ```, ```removeDBData == false ```. 

Once again, I'd go with an ```else``` block in line 523.

davivel:
@LukeOwncloud , please, see my comments and make a new PR onto ```master ``` branch. Thanks a lot :)

LukeOwncloud:
Very true!

Issue #897, opened 2015-02-24
Not displaying server side video thumbnails


fxshpac:
Running owncloud core version 8.0.0-5 and android version 1.7.0, I want to report that, in gallery view, the thumbnails for videos are still showing default icon, even after I enabled OC/Preview/Movie on server side. The web interface displays them just fine. Therefore I suspect it's the Android app which is not taking advantage of the API to retrieve these thumbnails for videos.

PR: ~~https://github.com/owncloud/android/pull/1035~~ https://github.com/owncloud/android/pull/1780

masensio:
Hi @fxshpac,
sorry but ownCloud server doesn't provide thumbnails for videos.

Show thumbnails for videos is an improvment for our app. 
What do you think @jancborchardt about this?

jancborchardt:
Yeah, video thumbnails would be very good. @DeepDiver1975 anything that needs to be added to the server to enable this?

tobiasKaminsky:
I have installed ffmpeg and ffmpegthumbnailer.
After that the server can generate thumbnails.
With one very little change the app can also display these thumbnails.
A PR for this is coming soon.

tobiasKaminsky:
PR is currently not possible as I ave committed it directly.
Sorry for this :/

tobiasKaminsky:
PR: #1035

tobiasKaminsky:
Remaining problem is that in gridView the user cannot distinguish between image and video if both have a thumbnail.
@jancborchardt Do you have an idea? Maybe display a film strip like this http://www.vorlagendrucken.de/wp-content/uploads/2014/01/Einladung-Gutschein-Filmstreifen_1.jpg

jancborchardt:
The thumbnail for videos should have a play button overlayed: https://github.com/owncloud/core/blob/master/core/img/actions/view-play.png (SVG in the same folder), that’s the common way to distinguish videos from images.

tobiasKaminsky:
Ok. I will add this.
Is really better than film strips.
Any suggestion about the size? Half of the size, centered?

jancborchardt:
Yeah, half the size and centered sounds good.

jancborchardt:
Maybe third the size – experiment and see what looks good.

tobiasKaminsky:
![video_overlay_big](https://cloud.githubusercontent.com/assets/5836855/8878835/b3104cf4-322d-11e5-8050-2c2bf1394a2d.jpg)
[...]

PR #1035, opened 2015-06-27
enable video thumbnail

tobiasKaminsky:
Requires ffmpeg and ffmpegthumbnailer on server.


tobiasKaminsky:
Update: on server side ffmpgthumbnailer is not needed, but:
'enabledPreviewProviders' => array(
        'OC\Preview\Movie'
   )
in config.php

tobiasKaminsky:
PR for #897

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -176,8 +180,20 @@ protected Bitmap doInBackground(Object... params) {
...
+
+                    if (((OCFile) mFile).isVideo()){
+                        thumbnail = addVideoOverlay(thumbnail);
+                    }
                 }  else if (mFile instanceof File) {
                     thumbnail = doFileInBackground();
+
+                    String url = ((File) mFile).getAbsolutePath();
+                    FileNameMap fileNameMap = URLConnection.getFileNameMap();
+                    String mMimeType = fileNameMap.getContentTypeFor("file://" + url);
```
davivel: Please, use ```FileStorageUtils.getMimeTypeFromExtension(...)``` instead. That way we avoid introducing ```java.net.URLConnection``` in the imports. Let's prevent the temptation of using other network libraries.

davivel:
The 'play' PNG seems very small for high resolution screens.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -362,6 +378,51 @@ public static ThumbnailGenerationTask getBitmapWorkerTask(ImageView imageView) {
         return null;
     }
 
+    public static Bitmap addVideoOverlay(Bitmap thumbnail){
+        Bitmap playButton = BitmapFactory.decodeResource(MainApp.getAppContext().getResources(),
+                R.drawable.view_play);
+
+        Bitmap resizedPlayButton = Bitmap.createScaledBitmap(playButton,
```
davivel: Thumbnails are saved always with the same resolution in a given device, right? In that case, the scaled load of the 'play' image and the calculation for its coordinates should be done only once, and kept in static fields.

davivel:
Reviewed; please, see my comments.

AndyScherzinger:
@davivel and @tobiasKaminsky I just pushed a play icon for all the resolutions in standard 24dp sizes. Hope that is fine with you @tobiasKaminsky me simply committing this to your branch.
Please check if you are happy with the new icon size on higher resolution devices.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
AndyScherzinger: Since I added resolution dependent play icons for all sizes the scaling shouldn't be needed anymore and the icon should be sharp on all device sizes.

tobiasKaminsky: The problem is that I need the exact coordinates of the three points forming the triangle to compute the exact visual center of the play button (which is not the center of the image).
So therefore I had modified the original image and cropped all outer space of the play button.
If we do this on all play buttons, I guess it is working.

tobiasKaminsky: @AndyScherzinger can I crop the images?
Scaling (or at least resizing) is still necessary as the visual center of the play button is not the center of the image.

AndyScherzinger: @tobiasKaminsky sure, go ahead. The pngs are the standard ones provided by googles material design icon repository

tobiasKaminsky: > Thumbnails are saved always with the same resolution in a given device, right? In that case, the scaled load of the 'play' image and the calculation for its coordinates should be done only once, and kept in static fields.

As video thumbnails can occur in file *and* grid view, the computation must be in this function, as it depends on the thumbnail size.

tobiasKaminsky:
@davivel fixed and added comments to your review

jesmrec:
Dear friends @owncloud/android-developers, this branch needs a rebase with master. Does everybody volunteer? :)

davivel:
I'll do it, @jesmrec .

davivel:
Rebased; replacing with a new PR to master

PR #1037, opened 2015-06-28
HOTFIX: wrong textview id lead to no folder name

tobiasKaminsky:
wrong textview id lead to "textview" instead of folder name.
@davivel


davivel:
Thanks, @tobiasKaminsky , but this was already fixed in the release-1.7.2 branch. we are deeply testing there the code for the incoming release, and fixing bugs there also.

tobiasKaminsky:
Ah. I never realized this branch.
I thought all  updates would be integrated in develop and from that branch the 1.7.2 release would just best compiled.

PR #1039, opened 2015-06-28
Updating the target and min SDK versions

jab416171:
Setting the min sdk version to 15 will only eliminate 5.9% of users. If this is too many, then feel free to reject this PR and I'll set it to 10.

https://developer.android.com/about/dashboards/index.html

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -36,8 +36,8 @@
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
     
     <uses-sdk
-        android:minSdkVersion="8"
-        android:targetSdkVersion="19" />
+        android:minSdkVersion="15"
+        android:targetSdkVersion="22" />
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" >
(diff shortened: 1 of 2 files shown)


masensio:
Hi @jab416171,

We have changed the minSdkVersion for the release 1.7.2. Now 14 is the value.

Also, we have planned to increase the targetSdkVersion and buildToolVersion when we will introduce Material Design.

So I'll close this PR without merging.

Thanks for your contribution.

jab416171:
@masensio 14? Why not 15? Nobody is on 14 (less than 0.1%).

Is there any harm in updating the targetSdkVersion and buildToolVersion now? Why wait until the material design work is done?

masensio:
@jab416171,

We've decided to do step by step:
- Remove ActionBarSherlock, add android-support-v7
- Change targetSdkVersion to 14, removing Android 2 support 
- Adapt the app to Material Design --> increase the buildToolVersion.

If we increase everything at the same time, we need to applied the new Material Design Theme to the app. We have preferred to do step by step, testing every change carefully.

Thanks for your ideas.

PR #1044, opened 2015-07-01
Resized images

tobiasKaminsky:
Every image is loaded in a reduced version. The width and height is determinated by the screen.
If the user wants to have the real file, he has to long-click and select download.

The benefit is a very small, but yet usable preview of the image.
It can be used to send via whatsapp, or as a file to another app (like k9) with the correct filename.

Remaining problem:
https://github.com/owncloud/core/issues/16250


tobiasKaminsky:
When the reduced image is shown we could add a FAB to download the original file.
@AndyScherzinger Do you want to do this? And maybe do the code review / functionality check?

tobiasKaminsky:
@davivel Can we schedule this for the next milestone or the one after that?
I think this is a very big improvement to save data traffic, especially when using mobile.

joeplus:
This would be huge and speed up the thumbnail loading significantly I can assume, in addition to the saved traffic!

tobiasKaminsky:
For now this PR depends on galleryplus for OC < 8.2 and gallery for OC >= 8.2.
By OC 9.0 we can change to the new preview webdav api.

AndyScherzinger:
@tobiasKaminsky As soon as #1100 is merged to master I can add a FAb here, in case this is fine with @jancborchardt :)

elpraga:
@tobiasKaminsky I am not 100% sure if we are talking about the same thing, but I guess You know better as You are more involved in this thread.

My original idea mentioned in #1278  was to show enlarged thumbnail and a spinning wheel instead of just a spinning wheel and a black background as for now while the larger preview is being downloaded / generated by the server (when we just look through pictures in fullscreen)
- in a similar way how google does it while we preview image search on Android.
This would give a user at least an idea what preview is he/she waiting for right now.

tobiasKaminsky:
@elpraga see #1278 
There we can discuss it better :)

tobiasKaminsky:
@rperez can we have this in the milestone after the current one?
As far as I have read in beta this is one of the most valuable improvements for the majority of the user.

rperezb:
@tobiasKaminsky curious by this comment: https://github.com/owncloud/android/pull/1044#issuecomment-147158641
Is it needed to have enabled the galleryplus app or oc9 ?

rperezb:
@jancborchardt what do you think about including this on the product? By default not downloading the full resolution image? 
cc @MTRichards

jancborchardt:
@tobiasKaminsky I’ve been using this in practice for quite some time in the Beta app. And I have to say it’s a bit unexpected. It’s great that the bigger thumbnail immediately shows so I can just view it. But as soon as the file is opened, it should download in the background. That’s somehow just expected.

Right now every time I zoom into a picture, I have to wait additionally for it to be downloaded. But opening a picture is a desire to view it, hence ownCloud should directly download it.

tobiasKaminsky:
Currently I am using the owncloud as a giant photo album and it happens quite often that I show 50-100 pictures while I am not in WLAN. With this PR the transferred data is not 7MB per image but ~100Kb and therefore saves really really much traffic, which is expensive in germany ;)
That was the main reason for creating this PR.
Of course, @jancborchardt, it happens that I do need to zoom in, but this is in 1:50 cases...

I remember that we talked about this quite some time ago and you proposed that the image could be auto-downloaded while zooming in at a certain level.
Maybe it is also an idea to have this as an option.
But I do think that it is quite a big improvement for the user to have an image nearly immediately which also does not take much space.

tobiasKaminsky:
@rperezb 
> on the other hand, what happens when you download a folder? by default, if there are images, they are downloaded reduzed? if I wanted to download a folder with the highest resolution, could I?

Downloading is working just the way it is now.
The difference is that a normal click on an image does not download the whole file but just load the resized/screen adapted version of it and stores this resized image in cache.

> Is it needed to have enabled the galleryplus app or 

OC < 8.2: galleryplus
OC >= 8.2: gallery
OC > 9: webdav api (not working now)
The main reason behind this is that a thumbnail API with custom size is needed.

jancborchardt:
@tobiasKaminsky right, then let’s see what we can do to enhance this behavior. Download on zooming in is definitely good, but I feel this might not be enough. I would say the download should happen at already the tiniest amount of zoom, not only at a certain level. Is that ok? And maybe let’s also do it when a picture was viewed longer than x seconds, say 3?

elpraga:
I am not so sure @jancborchardt if automatic downloading intermediately after starting a zoom is a good idea and I certainly disagree that starting a download after viewing the image for 3 secs automatically is a  way to go.

I have been using both beta and and a legacy app for a long time, and I love the behavior of beta concerning the images, as it allows me to comfortably view images and to save valuable space on my mobile device. 

Mobile data consumption and using limited space are the two main concerns I have on my mobile device, and I don't think it would be good to download some files automatically. The original photos are usually much larger then the previews and they would be hidden in a directory tree and it would make cleaning up those unintentionally downloaded original photos to free space unnecesarilly time consuming and hideous.   

Most of the times viewing the previews of photos in full screen is sufficient (even on a tablet), and that is why I do not think that starting the download automatically after viewing the image for 3 ses would be a good idea, and many times I did a small zoom, but not always I wanted to download the original photo. But I do understand that it should be clear for the viewer that he is seeing a preview and not a pore quality original.

What about showing a clickable text: "Do you want to download the original full size photo?" (obviously not as long, I wrote is so long to explain clearly the idea) when zooming starts? Is it technically doable?

What do you think about this?

tobiasKaminsky:
There is something like that: https://github.com/JohnPersano/SuperToasts
toast

This could say "download full image: yes / no", and will be shown after zooming in with e.g. 2x .
After the second it will show "never", just like the translate box in firefox. Same for "yes" -> "ever".
Then a toast can be shown: Behaviour can be changed in settings.
In setting: click on image downloads:
- full image to mobile
- resized image to cache

I strongly disagree on downloading anything automatically.

@jancborchardt I know that you do not like new settings, but is this something worth to put in?

AndyScherzinger:
As an alternative to a "SuperToast" you could now also just make use of the Snackbar, like we now use it for hinting the user for granting the permission, see screenshot: https://github.com/owncloud/android/issues/1343#issuecomment-167681757

tobiasKaminsky:
Snackbar seems much better as it is not another lib!
:+1: for the hint.
Waiting for @jancborchardt @davivel @rperezb for the final go

rperezb:
@jancborchardt is the one here
thx

tobiasKaminsky:
If this PR is coming the default cache size should be increased to 100Mb, I think.

jancborchardt:
Yeah, Snackbar is definitely the way to go here. @AndyScherzinger has to know ;)

tobiasKaminsky:
@jancborchardt then you are fine with the idea of resized image idea? If so I will add the proposed enhancements and hope this will get merged sooon :)

jancborchardt:
Yeah, it definitely provides for a faster viewing experience. We just need to iron out the kinks. :)

Maybe the default previews should also be a tad bigger as it looks strange on HiDPI, and also it should maybe not be square bit same aspect ratio as the original. Or are both those things caused by the fact that it's essentially the file thumbnail?

cc @owncloud/thumbnails @owncloud/previews

tobiasKaminsky:
There are two steps:
In list/grid view load the thumbnail (rectangle very small image)
Clicking on an image shows preview. Until the preview image, that is exactly the size of the display, is loaded the rectangle thumbnail is shown with an spinning wheel on top of it.

So, if you see a square image it is just because the resized, screen adapted image is not fully loaded.

tobiasKaminsky:
@all please move to #1599 as the current one is against develop and not master.

davivel:
This is not active any more, #1599 is the place to go -> closing.

Issue #786, opened 2014-12-07
Long click on a file should allow to mark it as a favorite


tobiasKaminsky:
If you want to mark multiple files as favorite it is time consumption to first open the detail page and then select "keep file uptodate".
@jancborchardt What do you think?

purigarcia:
It is done in #1050 .

davivel:
Merged into develop, will be released with version 1.8

PR #1050, opened 2015-07-09
Favourites context menu

purigarcia:
This PR will fix this issue #786 .
There is a bug. Menu is not refreshed in the gallery neither in the preview of the video when Unfavorite is clicked. @davivel , @masensio could you give me some tips to fix this?
It is pending to change name of variables from keepInSync to favorites. But I's rather wait until the bug is fixed.
Also, we should take into account transifex if we merge that. british english, spanish

### DIFF
--- a/src/com/owncloud/android/files/FileMenuFilter.java
+++ b/src/com/owncloud/android/files/FileMenuFilter.java
@@ -216,14 +216,14 @@ public class FileMenuFilter {
         }
 
         // FAVORITES
-        if (mFile == null || downloading || uploading || mFile.isFolder() || mFile.keepInSync()) {
+        if (mFile == null || downloading || uploading || mFile.isFolder() || mFile.isFavorite()) {
             toHide.add(R.id.action_favorite_file);
         } else {
             toShow.add(R.id.action_favorite_file);
         }
 
         // UNFAVORITES
-        if (mFile == null || downloading || uploading || mFile.isFolder() || !mFile.keepInSync()) {
+        if (mFile == null || downloading || uploading || mFile.isFolder() || !mFile.isFavorite()) {
             toHide.add(R.id.action_unfavorite_file);
         } else {
             toShow.add(R.id.action_unfavorite_file);
--- a/src/com/owncloud/android/files/FileMenuFilter.java
+++ b/src/com/owncloud/android/files/FileMenuFilter.java
@@ -216,14 +216,14 @@ public class FileMenuFilter {
         }
 
         // FAVORITES
-        if (mFile == null || downloading || uploading || mFile.isFolder()) {
+        if (mFile == null || downloading || uploading || mFile.isFolder() || mFile.isFavorite()) {
             toHide.add(R.id.action_favorite_file);
         } else {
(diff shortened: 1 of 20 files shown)


davitol:
:+1:

davivel:
Sorry, but the bug mentioned by @purigarcia is still pening; so this can't be approved by QA.

davivel:
The bug mentioned in the PR is not reproducible. Back to...

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -215,6 +215,20 @@ private void filter(List<Integer> toShow, List <Integer> toHide) {
             toShow.add(R.id.action_send_file);
         }
 
+        // FAVORITES
+        if (mFile == null || downloading || uploading || mFile.isFolder() || mFile.keepInSync()) {
```
davivel: Please, rename the method ```OCFile.keepInSync()``` to ```OCFile.isFavorite()```

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -223,6 +224,24 @@ public void syncFile(OCFile file) {
             mFileActivity.startService(intent);
         }
     }
+
+    public void toggleKeepInSync(OCFile file, boolean isFavorite) {
```
davivel: Please, rename ```toggleKeepInSync(...)``` to ```toggleFavorite```; ```setKeepInSync(...)``` to ```setFavorite(...)``` , etc..

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -762,6 +768,30 @@ private void onSynchronizeFolderOperationFinish(
...
+                Intent i = new Intent(this, ConflictsResolveActivity.class);
+                i.putExtra(ConflictsResolveActivity.EXTRA_FILE, syncedFile);
+                i.putExtra(ConflictsResolveActivity.EXTRA_ACCOUNT, getAccount());
+                startActivity(i);
+
+            }
+
+        } else {
+            if (operation.transferWasRequested()) {
+
```
davivel: What's the target of this blank block?

I would negate the condition in the ```if``` to apply the code in ```else```directly.

src/com/owncloud/android/files/FileMenuFilter.java
tobiasKaminsky: Is this a good idea?
As in /core favorite files are something different. And maybe we want to support them in the future (and have them called "isStarred)? Then we would have different namings for same thing.

davivel:
Reviewed, please take a look to my comments.

src/com/owncloud/android/files/FileMenuFilter.java
davivel: That's something to consider, but right now we have no idea when the integration of the server-side stars will be prioritized (could be soon, could be in two years). On the while I'd prefer the code is as easy to read as possible, a bit of internal coherence will help.

davivel:
:+1:

PR #1052, opened 2015-07-09
change the subject when sharing in saml

purigarcia:
When the authentication is done via shibboleth, the subject of the email when sharing a file/folder by link  was "null shared X with you". 
Searching for a solution, I found that in the class OwnCloudSamlSsoCredentials we have this method 
public String getUsername() {
		// its unknown
		return null;
	}
And it is the reason why is null.
As if seen that it's unknown I decided to change the subject in shibboleth for "x has been shared with you". 
If it is not a good solution, decline the pr

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -341,6 +341,7 @@
 	<string name="with_you_subject_header">with you</string>
     
 	<string name="subject_token">%1$s shared \"%2$s\" with you</string>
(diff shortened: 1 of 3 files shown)


davivel:
@masensio , conflicts solved, ready to review.

masensio:
Done.

masensio:
Merged

PR #1054, opened 2015-07-09
Thumbnail for OC7

zmatsuo:
My owncloud's version is 7.0.6.2.
owncloud android app can't display thumbnails until download.
But web browser can display thumnails.

So, when app can't use the web api to make thumbnails,
app use the preview.png for web browser.

If you hide owncloud version (ex. securty reason?),
app can download and display thubnails.

### DIFF
--- a/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -76,7 +76,8 @@ public class ThumbnailsCacheManager {
 	private static final int THUMBNAIL_GET_TYPE_PREVIEW_PNG = 2;
 	private static final int THUMBNAIL_GET_TYPE_NO_WAY = 3;
 	private static int mThumbnailGetType = THUMBNAIL_GET_TYPE_GUESS;
-
+	private static Object mGuessLock = new Object();
+	
     public static Bitmap mDefaultImg = 
             BitmapFactory.decodeResource(
                     MainApp.getAppContext().getResources(), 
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
I doubt that this will be merged as sooner or later the support for OC 7 will be dismissed.
But of course it is a good idea to support older version...

cmonteroluque:
Jan 2016 is end of support for oC 7. @davivel @masensio what do you think?

zmatsuo:
I use debian stretch(testing).
This distro includes owncloud.
And this owncloud's version is 7.0.7
(https://packages.debian.org/stretch/owncloud)

So, I think, many user are still using OC7.
I believe supporting for OC7 is very cool for many user.

I make PR #1065.
If file is not downloaded and thumbnail is too small in list view,
We can see bigger thumbnail.

tobiasKaminsky:
As owncloud has its own update manager I think it is the best to install it independent of the distro package manager. 
So you can get OC8 just now ;)
(And as the cloud is 24/7 connected to the internet it should always be the latest stable version available due to security improvements, I think)

The PR #1065 is a very good idea :+1:

zmatsuo:
My owncloud server is updated 8,so I don't need this PR.

I could not understand update manager.
Because Debian's owncloud updater app is removed.
(I found http://changelog.debian.net/owncloud 5.0.3+dfsg-1)

I uninstall distoro's OC7 and install OC8 from owncloud.org.
So I not need this PR.

Debian's owncloud version is still 7.0.10.
(https://packages.debian.org/stretch/owncloud)
I guess meny users are still using OC7...

Thank you.

Issue #1041, opened 2015-06-29
Sort in uploader


masensio:
Add the feature to sort in the uploader, in order to avoid a long scroll  when selecting the most recent files for uploading.

Related with https://github.com/owncloud/android/issues/1040

tobiasKaminsky:
You mean when adding files from within the app and selecting "files"?
(UploadFilesActivity class)

masensio:
Yes. Maybe the title is some confused.
I mean to add sort feature when you are selecting files to upload, inside the app.

purigarcia:
Hi @tobiasKaminsky, are you implementing this?

tobiasKaminsky:
I will do it tomorrow.

tobiasKaminsky:
Done.
I will upload it tomorrow.

sd1998:
@davigonz I would like to work on this.

michaelstingl:
@sd1998 sure, you're the sorting-expert now 👍

davigonz:
Already implemented via https://github.com/owncloud/android/pull/2305

PR #1058, opened 2015-07-11
add sort to UploadFileActiviy

tobiasKaminsky:
As discussed in #1041


davigonz:
No longer needed, already implemented via #2305

Issue #915, opened 2015-03-06
Conflict dialog wording confusing


PVince81:
#### Steps
1. Upload file "test.txt" directly to server
2. Open Android client
3. Pick "Keep in sync" for that file and wait for download
4. Go offline (airplane mode)
5. In the web UI, change "test.txt" and save
6 On mobile (still offline), edit "test.txt" too and save
7. Go back online
8. Long press on "test.txt" and pick "Refresh file"

At this point you get a dialog that says:
"Remote file "test.txt" is not synchronized with local file. Continuing will replace content of file on server" with three buttons: "Don't upload", "Keep both" and "Overwrite".

I want to overwrite the local file, so should I pick "Don't upload" or "Overwrite" ? Does "Overwrite" mean "Overwrite server file" or "Overwrite local file" ?
"Don't upload" could also mean "Do not overwrite server version but overwrite local version".

This is very confusing and even scary as I'm afraid to lose the wrong changes.

Also the dialog mentions "continuing" but there is no "continue" button.

How about changing the labels to "Use local version", " Use server version" and "Keep both" or something more explicit ?

To be discussed with @owncloud/designers

PVince81:
I just tried "Don't upload" and it seems to just cancel ?
The local content is still there, it was not overwritten by the server content

I'm still too scares to tap "Overwrite" because it is not clear whether it will overwrite the local content or the remote content... Also this means there is a missing option. If one overwrites the local file, where is the option to overwrite the remote file ?

PVince81:
Any update on this ? I recently deleted the wrong file because I misunderstood the options.

A short term fix would just to fix the wording of the labels.
@jancborchardt what do you think ?

jancborchardt:
Good call @PVince81. The three buttons should be:
* Use local version
* Use server version
* Keep both

@rperezb @davivel

tobiasKaminsky:
+1
This should be very easy done via transifex?

jancborchardt:
No, the original text needs to be changed here in the source code, right?

tobiasKaminsky:
Ah yes.
You are right.
Shall I do this? Or do we wait for approval by @rperezb and @davivel ?

jancborchardt:
@tobiasKaminsky I’d say do it, with the text I suggested above and then we can discuss in the PR. Pretty sure it’s better than it is now.

masensio:
Fixed in develop.

PR #1068, opened 2015-07-17
Conflict dialog: new wording

tobiasKaminsky:
Acording to #915


davivel:
Only the file res/values/strings.xml should be changed. Changing the other translation files will make the automatic fetch&merge from Transifex to fail.

jancborchardt:
@tobiasKaminsky sorry for maybe wrong instructions – of course only the one source file should be changed since the translations are handled via Transifex. Can you change it as @davivel commented? :)

tobiasKaminsky:
No problem :)
I was just to excited as this was such an easy PR, that I missed the transifex system.
I hope, everything is right now?

jancborchardt:
Not sure why `res/values-tzl/strings.xml` is deleted as it seems unrelated. But the change looks good otherwise.

@davivel @masensio please review! :)

tobiasKaminsky:
This file prevents all of my current branches from building.
So I have to delete it every time...
And I must have forgotten to exclude it.

jancborchardt:
@tobiasKaminsky ah ok, no problem. Just amend and force push it. :)

davivel:
Mmm... the code seems OK, but the button 'Use local version' is not really doing that, it's just cancelling. 

I'd rather that 'Use local version' removes the local copy the of the file and requests a download from the server. That requires modifying ConflictsResolveActivity#conflictDecisionMade(). I would add a fourth value to Decision, such es Decision.LOCAL, that should be sent from the button in ConflictsResolveDialog instead of Decision.CANCEL (but DECISION.CANCEL  is still used by the cancel() method of the dialog).

@tobiasKaminsky , could you go for this extension?

jancborchardt:
> I'd rather that 'Use local version' removes the local copy the of the file and requests a download from the server.

That was a typo, right? You mean »Use server version« would delete the local copy?

davivel:
Yes, sorry. Thanks, @jancborchardt

tobiasKaminsky:
@davivel I will implement it.

tobiasKaminsky:
conflict file

@jancborchardt as we are already discussing the text...
"Continuing will replace content of file on server" is not right if one selects "use local version".
I think there must be a question to which the "answers" can fit, such as: "What to do?"

davivel:
True, all the message should be updated.

Something like "Remote and local versions of /0.txt cannot be automatically synchronized. What do you want to do?"

cc @jancborchardt for better wording

tobiasKaminsky:
@davivel I hope that I have used the correct downloading method.
Please review :)

davivel:
Perfect, pal.

:+1:

cc @purigarcia , ready for QA.

purigarcia:
issue validated, but I would not merge it until the message is update.
cc @jancborchardt @tobiasKaminsky @davivel @masensio

davivel:
Please, @purigarcia , retest. The bug you found before was not due to this PR, but it's making the operation here useless. I pushed the fix.

cc @masensio for an extra code review.

davivel:
And the message is updated now too. Let's try to merge in this point.

cc @purigarcia , @masensio

pierrejochem:
Wich? It should be Which or not?

tobiasKaminsky:
In strings.xml it is correct: "which"

pierrejochem:
okay :)

purigarcia:
validates! qa approved

jancborchardt:
:+1:

masensio:
Reviewed

PR #919, opened 2015-03-11
Reliable uploads

LukeOwncloud:
Even though the branch is being ignored and is thus stale already, for completeness sake: Here a PR for my proposal for reliable uploads: An upload manager which handles both instant and normal uploads and retries failed uploads.

Fixes: https://github.com/owncloud/android/issues/340

Discussion here:
https://github.com/owncloud/android/issues/762

Preview screenshots here:
https://github.com/owncloud/android/issues/765

Upload manager could be used to easily solve a number of issues and to provide requested features, including upload on wifi/charging only, rename on upload, instant upload with delay, extend instant upload to other folders

Issue #819, opened 2014-12-24
Avoid duplicating files


tobiasKaminsky:
Based on #346, this summarizes all ideas to get a clean idea how to implement this long awaiting feature.
I will update the description when there is a new idea or something is unclear, so that this is up to date. 
First the way to do is shown, then all possible problems or scenarios are described...

**Desired behaviour:**
When uploading a file through owncloud app the file get uploaded to the owncloud server and remains on the place on the device. In the database of the app an entry is created which stores the local path.
When opening/downloading a file the app first tries to open and verify the file in the local path that is stored in the database. If the file is there and up-to-date it will be shown. If it is not there, the file will be downloaded and be stored in the local owncloud folder structure.

**TODO**
The user needs to be informed that the file is *shared* with owncloud. This means that the original file keeps untouched, even if the remote file changes.

**File gets deleted on server:**
The file also get deleted from the database and from the local owncloud folder.

**File gets deleted only on device (via ownCloud App):**
This option deletes the file from local owncloud folder.

**File gets changed on server:**
When opening the file with owncloud app it is determined if the file is up to date. Otherwise it will be downloaded and be added to local owncloud folder.

**File gets changed on device:**
As now the file needs to be selected in owncloud app and choose "refresh file". This will upload the file change to the server --> nothing special here.

**File is "kept-to-sync" on one account:**
File is kept in local owncloud path on device, but gets updated as soon as the server file has been changed.

**File is "kept-to-sync" on multiple accounts:**
If there is a change on the server the local owncloud file will be updated. But none of the other files will be modified. Example: There are three distinct files to work with: $account1/test.txt, $account2/test.txt, and /data/test.txt. So if $account1/test.txt is updated on the server, then the update should be downloaded and displayed if the user opens $acount1/test.txt. However, if the user locally modifies /data/test.txt, neither the account1 nor account2 copy should be updated.

I hope I have summarized this correctly. If not, or something is unclear, please feel free to comment.
Please also comment if this should be done this way.

strugee:
> @jancborchardt You are right. But if the files are not linked then you will end up with different versions of the same file like we have it now.

This (IMHO, obviously) is the best behavior. Remember, the way that the user will generally be putting things into ownCloud on their Android device is the "Share" action. "Share" implies creating a copy. This is what basically all other apps do: if you share e.g. a picture over SMS, deleting the message doesn't delete the original image. If you share a picture on a social network, deleting the original photo won't delete the post on the social network. And vice versa.

[...]

PR #1069, opened 2015-07-17
Uploader: copy file, move to OC folder, delete from device

tobiasKaminsky:
while uploading a file from another app the user can choose:
- copy file to server and copy to local folder on device (current behaviour)
- copy file to server and move file to local folder on device
- copy file to server and delete file on local device


jospoortvliet:
How does this relate to https://github.com/owncloud/android/pull/919 if I might ask?

tobiasKaminsky:
Maybe you mean a different issue? Maybe #819?
It is not related. It should improve the current situation until it get really (and better) solved.

jospoortvliet:
I did mean #919, being a big rewrite of the uploader - was just curious...

jospoortvliet:
Can you give me a reasonable use case for option 1 and 2?

I see two reasons to upload to ownCloud:

* you want to save storage space and keep your data on your cloud. For this, you want option 3 (this is the common case of 'upload pictures automatically to ownCloud' for most ppl)
* you want to keep your file synchronized to your cloud to have it also on other devices or backed up safely (there's no option for that, I see)

For the first case, yes, upload and delete.
For the 2nd case, you need the file to be kept in sync. 

I would suggest to make option 1 default and offer the option to keep files synced locally for those few people who have a phone with 128gb of storage ;-)

tobiasKaminsky:
My option 1 is only for backward compatibility as it is the current behaviour.
option 2 is for those people who do not want to get files duplicated, but still want them on their phone.
For instance: I move my camera photos with this option to my owncloud. So all files are being moved from DCIM to sdcard/owncloud (and being uploaded). But still I can access them via gallery.
If the phone is running out of space I can delete the folder via our app.

Real synchronizing (2 way) is not covered by this PR and was not being planned.
This is something that must be discussed and really carefully planned in a new issue.

tobiasKaminsky:
This is part of #1168, so closing this one.

PR #1072, opened 2015-07-22
Automation UI test delete folder with downloaded files inside

purigarcia:
Some refactor has been done also


purigarcia:
Delete it because it is old-fashioned

PR #1078, opened 2015-07-28
Check if ETag for remote folder is null, i.e. it was not received from the server

gstraube:
When the ETag is missing for the remote folder, the app crashes with a null pointer exception. In my case, the absence of ETags was caused by having server encryption enabled.


davivel:
Encryption disables ETags? I need assistance here. @danimo , @dragotin , @PVince81 , @DeepDiver1975 , is this expected to happen?

PVince81:
Encryption doesn't disable etags. If it did, my own personal server that runs 8.1 with encryption wouldn't work with the Android app, but it does.

It must be some other bug. Please raise a ticket in https://github.com/owncloud/core with details about your server config. Please use the issue template: https://raw.githubusercontent.com/owncloud/core/master/issue_template.md

gstraube:
It's probably not a problem in newer versions (I installed the server via Apt on Debian and ended up with a rather outdated installation) but I'll see if it comes up in version 8.1. 

I just thought it might be useful to check for a null value instead of letting the app crash, especially since this has been a problem in the past (see e.g. https://github.com/owncloud/client/issues/1097, https://github.com/owncloud/client/issues/1279 and https://github.com/owncloud/core/issues/6396 and the discussion on https://forum.owncloud.org/viewtopic.php?f=14&t=16800&start=20).

davivel:
Sure, just wanted to be sure that nothing else should be checked. ETags are essentials for the synchronization process.

The code seems :+1:. After logging the error, the method will return reporting that there are changes in the server, even if there are not, what is better than the opposite. 

@gstraube , this is your first contribution to ownCloud, right? We need you send a signed copy of the contribution agreement in https://owncloud.org/contribute/agreement/ to @karlitschek before accepting your PR.

Thank you very much.

gstraube:
Great, thank you! I emailed the signed contributor agreement to @karlitschek.

davivel:
Thanks, @gstraube .

:+1:

Call for QA, @purigarcia

purigarcia:
@gstraube could you update the branch with our develop before testing? Thanks in advance

gstraube:
Done!

purigarcia:
@gstraube , thanks
Approved by qa
@masensio @davivel ready to merge

masensio:
Reviewing

masensio:
Code Review done.
Thanks @gstraube for your contribution.

PR in develop, master updated with develop code

gstraube:
Awesome, thank you all!

PR #1079, opened 2015-07-29
Automation Ui test check there are not false positive

purigarcia:
Included delete folder with contents inside.


purigarcia:
this is old-fashion ;)

Issue #1080, opened 2015-07-30
Grid view not shown if picasa.ini is present


etiess:
I use Picasa on my Windows desktop: it creates a special file in each folder called ".picasa.ini". This file is used to store useful information (people in the pictures, modification, ...). So, IMO, it's useful that ownCloud syncs it.

Unfortunately, this prevents the Android app to consider the folder "full of pictures, and only pictures".

As a consequence, the grid view is not shown. Instead of that, we have a file list, with thumbnails and the name of the folders.

Would it be possible, for example, to exclude files starting with a point, when the app wants to know if there are only pictures in the folder?

Config : 
Ubuntu 15.04   3.19.0-25-generic
Server version: Apache/2.4.10 (Ubuntu)
PHP 5.6.4-4ubuntu6.2 (cli) (built: Jul  2 2015 15:29:28)
mysql  Ver 14.14 Distrib 5.6.25, for debian-linux-gnu (x86_64) using  EditLine wrapper
Android ownCloud app: 1.7.2
Android version : 4.2.1
ownCloud server: 8.1.0

tobiasKaminsky:
On the first implementation the threshold was 80% pictures to switch to grid view.
Now it is 100% pictures.
But afair there should be a setting sooner or later.

But (again) you are right, that files starting with a dot should be ignored for computing the threshold.
@davivel Is this ok for you? Then I would implement this. Should be just one or two lines ;)

The customized threshold in the preferences we can discuss in a new issue.

etiess:
Great! I agree with your 2 complementary solutions:
1. Best "straight to the point" solution: ignore files starting with a dot for computing the threshold
2. Wider solution with manually adjustable threshold: https://github.com/owncloud/android/issues/1081 

In order to complete solution 1, would it be possible to look at the files attributes? For example, I would propose to ignore the files with an attribute "hidden" in the threshold calculation.

Finally, I would make another proposal (which may have been already discussed: in this case, sorry!). What about putting a button to manually switch from list view to grid view? This button would be available on the web interface and in the android app. It would work like this:
- The user opens a folder
- Threshold calculation decides to show the folder in list or grid view
- The user can toggle the button to change the view

I think I would prefer this solution to the adjustable threshold (it is complementary though). As an improvement of this feature, ownCloud could remember the user's choice.

Do you want me to open a new issue to discuss this?

masensio:
Hi @etiess,

Thanks for your feedback.

[...]

Issue #894, opened 2015-02-22
1.7.0: Grid view for pics should be optional / additional!


deMattin:
The new grid view in 1.7.0 might be a great feature if you are able to select it like gallery in the owncloud web-app.
But it should not be the forced AND ONLY view in a file-client like the android app.

In my opinion the new grid view of 1.7.0. is very problematic and mostly a step backwards because you are not able to see the name of the file in this or another view.
Further more this view is forced automatic active in all folders with only pictures in it and mostly only shows "placeholders" for the pics (on a Owncloud 7.x Server) and no additional filenames.

Is there any chance to switch off this view generally and get back the old view with mini thumbnail and the name of the file?
With the mini-thumbnails it is not essential, if there exists a thumbnail or if there's only a placehoder for it.

In the google store you see some (many) other comments claiming the "broken" view.
But even if the previews will work, a gallery style view should be an ADDITIONAL option and not the only possibility to view this type of files in a list and to decide, if you want to download the whole file on your android.

rperezb:
thanks for your comment,

@jancborchardt what do you think?
FYI @MTRichards

rensgr:
agree fully.  this stupid interface looks like this: 
![stupid-view](https://cloud.githubusercontent.com/assets/4196449/6326621/ed1e78fe-bb53-11e4-88cd-6816f0e0e578.jpg)

LukasReschke:
@jancborchardt

rensgr:
while it should have looked like this:::  (yes, that is your competition......)
![proper-view](https://cloud.githubusercontent.com/assets/4196449/6326645/20d89b2a-bb54-11e4-8846-e3a0a3df325e.jpg)

jancborchardt:
The first question is why the placeholders are shown and not the thumbnails of the actual images. @rperezb @davivel?

Secondly yes – after the feedback we received through the Play Store, via email and here, it is indeed better to have a view setting in the options dropdown top right. That should remember its setting per folder. @MTRichards @rperezb @davivel ok?

MTRichards:
yep.

zebtom:
I would like to add that the app became unstable on my device (samsung galaxy S4). In a typical photo gallery (200 pics, 2-3Mo each), scrolling up/down lags and the app crashes when scrolling too fast in a gallery view.

nickvergessen:
(how is an "empty"/unusable file list an enhancement?!)

LukasReschke:
[...]

PR #1082, opened 2015-08-02
ignore hidden files also if they are images for counting

tobiasKaminsky:
As discussed in #1080


davivel:
:+1: 

@jancborchardt , what do you think about this tweak?

Anyway, this is a patch. The key change for me is #894

davivel:
Needs QA; cc @purigarcia

jancborchardt:
Yes, hidden files should be ignored always.

etiess:
Great @tobiasKaminsky ! I looked at your PR: I understand that you consider a file "hidden" if it starts with a dot. This solves my issue with .picasa.ini

But would it be possible to ignore every hidden files? Even the ones marked as hidden by the OS (Windows or others) and that don't start with a dot? Is this file property is accessible by ownCloud?

tobiasKaminsky:
@etiess AFAIR the support for Windows Server is dropped in OC8.1.
Linux does not have hidden files like Windows. The only way is to use a ".". But this is more a convention as a real file property like in Windows.
So AFAIK this property is not in OC.

purigarcia:
well done, @tobiasKaminsky 
@masensio , @davivel ready for merge

masensio:
Reviewing

masensio:
Code Review done.
Thanks @tobiasKaminsky 
PR into develop and master updated too.

Issue #1081, opened 2015-07-30
Change list vs grid view


tobiasKaminsky:
> I don't know if it's stated in #894, but the grid setting, when enabled, should be propagated down the tree of files to all descendent folders.
If set in root fold, all the children will be shown in grid mode too.
Then, if list view is selected for a subfolder, it will applied (and remembered) to that subfolder and all its children.

>Think in a 3-valued property: "list", "grid" and "inherited"; the default value for every folder is "inherited", with the exception of the root folder, where "inherited" makes no sense and is promoted to "list". The user only can choose in the UI between "list" or "grid", "inherited" is for internal use.

OLD:
The threshold was 80%, now it is 100%.
I think this should be adjustable as each user may have a different wish.

@jancborchardt What do you think? A slider with value?
https://www.google.com/design/spec/components/sliders.html#sliders-continuous-slider
--> With an editable numeric value

davivel:
I wouldn't go this way. Sound like lots of noise about something that should be simple.

I'd just go for #894

tobiasKaminsky:
I think both ideas are additional, not complementary.
Imagine that you always want to have grid view. With #894 you would have to switch the view for every single folder to grid view.
With the threshold you can manage this by setting it to 0%.
(But maybe it is just me that is using our app mainly as an image browser due to my very small internal disk space)

davivel:
I don't know if it's stated in #894, but the grid setting, when enabled, should be propagated down the tree of files to all descendent folders.

If set in root fold, all the children will be shown in grid mode too.

if then, list view is selected for a subfolder, it will applied (and remembered) to that subfolder and all its children.

jancborchardt:
@davivel yup!

tobiasKaminsky:
This is a very good idea and would fit perfectly for my use case.
How can this be shown and explained to the user?

New folders would have automatically the setting of the parent folder?

jancborchardt:
It doesn’t need to be shown/explained I’d say, at least not in the first iteration.

davivel:
> It doesn’t need to be shown/explained I’d say, at least not in the first iteration.

Agreed.

Think in a 3-valued property: "list", "grid" and "inherited"; the default value for every folder is "inherited", with the exception of the root folder, where "inherited" makes no sense and is promoted to "list". The user only can choose in the UI between "list" or "grid", "inherited" is for internal use.

jancborchardt:
Good call!

tobiasKaminsky:
[...]

PR #1099, opened 2015-08-16
Switch list vs grid

tobiasKaminsky:
As discussed in #1081

I have used SharedPreferences for this and not a sqlite db.

Should we have a non blockable information on first change like this "the view settings is remembered for this and all subfolder, if they are not set otherwise".

@davivel @jancborchardt 

(minor side note: I had accidental pushed the commits direct to develop, but have reverted this)
Sorry for this...


tobiasKaminsky:
Another point is: do we have to care about deleting entries in the SharedPreferences? And if so, how do we do it?

jancborchardt:
Please review also @AndyScherzinger @Stoyicker @Kernald :)

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -373,4 +376,65 @@ public static Point getScreenSize(Activity caller) {
...
+        }
+    }
+
+    public static void setViewMode(OCFile file, boolean setGrid){
+        SharedPreferences setting = MainApp.getAppContext().getSharedPreferences(
+                "viewMode", Context.MODE_PRIVATE);
+
+        SharedPreferences.Editor editor = setting.edit();
+        editor.putBoolean(file.getRemoteId(), setGrid);
+        editor.commit();
```
Kernald: Synchronicity is not important here, you should call `editor.apply()` instead of `editor.commit()`

Kernald:
Removing an entry in `SharedPreferences` is really simple: http://developer.android.com/reference/android/content/SharedPreferences.Editor.html#remove(java.lang.String)

We should probably remove corresponding entries when a folder is removed. The back-end file is loaded once when it's first needed then kept in memory, if the user switch the view mode in a lot of folders, the initial load could take long. I never experienced this kind of issue, and don't really know if the impact is really important.

Another idea to keep this file the smallest possible: instead of storing both list and grid state, remove the entry when the user switches back to the default (list).

tobiasKaminsky:
With this implemention there is only a switch between grid and list view possible.
But it is not possible to manually delete the setting for a specific folder.
So maybe it should be something like that:
- grid
- list
- inherited
And the third option would delete the specific setting.

AndyScherzinger:
Looks good to me, only one thing: I would recommed/prefer to switch the menu icon so you wouldn't need the text (done like this in Google's apps as well imho) to distinguish the action but could (also) show it visually :smiley: 

so I would recommend using:
For show list view
https://www.google.com/design/icons/#ic_view_list
and for grid
https://www.google.com/design/icons/#ic_view_module

tobiasKaminsky:
Do you mean, that the action is always visible? (Like currently "upload file" and "create folder")?
For me there is no icon shown in the menu next to the text.

But these are great icons.

switch

AndyScherzinger:
Yes, I meant always shown but that probably only makes sense after my fab speed dial implementation has been tested and merged and thus the space in the action bar has been freed up. 
Am 19.08.2015 19:28 schrieb tobiasKaminsky <notifications@github.com>:Do you mean, that the action is always visible? (Like currently "upload file" and "create folder")?
For me there is no icon shown in the menu next to the text.

But these are great icons.

tobiasKaminsky:
@AndyScherzinger 
I have put those files in the corresponding folders: 
drawable
drawable-mdpi
drawable-hdpi
drawable-xdpi
But still the xml code says: no file:
 < item
        android:id="@+id/action_switch_view"
        android:icon="@android:drawable/ic_view_list"
        android:orderInCategory="2"
        app:showAsAction="always"
        android:title="@string/action_switch_grid_view" />

Do you have any idea?

Kernald:
@tobiasKaminsky : it should be `@drawable/ic_view_list` instead of `@android:drawable/ic_view_list` if the file is in the application's folders.

By the way, the `drawable` folder without density qualification should not contain PNG assets, only scale-independent ones (XML variants of PNG assets, shapes...).

Kernald:
And as a side note: in addition to xhdpi, we probably should provide xxhdpi assets, most of high end devices are in this density bucket now.

tobiasKaminsky:
drawer2
drawer1

@Kernald Thanks for your hint.
One more question: https://www.google.com/design/icons/#ic_view_list has the icon in 18 24 36 48dp.
How do I know which size has to be in which drawable-??? folder?

Kernald:
@tobiasKaminsky : if you download, for example the 24dp version, you'll see that you get a zip with the corresponding drawable-bucket folders. The question is: where do you plan to put your icon? For an `ActionBar`/`Toolbar`, as mentioned [here](https://www.google.com/design/spec/layout/metrics-keylines.html#metrics-keylines-touch-target-size), you should use the 24dp one.

jancborchardt:
@tobiasKaminsky I’m not sure we need to put the switcher that present.

tobiasKaminsky:
You are the design/UI guy  ;)
So it is up to you.

jancborchardt:
@tobiasKaminsky hehe, ok :) I would put it in the dot-dot-dot menu for now. :)

tobiasKaminsky:
As you wish :) :)

AndyScherzinger:
Sounds good to me too. I would recommend putting it on the bar out of the overflow menu as soon as  (or in case) we finish/merge/agree-on the FAB implementation since the upload/create actions would then be removed from the action bar.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1200,6 +1246,15 @@ public void onBrowsedDownTo(OCFile directory) {
         cleanSecondFragment();
         // Sync Folder
         startSyncFolderOperation(directory, false);
+
+        MenuItem menuItem = mOptionsMenu.findItem(R.id.action_switch_view);
+
+        changeGridIcon();
+        if (DisplayUtils.isGridView(directory, getStorageManager())){
```
AndyScherzinger: same code fragment as line 325ff

AndyScherzinger:
Code review done! :+1:

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Right!
I have moved this to a separate function.
Thanks!

davivel:
Reviewing this now.

davivel:
> I have used SharedPreferences for this and not a sqlite db.

Why?

AndyScherzinger:
I guess both ways would work (preferences Vs. DB) but he probably considered this being a setting/preference and thus used SharedPreferences to implement this :)

davivel:
> - grid
> - list
> - inherited

That third option seems created just for techies. I can imagine my mother browsing their files in the OC app and switching between grid and list without no fear, but sure she would call me by phone to wonder what does that of 'inherited' mean. And I'm not sure I would get to explain it to her.

@jancborchardt , what do you think of this?

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -321,6 +322,8 @@ private void initFragmentsWithFile() {
                     startTextPreview(file);
             }
 
+            switchLayout(getFile());
```
davivel: There is a local variable ```file``` set to ```getFile()``` 13 lines above; I would use it instead of calling ```getFile()``` again.

tobiasKaminsky:
The third option is not valid anymore.
There is only "switch to grid view" or "switch to list view"

tobiasKaminsky:
SharedPreferences was a lot easier to implement as I know how to do it, whereas using a sqlite db is new to me.
If you think that it is better to use a db, I/we can try to change it.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: As it is only a quick change, maybe you can do it, if it otherwise will prevent this PR from QA?
(I can do it on sunday or monday afternoon)

AndyScherzinger:
I can also sidewing an sqlite implementation in case it will be decided to go for a db based solution

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1200,6 +1250,11 @@ public void onBrowsedDownTo(OCFile directory) {
         cleanSecondFragment();
         // Sync Folder
         startSyncFolderOperation(directory, false);
+
+        MenuItem menuItem = mOptionsMenu.findItem(R.id.action_switch_view);
```
davivel: Variable not used; please, remove this line.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -497,6 +509,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
     public boolean onCreateOptionsMenu(Menu menu) {
         MenuInflater inflater = getMenuInflater();
         inflater.inflate(R.menu.main_menu, menu);
+        mOptionsMenu = menu;
+
+        MenuItem menuItem = mOptionsMenu.findItem(R.id.action_switch_view);
```
davivel: Variable not used; please, remove this line.

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
```diff
@@ -127,6 +127,13 @@ protected void switchToListView() {
             mCurrentListView = mListView;
         }
     }
+
+    public boolean isGridView(){
```
davivel: Very minor. Why the change from ```isGridMode()``` to ```isGridView()```?

src/com/owncloud/android/ui/fragment/FileDetailFragment.java
```diff
@@ -217,6 +217,24 @@ public void onPrepareOptionsMenu(Menu menu) {
             item.setVisible(false);
             item.setEnabled(false);
         }
+
+        item = menu.findItem(R.id.action_switch_view);
```
davivel: This shouldn't be this way. In tablets in landscape mode, when a file is shown in the right pane, the options for the current folder still make sense. Please, undo this.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -290,6 +290,18 @@ public void onPrepareOptionsMenu(Menu menu) {
             item.setVisible(false);
             item.setEnabled(false);
         }
+
+        item = menu.findItem(R.id.action_switch_view);
```
davivel: Same comment here as in ```FileDetailFragment```

src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -74,6 +75,8 @@ public OCFile createFromParcel(Parcel source) {
 
     private boolean mIsDownloading;
 
+    private boolean mShowGridView;
```
davivel: Never used, right? Please, remove it.

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -245,6 +258,67 @@ public static Point getScreenSize(Activity caller) {
...
+            return !setting.contains(parentDir.getRemoteId());
+        }
+    }
+
+    public static void setViewMode(OCFile file, boolean setGrid){
+        SharedPreferences setting = MainApp.getAppContext().getSharedPreferences(
+                "viewMode", Context.MODE_PRIVATE);
+
+        SharedPreferences.Editor editor = setting.edit();
+        editor.putBoolean(file.getRemoteId(), setGrid);
```
davivel: I'd rather use ```file.getId()``` as the index. In theory ```file.getRemoteId()``` has information enough to not collide even through different server, but ```file.getId()``` does not depend on theory, it's unique for sure.

davivel:
> The third option is not valid anymore.

I'm sorry, I saw it later in the code.

> If you think that it is better to use a db, I/we can try to change it.

> I can also sidewing an sqlite implementation in case it will be decided to go for a db based solution

Thank you both. Maybe it isn't worth it. My main concern if how nice the solution will scale if a user sets different values for lots and lots of folders, but probably this is not a realistic case. There are changes more important to do in the persistence layer, so let's keep it this way. If tests reveal performance problems, will look into it again.

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -245,6 +258,67 @@ public static Point getScreenSize(Activity caller) {
     }
 
     /**
+     * Determines if user set folder to grid or list view. If folder is not set itself,
+     * it finds a parent that is set (at least root is set).
+     * @param file
+     * @param storageManager
+     * @return
+     */
+    public static boolean isGridView(OCFile file, FileDataStorageManager storageManager){
```
davivel: I wouldn't do this in a static method. An ```Activity``` or a ```Fragment``` would be a better place.

Even further, did you try to implement the code added to ```FileDisplayActivity``` in ```OCFileListFragment``` instead? That would result in a better structure IMHO. The feature is contained in the scope of ```OCFileListFragment``` responsibility. 

The addition of the menu item in the ```ActionBar``` should be possible also from ```OCFileListFragment```.

Would you give it a try?

davivel: Very minor: ```isGridView``` could be confused with the meaning of ```isGridView``` (or even ```isGridMode```) in ```ExtendedListFragment```. 

What about ```isGridPrefered```, or something similar?

davivel:
Review finished. @tobiasKaminsky , have a look to my comments, please.

Thanks.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Done

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Done

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Done for the menuItem.
As I recall correctly it is important to have the mOptionMenu updated.

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
tobiasKaminsky: Do not know anymore. Maybe just because it was implemented on different days ;)

tobiasKaminsky: I will change it to gridView as it is a view/layout, ok?

src/com/owncloud/android/ui/fragment/FileDetailFragment.java
tobiasKaminsky: Done

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
tobiasKaminsky: Done

src/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: Done

src/com/owncloud/android/utils/DisplayUtils.java
tobiasKaminsky: The "problem" is that .putBoolean expects a string, so I have to convert it.
But...done ;)

tobiasKaminsky:
I have also updated it to the latest master.

tobiasKaminsky:
@davivel 
Everything is now included and up to date. :)

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
davivel: :+1:

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
davivel: Yes, handling of mOptionsMenu is right.

src/com/owncloud/android/utils/DisplayUtils.java
davivel: Good catch :)

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -567,12 +568,38 @@ public void onClick(DialogInterface dialog, int which) {
                 builder.create().show();
                 break;
             }
+            case R.id.action_switch_view:{
+                if (isGridView()){
+                    item.setTitle(getApplicationContext().getString(R.string.action_switch_grid_view));
```
davivel: ```getAppicationContext()``` is not really needed. ```FileDisplayActivity``` is a ```Context```, so ```getString``` is available in ```this```.

Or am I missing something?

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -567,12 +568,38 @@ public void onClick(DialogInterface dialog, int which) {
...
+                return true;
+            }
         default:
             retval = super.onOptionsItemSelected(item);
         }
         return retval;
     }
 
+    public void setViewMode(OCFile file, boolean setGrid){
+        SharedPreferences setting = MainApp.getAppContext().getSharedPreferences(
```
davivel: Same here, ```getSharedPreferences``` should be available directly in the ```FileDisplayActivity```

davivel:
Thanks a lot, @tobiasKaminsky .

Just a couple of extra comments. Should be the last, from my side. Almost there.

davivel:
@tobiasKaminsky , if you don't mind, I'll make those final changes by myself; this is now between the top priorities and QA is willing to start with it :) .

davivel:
Scheduled for next release

tobiasKaminsky:
Please feel free to do it.
I can do it in ~5 hours.

Great to see this progress :)

davivel:
This is :+1: for me now.

@tobiasKaminsky , review my last changes.

@jesmrec , if @tobiasKaminsky is OK with the last changes, this is ready to start QA.

tobiasKaminsky:
Just a quick look, but it looks good to me :+1:
So, QA can start :)

davivel:
Rebasing on master...

davivel:
This was trickier to rebase than expected; @jesmrec, part of the tests should be repeated to check that nothing was wrong

jesmrec:
The tests were passed again succesfully.

rperezb:
@carlaschroder new feature for the Android app, being possible to decide the list mode between grid and list one

PR #1105, opened 2015-08-21
switch from legacy infrastructure to container

tobiasKaminsky:
Restrictions, but they should not affect us:
- Using sudo isn’t possible (right now)
- Databases don’t run off a memory disk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: android
 android:
   components:


tobiasKaminsky:
Benefits:
- Builds start in seconds
- Faster builds
-  More available resources

AndyScherzinger:
travis output look good to me :+1:

tobiasKaminsky:
Yeah and it seems to be faster ~2min to 1min30s :)

davivel:
:+1:

Thank you very much.

PR #1107, opened 2015-08-23
Add `android:filterTouchesWhenObscured`

LukasReschke:
To prevent tapjacking like attacks we should explicitly enable this attribute. Fore more details see https://blog.lookout.com/look-10-007-tapjacking/ and https://developer.android.com/intl/en/reference/android/view/View.html#Security

Note that this does not affect all Android devices as for example Samsung started with Ice Cream Sandwich to mitigate this issue on OS level. This PR also does not offer a complete protection but already a pretty suitable one. (for example the ActionBar navigation in fragments still can be invoked, but account settings are at least protected now)

To test this I can recommend the https://github.com/mwrlabs/tapjacking-poc application.

Fixes https://github.com/owncloud/security-tracker/issues/165

### DIFF
--- a/res/layout-land/account_setup.xml
+++ b/res/layout-land/account_setup.xml
@@ -17,11 +17,13 @@
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-                android:layout_width="match_parent"
-                android:layout_height="match_parent"
-                android:background="@color/login_background_color"
-                android:padding="8dip"
+<RelativeLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/login_background_color"
+    android:padding="8dip"
+    android:filterTouchesWhenObscured="true"
     >
 
     <LinearLayout
--- a/res/layout-v11/activity_row.xml
+++ b/res/layout-v11/activity_row.xml
@@ -2,7 +2,7 @@
 <!--
   ownCloud Android client application
 
-  Copyright (C) 2016 ownCloud GmbH.
+  Copyright (C) 2015 ownCloud Inc.
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License version 2,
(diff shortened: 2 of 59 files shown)


LukasReschke:
@MTRichards @rperezb THX

rperezb:
thanks @LukasReschke

davivel:
The code seems fine :+1:

But, note for @purigarcia: we should test if this flags interferes with the operation of the automatic regression tests in progress, testing with an emulator without UI.

LukasReschke:
Can anybody with an real Android phone test if this interferes with https://play.google.com/store/apps/details?id=com.touchtype.swiftkey&hl=en or https://play.google.com/store/apps/details?id=com.gau.go.launcherex&hl=en? – I did hear some rumor.

Much appreciated!

tobiasKaminsky:
@LukasReschke I have tried it with swiftKey. 
I could add a new folder while typing with swiftKey.
But I am on 5.1 ;)

LukasReschke:
@tobiasKaminsky Are you by any chance using a Samsung device? Would be awesome to have this tested on a non-samsung device as well since Samsung has some custom patches which affect this *shudder*

Just wanting to ensure that this breaks nothing. If it does we can always go with an accepted risk :)

masensio:
@LukasReschke no worries. @purigarcia will test this Pull Request with a non-samsung device and also its important to the automation tests. So she will do it.

masensio:
Hi,
I've tested Create a folder, rename a folder  with  Nexus 5 (4.4.2), Samsung Galaxy tab (4.2.2) and Nexus 9 (5.0.1) and  SwiftKey and Launcherex, everything works properly until now.

I'm checking if the automated tests work fine. Cross fingers!
-

davivel:
As discussed internally, and due to problems with the automatic tests, this will be delayed to next release.

cc @MTRichards , @cmonteroluque , @rperezb

LukasReschke:
Let's consider this an accepted risk as of now then and document this as such. – This is aligned with other vendors such as Dropbox who decided the same.

davivel:
@LukasReschke , the problem I talked about is not related with the PR itself and needs to be fixed anyway.

I understand this is not critical, but can keep it around 'til the tests are fixed.

davivel:
I'd recommend to delay this issue to have a fast bug-fixing release in 2.0.1. Far from being critical, IMHO.

cc @rperezb @MTRichards @cmonteroluque

cmonteroluque:
ok @davivel

davivel:
I'm very sorry to say this will have to go to be delayed one release. See some more details in #1750 (comment)

davivel:
Rebased and applied to layouts added from the creation of the PR.

davivel:
I can confirm that the flag doesn't prevent that automated tests interact with the UI. Tested with Espresso.

That should be enough to consider it tested. Do you agree, @jesmrec ?

jesmrec:
right @davivel, the main problem would be that we can not perform the Espresso UI tests. If we can, go ahead

Issue #1110, opened 2015-08-24
Accentuated letter are not sorted alphabetically


etiess:
In a folder list and in a file list, accentuated letter are not considered properly to sort the folders/files.

For example, the letter "e" with accent "é" should be considered as a normal "e". But a folder or a file starting with "é" is sent to the end of the list, after "z".

Example:
- Étudiant should be before Radio
- Fé should be before Fr

It's the same for other accentuated letters (I didn't test all of them): í, à, ô, ù, ñ, ...

masensio:
Hi @etiess,

Thanks for your ideas.

The current method to sort the filenames is using an Alphanumeric Comparator, but it doesn't take into account the localization.
To improve the sorting method we can take into account the locale, in this way we can sort accentuated letter properly. 
I think it is a good improvement, but I can't say when it will be available.

etiess:
I understand. I don't know if it can help, but I found a workaround here: http://stackoverflow.com/questions/16282083/how-to-ignore-accent-in-sqlite-query-android
I quote: 


Generally, string comparisons in SQL are controlled by column or expression COLLATE rules. In Android, only three collation sequences are pre-defined: BINARY (default), LOCALIZED and UNICODE. None of them is ideal for your use case, and the C API for installing new collation functions is unfortunately not exposed in the Java API.

To work around this:

    Add another column to your table, for example MOVIE_NAME_ASCII

    Store values into this column with the accent marks removed. You can remove accents by normalizing your strings to Unicode Normal Form D (NFD) and removing non-ASCII code points since NFD represents accented characters roughly as plain ASCII + combining accent markers:

    String asciiName = Normalizer.normalize(unicodeName, Normalizer.Form.NFD)
        .replaceAll("[^\\p{ASCII}]", "");

    Do your text searches on this ASCII-normalized column but display data from the original unicode column.

tobiasKaminsky:
![locale](https://cloud.githubusercontent.com/assets/5836855/9467641/5c8d7ce8-4b3d-11e5-8e36-2e1f7361f077.jpg)

That was an easy one :)

etiess:
Wao ;) Great!

tobiasKaminsky:
PR is in #1117

etiess:
@tobiasKaminsky I just installed the last version of the android app: looks great. (v. 1.8.0)

This issue is still present on my mobile: is your PR integrated in this version? Thank you

AndyScherzinger:
[...]

PR #1114, opened 2015-08-25
try fix #1113

tobiasKaminsky:

Issue #1113, opened 2015-08-25
png background is shown in the folder


purigarcia:
- STEPS TO REPRODUCE IT
  There is an account which has two folder:
   -a (Inside folder a, there are two png without background)
   -Documents

   1. login
   2. Access to folder "a"
   3. Thumbnails without background is shown
   4. go back to root

- CURRENT BEHAVIOR
    -- The background for png are shown in the folder 

- EXPECTED BEHAVIOR
    -- background is only shown in png images, not in other icons

- DEVICE
      -- Motorola e v.4.4.4

This issue came from #1056

@tobiasKaminsky could you take a look? thanks

purigarcia:
![background](https://cloud.githubusercontent.com/assets/4538201/9466401/ea57462a-4b34-11e5-8766-850853f46959.png)

tobiasKaminsky:
@purigarcia Strange, this is not happening to me.
Can you try #1114 ?

PR #1117, opened 2015-08-25
Sort accents

tobiasKaminsky:
ref: #1110


Kernald:
While your fix to #1110 looks fine to me (and that's the really important part of this comment!), I have two remarks: your two first commits (6df6548 and 69e7869) are unrelated to this change, and you made a lot of space/newline changes. It makes reviewing more complicated, it would be great to avoid it (or at least, change it in a specific commit, maybe? I don't know the usage in this repository).

tobiasKaminsky:
You are right.
But I doubt that this minor optical clean up would be merged if it is a single PR ;)

And I do not like the new lines before {, e.g:

if (test == 0) 
{ 
do this
}

is nasty and should be:
if (test == 0){
  do this
}

Kernald:
I agree totally on the style ;-) If this happens again, maybe you should split the changes: a first commit to change the style, then the actual code change… As I said, I don't know if there's an usual way to work with these kinds of things at OwnCloud, it's just a personnal feedback :-)

AndyScherzinger:
I also reviewed the code and it looks absolutely fine to, imho ready for QA then: @davivel @rperezb 

+1

rperezb:
Thanks!

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -326,6 +326,8 @@ public View getView(int position, View convertView, ViewGroup parent) {
                         Drawable backrepeat = mContext.getResources().
                                               getDrawable(R.drawable.backrepeat);
                         fileIcon.setBackground(backrepeat);
+                    } else {
+                        fileIcon.setBackground(null);
```
masensio: Is this line necessary for the fix?

I think this line is from another fix (https://github.com/owncloud/android/pull/1113 and https://github.com/owncloud/android/pull/1114). Besides at the end this line had been removed in this fix.

Maybe, if you update this PR with the last version of master branch this line will disappear.

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -335,6 +337,8 @@ public View getView(int position, View convertView, ViewGroup parent) {
 
             } else {
                 // Folder
+                fileIcon.setBackground(null);
```
masensio: Same comment as in the previous line.

tobiasKaminsky:
@masensio You are right. This is now replaced :)

masensio:
Thanks @tobiasKaminsky for the changes.
So PR Ready to be tested :)

rperezb:
@Dianafg76 can you please take care of QA this? 
thx!

Dianafg76:
I tested and it is working OK
Smsung Galaxy S3 v 4.1.2

masensio:
thanks @Dianafg76

PR #1119, opened 2015-08-28
Strip away index.php/apps/files to allow copy pasting from the server

nickvergessen:
When copying the server URL from the browser windows, it is something like
`http://localhost/ownCloud/master/core/index.php/apps/files/`

This patch removes trailing `index.php` and other `/index.php/apps/*` parts for the user

I will make the same patch for the desktop client soon. It lowers the burden for the user a bit.

### DIFF
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -742,7 +742,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mOAuthTokenEndpointText.getText().toString().trim());
         
         getServerInfoIntent.putExtra(
-                OperationsService.EXTRA_OAUTH2_QUERY_PARAMETERS, 
+                OperationsService.EXTRA_OAUTH2_QUERY_PARAMETERS,
                 queryParameters);
         
         if (mOperationsServiceBinder != null) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Only saw this now – good call @nickvergessen :+1:

masensio:
Hi @nickvergessen,

thanks for your contribution! 
Code review done. :+1: 
This issue will be tested as soon as possible. :smile:

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1143,6 +1145,17 @@ private String normalizeUrlSuffix(String url) {
         return url;
     }
 
+    private String stripIndexPhpOrAppsFiles(String url, EditText mHostUrlInput) {
```
przybylski: Can't this be achieved using URL class? It looks like it can be. Something like:
URL u = ...;
return u.toString().substring(0, u.toString().length() - u.getPath().length()-1);
?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1143,6 +1145,17 @@ private String normalizeUrlSuffix(String url) {
         return url;
     }
 
+    private String stripIndexPhpOrAppsFiles(String url, EditText mHostUrlInput) {
+        if (url.endsWith("/index.php")) {
```
przybylski: What about: http://hostname/index.php/ this is valid URL and will probably fail?

przybylski:
sorry my comments were invalid ;)

rperezb:
Validated! it's working as expected.

@masensio @davivel can you please merge? 

thx

masensio:
Merged!!

PR #1142, opened 2015-09-07
Disable file access from the web view

LukasReschke:
This is probably not required here and good practise as a further hardening.

Fixes https://github.com/owncloud/android/issues/1143


masensio:
Thanks @LukasReschke  for the fix.
Before merging this PR we need to ensure the app works properly if Saml is enabled as authentication method. 

Ready to be tested.

davivel:
@rperezb , probably we should push this up in priority with the other tiny PRs to merge next.

LukasReschke:
Yeah. Would be great to have. THX.

rperezb:
:+1:

davivel:
:+1:, waiting for Travis...

davivel:
In! Thanks, @LukasReschke

PR #1124, opened 2015-08-30
New filetype icons

Kernald:
This uses the icons referenced in #1118 . As mentioned in this issue, the ldpi variant has been scaled down to 32\*32px (from 36\*36px) to keep a multiple of 16px.

Here's the result, on an xhdpi screen:
device-2015-08-30-222849

While testing this, I also noticed an issue on the file list adapter: checker background was added on PNG items, but the background wasn't reset on other items. This is visible on the screenshot above on the folder and the PDF file (I fixed this after I took the screenshot).

Issue #1118, opened 2015-08-28
Use new filetype icons


jancborchardt:
From core master: https://github.com/owncloud/core/tree/master/core/img/filetypes

Preview:
![filetype icons new](https://cloud.githubusercontent.com/assets/925062/9544685/f5bf92e2-4d82-11e5-85cf-96d333c73eb6.png)

They are 16*16px but of course scale them up. Please only in multiples of 16px, for example 48px.

This should be done in sync with the release of 8.2 server. cc @rperezb @MTRichards @davivel @masensio

ggdiez:
Hey @jancborchardt 

These new icons looks nice. :+1: 

I can not wait to change the current ones for these.

jancborchardt:
cc @tobiasKaminsky @AndyScherzinger @Stoyicker @Kernald too :)

AndyScherzinger:
Nice! This should simply be replace the actual png files with these and that's it :)

Kernald:
What about the ldpi version? Currently, the images are 36*36 px - which is not a multiple of 16. Should I upscale them to 48 px?

jancborchardt:
Hm, damn. It’s pixel-perfect at 16*16. Maybe then it’s better to just use 32*32 instead of 36?

JasperWeiss:
@AndyScherzinger can you add these to your material design work for android?
when is material design going to be merged? can't wait :)

Kernald:
@JasperWeiss I generated all the icons in the right size, I'm just waiting for an answer regarding the ldpi size to open a PR.

jancborchardt:
@Kernald as I said: Use 32 * 32 instead of 36. :)

Kernald:
Alright, I'm not at home right now, I'll open the PR later today.

jancborchardt:
Great stuff, thank you! :)

rperezb:
thanks @Kernald !
@masensio will review it :smile:

AndyScherzinger:
should further icons be added? The screenshot of the webapp shows the gear (for bin) and the "<>" for css. This would also be something that could potentially be added to the android app :)

Kernald:
For reference, if additional icons are planned: the mime-type/icon equivalency table from core is [here](https://github.com/owncloud/core/blob/master/core/js/mimetypelist.js).

AndyScherzinger:
[...]

PR #1147, opened 2015-09-09
Add additional filetype icons

Kernald:
First step of #1118 was done in #1124: replace existing icons by new ones. This PR adds new icons for additional filetypes:
 - application
 - calendar
 - vcard
 - code
 - external folder


AndyScherzinger:
Sweet! Thanks @Kernald!

@davivel @jancborchardt @masensio I merged it into the proper feature branch and will soon start implementing the 8.2 icon logic to incorporate the new icons - stay tuned :dancer:

AndyScherzinger:
@Kernald do you know where to find the logic (in the web app) to calculate the right icon (besides the file-type list), to get an idea when to choose the calendar icon for example? :)

Kernald:
I think it's [here](https://github.com/owncloud/core/blob/master/core/js/mimetype.js#L76). I should have kept the same filenames, then… you have the mimetypes to look for in the last array of [this file](https://github.com/owncloud/core/blob/master/core/js/mimetypelist.js#L85).

jancborchardt:
Btw @Kernald @AndyScherzinger @davivel do you compress images, or does the compiling step have something for that anyway? Otherwise I can recommend Trimage: http://trimage.org/

AndyScherzinger:
According to http://developer.android.com/guide/topics/graphics/2d-graphics.html#drawables-from-images there is an optimization via the aapt tool within the build process. I don't know if this is as efficient as a dedicated optimization upfront though.

AndyScherzinger:
@Kernald no worries about the file names. The Android app has a slightly different, smarter implementation which has a detection algorithm which checks the mimetype to find an icon but also has a fallback if that doesn't work and then guesses the icon based on the file extension.

Kernald:
@jancborchardt I didn't compress then in any way, they are straight out of Inkscape. And as @AndyScherzinger mentionned, aapt crunches them during the build. So they may take up some "useless" space on the repo, but not in the final APK. I can trim them if needed, of course.

jancborchardt:
@Kernald if you have some free time, feel free to crunch all the PNGs through Trimage then. :) http://trimage.org – simple drag & drop and it replaces the originals.

Kernald:
@jancborchardt I'll do it, probably this week-end ;-)

PR #1155, opened 2015-09-14
Update setup, changes from the issue #1154

masensio:


davivel:
:+1:

Documentation, doesn't need QA.

Issue #1160, opened 2015-09-16
Refresh icon/action on URL textfield


AndyScherzinger:
With the new (material design) implementation buttons can't simply bind java methods via configuration anymore but need to do this is the way we already did it except for one instance in the login screen, when you hit the refresh button (for example when you entered an invalid owncloud url).

Bug number 2 (I know this should be separated..) is a UI glitch, or rather two on the login screen:
* the refresh button is oversized in the landscape mode
* the hint text has a size glitch when you refresh on an invalid address (spinning icon Vs. warning icon)

Unfortunately I don't have the time now to fork/PR but I have the fix already on the other branches for #1090 and #1100 

e.g.
https://github.com/owncloud/android/commit/044c597e2bf81c40340a9e43bcb9c965d4792dbd
https://github.com/owncloud/android/commit/549284e4f97481b3bcffb2528925fc28525ff9cf
https://github.com/owncloud/android/commit/d78077de4da35f0104457c60675421c3f01ba396

hopefully I will find the time to open up a PR tomorrow. 

**Update**
The crash bug isn't present on 1.8/master just on my feature branches. The remaining two issues have been fixed on PR #1163

AndyScherzinger:
imho closed since #1163 has been merged. @masensio @davivel

davivel:
:+1:

Issue #192, opened 2013-06-19
(As an ownCloud Android user I want) input elements to keep their size and position


jancborchardt:
Currently the big refresh icon increases the size of the Server address input field, and hence also pushes everything else down.

Instead, the input field should keep its size. The icon could be a bit smaller (tappable area should still be big enough though!). It’s your call – maybe there’s a standard Android refresh-button-inside-of-input-field element?

davivel:
Sorry, but this is not fixed yet. Checked with Nexus 5, Android 5.0.1. In case of doube, enable "Show layout bounds" in the developer settings of the device.

tobiasKaminsky:
@AndyScherzinger maybe you can verify if this is still the problem after your fabulous material update?

AndyScherzinger:
Problem still existed and has been fixed in PR #1163 (which I targeted for 1.8 now) - merge depends on @rperezb @davivel and @masensio I guess

masensio:
PR #1163 merged so this issue is fixed

PR #1162, opened 2015-09-17
Fix refresh click on login screen

AndyScherzinger:
As mentioned in #1160 there were some problems on the login screen, while the crash not being present on the 1.8 branch (but on my actual material branches, sorry for that). Two issues remained, being:
* refresh icon size on landscape
* hoping server status text

Both are fixed with this PR and my guess would be that they both were present in 1.7.2 but I haven't tested this. At least for the refresh icon I found an issue: #192

This PR is a UI change only, no code changes as in java, just xml fixes.

Before:
device-2015-09-17-112526

After:
device-2015-09-17-113956

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -17,8 +17,8 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
  -->
 <manifest package="com.owncloud.android"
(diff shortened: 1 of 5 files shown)


AndyScherzinger:
Wrong target branch...

PR #1163, opened 2015-09-17
Fix refresh click on login screen

AndyScherzinger:
As mentioned in #1160 there were some problems on the login screen, while the crash not being present on the 1.8 branch (but on my actual material branches, sorry for that). Two issues remained, being:
* refresh icon size on landscape
* hoping server status text

Both are fixed with this PR and my guess would be that they both were present in 1.7.2 but I haven't tested this. At least for the refresh icon I found an issue: #192

This PR is a UI change only, no code changes as in java, just xml fixes.

Before:
device-2015-09-17-112526

After:
device-2015-09-17-113956

### DIFF
--- a/res/layout-land/account_setup.xml
+++ b/res/layout-land/account_setup.xml
@@ -82,7 +82,7 @@
 	        		android:id="@+id/hostUrlFrame"
 					android:layout_width="match_parent"
 					android:layout_height="wrap_content"
-					android:layout_marginBottom="10dp"
+					android:layout_marginBottom="0dp"
 			        >
 					<EditText
 						android:id="@+id/hostUrlInput"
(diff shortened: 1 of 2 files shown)


davivel:
Thanks, @AndyScherzinger . 

About #192, you already fixed it in portrait with the first Material Design PR; so, although a "side fix", it's fine to fix it also for landscape in this release.

AndyScherzinger:
@davivel you are right about the icon, since I replaced the icon set for material ones, and thus had icons in the right size so the field resizing didn't happen :) Besides that the landscape mode also used another icon and another hard coded size definition.

The fix I did on the side (and maybe introduced with material) was the jumping server status text (take a random url and hit the refresh button) wenn the status text is replaces it switches from the spinning icon and the warning sign icon. The later as all material icons seems to be a bit larger in size with in turn leads to a jumping text during the refreshes (quite a detail, but a glitch nonetheless).

AndyScherzinger:
pinging @masensio  :)

masensio:
Hi @AndyScherzinger,

we have passed the tests about the login and more for checking this fix. Everything related wth this fix works fine. So I'll merge this PR and continue with the tests on release1.8 branch.

Thanks for your fix.

AndyScherzinger:
Nice! You are welcome! :)

PR #1090, opened 2015-08-07
Material buttons and checkboxes

AndyScherzinger:
Started a followup branch ```material_buttons``` and PR to #1076 as per discussion with @davivel 

Things being implemented on this branch (already):
* material design buttons --> DONE
* material design check boxes on multi-select lists --> DONE

Things I would start working on (subject to discussion):
* margin/padding according to material guidelines --> DONE
* recycler views instead of list views --> POSTPONED see #984
* replace third party grid implementation with android's grid layout implementation-> POSTPONED see #984

pinging @davivel @jancborchardt @tobiasKaminsky @masensio for the discussion

PR #1168, opened 2015-09-21
Avoid duplicate files

tobiasKaminsky:
As discussed in #819
Reported issues:
- [x] wording https://github.com/owncloud/android/issues/1311
- [x] setting view not updated https://github.com/owncloud/android/issues/1312
- [x] failed upload https://github.com/owncloud/android/issues/1313
- [x] error preview https://github.com/owncloud/android/issues/1319


res/values/attrs.xml
```diff
@@ -11,4 +11,18 @@
 
 	</declare-styleable>
 
+	<string-array name="pref_behaviour_entries">
+		<item>do nothing</item>
```
masensio: I think it is better to define a string per item and use them in the string-array item instead of using a literal. 
Besides, we are using Transifex to translate into different languages and it takes the file strings.xml to do the translation.

tobiasKaminsky: You are absolutely right, I have changed it.

tobiasKaminsky:
@all: Please review

tobiasKaminsky:
behaviour
Behaviour now has a summary.

tobiasKaminsky:
Behaviour now has a default value.

tobiasKaminsky:
@AndyScherzinger Regarding the dialog, it is not happening to me, so I cannot verify if.
But this is a ListPreference not a Dialog.
Do you have an idea?

tobiasKaminsky:
upload

@AndyScherzinger 16dp as padding on the left. Do I have to put this also on the right?

AndyScherzinger:
Yes, it should also be on the right even tough I guess we will never hit that end :)

AndyScherzinger:
As for the dialog, not really. This only happens on pre-lollipop devices where one needs to explizitely replace dialogs with the appCompat ones. Not sure if and how that could be achieved for a preference screen and its list preference in particular.

AndyScherzinger:
I pushed a custom list preference implementation that should take care of it. 
The Dialog looks fine now. I tested it on an Android 15 emulator which doesn't change the value on the preference screen right after choosing an option but it gets persisted (close preferences, open them again and you see the new value). I did the same test with the original ListPreference with the same results, so this might be due to the emulator.

@davivel @masensio can you by any change test this on a real device to make sure my latest push to this branch works for the List Preference?

AndyScherzinger:
Added some more styleing, see screenshot
device-2015-09-27-204223

AndyScherzinger:
@jancborchardt any comment on the paddings? The bottom elements are now in line with the material style guide (16dp) and thus are perfectly aligned with the icon in the action bar. Unfortunately the icons of the list items are slightly larger than the others and are thus center aligned with the actionbar icons (on the left) and the bottom "bar" icons. It looks a bit out of balance but I don't have any idea how to "fix" this visual inbalance :(

jospoortvliet:
I believe that Jan is on holiday so we'll have to figure things out by ourselves for a bit ;-)

I like how it looks now - perhaps we can merge it this way and create an issue to fix it later on. That creates a nice junior job for somebody new to our Android app!

AndyScherzinger:
Sweet! Than it is already in place in this PR.
@tobiasKaminsky Hope I is okay that I pushed this to your PR/branch :)

tobiasKaminsky:
@AndyScherzinger Thank you for your help with this PR.
It is great to see cooperation on an issue :)

simonbuehler:
when can this be merged into master? really would like to have that in the store app

AndyScherzinger:
@simonbuehler basically depends on the ownCloud team and their resource situation, any comment @rperezb @jancborchardt @davivel @masensio ?

tobiasKaminsky:
According to github, his is planned for the current milestone: https://github.com/owncloud/android/milestones/1.8.1-current
The release date is set to 10.11.15.
But no promises that it really will be in, it is more a plan than a hard rule ;)

jancborchardt:
Nice! Just a few details:
* [ ] the radio options (copy/move) should move together, and be centered. Otherwise Copy seems to belong to the Cancel button, and move to the Upload button – which is not the case.
* [ ] additionally the upload button should be colored with the primary action button style (like the log in button) to make it clear as the primary action.

AndyScherzinger:
@jancborchardt not sure I fully understand how to change the checkboxs' layout, can you provide a screenshot of some kind.

As for the button. Impossible to do so, since the theming to mark buttons as primary actions is part of #1090 which hasn't been merged yet, so adding it here would make this PR completely depending on #1090 so imho we should open a new ticket to document this and reference the two precondition PRs :)

tobiasKaminsky:
@jancborchardt I think @AndyScherzinger needs your input here...

tobiasKaminsky:
@jancborchardt (as I know you are online right now) Please have a look and try to clarify @AndyScherzinger question.

tobiasKaminsky:
New issue regarding colorify primary action button: #1204

AndyScherzinger:
Still an open question on the radio buttons for moving on with this PR.

jancborchardt:
Sorry for the delay! :) This is what I mean:
8eab84b0-6558-11e5-950a-a235ec886544
Even better though would be to have it integrated in the »Upload« button with a split button (like http://getbootstrap.com/components/#btn-dropdowns-split ) if that’s possible. By default it says:
»Upload (copy)«
And then in the arrow dropdown it says »Copy file« and »Move file«. What do you think? Because the very prominent radio buttons there are a bit strange and take a lot of space.

AndyScherzinger:
Ah, got it :)
@jancborchardt I can implement the radiobutton alignment as you have shown on the screenshot. The split button is no standard UI element on the Android unfortunately, so that would mean writing our own button implementation (couldn't find any library implementing such an element)

AndyScherzinger:
@tobiasKaminsky @jancborchardt like I said before this split button would be quite the project, the center alignment has been committed to the branch and thus I would stop working on this PR :)

jancborchardt:
Yeah, sounds good! A split button would indeed be a different project. ;)

AndyScherzinger:
Code looks good to me :+1: 

anyone else for review? @davivel @masensio 

Important for the testing is also the visualization of the chosen list preference in the preference screen (it is persisted, that's been tested).

jancborchardt:
cc @tombrossman for testing since he also mentioned the issue with the duplicate files. Tom, would be awesome if you can test this. :)

@tobiasKaminsky @AndyScherzinger is the beta already available somewhere, as an apk or something? The goal is to get it to F-Droid and the Play store, right?

tombrossman:
Sure I'll test it. Just need that apk once it's ready.

tobiasKaminsky:
https://github.com/owncloud/android/tree/beta/apks
Latest is currently: owncloud-beta-20151109.apk

jospoortvliet:
@tombrossman and, working?

/me is curious :grinning:

tobiasKaminsky:
@jospoortvliet You can also test the beta version if you want. :) :)
The more tester we have the fewer bugs will finally be there when the PRs are being tested/integrated by the core team.

tombrossman:
@jospoortvliet @jancborchardt Okay I've got ownCloud server installed again with the current latest owncloud-beta-20151118.apk running on a Nexus 4.

I have two problems so far, but the good news is that it looks like there are no longer duplicates filling up the storage on my mobile. The mobile settings I chose are Behaviour > do nothing, and Instant picture uploads > on.

The problems now are:
 - The mobile client displays the error notification "Upload failed  The operation couldn't be completed, server is.."
 - The mobile client repeatedly uploads the same image, they are appearing as photo.jpg, photo(1).jpg, photo(2).jpg, etc. on the server and on my desktop. This only happens with new photos I take, fortunately.

I haven't a clue how to read the full notification or how to check the mobile client logs, but I'm sure these will be helpful once I find them. I'll try to learn how over the weekend. I posted a [gist of the server logs from today](https://gist.github.com/tombrossman/80dccef0b79e97bdfc85) but I don't see anything related to image uploads so I don't know if that is useful information.

rperezb:
JFYI @masensio is currently reviewing this one! yay!!

masensio:
Hi @tobiasKaminsky,

I'm reviewing again this PR. 
It will be very useful if you update this branch with the lastest version of master branch. Could you update this PR?
 At this moment this PR has conflicts that must be resolved.

Thanks.

res/values/strings.xml
```diff
@@ -364,5 +364,14 @@
...
     <string name="file_list__footer__files_and_folder">%1$d files, 1 folder</string>
     <string name="file_list__footer__files_and_folders">%1$d files, %2$d folders</string>
+    <string name="prefs_instant_behaviour_dialogTitle">Upload file to server and ...</string>
+    <string name="prefs_instant_behaviour_title">Behaviour</string>
+    <string name="upload_copy_files">Copy file</string>
+    <string name="upload_move_files">Move file</string>
+
+    <string name="pref_behaviour_entries_do_nothing">do nothing</string>
+    <string name="pref_behaviour_entries_copy">copy file to OC folder</string>
+    <string name="pref_behaviour_entries_move">move file to OC folder</string>
```
masensio: I think is better to put the appname as a parameter, as in other strings, with %1$s. Just in case, we need to change it someday.
Then, to use String.format(getString(R.string.pref_behaviour_entries_copy), getString(R.string.app_name)); instead of getString(R,string.pref_behaviour_entries_copy).

(The same comment for pref_behaviour_entries_move)

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -122,6 +123,25 @@ private void handleNewPictureAction(Context context, Intent intent) {
         i.putExtra(FileUploader.KEY_UPLOAD_TYPE, FileUploader.UPLOAD_SINGLE_FILE);
         i.putExtra(FileUploader.KEY_MIME_TYPE, mime_type);
         i.putExtra(FileUploader.KEY_INSTANT_UPLOAD, true);
+
+        // instant upload behaviour
+        SharedPreferences appPreferences = PreferenceManager.getDefaultSharedPreferences(context);
```
masensio: This block is three times in the same file. 
Could you include this code in a method and use it instead of repeat it?

masensio:
@tobiasKaminsky, I've finished the code review. 
Please, see my comments.

I'll review it again when the branch will be updated with master branch, just in case.

res/values/strings.xml
tobiasKaminsky: This is not used in java code but in attrs.xml:16.
I do not know how to replace parts of the string within in an xml file.

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
tobiasKaminsky: This is done.

tobiasKaminsky:
@masensio Thank you for your code review. Apart of the the string ressource mention of you I have updated the PR and extracted the duplicated code part.
Again ready for review :)

jospoortvliet:
@tobiasKaminsky your last commit fixed the problem @tombrossman found, didn't it? The duplicate files? Nice work!

res/values/strings.xml
masensio: Maybe the easiest solution is to define the strings as:
<string name="pref_behaviour_entries_do_nothing">Do nothing</string>
<string name="pref_behaviour_entries_copy">Copy file to app folder</string>
<string name="pref_behaviour_entries_move">Move file to app folder</string>

We avoid to put the appname in the strings, because if we change the appname we avoid to review every appname reference in the xml files.

masensio:
Hi @tobiasKaminsky,
I'm reviewing the final implementation of this feature.

Related with the User Inteface, the uploader with move and copy option and the uploader for out of app files are not regular, you can see here:

Out of app uploader:
screenshot_20151123-173611

New uploader: 
screenshot_20151123-171723

On the other hand, the new uploader has two lines above the radio button. 
IMHO the radio button in this view gives an strange appereance. 
I was thinking about it and I'll send here my idea:
default group - page 1_1

Change cancel button by an spinner with the different options to do. The option selected in settings by default will be selected when the uploader is opened.
This is only an idea, open to discussion of course.
What do you think about it ?

@jancborchardt @tobiasKaminsky @AndyScherzinger @davivel @rperezb @jesmrec @MTRichards

tobiasKaminsky:
Hmm. It is an idea, but I am unsure if I like it more than the current way.
For me a spinner is a quite unusual thing for an action. For me it is more a fixed setting, like in settings of android.
Beside that, where would you put the cancel button? Or is "back" the only way to cancel? (which would be fine for me).

przybylski:
I'm with @tobiasKaminsky on this one. Radios are move obvious then a spinner

jancborchardt:
Are split buttons a thing on Android? http://getbootstrap.com/components/#btn-dropdowns-split That is, the upload button could be named

> Upload (move)

and be split, with the choice between `Upload (move)` and `Upload (copy)`.

jancborchardt:
The reasoning being that switching between Move and Copy won’t be something which is switched between a whole lot, hence it shouldn’t be such a present UI element taking away from the space the list uses.

przybylski:
@jancborchardt they are implementable but they are not a standard control so users might be confused and mistake them for a popup menu button

rperezb:
@tobiasKaminsky 

I have found a possible bug

Device: HTC Android 4.3
Server:  8.2.1

No matter if I select copy files or move file, at the end the file is always available on the device, file includes the green arrow, plus file is available on the ownCloud folder

davivel:
I'm having a look.

For the moment:

> Or is "back" the only way to cancel? (which would be fine for me).
- Back as cancel would be fine for me too.

@tombrossman, the issues you mention in the beta probably are not related to this particular pull request (PR). This contribution should have zero impact on instant uploads. There are better PR candidates to cause those problems.

- I like split buttons a lot, but would require more time to explicitly create the component; hard decision; probably it's better to do it later. Except for ... we discussed about this topic a lot, so we all know what "Copy file" and "Move file" mean. But I'm pretty sure that regular users will have no idea what's that about. And since the difference is not appreciable in the app, but outside the app, many people will be confused.

I don't have a solution yet. Thinking on it.

davivel:
@AndyScherzinger , there is no other way to fix the problem of padding without a custom class for the Preferences? Something based in styling for different Android versions?

Just a comment, I didn't really think in depth about the issue.

tombrossman:
@davivel no problem, I understand this patch targets the issue of dupes on the device and not on the server. It definitely solved that issue and I look forward to seeing it as new feature in a stable apk. I regret I can't help debug further just now but I am back to stable because all the duplicate uploads to the server were unacceptable (10x and counting when uninstalled...). Switching to the latest stable resolved the extra upload problems instantly. Thanks for all the work so far.

AndyScherzinger:
@davivel I am a little confused by the mentioning of me ?! ;)
The bottom area is using the standard 16dp padding but of course could use anything else but I didn't get the reference to this being a bug actually? Anymore details? Please? :smiley:

davivel:
Sorry, I misread some comments, it has nothing to do with the padding.

My concern is about the need of introducing src/com/owncloud/android/ui/dialog/OwnCloudListPreference.java . Seems like some comments about the dialog where it's introduced are not in the issue. Could you give me a clue about what happens in pre-Lollipop devices without it?

AndyScherzinger:
Puh, I might have to check it again. I think there was a compatibility issue and writing a list implementation could fix this but I don't remember right now what this is has been :(

davivel:
@AndyScherzinger , forget it, it's a minor question.

Thanks

AndyScherzinger:
Just checked the code again. It is simply due to design reasons. Using the standard list preference implementation means a "normal" Alert Dialog will be launched by androids implementation. My List Preference implementaiton basically overrides this behavior (nothing else) and launched an AppCompat Dialog so the Dialog is material designed like all the other dialogs. 

Hoping my explanation isn't to confusing.... So to answer your earlier question. Not using this custom implementation would mean that in the preference screen and only there the list preference would open in a design/theme depending on the Android version in contrast to all other dialogs.

davivel:
Understood, thanks.

AndyScherzinger:
Great, glad I could clarify :smiley:

AndyScherzinger:
@masensio the line isn't doubled (https://github.com/owncloud/android/pull/1168#issuecomment-158999850) I just added a grey line as a separator between the list and the bottom button area, so in some cases while scrolling it might leave the impression that there are 2 lines which is 1 for the bottom area and one is the list divider :smile:

masensio:
@AndyScherzinger, ok I understood this line is a separator. 
I think when this line seems a duplicated line is strange, maybe an option would be that this line was a little wider to be different than the others separators. It is only an idea, maybe this also seems strange.

AndyScherzinger:
@jancborchardt what do you think, regarding @masensio's idea for the separating line? (see screenshot https://github.com/owncloud/android/pull/1168#issuecomment-158999850)

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -122,9 +123,34 @@ private void handleNewPictureAction(Context context, Intent intent) {
...
 
+    private Intent addInstantUploadBehaviour(Intent i, Context context){
+        SharedPreferences appPreferences = PreferenceManager.getDefaultSharedPreferences(context);
+        String behaviour = appPreferences.getString("prefs_instant_behaviour", "NOTHING");
+
+        if (behaviour.equalsIgnoreCase("NOTHING")) {
+            Log_OC.d(TAG, "upload file and do nothing");
+            i.putExtra(FileUploader.KEY_LOCAL_BEHAVIOUR, FileUploader.LOCAL_BEHAVIOUR_FORGET);
+        } else if (behaviour.equalsIgnoreCase("COPY")) {
+            i.putExtra(FileUploader.KEY_LOCAL_BEHAVIOUR, FileUploader.LOCAL_BEHAVIOUR_COPY);
```
masensio: If I'm not wrong, is this option accesible at this point?
Same comment for behaviour.equalsIgnoreCase("DELETE")

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -101,6 +101,7 @@
     public static final int LOCAL_BEHAVIOUR_COPY = 0;
     public static final int LOCAL_BEHAVIOUR_MOVE = 1;
     public static final int LOCAL_BEHAVIOUR_FORGET = 2;
+    public static final int LOCAL_BEHAVIOUR_REMOVE = 3;
```
masensio: This option is never used

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
tobiasKaminsky: COPY and DELETE exist no longer, so we can remove them.

src/com/owncloud/android/files/services/FileUploader.java
tobiasKaminsky: Right, deleted.

masensio:
Hi @tobiasKaminsky,
thanks for your update. 
We need to test a bit before merging.

For me the code is good. :)

masensio:
PR Accepted :)

tobiasKaminsky:
:+1: 
This is great :)

AndyScherzinger:
Awesome! :+1:

enoch85:
:+1: wohoow! Just tested on beta, works great! Thank you!

jospoortvliet:
WHOOHOOOO

PR #1177, opened 2015-09-27
Add ownCloud color to notifications for lollipop

AndyScherzinger:
…for progress bar and icon background.

It is quite a details thing, see screenshot. The color setting on the NotificationCompat.Builder just has an effect on Lollipop devices and up.

Please Review:
Design: @jancborchardt 
Code: @davivel @masensio @tobiasKaminsky 

device-2015-09-27-212239


tobiasKaminsky:
Code looks good :+1:

masensio:
:+1: 
Ready to be tested

src/com/owncloud/android/notifications/NotificationBuilderWithProgressBar.java
```diff
@@ -55,7 +55,8 @@
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
             return new NotificationBuilderWithProgressBar(context); 
         } else {
-            return new NotificationCompat.Builder(context);
+            return new NotificationCompat.Builder(context).
+                    setColor(context.getResources().getColor(R.color.primary));
```
rperezb: @AndyScherzinger could it be possible to use this color https://github.com/owncloud/android/blob/master/res/values/setup.xml#L34 instead of the color.primary? The reason is because we have some branded apps and we need this change to bit there too.

@jancborchardt @MTRichards  what do you think? what Andy has modified is the background color of the notification image, I am suggesting to use the same one that it´s used on the top bar of the app

@AndyScherzinger please, don´t modify till we have their input :smile_cat: thanks!

AndyScherzinger: @rperezb primary is the action bar color :) - with the move to material design the primary color defines the color of the action bar. the color you referred to isn't used anymore in the app. So this is right now legacy code looking at android app but might still be used by other third party apps who haven't moved to material with the AppCompat implementation yet.

In case we want to define this for third party apps also would mean we would have to move the primary color etc to the setup.xml file

rperezb:
ummm, not sure, using this branch. If I modify the color: https://github.com/owncloud/android/blob/master/res/values/setup.xml#L34 the top bar is modified, for instance:

screen shot 2015-10-07 at 17 28 00

Yes, notifications are part of the app, however, we modify the image used on the notifications so that in some cases other images but the ownCloud regular ones are used. In this branch we are forcing to have the color.primary as background color and sometimes, this is not the expected one, we´d rather to have some kind of flexibility here. 
It´d be great if you use, at least, for this concrete option, the top bar color variable.

thanks!

AndyScherzinger:
Argh, now I found the issue.... this depends on the android version (could consider it a bug in my implementation). The actionbar uses the color-id you mentioned anything else doesn't (as in uses the primary color).

So my suggestion is: I will "map" the primary color (material design id) to the color-id you mentioned above and from there simply work with the primary, so whenever you change the color you mentioned above it will also work throughout the other definitions :)

rperezb:
sold! I talked offline with @MTRichards and he agrees on that color as a based one too :)

Great team work! (as usual)

MTRichards:
Thanks @AndyScherzinger and @rperezb

AndyScherzinger:
@rperezb @MTRichards I comitted the change thus you should now be able to change the color which would also have an impact on the notification color (completely untested atm). 
Now here's the BUT the material design defines some more colors/ids which aren't available in the setup.xml:
* you will only be able to change the main bars color (as shown on your screenshot)
* you cannot change the system bars color 
* you cannot change the secondary button color (pattern is part of #1090)
* you cannot change the checkbox color (part of #1090) since they are image based
* you cannot change the progress indicator below the actionbar they it is also image based + animation

So my proposal would now be:
* keep and merge this PR as it is now
* @rperezb opens a new issue which defines the things that you want to be able to have control over using screenshots hinting which elements you need color control for
* assign the ticket to me
* I will copy the neccesary color-ids into your setup file so you are able to manipulate all the defined elemtents accordingly :)
* last but not least, can you @rperezb clarify if and how the ids would have to be/stay in the setup.xml, just asking since the start/end color naming pattern for the actionbar doesn't make sense for material themend apps. My proposal here would be to keep them for now to be backwards compatible for third party apps, have the primary next to it and remove the lecagy ones at some point in time. :)

Any comments on this approach @rperezb @MTRichards ?

MTRichards:
A great teamwork I think.

-- 
Matt Richards
VP Products & Markets 
ownCloud, inc.

jancborchardt:
Nice! Just one tiny detail: The cloud is not centered visually in the circle. It needs to be moved a bit up and to the right – just a few pixels.

AndyScherzinger:
@jancborchardt I can't fix that ;) - The notification is a Android standard one so the positioning of the cloud depends on the positioning of the "cloud" in the png file itself. The icon within the notification is also the one used by android to be displayed in the system bar.

rperezb:
Thanks a lot @AndyScherzinger 
tested and it works just fine:

notification

lateral panel

Besides, in this branch you have modified the color of the text on the login screen:

login android
Which it´s perfect! it was something that we wanted to do :dancer: 

Then, I'll check your comment, https://github.com/owncloud/android/pull/1177#issuecomment-146917314

@masensio @davivel please review

AndyScherzinger:
Glad I could help. The login screen change is by definition. The text color is based on the colored primary. So as soon as you merge this branch and #1090 this will also impact the primary action button :)

jancborchardt:
Keep in mind there’s a difference between background color and primary action button color though. :)

* #1d2d44 (dark blue): header bar, background for this icon, etc.
* #35537a (lighter dark blue): primary action button (upload etc), probably heading text in settings (right?), and probably also sidebar part of where the username is shown

Input field text should not really be colored, that looks kind of strange, no?

AndyScherzinger:
Yes, you are right. :)
Input text is also defined by one of these colors by material default color assignment. Would have to check which one, but I think the dark blue is also set for the text field text color.

jancborchardt:
Shouldn’t the text field color just be default (black or off-black I guess)? Otherwise it’s going to look strange for some brandings I guess, especially when they have a light or neon-like header color (which I don’t recommend in the first place, but well).

AndyScherzinger:
So how to move on with this PR?

jancborchardt:
As soon as the input field text is fixed, it’s good to go from the design side. : )

AndyScherzinger:
@rperezb @Dianafg76 I changed the input field coloring according to the @jancborchardt's feedback. Sorry for taking that away from you.

In case this is something you need in a different way and discussed with @jancborchardt I can implement it, but in a different/new PR. Since bloating the PRs lowers the change to get the PR merged and also raises the time it usually takes to get the merged by the team.

rperezb:
Checked it.

The text letters are the default ones, black
@davivel @masensio can you please merge?

davivel:
Code ok :+1:

Thanks, @AndyScherzinger

AndyScherzinger:
You are welcome, always a pleasure @davivel :)

Issue #1166, opened 2015-09-20
Sync only one wifi continues after connection switch to mobile


tobiasKaminsky:
As @AndyScherzinger reported (https://github.com/owncloud/android/issues/41#issuecomment-138266840):
One issue I had with the wifi upload on my vacation was that the sync started being on wifi (as it should) but went on after I lost the wifi connection and further uploaded the files and thus impacted my data plan (being in a foreign country) pretty heavily with no possibility to cancel the uploads (I know their is already an issue opened for cancellations)

tobiasKaminsky:
@AndyScherzinger:
Imho a "simple" solution would be to check before every file if "wifi only" is activated and wifi connected and then upload the next file and else to stop (or pause).

tobiasKaminsky:
Additionally the upload must receive a broadcast intent if the connection switches from wifi to mobile data and cancel the upload.

AndyScherzinger:
Jepp, agreeing on the solution! :)

AndyScherzinger:
any knowledge on where this would need to be implemented in actual code base?

tobiasKaminsky:
In InstantUploadBroadcastReceiver, function handleConnectivityAction

AndyScherzinger:
Hmm

        if (!intent.hasExtra(ConnectivityManager.EXTRA_NO_CONNECTIVITY)
                && isOnline(context)
                && (!instantPictureUploadViaWiFiOnly(context) || (instantPictureUploadViaWiFiOnly(context) == isConnectedViaWiFi(context) == true))) {

--> checks the connection (type) but then simply starts uploading, then comes the loop:

Cursor c = db.getAwaitingFiles();
            if (c.moveToFirst()) {
                do {

So the code checks the connection and if it is appropriate starts uploading without ever checking the precondition (connectivity + user settings) again. So I guess we would have to kind of duplicate of duplicate the check and also do this check an the beginning of each loop iteration.

AndyScherzinger:
One thing I am not sure of is the reference to context. Would it be possible that context somehow gets terminated by Android's life cycle system and at some point in time when using it for checking for the wifi connection the code would crash? (I am not that firm with Android and weak/strong references and the context's life cycle when being used in a receiver instance)

tobiasKaminsky:
1. point: you are right. Within the loop there must be an additional check. This is easy and I did it locally.
But the problem is to cancel the current upload. This can be do via FileUploader.cancel(account).
[...]

PR #1191, opened 2015-10-07
Cancel upload on wlan exit

tobiasKaminsky:
ref #1166


AndyScherzinger:
haven't found the time yet, very Sorry!

@davivel @masensio @Dianafg76 @rperezb can you help @tobiasKaminsky out here since I most likely won't be able to test drive before end of next week :(

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -184,6 +197,11 @@ private void handleConnectivityAction(Context context, Intent intent) {
             Cursor c = db.getAwaitingFiles();
             if (c.moveToFirst()) {
                 do {
+                    if (instantPictureUploadViaWiFiOnly(context) &&
+                            !isConnectedViaWiFi(context)){
+                        break;
```
LukeOwncloud: At this point, can you be sure that the current upload is an instant picture upload?
What would happen if `instantPictureUploadViaWiFiOnly()==true` and `isConnectedViaWiFi()==false` but the users just decided to upload, e.g., their shopping list?

AndyScherzinger:
@tobiasKaminsky should we add this to the beta for test driving? Seems to make sense to me but you are the lead for the beta branch so I leave the decision to you

tobiasKaminsky:
Yes of course!
I will keep in mind. Either today or tomorrow afternoon.

tobiasKaminsky:
It is now in latest beta

davivel:
Starting review of this one.

src/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -372,6 +372,9 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             if (temporalFile != null && !originalFile.equals(temporalFile)) {
                 temporalFile.delete();
             }
+            if (result == null){
+                return new RemoteOperationResult(false, 404, null);
```
davivel: I need to review this execution path. Why 404, @tobiasKaminsky ?

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -177,6 +177,19 @@ private void handleConnectivityAction(Context context, Intent intent) {
             return;
         }
 
+        if (instantPictureUploadViaWiFiOnly(context) && !isConnectedViaWiFi(context)){
+            Account account = AccountUtils.getCurrentOwnCloudAccount(context);
+            if (account == null) {
+                Log_OC.w(TAG, "No owncloud account found for instant upload, aborting");
```
davivel: Better not including "owncloud" word in the log messages.

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -209,6 +211,21 @@ public void onDestroy() {
     public int onStartCommand(Intent intent, int flags, int startId) {
         Log_OC.d(TAG, "Starting command with id " + startId);
 
+        if (intent.hasExtra(KEY_CANCEL_ALL) && intent.hasExtra(KEY_ACCOUNT)){
+            Account account = intent.getParcelableExtra(KEY_ACCOUNT);
+
+            Log_OC.d(TAG, "Account= " + account.name);
+
+            if (mCurrentUpload != null) {
```
davivel: This block is duplicating the code in FileUploaderBinder.cancel(Account account); should be safe call mBinder.cancel(account) instead. 

Besides, should explicltly return after cacelUploadForAccount(account.name) with Service.START_NOT_STICKY. Right now the exit is done in the next if block, as a side effect of the extras not included in the Intent.  (I know this method is a mess with so many return points, all of them with the same value, but let's not rewrite that here, since there are many changes in FileUploader in the branch about the list of uploads).

davivel:
Seen. Please, @tobiasKaminsky, see the comments.

Thanks in advance.

src/com/owncloud/android/operations/UploadFileOperation.java
tobiasKaminsky: I do not know anymore.
Maybe it was just a placeholder while finding something better.
Do you have a suggestion?

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
tobiasKaminsky: So I just remove owncloud?

tobiasKaminsky:
@davivel answered your comments and updated to latest master.
As your questions are very small you can change them maybe by yourself to get the review faster done as I can only do it tomorrow afternoon (german timezone).

davivel:
No hurry, this will not go in current release, but in the next.

Thanks.

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
davivel: That's enough, yes, I'll do.

src/com/owncloud/android/operations/UploadFileOperation.java
davivel: LOL

OK, I'm checking. Thanks

davivel:
Reviewing again here...

src/com/owncloud/android/operations/UploadFileOperation.java
davivel: Mmmm... in theory, we should grant that ```result``` is never null at that point... that' s why it's not being checked.

davivel: The only way I can see that something could lead to ```result == null``` is that a ```java.lang.Error```occurred... what is pretty bad itself

davivel:
:+1: 

@jesmrec , ready for QA.

Issue #756, opened 2014-11-25
Make local storage size adjustable


tobiasKaminsky:
Currently the cache size is 10Mb, which may be a little small.
Do we want the cache size adjustable in Preferences?
Maybe also with the ability to clear it, just like Dropbox or Drive?

EDITED by @davivel

See https://github.com/owncloud/android/issues/756#issuecomment-240070815 below for updated scope.

tobiasKaminsky:
Clearing the cache is already possible via settings -> apps -> owncloud -> clear cache.
I do not know if an extra in-app setting is needed.

jab416171:
I think an extra in-app setting would be nice. In addition, downloaded files go to the SD card, so clearing the cache via settings doesn't delete the images on the SD card.

I think there should be a distinction between files that are kept in sync, and files that are temporarily downloaded for viewing. Temp files should be stored in cache, and files that are to be kept in sync should be stored on the SD card.

tobiasKaminsky:
This is something we had discussed in some other issue and would be the way google drive/dropbox is doing it.
The main benefit of our system is that other apps (like audio player) can use the downloaded files directly as they are stored in a normal file system.

jab416171:
If you just chose "Keep file up to date" for all of your music files, then audio players would still be able to use the downloaded files directly. Do we not want to store cached files (that are downloaded when we open the file) in the cache?

Also, and this is probably a separate bug, when uninstalling ownCloud it doesn't clean up its folders from the SD card.

tobiasKaminsky:
currently keep file up to date is not necessary for this. And from a user view it is not intended to have all files that one wants to use within other apps have to been checked via this checkbox.

Currently this is our intention: all opened files are stored in the filesystem and not in the cache. There are quite a few issues regarding this.

The second paragraph is definitive a new issue. Please create it.

davivel:
We should go further with this one, extend the LRU cache concept to the files synced in the local ```owncloud``` folder and let the user to decide the maximum local storage to be used to keep synced files (excluding those marked to be kept-in-sync). 

We shouldn't keep extending the settings to improve the use of network or storage only for images. Same concepts are applicable to all the files synced by the app.

I'm sure the idea of the global LRU has been commented in some other place, but I cannot find the issue, so I'll re-title this one.

0-wHiTeHand-0:
Hi,
[...]

PR #1195, opened 2015-10-12
Resize cache

tobiasKaminsky:
#756

10Mb is definitively too few for a cache when the user has many images, especially when #982 is merged.


tobiasKaminsky:
@AndyScherzinger @LukasReschke please review :)

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -237,6 +238,28 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+            Long cacheSize = ThumbnailsCacheManager.getMaxSize();
+            pCacheSize.setSummary(cacheSize + " Mb");
+            pCacheSize.setOnPreferenceChangeListener(new OnPreferenceChangeListener() {
+                @Override
+                public boolean onPreferenceChange(Preference preference, Object newValue) {
+                    String temp = (String) newValue;
+                    Long size = Long.decode(temp);
+                    if (ThumbnailsCacheManager.setMaxSize(size)){
+                        appPrefs.edit().putString("pref_cache_size", size.toString());
+                        pCacheSize.setSummary(size + " Mb");
```
AndyScherzinger: We use a capital B for Byte on the lists so for visual consistency I would use " MB" but I leave that to @jancborchardt

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -237,6 +238,28 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+        if (pCacheSize != null){
+            final SharedPreferences appPrefs =
+                    PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+            Long cacheSize = ThumbnailsCacheManager.getMaxSize();
+            pCacheSize.setSummary(cacheSize + " Mb");
+            pCacheSize.setOnPreferenceChangeListener(new OnPreferenceChangeListener() {
+                @Override
+                public boolean onPreferenceChange(Preference preference, Object newValue) {
+                    String temp = (String) newValue;
+                    Long size = Long.decode(temp);
```
AndyScherzinger: The two lines can be combined since temp isn't used anywhere else.

AndyScherzinger:
Looks good to me :+1:

src/com/owncloud/android/ui/activity/Preferences.java
tobiasKaminsky: You are right!
Mb is also official Megabit, which is not the same as MB = MegaByte...

Issue #1192, opened 2015-10-09
Date/time issue


Dreadeath:
How can I change date and time view in app below folder names? I have only time right now, which is not very useful
![screen](https://cloud.githubusercontent.com/assets/8825774/10394911/a950e29e-6e89-11e5-88bc-da5d019a5542.png)

Version 1.8.0 from f-droid.

AndyScherzinger:
At the moment there is no option to change this format. The app shows time being since the latest change which shows hours for the first day and after that shows the date (day/month). So this information will change in your app tomorrow :)

Dreadeath:
Some folders updated months ago :-\

AndyScherzinger:
hmm, that is strange.

Can you provide some kind of test account for your server to be able to debug this issue? 
What Server version are you using?

Dreadeath:
I did some tests, and when I change my smartphone's language to English, dates are displayed as it supposed to. So it is something localization-related I guess.

LukeOwncloud:
Confirmed. This is not a server issue. It's all about the system language. Try Russian, for example:
3 hours ago and 2 days ago seem to be correct. However, for older files just a timestamp as shown above is shown (I do not see any relation between that time and the actual change date.)

tobiasKaminsky:
FileListListAdapter: 522
private CharSequence showRelativeTimestamp(OCFile file){
        return DisplayUtils.getRelativeDateTimeString(mContext, file.getModificationTimestamp(),
                DateUtils.SECOND_IN_MILLIS, DateUtils.WEEK_IN_MILLIS, 0);
    }

I do not see any language specific or custom implementation from us...
So for now I assume this is a bug by android?

AndyScherzinger:
The further methods use the locale and plurals. Might be worth debugging. Not sure it is a Android bug. 
Quite sure there is a work around :) I'll see if I find the time to look into this.

AndyScherzinger:
@tobiasKaminsky @davivel @jancborchardt I found the "problem"...
https://github.com/owncloud/android/blob/master/src/com/owncloud/android/utils/DisplayUtils.java#L211
--> return dateString.toString().split(",")[0];

The ownCloud app's code is return the first part of the comma separated dateString. This works fine for English/German/etc. dates but not for e.g. Russian simply due to the fact that the Russian date format is <time>, <date> ... :(

[...]

Issue #188, opened 2013-06-19
(As an ownCloud Android user I want) more useful modified dates


jancborchardt:
Currently it’s very hard to figure out when a file was changed, because only absolute dates are shown. That means I have to check what time it currently is as well.

Instead of showing a full absolute date like »Jun 19, 2013 1:26:56 PM«, a relative modification date like »3 days ago« or »6 minutes ago« would be much more useful.

This is also how the web interface does it too. The exact algorithm is: https://github.com/owncloud/core/blob/master/core/js/js.js#L708-L731

PR #1200, opened 2015-10-15
fix for #1192

AndyScherzinger:
This PR contains the Fix for the date/time display for files and folders for certain locales as discussed in #1192 

@rperezb hope I labelled it correctly.
Please review @davivel @tobiasKaminsky @masensio

### DIFF
--- a/src/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/com/owncloud/android/utils/DisplayUtils.java
@@ -193,19 +193,7 @@ public class DisplayUtils {
         else if ((System.currentTimeMillis() - time) < 60 * 1000) {
             return c.getString(R.string.file_list_seconds_ago);
         } else {
-            // Workaround 2.x bug (see https://github.com/owncloud/android/issues/716)
-            if (    Build.VERSION.SDK_INT <= Build.VERSION_CODES.HONEYCOMB && 
-                    (System.currentTimeMillis() - time) > 24 * 60 * 60 * 1000   ) {
-                Date date = new Date(time);
-                date.setHours(0);
-                date.setMinutes(0);
-                date.setSeconds(0);
-                dateString = DateUtils.getRelativeDateTimeString(
-                        c, date.getTime(), minResolution, transitionResolution, flags
-                );
-            } else {
-                dateString = DateUtils.getRelativeDateTimeString(c, time, minResolution, transitionResolution, flags);
-            }
+            dateString = DateUtils.getRelativeDateTimeString(c, time, minResolution, transitionResolution, flags);
         }
 
         String[] parts = dateString.toString().split(",");
@@ -217,7 +205,7 @@ public class DisplayUtils {
             }
         }
         //dateString contains unexpected format. use localized, absolute date.
-        return DisplayUtils.unixTimeToHumanReadable(time);
+        return dateString.toString();
     }
 
     /**
--- a/src/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/com/owncloud/android/utils/DisplayUtils.java
@@ -207,8 +207,17 @@ public class DisplayUtils {
                 dateString = DateUtils.getRelativeDateTimeString(c, time, minResolution, transitionResolution, flags);
             }
         }
-        
-        return dateString.toString().split(",")[0];
+
+        String[] parts = dateString.toString().split(",");
+        if (parts.length == 2) {
+            if(parts[0].contains(":") && !parts[1].contains(":")) {
+                return parts[1];
+            } else if(parts[1].contains(":") && !parts[0].contains(":")) {
+                return parts[0];
+            }
+        }
+        //dateString contains unexpected format. use localized, absolute date.
+        return DisplayUtils.unixTimeToHumanReadable(time);
     }
 
     /**
--- a/src/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/com/owncloud/android/utils/DisplayUtils.java
@@ -210,10 +210,10 @@ public class DisplayUtils {
 
         String[] parts = dateString.toString().split(",");
         if (parts.length == 2) {
-            if(parts[0].contains(":") && !parts[1].contains(":")) {
-                return parts[1];
-            } else if(parts[1].contains(":") && !parts[0].contains(":")) {
+            if (parts[1].contains(":") && !parts[0].contains(":")) {
                 return parts[0];
+            } else if (parts[0].contains(":") && !parts[1].contains(":")) {
+                return parts[1];
             }
         }
         //dateString contains unexpected format. use localized, absolute date.
--- a/src/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/com/owncloud/android/utils/DisplayUtils.java
@@ -204,7 +204,7 @@ public class DisplayUtils {
                 return parts[1];
             }
         }
-        //dateString contains unexpected format. use localized, absolute date.
+        //dateString contains unexpected format. fallback: use relative date time string from android api as is.
         return dateString.toString();
     }
 


src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -207,8 +207,17 @@ else if ((System.currentTimeMillis() - time) < 60 * 1000) {
...
-        
-        return dateString.toString().split(",")[0];
+
+        String[] parts = dateString.toString().split(",");
+        if (parts.length == 2) {
+            if(parts[0].contains(":") && !parts[1].contains(":")) {
+                return parts[1];
+            } else if(parts[1].contains(":") && !parts[0].contains(":")) {
+                return parts[0];
+            }
```
LukeOwncloud: The `else if` branch is expected to be executed most of the times. Swapping `if` and `else if` branches would thus bring a tiny performance increase. (It is just two ``contains`` and marginal logic operators less, but this code is executed for *every* file in the current directory...)

AndyScherzinger: Good point, swapped it in commit https://github.com/owncloud/android/commit/e8eda963ae6aaf3ac73500c4b75a3c2d533b6176

AndyScherzinger:
For testing: Just launch the app in an emulator, log in, and check the dates on the folders files in the list view(s) for e.g. English and Russian (one of the languages that made this fix necessary) on pre 23 Androids.

masensio:
Code reviewed. 
+1.
No comments from my side.

So Ready for testing.

tobiasKaminsky:
Code looks fine :+1:

rperezb:
@AndyScherzinger could it be possible not to shown the time, just the date? so that we are following the current approach as well as other ownCloud clients one?

relative dates

AndyScherzinger:
@rperezb programmatically nearly anything is possible ;) Which language did you choose? The Problem with the format above is that the implementation (also pre 1.8.0) depends on a "," as a seperator between date and time which obviously doesn't work for your chosen locale. So in  order to fix this we might even have to work with regular expressions.
I am pretty bad at that :( @tobiasKaminsky @LukeOwncloud do you have any input on this? We seem to need another check which takes care of this format :disappointed:

rperezb:
yes! This is what we always say here, almost everything is possible...how much time it´s needed...that´s another matter :open_mouth: 

About the language,Czech (I think)

However, I have just noticed using the master branch, that we are also including the time within this languages, so perhaps, this is not so important since we are following the same pattern (we may consider this on another pr, if wanted)

AndyScherzinger:
Hmm, Czech shouldn't have Cyrillic letters I think. This PR is as of now an improvement for Russian and other languages which have a reverse format than for example en_EN.
So we could tackle this matter in a new PR since this PR fixes the issue in #1192 but as you said doesn't fix all (other) possible issues. Since it is still some time until the next release we can keep it within this PR for now and in case we don't get it fixed for the next release open a new issue for this matter so the partial fix which is quite annoying for Russian users doesn't get blocked for 1.8.1. Just an idea :)

rperezb:
It´s difficult to know the language :relaxed: when selecting it, it´s called on the language itself...
Checked, this pr fixed #1192 so, qa approved :+1: 
@masensio @davivel please consider merging

rperezb:
@AndyScherzinger about keeping this pr open, we'd rather to make small changes and, being possible, with the aim well defined so that it´s easy not only to develop it but also to qa it.
Feel free to work in another branch for this! thanks a lot for your willingness, very appreciated!

LukeOwncloud:
I just had a quick look. It seems that there is no DateFormatter in this project, yet, which only shows the date is a localized manner.  The (short) date format which is shown currently (for roman languages) comes from `DateUtils.getRelativeDateTimeString()` (Android library).

Just showing the date (not time) in a localized way should be possible with this code (not tested) (instead of last ``return``):

    Locale currentLocale = context.getResources().getConfiguration().locale;
    DateFormat df = DateFormat.getDateInstance(DateFormat.MEDIUM, currentLocale);
    return df.format(yourDate);

BTW, this PR uses ` DisplayUtils.unixTimeToHumanReadable()` which produces the output shown in the screenshot above. The same method is also used for dates in the future.

Quite difficult to get it right for all languages. Glad that you accept it as is! :)

AndyScherzinger:
New issue created for remaining, newly found date/time issue #1211

rperezb:
@AndyScherzinger I'm sorry, i have just found that this is not working on every device

This is the same account as shown on https://github.com/owncloud/android/pull/1200#issuecomment-150494691 on and Android 6 device set as language:

language

It´s not working
relative dates2

(Sorry for the confusion)

LukeOwncloud:
Isn't the last screenshot exactly what it is suppsed to be? Date in format "DAY.MONTH". What should it say instead?

rperezb:
@LukeOwncloud you are right, this is the Date.
However, I was expected something as on the first screenshot,  Day Month Year, @AndyScherzinger is this possible to be added?

tobiasKaminsky:
If you need my help, please specify more exact what it should look like.
Currently I am just confused ;)

LukeOwncloud:
https://github.com/owncloud/android/issues/188#issuecomment-59604126 shows a screen screenshot using whole output from `DateUtils.getRelativeDateTimeString()`. In the current version it is only the part in front of the comma (without time). Just taking this part is now creating the discussed problems.

@jancborchardt Would it be okay to use the whole output string as shown in the screenshot? This would simplify this matter a great deal.

The problem is:
Output of `DateUtils.getRelativeDateTimeString()`:

- always contains relative/absolute date + time
- does not always contain both parts in same order
- does not always contain a comma
- does not always contain a colon in the time part

rperezb:
@tobiasKaminsky 

This is how it´s shown on an Android 4 device:
relative dates

This is how it´s shown on an Android 6 device:
relative dates2

While yes, it´s shown the date instead of the time (which is the goal of this fix :) ) it´s shown within a different format, for instance, it does not include the year and this may lead to confusion too 

What @LukeOwncloud  is suggesting

screen shot 2015-10-26 at 9 06 22

LukeOwncloud:
Supplement to my previous post. I see four options to solve this issue:

- use the output of `DateUtils.getRelativeDateTimeString()` as is (Screenshots 1+3. Disadvantage: As mentioned above, output may vary for different languages and APIs. Advantage: Easy to implement, should show a nicely internationalized output)
- use the output of `DateUtils.getRelativeDateTimeString()` then try to find the date part and skip the time (This is was this PR does and what was done before -> Screenshot 2. Disadvantage: Does not work as so far as output of `DateUtils.getRelativeDateTimeString()` varies too much. Even if it works as intended, output may not be what we want (Screenshot 2). Advantage: Shorter than 1 (less characters) while still showing the important bits.)
- create custom output. Drawback: A lot of work and knowledge about all kind of languages required. Advantage: Output is exactly what we want.
- show absolute date only. (Drawback: Contradicts https://github.com/owncloud/android/issues/188. Advantage: Easy to implement.)

jancborchardt:
If it’s the current year, the year can be omitted. Also, don’t show seconds in the time display.

AndyScherzinger:
The hard part about this is what Luke already laid out is the fact that a custom implementation which also does this in a locale-ish way is quite a challenge since one would have to know the time formats throughout the locales and Android versions.
From what I understood the following behavior is desired for.
1. today (x... ago)
2. yesterday ("yesterday")
3. the day before yesterday ("vorgestern", only existing in certain locales)
4. any other day in the past in the actual year (day, month, time (hours, minutes)
5. any other day in the past in the years before (day, month, year, time (hours, minutes)

So in order to accomplish this we imho would need a custom implementation which would first check which of the described scenarios is given for the timestamp and then choose the appropriate formatting. 

Open issues I see as already mentioned by @LukeOwncloud :
* The given API from Google chooses the formatting so we would have to then parse, split, reassemble the representation which given the posts above is tricky due to the different formats chosen by the api on the different locales/android-versions
* Using a completely custom implementation would enforce exactly one formatting which would be confusing for users with a their expected formatting (best example would be US with MM/dd).

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -207,8 +207,17 @@ else if ((System.currentTimeMillis() - time) < 60 * 1000) {
...
+        String[] parts = dateString.toString().split(",");
+        if (parts.length == 2) {
+            if (parts[1].contains(":") && !parts[0].contains(":")) {
+                return parts[0];
+            } else if (parts[0].contains(":") && !parts[1].contains(":")) {
+                return parts[1];
+            }
+        }
+        //dateString contains unexpected format. use localized, absolute date.
+        return DisplayUtils.unixTimeToHumanReadable(time);
```
LukeOwncloud: Changing this line to `return dateString.toString();` would solve comment https://github.com/owncloud/android/pull/1200#issuecomment-151290744.

AndyScherzinger: How? Wouldn't it still be wrong for Russian (being just one of the known conflicting locales)

LukeOwncloud: `getRelativeDateTimeString` outputs something like `Dec 12, 4:12 AM` (except that date and time might be swapped, there is no comma, time might not contain colon, etc.) and if I interpret the documentation correctly it even shows the year for older dates. But there should be no seconds.
Actually, this function is pretty much what we want except that it shows additionally the time. So I think it is okay to use the whole output, if we are unable to determine what and where the time part is.

The seconds were a result from `unixTimeToHumanReadable` which outputs dates like ``Dec 31, 1969 4:00:00 PM``.

AndyScherzinger: But looking at @rperezb screenshot for Android 6 you end up with a weird result like "15.09" ?!

LukeOwncloud: I still do not think that it is weird. It is a date (hopefully in the correct format for the chosen local). And as @jancborchardt pointed out: "If it’s the current year, the year can be omitted."

AndyScherzinger:
changed the fallback date string as mentioned by @LukeOwncloud 
I also checked the formatting for Nexus5 Android 6 for Russian and it looks like this (perfectly fine!)
device-2015-10-29-102301

So I have no idea how @rperezb got the  formatting in her screenshot. Any details on this?
Else 15.9 looks bad imho but if that is how the respective Android version and its locale date formatting has chosen to render it, there is not much we can do about it if we want to support locale depending formatting ;)

jancborchardt:
Ah and btw – we have limited space on this screen because it’s mobile, and limited mental space anyway. So showing the relative date mainly should suffice. Having something like »3 days ago« is good, and adding »19:34« doesn’t add any particular value because it’s too detailed information in a view which should only give you an overview. It’s something which could be displayed in a detail view, if anything.

rperezb:
Checked with the latest changes...and it looks just great! (same language and device to the one used for the weird 19.5
english

other language

:+1: :+1: from my end
@davivel  @masensio

masensio:
:+1:  from my side

masensio:
PR in master branch

PR #1208, opened 2015-10-21
Share with users -- add share

davivel:
Another increment on the 'share with users' functionality. Added the operation to create new shares when a user is chosen in the list of suggestions.

Once again, not ready for formal QA yet.

@masensio, @AndyScherzinger , @tobiasKaminsky , @all : code reviews are welcome.

### DIFF
--- a/src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
+++ b/src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
@@ -116,7 +116,8 @@ public class CreateShareViaLinkOperation extends SyncOperation {
 
     public Intent getSendIntentWithSubject(Context context) {
         if (context != null && mSendIntent != null && mSendIntent.getStringExtra(Intent.EXTRA_SUBJECT) != null) {
-            if (getClient() == null || getClient().getCredentials().getUsername() == null) {
+            if (getClient() == null || getClient().getCredentials() == null ||
+                    getClient().getCredentials().getUsername() == null) {
                 mSendIntent.putExtra(
                         Intent.EXTRA_SUBJECT,
                         context.getString(R.string.subject_shared_with_you, mFileName)
--- a/src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
+++ b/src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
@@ -2,6 +2,7 @@
  *   ownCloud Android client application
  *
  *   @author masensio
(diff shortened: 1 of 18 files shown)


src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
```diff
@@ -121,30 +106,35 @@ public String getPath() {
...
-    public int getPermissions() {
-        return mPermissions;
+    public Intent getSendIntent() {
+        return mSendIntent;
     }
 
-    public Intent getSendIntent() {
+    public Intent getSendIntentWithSubject(Context context) {
+        if (context != null && mSendIntent != null && mSendIntent.getStringExtra(Intent.EXTRA_SUBJECT) != null) {
+            if (getClient() == null || getClient().getCredentials().getUsername() == null) {
```
AndyScherzinger: Not sure if this can happen but looking at the null checks there could still be a credentials is null scenario (in case getCredentials can be null)

src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    
+    public String getPath() {
+        return mPath;
+    }
+
+    private void updateData(OCShare share) {
+        // Update DB with the response
+        share.setPath(mPath);
+        if (mPath.endsWith(FileUtils.PATH_SEPARATOR)) {
```
AndyScherzinger: This Block can be shortened to ``share.setIsFolder(mPath.endsWith(FileUtils.PATH_SEPARATOR))``

AndyScherzinger:
Code looks good :+1: 
Just some minor comments/questions :smiley:

davivel:
Fixes from review applied. Thanks!

src/com/owncloud/android/operations/CreateShareViaLinkOperation.java
davivel: Probably it can't happen with the current code. But I also though that in the past, until the message composition made the app crash and we had to add the alternative subject without user name. Extra check will not harm.

masensio:
Code reviewed. :+1: 

Before merging it we need to solve the conflicts.

davivel:
Doing

davivel:
Got it, @masensio.

masensio:
Merged :smile:

PR #1216, opened 2015-10-26
Polishing shares: automatic keyboard in search and complete error messages

davivel:


masensio:
Code reviewed

src/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -201,12 +197,25 @@ public void onBackPressed() {
...
-        if (operation instanceof UnshareOperation) {
-            refreshUsersInLists();
-        } else if(operation instanceof CreateShareWithShareeOperation){
-            refreshUsersInLists();
-            // Clean action
-            getIntent().setAction(null);
+        if (operation instanceof UnshareOperation ||
+                operation instanceof CreateShareWithShareeOperation) {
+
+            if (result.isSuccess()) {
```
AndyScherzinger: No null check here for ``result`` in comparison to ``onGetDataShareWithFinish``

src/com/owncloud/android/utils/ErrorMessageAdapter.java
```diff
@@ -173,7 +173,11 @@ public static String getErrorCauseMessage(RemoteOperationResult result,
             }
         } else if (operation instanceof CreateShareViaLinkOperation ||
                     operation instanceof CreateShareWithShareeOperation) {
-            if (result.getCode() == ResultCode.SHARE_NOT_FOUND)  {      // Error --> SHARE_NOT_FOUND
+
```
AndyScherzinger: Unshare and createShare message calculation is basically the same code except for the chosen messages of course.

Would it make sense to put the message ids in a collection for example a hashmap with a key consisting of the message-id (maybe shortened and an key for the kind of operation.)? 
--> This way the code block could be shortened quite a bit :)

AndyScherzinger:
code reviewed :+1: 
Just 2 minor comments :smiley:

src/com/owncloud/android/ui/activity/ShareActivity.java
davivel: I'll change ```onGetDataShareWithFinish```. I prefer the app crashes in this case, helps to detect a development error sooner.

src/com/owncloud/android/utils/ErrorMessageAdapter.java
davivel: I wouldn't go with the collection. Those mappings are more difficult to read & maintain, and error messages may change quite easily.

About the other idea, I think I see your point. The messages about permissions need an overall review. But trying to refactor the messages around around their structure is dangerous because of the translations to languages with very different syntax. It's a topic we can revisit in a while, since we'll have to extend this dialog soon with the options to change permissions on the shares.

davivel:
Tiny update.

Thanks for the comments, @AndyScherzinger

AndyScherzinger:
You're welcome, always a pleasure! :)

PR #1219, opened 2015-10-27
Fix Select Local Version or Keep Both behaviour when solving file cof…

masensio:
…lict


rperezb:
Great! checked with 1 device (Android 6 and it's working!)
tricky one

AndyScherzinger:
code changes look good to me :+1:

davivel:
:+1:

Dianafg76:
I tested with Nexus 9 v6.0 and Android Galaxy S3 v 4 and is working OK

davivel:
So there we go...

PR #1100, opened 2015-08-17
Material FAB with speed dial implementation

AndyScherzinger:
This PR is for discussion and experimentally test driving a FAB implementation.

Nothing to see yet but I wanted to open a PR in order to check for build compliance and discuss the progress and changes/adaption of any feedback. Incorporating the discussion on #983

The branch is based on latest ````master```` as in release 1.9.1

PR #1220, opened 2015-10-27
Move to compile target Marshmallow (Android 6 - v23)

AndyScherzinger:
This is a PR ready to test drive and review and so on, probably some further discussion...

*Beware: This Branch is based on* ``material_fab`` - works for gradle and for ant

What has been done so far:
* Bumped compile target to 23
* Bumped AppCompat to 23.1.1
* Bumped build tools to 23.0.2 - same for the travis build
* Fixed MediaService implementation
    * Moved on to NotificationCompat.Builder (as it is done in all aother placed within the app) since the ``Notification.setLatestEventInfo()`` has been removed with Android M, see http://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-notifications
    * Also added the notification background coloring as done on master for other notifications and also added the material design "play button" icon for the notification/system bar
* Fixed Apache HTTP
    * With Android M, Apache HTTP has been removed from Android (!). I added the Android suggestion in case you "have" to stay with apache http, see http://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client - This could be refactored in a future, separate PR
* Added AppCompat Design lib 23.1.1
* ONGOING: v23 permission API calls for write access to external storage (necessary since compikle target has been bumped to 23) 

Pinging @davivel @masensio @rperezb for the Upgrade fact and the technical debt (which should be fixed in the future, see apache http)

Pinging @tobiasKaminsky @LukeOwncloud @stoyicker for community code review

Pinging @jancborchardt due to the fact that setting that background color (now also) for the Media player Notification is a UI change - a minor one, but a change nonetheless.


AndyScherzinger:
@masensio @davivel any idea why travis is failing?
``Error: Target id 'android-23' is not valid. Use 'android list targets' to get the target ids.``

Any idea how this can be achived (getting travis to know about Android M)?

AndyScherzinger:
@masensio @davivel Sorry for the disturbance, travis build is fixed. I hope the changes to the build tools version and bump to 23 doesn't conflict with your internal build pipelines. Nevertheless it is important and good to know that appache http is no more with Android M except when added as a legacy jar (done in this PR).

@tobiasKaminsky would this be something for the ``beta`` branch? It is a minor change in the owncloud code both still a new AppCompat version and compile target which is always a critical change.

AndyScherzinger:
depends on  #1100

tobiasKaminsky:
@AndyScherzinger Does this mean that the current version is not running on android 6?

AndyScherzinger:
It is perfectly running on 6, but as soon as you bump the compile target to 23 stuff needs to be fixed to compile again. So this is basically a staying on top of the AppCompat Lib maintenance  PR

AndyScherzinger:
And besides it is simply to stay on top of the Android target and to be able to use the latest AppCompat since it usually is targeted against the latest Android version. Targeting the latest Android version also means that the apps standard UI elements will be rendered in based on the Android version. If you target a lower version like ICS it would mean that even on never Android version the elements would be rendered with the ICS styling (except you manually changed styling in the theme file). This can usually be seen for buttons but not in the owncloud app since I moved to the AppCompat implementation of them to be able to implement Material Design for 4.0 and up.

AndyScherzinger:
permissions api discussion to be done in #1343

masensio:
Hi @AndyScherzinger,

Thanks for working in this PR. It is very interesting for us.
We want to merge this change on master branch, so it would be better if this PR were independent of the FAV branch, because we prefer to prioritaze this PR before the FAV PR #1100.
So, could you make this PR independent of FAV PR?

Also this branch has conflicts with master, could you rebase with master?

AndyScherzinger:
Hi @masensio sure thing, I can do that, but this PR is still under development since up to now I only added the changes necessary file access but not for contacts even though I don't know where the contacts would be accessed, but I can add that code change, probably over the weekend.

I'll try to rebase but maybe I need to do the changes again, because whenever I tried to rebase something goes wrong (I still don't fully know how to rebase :disappointed:)

Please be aware that this branch "only" has a initial rudimentary implementation checing for the permissions when the app starts. I does not (yet) check for the permission in all the places where they are needed which we should do at some point since with Marshmallow I can revoke permission at any time even while the app is running. But that is something which I would consider mid-priority not high. :)

This PR would also fix #1401 :grin:

masensio:
Hi @AndyScherzinger,

I know when you rebase you have to fix the conflicts step by step, but the repository history is cleaner at the end. I suppose you know the steps to start a rebase, and your problems with that process are the conflicts founded in the middle of the process. You need to fix these conflicts and continue with the rebase. If I can help you with thisin any way please let me know.

On the other hand, I know this PR is not ready to review, I've seen the ```Developing``` label, but I've seen it is not ```master``` branch dependent and I prefer tell you that we need this PR independent of FAB branch.

AndyScherzinger:
@masensio Yes please. I have never successfully done a rebase (resolving conflicts isn't the issue) but I simple don't know how to do this (with tortoiseGit) So if you have any hints for me. I googled but I just don't get it :cry:

AndyScherzinger:
I can even name all the commits that need to be kept if that is of any help...

davivel:
Hi, @AndyScherzinger . Joining here to bear a hand. Let me have a look.

davivel:
>  I only added the changes necessary file access but not for contacts even though I don't know where the contacts would be accessed,

I don't understand, why do you want to add permission to access contacts? The app doesn't need nor use it.

AndyScherzinger:
@davivel Sweet! :) - Still up for discussion is if we need to redo the changes I mentioned since @masensio wants the PR to be based on master not on #1100 
On the other hand @MTRichards wrote that #1100 is in the merge pipeline but open issue imho is the lib discussion. If the lib is fine for now than #1100 can be merged, and than this one is on par again with master.

For now this PR only implemented the file system permission check, open issue here is the contacts permission check.

AndyScherzinger:
@davivel I don't want to add the permission! But if you open up the app permission in the system settings you will see that for some reason the app defines permissions for file access and contacts. That hasn't been added by me. I just moved to the new permission implementation, so the branch gives you the implementation to ask for the file access permission.

davivel:
> @masensio wants the PR to be based on master not on #1100  
> On the other hand @MTRichards wrote that #1100 is in the merge pipeline but open issue imho is the lib discussion.

Yes, the point is that we need the lib update before than the FAB. The target is fixing #1401, that you researched about. We'd prefer to invert the order because testing on FAB will be more extensive, while the upgrade to v23 is mainly technical and will not require QA time.

davivel:
OK, I understand. That's weird... will check it.

davivel:
In any case, is it strictly needed to add the dynamic permission handling to get the compilation target in v23 working?

AndyScherzinger:
Sure thing, sounds fine to me, while I have to say in order to do it right (permission implementation) it would need quite the QA since on Android M I can revoke any granted right at any point in time, so permission checks would need to be implemented at all places where it is needed.

Thanks for the collaboration on this issue! :+1: 

Regarding you question: Yes dynamic permission handling is the way of the 23 API. For a quick win scanario I think we can start with simply having the start check I already implemented and not having check in all the other places, so whenever a user revokes permissions things won't work again until he restarts the app and get's asked again for the permission (this is all already implemented on this branch).

davivel:
Sounds fine. In any case, we'll have to push up the rest of permission handling in priority. D**n dependencies...

cc @rperezb

davivel:
OK, found the reason. GET_ACCOUNTS permission is in the CONTACTS group, and the interaction of the user is to approve or deny permission groups, that's why it's accessible in settings.

davivel:
Reference: http://developer.android.com/intl/es/guide/topics/security/permissions.html#normal-dangerous

davivel:
I transplanted ```target_marhmallow``` from ```material_fab``` to ```master```. Pushed it in a different branch, new PR is  #1473, still in progress.

davivel:
Keeping branch alive some time just in case I screwed something in the rebase process.

PR #1221, opened 2015-10-28
check status of external storage before file write/update

templateK:
check status of external storage before file write/update

### DIFF
--- a/src/com/owncloud/android/utils/FileStorageUtils.java
+++ b/src/com/owncloud/android/utils/FileStorageUtils.java
@@ -54,7 +54,7 @@ public class FileStorageUtils {
     public static Boolean mSortAscending = true;
 
     public static final boolean isStorageWritable() {
-        return Environment.getExternalStorageState(sdCard) == Environment.MEDIA_MOUNTED;
+        return Environment.getExternalStorageState() == Environment.MEDIA_MOUNTED;
     }
 
     public static final String getSavePath(String accountName) {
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -52,10 +52,15 @@
     public static Integer mSortOrder = SORT_NAME;
     public static Boolean mSortAscending = true;
 
-    
+
     //private static final String LOG_TAG = "FileStorageUtils";
 
+    public static final boolean isStorageWritable() {
+        return Environment.getExternalStorageState() == Environment.MEDIA_MOUNTED;
```
AndyScherzinger: This should imho be implemented with an equals since you are comparing Strings and to make it null safe I would suggest:

``return Environment.MEDIA_MOUNTED.equals(Environment.getExternalStorageState());``

AndyScherzinger:
Hi @templateK 
Thanks for your PR. Can you give some more details on the reason for the PR as in what is the impact of it. I looked through the code changes and found just one minor issue with a String comparison.

In case this is your first PR to any ownCloud repo, please fill out the contributor agreement, see https://owncloud.org/contribute/agreement/ , sign it and send it to @karlitschek and probably to @masensio 
The agreement is a precondition in order to contribute code to the owncloud repos.

templateK:
In file writing functions at  FileDataStorageManager class,
(```saveFile, saveFolder, moveLocalFile, copyLocalFile, updateSharedFiles ```)
they only checks  __existence__ of file on external storage.

Android external storage has many [state]( http://developer.android.com/reference/android/os/Environment.html) for example, __not-mounted, read-only__ etc.

So, I added  condition that checks __writable__ status of external storage to these functions.

templateK:
I decide that I won't contribute to this project. Feel free to close this PR.

tobiasKaminsky:
AFAIR you can also say that your code is under a special license, if you do not want to sign anything...

jesmrec:
closing... not scheduled

Issue #470, opened 2014-04-18
Unable to connect with SSL


ConiKost:
I've installed the ownCloud Client on my Android phone (Nexus4), Version 1.5.5. Owncloud itself is running on Version 6.0.2.

My Windows and Linux client can connect without any problems. But I am unable to connect with my android phone.

When typing my URL: https://cloud.domain.tld/ it just says "An unknown error occurred". That's it. What can I do?

https://cloud.domain.tld/status.php can be reached without any problems from my phone in the browser. I can't find any connection attempt on my webserver from my android phone..

privong:
For what it's worth, I am also experiencing issues synchronizing contacts and my calendar with owncloud using davDroid. This leads me to suspect that the fault is either the the nginx configuration or with an underlying android library. Unfortunately, I have not figured out how to differentiate between the two. Given my nginx configuration is unchanged (since before the problem arose), I supsect android, but I have no proof.

privong:
Correction, it seems this was an SNI issue, for me. I disabled all other SSL servers in my nginx configuration, restarted the server, and now I can connect with both the android owncloud client and with davdroid. So, perhaps this issue (at least for me) was related to #300 and there not (yet) being support for SNI?

davivel:
@privong, the support of SNI in the app is limited, conditioned by the smartphone. What model is yours? What Android version does it hold?

davivel:
@faulkner93 , @InfoSec812 , @Ruelam - the issue with the trailing slash is well-known; it should be fixed with the current released version (1.6.2), though we are still discussing if that's true in other issue. Sorry for the inconvenience.

davivel:
@R0000, that's quite strange. There must be some rule in your server configuration that is enforcing lowercase in some way. 

The app is not transforming the case-state of the URL. Even if it did it, your use case does not fit in a possible bug, since there is no way that the app knows wether precisely the 'C' character should be lowercase or uppercase.

Please, take a look to your web server configuration.

davivel:
@poschd , did you check the Nginx configuration? I think it suffered some update in the change from OC 6 to OC 7. Please, take a look to: http://doc.owncloud.org/server/7.0/admin_manual/installation/nginx_configuration.html

Other possibilities: do you use SNI in your server? What model is your device? What Android version?

davivel:
@brantg , do you use SNI in your server? What model is your device? What Android version?

davivel:
@ConiKost, is this still a problem for you?

[...]

PR #1226, opened 2015-10-30
Add support for new security protocols on older android versions

SandroMachado:
With this pull request the application should update its own [security provider](https://developer.android.com/intl/ko/training/articles/security-gms-provider.html#patching) enabling by default the latests security protocols and increase the overall app security.
The workaround introduced by https://github.com/owncloud/android-library/pull/43 can also be removed. Updating the security provider also fix the android issues with TLS on API levels 16 to 20.

This PR should also close the issues related to problems with security protocols like https://github.com/owncloud/android/issues/470 

Travis integration is broken because it is necessary to add support for the google play services `compile 'com.google.android.gms:play-services-base:8.1.0'`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ dependencies {
     compile project(':owncloud-android-library')
     compile 'com.jakewharton:disklrucache:2.0.2'
     compile 'com.android.support:appcompat-v7:22.2.1'
+    compile 'com.google.android.gms:play-services-base:8.1.0'
 }
 
 android {
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -59,6 +59,10 @@ import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import com.google.android.gms.common.GooglePlayServicesNotAvailableException;
+import com.google.android.gms.common.GooglePlayServicesRepairableException;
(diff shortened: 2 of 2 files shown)


tobiasKaminsky:
Requiring Google Play Services is, for me, a no go.
Why is this necessary? Can we omit this dependency?

SandroMachado:
We need the Google Play Services Base dependency because it is the only way to update the application context with the latest updates for the Security Provider. We only need the base dependency, so it is not the full google play services library.

Since the application is available through the Google Play Store, most part of the users will have the google play services and should be seamless to them the update. If the user does not have the google play services, the app will also work without breaking. This PR will increase the support and the security of the application.

kha0S:
:+1: for merging. This will disable insecure encryption methods.

SandroMachado:
This contribution is MIT licensed.

AndyScherzinger:
@SandroMachado the build is broken. Most likely due to the missing exploded jar/aar for the ant build and its corresponding build scripts. This would need to be fixed in order to be able to merge this in case @davivel and @masensio (core team) decide that this should be merged.

SandroMachado:
@AndyScherzinger thanks for the feedback.

AndyScherzinger:
@SandroMachado you are very welcome! Also I saw that the PR has merge conflicts which would need to be resolved. But as I said you probably should wait for the feedback from @davivel and @masensio before putting any more effort into it.

jesmrec:
TLS 1.2 already supported

PR #1236, opened 2015-11-01
Streaming

tobiasKaminsky:
This re-enables support of streaming:
- first it is tried with our internal player
- if this fails it allows the user to open the file with an external app
- with long press on video/audio -> streaming the user can directly choose an external app

Remaining security problem:
- the password is maybe being exposed to the user, when streaming with an external app. This depends on the app, if it shows the full url. A warning is shown to the user prior to opening the external app.


malkomich:
Sorry @tobiasKaminsky but I can't get it working :(

Have you tested it?

davivel:
We shouldn't trust on external apps for streaming, not with the current handling of credentials. Sharing the password with other apps is not an option.

Taking this into account, rebasing on master is too hard to be worth. Sorry, but can't go in.

Issue #880, opened 2015-02-17
Simple screen to share with users / groups


Bugsbane:
Currently, my main usecase that I want ownCloud for is to share photos and videos with family overseas quickly, as I take them. If I want to share a photo just after I take it with my phone though, I need to:

1. Go to the gallery
2. Click share
3. Choose a pre-existing folder (no new ones!)
4. Upload
5. Log in to ownCloud via the browser
6. Navigate to the folder where it is
7. Find the individual file
8. Click share
9. Select / type the appropriate user(s) / groups or public link
10. If I'm sharing with someone who isn't a user on my ownCloud, copy the link
11. Switch to the email / sms app
12. Send a message with the public link

This is just to share a file, surely one of the core use cases of the ownCloud app. Accordingly, I rarely actually bother to share, and if I do, it's not until way later when I'm at home at my computer, with the full ownCloud interface.

Instead, I propose that we have a "Share" screen, described below. This would be accessible from 2 places. First, in the menu where the user long tap and holds a file / folder while browsing. Second, via a second button next to the "Upload" button when using Android's share feature. The second button would be "Upload and share" which would, obviously, upload as it does now, and then open the share options page.

The share page would have the following 3 radio button options:

1. A list of users / group names with checkboxes to choose who to share with. If there are more than, say 10, it would show the 10 users / groups most recently shared with and a search / filter field for the user to find a pre-existing person on that ownCloud instance (limited to people in the same groups if "Limit sharing to the same groups" is active, unless the user is an admin)..

2. Generate a public link, only.

3. An email field to share with non users (generate a public link and send that email address a message with it).

"Generate public link only" would be the default the first time, but the chosen options would be remembered when the user shares next time.

masensio:
Hi @Bugsbane,

thanks for sharing with us your ideas about sharing feature. We will take them into account for the future improvements of the sharing .

At the moment, Android app provides you a way to share your files by link. You can copy the link or select an app to share with link with others.

Bugsbane:
Thanks @masensio 

Yes, I understand the ability to create a public link, however my main reason for using ownCloud is precisely because I don't want to make my private data available to the rest of the world in the way that using a public link would.

Anyway, thanks for the reply.

masensio:
Hi @Bugsbane.
I understand your problem, but don't worry.
In the near future we will implement other features about sharing as: share with password protection, share with groups ... as in the browser.

I hope this is a good news for you :)

Bugsbane:
[...]

PR #1252, opened 2015-11-04
Share with users & groups

davivel:


davivel:
Last commits here are mine, so @masensio should review.

Other reviews will be also very appreciated. cc @AndyScherzinger @tobiasKaminsky @LukeOwncloud

davivel:
Implements #880

davivel:
Depends on https://github.com/owncloud/android-library/pull/90

AndyScherzinger:
Looks good to me just two minor things regarding dimens.xml
* where is fab_margin used (only fab implementation I am aware of is in the respective feature branch)
* dimens.xml contains dimen properties which at least up to now have been put into dims.xml thus they should be kept in one file

besides that :+1:

davivel:
Thanks, @AndyScherzinger , I will check it.

davivel:
Totally unneeded file. We started the new activity testing the Dialog theme from code generated by an Studio template, the file was introduced then.

Good catch!

AndyScherzinger:
Glad I could help! :)

masensio:
:+1:  :)
Waiting for travis

Issue #1246, opened 2015-11-04
Add build date and git version to beta build


LukeOwncloud:
@Wikinaut wrote:

> @LukeOwncloud here https://gerrit.wikimedia.org/r/#/c/248699/ is a code patch (fully tested!) which I recently developed for the Wikipedia App, it adds variant and build name, and git version hash to the apk name.
> 
> Perhaps you want me to formally send a Pull Request for oc/android, then let me know, and I will supply it very quickly.
> 
> Changes are required in two files see patch set 10 :
> 
> * https://gerrit.wikimedia.org/r/#/c/248699/10/app/build.gradle
> * https://gerrit.wikimedia.org/r/#/c/248699/10/app/src/main/res/values/strings_no_translate.xml

LukeOwncloud:
That reminds me of: https://github.com/owncloud/android/issues/763 (show build date)
We decided not to include it as it is very technical (and the issue about reproducible builds (which I think now, is not valid.))

@tobiasKaminsky and @AndyScherzinger What do you think? Should we include @Wikinaut's patch? I'd say just showing the build date would be fine (https://github.com/owncloud/android/issues/763). Actually, just uniquely naming the apk would be fine with me as well.

Wikinaut:
@LukeOwncloud thanks.

Wikinaut:
by the way, I am now and then in the chat, but this appears seldomly be used by owncloud developers.

AndyScherzinger:
I think it would be a good idea for the beta app but would rather like the version name (not the numeric number) to be used thus it would simply be displayed in the settings screen, kind of like the build number discussion in the referenced ticket so basically the baseline version 1.8.0 - builddate and I guess the date would be fine since that is probably enough to find out which build it is and if a bug has been fixed already. What do you think @tobiasKaminsky ?

Wikinaut:
@AndyScherzinger can you come to the chat now?

AndyScherzinger:
@Wikinaut Sorry, I can't, at least not during the regular work hours

tobiasKaminsky:
The build date will be visible in Settings under "owncloud beta android app".
The filename is generated by me in this format: owncloud-android-20151106.apk.
This is required for the built-in update system.
So therefore I close it.

LukeOwncloud:
Does not work with `owncloud-beta-20151105.apk`. Menu entry shows string "Version owncloud beta" where the actual version should appear.

Wikinaut:
[...]

PR #1253, opened 2015-11-04
improve apk filename

Wikinaut:
Here I send you a formal Pull request, for those who want to try with improved filenames. The filename generation can be further modified, of course. Please let me know, what you want.

issue #1246 
filenames will look like example ownCloud-release-1.8.0.10800000-7856cafe.apk

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,16 @@ dependencies {
 android {
     compileSdkVersion 22
     buildToolsVersion "22.0.1"
+
+
+    defaultConfig {
+        // Define your application name here.
+        // It must neither be present in /res/values/strings.xml
+        // nor in /res/values/string_no_translate.xml
+        // nor in /res/values/setup.xml
+        resValue 'string', 'app_name', 'ownCloud'
+    }
+
     sourceSets {
         main {
             manifest.srcFile 'AndroidManifest.xml'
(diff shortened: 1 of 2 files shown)


Wikinaut:
The Travis says:

```
 [aapt] /home/travis/build/owncloud/android/res/layout/drawer.xml:86: error: Error: No resource found that matches the given name (at 'text' with value '@string/app_name').
```
So, this is in need to be modified because the app_name cannot be setup in the xml any longer, but is defined in build.gradle. Rationale: see links/sources in the build.gradle.

Wikinaut:
Clarification: my patch is tested and works locally in the Android Studio. It's a matter of adapting the continuous integration tests - if you want to apply this easy patch.

AndyScherzinger:
It is a build failure, simply due to the fact that travis is using the ant target, thus if to try to build the app from the command line using ant (after calling ``setup ant``)  it will fail...

Wikinaut:
@AndyScherzinger Is there something _I_ can improve ? (I don't think so, but pls. let me know.)

AndyScherzinger:
if you checkout the branch, open a console and do 
* setup_env.bat ant
* ant clean
* ant debug

you'll get the same error thus you need to somehow add the changes you made to gradle to ant to (didn't check how) because ownCloud does use both build systems. :(

Wikinaut:
@AndyScherzinger Thanks for the hints, I did not know about that. I did what you suggested, got indeed the same message, and need now to find a way to fix the error message.

Let me know, if you (or someone else) find a way to fix this.

Currently, I do not know, how.

AndyScherzinger:
@davivel @masensio any input on ant? :)

tobiasKaminsky:
Just a question to understand the purpose of this PR.
I have to sign every generated apk as otherwise the app cannot be upgraded from version to version.
For the built-in update process I will use -YYYY-MM-DD.apk once I have rewritten the update check.

For f-droid: I guess they have their own apk naming convention. And as far as I know/read they generate new version if there is a new tag on github. So for every "release" I do, I will:
- sign the apk and put in apks/ as it is currently
- add a tag to this branch to get f-droid update their apk

Wikinaut:
@tobiasKaminsky the PR just makes a reasonable filename, this is the only change.

tobiasKaminsky:
Please see #1246.
I do not think that this is necessary.
But thank you for this idea!

PR #1274, opened 2015-11-09
Read capabilities for showing share options

masensio:


src/com/owncloud/android/operations/GetCapabilitiesOperarion.java
```diff
@@ -0,0 +1,49 @@
...
+ * Get and save capabilities from the server
+ */
+public class GetCapabilitiesOperarion extends SyncOperation {
+
+    @Override
+    protected RemoteOperationResult run(OwnCloudClient client) {
+        GetRemoteCapabilitiesOperation getCapabilities = new GetRemoteCapabilitiesOperation();
+        RemoteOperationResult result = getCapabilities.execute(client);
+
+        if (result.isSuccess()){
```
AndyScherzinger: Didn't check the mplementation behind this, but would it make sense to null check result, result.getData() and result.getDate().get(0) is this null safe by design?

src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -154,6 +155,57 @@
                 ProviderTableMeta.OCSHARES_ACCOUNT_OWNER);
     }
 
+    // Projection of Capabilities table
```
AndyScherzinger: what is the purpose of this map? It basically projects Strings onto themselves.

AndyScherzinger:
Looks good, just had two questions :) :+1:

src/com/owncloud/android/providers/FileContentProvider.java
masensio: The Capabilities Table projection is required for  the query method of FileContentProvider for accessing to a row of the Capabilities Table.

src/com/owncloud/android/operations/GetCapabilitiesOperarion.java
masensio: Maybe with the current implementation result.getData never is null, but an additions check never hurts to avoid future problems.

masensio:
@AndyScherzinger thanks for your code review.

src/com/owncloud/android/operations/GetCapabilitiesOperarion.java
davivel: AFAIR, result.getData() can't be null, but result.getData().get(0) may be, at least in a general sense.

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -297,6 +297,9 @@ public View getView(int position, View convertView, ViewGroup parent) {
                     } else if (file.isDown()) {
                         localStateView.setImageResource(R.drawable.local_file_indicator);
                         localStateView.setVisibility(View.VISIBLE);
+
+                    } else {
+                        localStateView.setVisibility(View.INVISIBLE);
```
davivel: What's this change about?

src/com/owncloud/android/providers/FileContentProvider.java
davivel: I think the point of @AndyScherzinger is, this is not needed, nor for this table, nor for the others. I made a fast test commenting out the calls to ```.setProjectionMap``` that pass these maps as a parameter, and the queries work perfectly without it. As he said, we are just mapping the names to themselves again, so it results in no change.

Let's take this PR as a chance to fully remove ```mFileProjectionMap```, ```mOCSharesProjectionMap``` and ```mCapabilitiesProjectionMap```. Don't worry about QA, if this could break anything, will be easy to see in the regression tests.

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
masensio: This line was introduced when we solved some conflicts.

davivel:
:+1:

Waiting for QA

masensio:
Code is now squashed.
Changes from code review made too, including remove unneccessary Projections Maps for DB tables.

PR #1281, opened 2015-11-12
Share with password and expiration time  getlink and more

masensio:


src/com/owncloud/android/ui/dialog/ShareLinkToDialog.java
```diff
@@ -142,19 +141,13 @@ public void onClick(DialogInterface dialog, int which) {
                             ComponentName name=new ComponentName(
                                     actInfo.applicationInfo.packageName, 
                                     actInfo.name);
-                            mIntent.setComponent(name);                               
+                            mIntent.setComponent(name);
 
-                            if (sendAction) {
-                                dialog.dismiss();    // explicitly added for Android 2.x devices
+                            // Send the intent
+                            dialog.dismiss();    // explicitly added for Android 2.x devices
```
davivel: We should try to remove this, now that Android 2.x devices are not supported anymore.

davivel:
:+1:

Issue #906, opened 2015-03-02
Internal music player does not stop playing when file is deleted


smu:
### Steps to reproduce:

1. Download a sound file 
2. the internal player starts to play the track
3. return to file list
4. player continues to play in the background
5. delete the file (locally and remote)

### Expected behaviour
The player should stop.

### Actual behaviour
Instead the player is still playing and the only possibility to stop it is to kill the owncloud app. At least I could not find another solution.

### Environment data
owncloud server: 8.0.0.7
owncloud-app: 1.7.0

masensio:
Hi @smu 

thanks for your feedback and for details of this bug.

We will take into account in future bugfixing about media files, but I don't know when exactly. 
Any volunteer?

davivel:
Fixed, will be released in version 2.2.0

PR #1292, opened 2015-11-18
Internal music player does not stop playing when file is deleted

rishabh7m:
Bug fix for issue #906 

edited by @jabarros

- [x] Create 'internal_music_player' branch from master in android repository @jabarros
- [ ] Merge 'internal_music_player' branch into master
- [x] Download rishabh7m:master branch changes
- [ ] CR changes @jabarros (WIP)

### DIFF
--- a/src/com/owncloud/android/media/MediaService.java
+++ b/src/com/owncloud/android/media/MediaService.java
@@ -63,6 +63,7 @@ public class MediaService extends Service implements OnCompletionListener, OnPre
     /// Intent actions that we are prepared to handle
     public static final String ACTION_PLAY_FILE = MY_PACKAGE + ".action.PLAY_FILE";
     public static final String ACTION_STOP_ALL = MY_PACKAGE + ".action.STOP_ALL";
+    public static final String ACTION_STOP_FILE = MY_PACKAGE + ".action.STOP_FILE";
 
     /// Keys to add extras to the action
     public static final String EXTRA_FILE = MY_PACKAGE + ".extra.FILE";
@@ -244,9 +245,23 @@ public class MediaService extends Service implements OnCompletionListener, OnPre
             
         } else if (action.equals(ACTION_STOP_ALL)) {
             processStopRequest(true);
+        } else if(action.equals(ACTION_STOP_FILE)) {
+            processStopFileRequest(intent);
         }
 
         return START_NOT_STICKY; // don't want it to restart in case it's killed.
+    }
+
+    private void processStopFileRequest(Intent intent) {
+        OCFile file = intent.getExtras().getParcelable(EXTRA_FILE);
+        try {
+            if(file.equals(mFile)) {
+                processStopRequest(true);
+            }
+        }
+        catch (NullPointerException e) {
+            Log_OC.e(TAG, "NullPointerException while stopping on OCFile ", e);
+        }
     }
 
 
--- a/src/com/owncloud/android/ui/dialog/RemoveFileDialogFragment.java
+++ b/src/com/owncloud/android/ui/dialog/RemoveFileDialogFragment.java
@@ -22,20 +22,22 @@ package com.owncloud.android.ui.dialog;
 
 /**
  *  Dialog requiring confirmation before removing a given OCFile.  
- * 
+ *
  *  Triggers the removal according to the user response.
  */
 
 import android.app.Dialog;
+import android.content.Intent;
 import android.os.Bundle;
 
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
+import com.owncloud.android.media.MediaService;
 import com.owncloud.android.ui.activity.ComponentsGetter;
 import com.owncloud.android.ui.dialog.ConfirmationDialogFragment.ConfirmationDialogFragmentListener;
 
-public class RemoveFileDialogFragment extends ConfirmationDialogFragment 
+public class RemoveFileDialogFragment extends ConfirmationDialogFragment
 implements ConfirmationDialogFragmentListener {
 
     private OCFile mTargetFile;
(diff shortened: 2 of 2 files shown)


przybylski:
Hmm there are two scenarios which might be problematic here.
1) having two audio files A and B, play file A and delete file B remotely and locally, won't that cause to stop playing of file A ?
2) If external app plays sounds and you delete a file remotely and locally won't that cause to stop playing current sound?

przybylski:
Ok, second scenario is invalid because external apps cannot use owncloud MediaService but the first one stays valid.

rishabh7m:
@przybylski Hi, I have fixed the 1st issue with this commit.

src/com/owncloud/android/media/MediaService.java
```diff
@@ -244,11 +246,25 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
 
+    private void processStopFileRequest(Intent intent) {
+        OCFile file = intent.getExtras().getParcelable(EXTRA_FILE);
+        try {
+            if(file.equals(mFile)) {
+                processStopRequest(true);
+            }
+        }
+        catch (NullPointerException e) {
+            e.printStackTrace();
```
przybylski: I think instead of printing stack you should write to log @davivel @tobiasKaminsky  any thoughts?

przybylski: Also http://www.javapractices.com/topic/TopicAction.do?Id=129

AndyScherzinger: I second that, should definitely be a log.Call :+1:

rishabh7m:
@przybylski @AndyScherzinger I have added log statement in latest commit.

src/com/owncloud/android/media/MediaService.java
```diff
@@ -244,11 +246,26 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
+    private void processStopFileRequest(Intent intent) {
+        OCFile file = intent.getExtras().getParcelable(EXTRA_FILE);
+        String remotePath = intent.getExtras().getString(EXTRA_FILE_REMOTE_PATH);
+        try {
+            if(file.equals(mFile)) {
+                processStopRequest(true);
+            }
+        }
+        catch (NullPointerException e) {
+            Log_OC.e(TAG, "NullPointerException playing " + remotePath, e);
```
przybylski: Isn't it more of as "NullPointerException while stopping"  ?

przybylski:
Why do you need Remote path anyway? It can be get from OCFile. Beside if someone wouldn't provide OCFile as a parcelable it is highly likely that remote path also won't be included. IMO this additional argument is redundant.

src/com/owncloud/android/media/MediaService.java
rishabh7m: Yup, you are right. I will change that :+1:

rishabh7m:
I think if OCFile will be null then OCFile.getRemotePath() will again throw NullPointerException. That's why I included it.

przybylski:
Yeah, null OCFile won't even be included because in onConfirmation you are doing the check for null, It is a good practice that you are catching null ptr exception, but if REMOTE_PATH is just there for logging purposes then I don't find it very useful.

rishabh7m:
Will log statement -  Log_OC.e(TAG, "NullPointerException while stopping file "); will be fine ?

przybylski:
NullPointerException while stopping on OCFile is more descriptive :)

rishabh7m:
Okay, I will fix that :)

przybylski:
Great! :) but now REMOTE_PATH is completely redundant :)

rishabh7m:
Sorry, forgot about that. Removing it.

przybylski:
Code looks good to me.
Please squash your commits and you have green light from me :)

przybylski:
@tobiasKaminsky IMO this is ready for merge, I belive it should first go to beta right ?

przybylski:
@rishabh7m Thank you for your contribution and the will to cooperate with annoying reviewers :)

przybylski:
@rishabh7m One more thing, could you maybe provide some meaningful commit message :)

AndyScherzinger:
Hi @rishabh7m thanks for contributing and great to have you here. One question since you don't have a "collaborator" tag on your postings. Did you already sign the contributor agreement https://owncloud.org/contribute/agreement/ and send it to @karlitschek and @masensio ? Is this a necessary in order to get your pull requests merged to the code base by the core dev team.
And again, thanks for your contribution! Always great to have new people around pushing the app forward!

rishabh7m:
Hi @AndyScherzinger , I have not signed the agreement earlier. There is an employer field in the agreement but I am a student, so will it be fine to leave that empty?

rishabh7m:
@przybylski lol, no one was annoying :)

AndyScherzinger:
@rishabh7m yeah, just leave it empty then, should be fine :+1:

src/com/owncloud/android/ui/dialog/RemoveFileDialogFragment.java
```diff
@@ -93,11 +95,16 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
         ComponentsGetter cg = (ComponentsGetter)getActivity();
         FileDataStorageManager storageManager = cg.getStorageManager();
-        if (storageManager.getFileById(mTargetFile.getFileId()) != null) {
+        OCFile ocFile = storageManager.getFileById(mTargetFile.getFileId());
+        if (ocFile != null) {
             cg.getFileOperationsHelper().removeFile(mTargetFile, false);
+            Intent i = new Intent(getActivity(), MediaService.class);
+            i.putExtra(MediaService.EXTRA_FILE, ocFile);
+            i.setAction(MediaService.ACTION_STOP_FILE);
+            getActivity().startService(i);
```
malkomich: It would be great if you set this behaviour also for a local remove, since it makes sense that if you want to remove the file locally, you don't want to keep it in the buffer of the media service.

P.D: The local remove listener is currently set in the negative button (which listener mehotd is onCancel). Extract to a new method to reuse the MediaService call.

davivel:
Hi, @rishabh7m. Sorry for joining so late, and thanks for your patience.

This PR seems really fine, thank you very much. 

I think @malkomich is right, makes sense stopping the reproduction also when the deletion is only local (method ```onCancel``` in ```RemoveFileDialogFragment```. Could you create a private method with the code stoping the media service and call when any of the deletions is confirmed? 

I will help you then with the rebase on master.

jabarros:
Hi everybody,
We have just opened a new PR, #1814 in order to be able to do the rebase and apply the CR changes proposed, so let me close this one.
If you want to do any comment, please, add it into the new PR thread.

PR #1296, opened 2015-11-19
Don't allow to remove passcode without entering it. Don't allow to se…

przybylski:
…t incomplete passcode,

### DIFF
--- a/src/com/owncloud/android/authentication/PassCodeManager.java
+++ b/src/com/owncloud/android/authentication/PassCodeManager.java
@@ -74,7 +74,7 @@ public class PassCodeManager {
                 ){
 
             Intent i = new Intent(MainApp.getAppContext(), PassCodeActivity.class);
-            i.setAction(PassCodeActivity.ACTION_REQUEST);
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
Merged into beta

davivel:
Code is fine, though I was pretty confused by the change of names  for PassCodeActivity.ACTION_ constants.

@przybylski , would you change PassCodeActivity.ACTION_REQUEST_WITH_RESULT to PassCodeActivity.ACTION_SET_WITH_RESULT ? 

Other than that, this is :+1: to be tested.

przybylski:
But calling with ACTION_REQUEST_WITH_RESULT doesn't set the passcode in PasscodeActivity. I just return the passcode from the user to caller and let the caller decide what to do. So changing REQUEST to SET is confusing

davivel:
@przybylski , OK, see your point. Let's keep it as is.

davivel:
Ready to go.

Thanks a lot, @przybylski .

PR #1293, opened 2015-11-19
External SD2

przybylski:
Implement mechanism with @tobiasKaminsky to allow users to specify their owncloud storage path.
Together with migration mechanism.
https://github.com/owncloud/android/issues/10

PR #1302, opened 2015-11-22
Correctly handle empty directory case when browsing local files

przybylski:
Merge only possible after #1168 and #1293


tobiasKaminsky:
Isn't this something that can be committed directly to your existing PR "externalSD2"?

przybylski:
I could, but it isn't directly related to externalSD2. It looks like this bug was there before, it revealed itself after putting together 'storage path' and 'dont duplicate files' features.

tobiasKaminsky:
Ok. I have updated the initial description.

przybylski:
Not needed anymore

PR #1303, opened 2015-11-22
Remove radios for copy or move in while searching for target dir in migration wizard

przybylski:
This should also be cherry-picked after merging 'do not duplicate files' (#1168) and 'change storage' (#1293) to master


tobiasKaminsky:
I have updated the description and linked both PRs to it. 
So this PR mentioned in the related PRs.

PR #1321, opened 2015-12-02
Force ant and android studio to use Java7

przybylski:


tobiasKaminsky:
This should also be used for travis, or?

przybylski:
travis is using ant so it will work there.
Branch whatsNew was failing at first because of old Java version with lack of diamond operator support

tobiasKaminsky:
In https://github.com/owncloud/android/blob/beta/.travis.yml
ant clean -Djava.source=7 -Djava.target=7
ant debug -Djava.source=7 -Djava.target=7
is used.
I was referring to this.
Maybe it is the same as your code?

przybylski:
Yeah, this is the same but put in a properties file, IMO this is better place than putting it to .travis.yml

tobiasKaminsky:
As long as it is working :)

AndyScherzinger:
I did that back then, but @przybylski's solution is better placed :)

AndyScherzinger:
@tobiasKaminsky this looks good to me and if it also looks good to you I would flag it ready for review :)

przybylski:
any objection on merging this?

tobiasKaminsky:
Not from me.
What about @davivel @masensio?

tobiasKaminsky:
@przybylski I do not know if we (community contributor) are allowed to do merges...

przybylski:
The speed of adoption of new code is killing me :P

davivel:
Who merged this?

davivel:
> Branch whatsNew was failing at first because of old Java version with lack of diamond operator support

So, is this the only reason to include this change?

What other implications will have it in builds?

davivel:
> any objection on merging this?

In this repository nothing should be merged without QA approval, or without a explicit statement from me, @masensio or @rperezb that doesn't need QA.

davivel:
Ready for discussion again in #1378

Issue #971, opened 2015-05-05
CHANGELOG.md hasn't been updated for 1.9


JohnVeness:
File CHANGELOG.md hasn't been changed when 1.7.1 was released, and still refers to 1.7.0. This matters because users who download Owncloud from F-Droid (https://f-droid.org/repository/browse/?fdid=com.owncloud.android) are linked to https://github.com/owncloud/android/blob/HEAD/CHANGELOG.md which is currently out of date.

masensio:
Hi @JohnVeness,

sorry for the delay.
Changelog is updated now.
https://github.com/owncloud/android/blob/HEAD/CHANGELOG.md

JohnVeness:
Same problem with 1.9.0

tobiasKaminsky:
cc @masensio

masensio:
Ups, sorry I forgot to update the changelog here

masensio:
Changelog updated again :smile: 
You can see at:
https://github.com/owncloud/android/blob/HEAD/CHANGELOG.md

PR #1340, opened 2015-12-11
Update Changelog for 1.9 version

masensio:


davivel:
Ups, we missed this in the release branch. Good catch, @masensio 

:+1:

masensio:
I forgot, but the issue https://github.com/owncloud/android/issues/971 reminded me

PR #1359, opened 2015-12-18
#1338: Correctly use delayUntil property of SyncResult

### LINKED ISSUE #1338: ownCloud 1.9.0 constantly syncing
Since 1.9.0 the Android OC client is constantly syncing and keeping the device awake when the auto-sync option under "Accounts" is checked. I've got it installed on two 5.1.1 devices and this behaviour occurs on both. Once it has finished scanning the whole folder structure it starts again after 30-60 seconds, keeping the device awake (it never goes do deep-sleep), draining the battery and generating permanent traffic.
This didn't occur with 1.8.0. Because of that I had to disable auto sync. 
Does anyone have this problem as well?

przybylski:
Sets a minimum time of three hours between automatic synchronizations. 

Should be enough to fix #1338.

### DIFF
--- a/src/com/owncloud/android/syncadapter/FileSyncAdapter.java
+++ b/src/com/owncloud/android/syncadapter/FileSyncAdapter.java
@@ -154,7 +154,7 @@ public class FileSyncAdapter extends AbstractOwnCloudSyncAdapter {
         mForgottenLocalFiles = new HashMap<String, String>();
         mSyncResult = syncResult;
         mSyncResult.fullSyncRequested = false;
-        mSyncResult.delayUntil = 60*60*24; // avoid too many automatic synchronizations
+        mSyncResult.delayUntil = (System.currentTimeMillis()/1000) + 3*60*60; // avoid too many automatic synchronizations
 
         this.setAccount(account);
         this.setContentProviderClient(providerClient);


davivel:
Code is:+1: .

davivel:
QA needed, should be easy. cc @Dianafg76 , @jesmrec

AndyScherzinger:
Awesome @przybylski :+1:

tobiasKaminsky:
Just as I am curious: this was added 15.09.11 according to git.
So why is this now such a big problem?

przybylski:
AFAIR it was always a problem

tobiasKaminsky:
ok. I just thought as version 1.9 is explicitly mentioned.

davivel:
> AFAIR it was always a problem
Yes.

> ok. I just thought as version 1.9 is explicitly mentioned.
Seems that for some people the problem is worse with 1.9, probably by some of the changes in the refresh of folders introduced with the folder synchronization. Other improvements can be done.

But this fix itself should have a great impact.

Dianafg76:
@davivel I tested this issue with Android 4 and Android 6 and It sync every 3hrs. So is working OK

davivel:
So, going in.

PR #1360, opened 2015-12-18
[Beta] Fix npe in instant uploads

### LINKED ISSUE #1341: "Unfortunately, Owncloud has stopped" after photo when AutoUpload is enabled
After a photo is taken, I get the above behavior and the photo never gets transferred to the server.  This occurs over cellular, as well as wifi.  It even happens when wifi-only is enabled and the phone is not connected to wifi.  This is only occurring on our 50 or so Verizon Kyocera Brigadier phones.   Our other phones (mostly Motorolas) work fine.  If there is anything I can do to help solve this, please let me know.

domneu

davivel:
Should fix #1341


davivel:
cc @tobiasKaminsky

PR #1361, opened 2015-12-18
[Beta] Fix NPE in instant uploads

davivel:
Prevents  #1341


davivel:
cc @tobiasKaminsky

tobiasKaminsky:
@davivel 
Is this only Beta related? If not, this PR should be against master and I will include it also in beta.

tobiasKaminsky:
Included in beta.

davivel:
Sorry for the lack of response.

The idea was that @domneu had a chance to test the bug fix sooner, since to merge into beta QA needs to approve first. Besides, we are not able to reproduce the bug with our devices.

tobiasKaminsky:
ok. no problem.
I already included in in beta.

Issue #296, opened 2013-11-02
Enhancement: Use Android 4.4 Document Provider API


pulser:
The document provider API in Android 4.4 (Kitkat) offers a framework for storage providers, such as OwnCloud, Google Docs, Box etc. 

http://developer.android.com/about/versions/kitkat.html#44-storage-access

This looks like a useful way to permit users to access their OwnCloud files quickly, and make use of OwnCloud as their default storage location, throughout the OS, and permits the provider to be written once, but used by any client application that uses this API.

Implementation documentation is available at http://developer.android.com/reference/android/provider/DocumentsProvider.html

strugee:
@davivel did you ever look into this? I share @alerque's concerns.

davivel:
@strugee , sure we looked into it. Unfortunately, looking does not produce any result similar to writing code.

This will be done when we can book the time for it. There are several reasons why it's very interesting to have it, but we are a small team an have to address many top priorities.

Code contributions in this and related issues are very welcome.

strugee:
@davivel should have been clearer. By 'it', I meant this issue's priority, not the issue itself. Sorry for the confusion (and thanks, as always, for your hard work).

haseebeqx:
ok i am working on it.  but when i used the existing content provider it shows permission denied since the `android:exported` is false and `android:grantUriPermissions` is not set. what should i do now. set `android:exported` to true or `granturipermission` or any other options ?.

 currently i managed to implement this feature with exported content provider and it is working fine

dschuermann:
Maybe it helps that there is a fully working example for Document Providers here: https://android.googlesource.com/platform/development/+/master/samples/Vault/

haseebeqx:
i already have a working model .

but the problem is in kitkat the document provider works from another app called documents inside system. so there are issues with current content provider and localbroadcasts. i can not use the current  file access provider to look at data since it is not exported and android:grantUriPermissions not set.

karliwalti:
May I additionally support the integration of this. Microsoft office is now supporting "Other cloud storage" and all major service providers are supported (tested with Drive, Box, Dropbox, and OneDrive ). To work on the go from the cloud it would be essential

AndyScherzinger:
Setting exported and android:grantUriPermissions to be able to implement this seems fine to me, what do you think @davitol @masensio @tobiasKaminsky ?

davitol:
@davivel :santa:

codinguser:
[...]

PR #1368, opened 2015-12-19
#296: Add documents provider

przybylski:
--------- edited by @jabarros
TASKs:
- [X] [QA] Create 'document_provider' branch in QA Repo @jesmrec
- [X] [QA] Merge 'document_provider' branch in QA Repo
- [X] [QA] Create test plan @jesmrec
- [X] [QA] Validate test plan @jesmrec

----------

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -129,16 +129,28 @@
             android:name=".providers.FileContentProvider"
             android:authorities="@string/authority"
             android:enabled="true"
-            android:exported="false"
+            android:exported="true"
             android:label="@string/sync_string_files"
             android:syncable="true" />
 
         <provider
             android:name=".providers.UsersAndGroupsSearchProvider"
-            android:authorities="com.owncloud.android.providers.UsersAndGroupsSearchProvider"
+            android:authorities=".providers.UsersAndGroupsSearchProvider"
             android:enabled="true"
             android:exported="false"
             android:label="@string/search_users_and_groups_hint" />
+
+        <provider
+            android:authorities="@string/document_provider_authority"
+            android:name=".providers.DocumentsStorageProvider"
+            android:exported="true"
+            android:grantUriPermissions="true"
+            android:permission="android.permission.MANAGE_DOCUMENTS"
+            android:enabled="@bool/atLeastKitKat">
+            <intent-filter>
+                <action android:name="android.content.action.DOCUMENTS_PROVIDER" />
+            </intent-filter>
+        </provider>
 
         <activity
             android:name=".authentication.AuthenticatorActivity"
--- a/src/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -1,0 +1,208 @@
+/**
+ *   ownCloud Android client application
+ *
+ *   @author Bartosz Przybylski
+ *   Copyright (C) 2015  Bartosz Przybylski
+ *   Copyright (C) 2015 ownCloud Inc.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License version 2,
+ *   as published by the Free Software Foundation.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.owncloud.android.providers;
+
+import android.accounts.Account;
+import android.annotation.TargetApi;
+import android.content.ContentResolver;
+import android.content.Intent;
+import android.content.res.AssetFileDescriptor;
+import android.database.Cursor;
+import android.graphics.Point;
+import android.os.Build;
+import android.os.CancellationSignal;
+import android.os.ParcelFileDescriptor;
+import android.provider.DocumentsProvider;
+
+import com.owncloud.android.authentication.AccountUtils;
+import com.owncloud.android.datamodel.FileDataStorageManager;
+import com.owncloud.android.datamodel.OCFile;
+import com.owncloud.android.files.services.FileDownloader;
+import com.owncloud.android.providers.cursors.FileCursor;
+import com.owncloud.android.providers.cursors.RootCursor;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Vector;
+
+@TargetApi(Build.VERSION_CODES.KITKAT)
+public class DocumentsStorageProvider extends DocumentsProvider {
+
+    private FileDataStorageManager mCurrentStorageManager = null;
+    private static Map<Long, FileDataStorageManager> mRootIdToStorageManager;
+
+    @Override
+    public Cursor queryRoots(String[] projection) throws FileNotFoundException {
+        initiateStorageMap();
+
+        final RootCursor result = new RootCursor(projection);
+
+        for (Account account : AccountUtils.getAccounts(getContext()))
+            result.addRoot(account, getContext());
+
+        return result;
+    }
+
+    @Override
+    public Cursor queryDocument(String documentId, String[] projection) throws FileNotFoundException {
+        final long docId = Long.parseLong(documentId);
+        updateCurrentStorageManagerIfNeeded(docId);
+
+        final FileCursor result = new FileCursor(projection);
+        result.addFile(mCurrentStorageManager.getFileById(docId));
+
+        return result;
+    }
+
+    @Override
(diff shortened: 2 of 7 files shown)


tobiasKaminsky:
Have you tested it?
Then I can do a code review and add it in beta?

przybylski:
works fine with multiple accounts, on nexus 4

tobiasKaminsky:
Maybe this would be a good use case to show it in the "what's new" slide show?

tobiasKaminsky:
Included in beta.
Please test @all

przybylski:
could be, but 'whats new' should be also included, this requires some addtional work from graphics people

tobiasKaminsky:
Yeah, then we should enforce this...
Who is responsible for the artwork?

przybylski:
No idea :D @AndyScherzinger said that he would take a look on images and texts

tobiasKaminsky:
@przybylski 
https://github.com/owncloud/android/issues/296#issuecomment-175223050 said:
The beta client seems to support opening via SAF, but not saving.

przybylski:
I know, for now I want to provide opening, saving will come in other PR

tobiasKaminsky:
@przybylski the current version shows only downloaded files, right?
It would be cool to have every file on the server and download them if necessary?
Or is it possible to show the download state to the user?

przybylski:
Last time I tested it, it was working just as you described it. Showing all the files and download then if necessary

tobiasKaminsky:
Maybe I have messed something up in beta.
Can you try it there?

przybylski:
The code looks fine on beta branch. Is it not downloading the file?
Currently my android device is far away from me so I cannot test it on the real data :/
The one confusing thing is that there will be no progress indicator on the activity but on the notification bar. But this is something that system is forcing so there si nothing we can do about it.

tobiasKaminsky:
The files / folders that are not already downloaded are not even shown, so I cannot click them.

przybylski:
Ok, I can see the problem,
when you were rebasing for beta branch you updated getFolderContent call to include true flag (only on device), it should be false on beta.

tobiasKaminsky:
Thanks :)
Now it is working! I will add it to beta soon.

Is it possible to show the download state?

przybylski:
No, DocumentProvider doesn't support progress indicator. The only download state is available in a notification bar, and it is normally invoked when ownCloud is downloading file..

tobiasKaminsky:
Ah, no, I meant the state of a file: downloaded on device / not downloaded.
As it might be interesting to see this in order to avoid triggering a big download when selecting a file.

przybylski:
There is no built in mechanism in DocumentsProvider to indicate this :/

AndroidManifest.xml
```diff
@@ -112,17 +112,29 @@
             android:name=".providers.FileContentProvider"
             android:authorities="@string/authority"
             android:enabled="true"
-            android:exported="false"
+            android:exported="true"
```
malkomich: Is it necessary to make this existing Provider public?

src/com/owncloud/android/providers/cursors/FileCursor.java
```diff
@@ -0,0 +1,61 @@
...
+        final int iconRes = MimetypeIconUtil.getFileTypeIconId(file.getMimetype(), file.getFileName());
+        final String mimeType = file.isFolder() ? Document.MIME_TYPE_DIR : file.getMimetype();
+        final String imagePath = file.isImage() && file.isDown() ? file.getStoragePath() : null;
+        int flags = imagePath != null ? Document.FLAG_SUPPORTS_THUMBNAIL : 0;
+
+        newRow().add(Document.COLUMN_DOCUMENT_ID, Long.toString(file.getFileId()))
+                .add(Document.COLUMN_DISPLAY_NAME, file.getFileName())
+                .add(Document.COLUMN_LAST_MODIFIED, file.getModificationTimestamp())
+                .add(Document.COLUMN_SIZE, file.getFileLength())
+                .add(Document.COLUMN_FLAGS, flags)
```
malkomich: Setting the "FLAG_SUPPORTS_THUMBNAIL" flag does not imply implement both "buildSearchDocumentsUri" and
"querySearchDocuments" methods?

http://developer.android.com/intl/es/reference/android/provider/DocumentsContract.Root.html#FLAG_SUPPORTS_SEARCH

I'm not sure about it but maybe both are expected (not only one) when that flag is set. Please correct me if I'm wrong ;)

src/com/owncloud/android/providers/cursors/RootCursor.java
```diff
@@ -0,0 +1,62 @@
...
+    public void addRoot(Account account, Context context) {
+        final FileDataStorageManager manager =
+                new FileDataStorageManager(account, context.getContentResolver());
+        final OCFile mainDir = manager.getFileByPath("/");
+        newRow().add(Root.COLUMN_ROOT_ID, account.name)
+                .add(Root.COLUMN_DOCUMENT_ID, mainDir.getFileId())
+                .add(Root.COLUMN_SUMMARY, account.name)
+                .add(Root.COLUMN_TITLE, context.getString(R.string.app_name))
+                .add(Root.COLUMN_ICON, R.drawable.icon)
+                .add(Root.COLUMN_FLAGS, Root.FLAG_SUPPORTS_SEARCH);
```
malkomich: Same as with the thumbnails flag. Method "getDocumentThumbnail" left, not sure if is expected for the Provider to have it implemented.

http://developer.android.com/intl/es/reference/android/provider/DocumentsContract.Document.html#FLAG_SUPPORTS_THUMBNAIL

AndroidManifest.xml
przybylski: yes, documents provider is queried by external application, and at that point its outside owncloud permission so exported content provider is required.

src/com/owncloud/android/providers/cursors/FileCursor.java
przybylski: No, you need to tell the caller that it is safe to call those methods.
In default implementation exception is thrown which is crashing the access to resources

src/com/owncloud/android/providers/cursors/RootCursor.java
przybylski: similar as above

src/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -0,0 +1,208 @@
...
+            result.addFile(f);
+
+        return result;
+    }
+
+    private void updateCurrentStorageManagerIfNeeded(long docId) {
+        if (mCurrentStorageManager == null ||
+                (mRootIdToStorageManager.containsKey(docId) &&
+                        mCurrentStorageManager != mRootIdToStorageManager.get(docId))) {
+            mCurrentStorageManager = mRootIdToStorageManager.get(docId);
```
malkomich: It tries to get a map element in an unsafe mode, because the map could be not instantiate and results in a NullPointerException.

src/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -0,0 +1,208 @@
...
+        return result;
+    }
+
+    @Override
+    public Cursor queryDocument(String documentId, String[] projection) throws FileNotFoundException {
+        final long docId = Long.parseLong(documentId);
+        updateCurrentStorageManagerIfNeeded(docId);
+
+        final FileCursor result = new FileCursor(projection);
+        result.addFile(mCurrentStorageManager.getFileById(docId));
```
malkomich: Here it is been generated sometimes a NullPointer exception, because `mCurrentStorageManager` is found in the StorageManager Map with a different id of the current storage managers.

davivel:
Hi, @przybylski . 

Should we understand you don't want to contribute this code to ownCloud anymore?

Thanks in advance.

przybylski:
It's not that I don't want to, but the fact that such simple pr was opened for over half a year got me discouraged.

src/com/owncloud/android/providers/DocumentsStorageProvider.java
przybylski: The map is always instantiated because of how DocumentsProvider is calling methods in this class. It is guaranteed that `queryRoots` will always be called before anything else and it is the place where the map is instantiated.

src/com/owncloud/android/providers/DocumentsStorageProvider.java
przybylski: While testing I never ran into such scenario

davivel:
I can understand that. In any case, closing it doesn't seem a way to get it merged faster, does it? 

Please, consider reopening it. Your contributions to this project always were more than important.

przybylski:
It didn't got it faster, BUT it got your attention so it kinda worked ;)

jabarros:
@jesmrec this PR is prepared to be tested :)
@przybylski maybe a rebase on master is needed in order to have the branch updated. Could you do that?

jesmrec:
👍  @jabarros  a test plan will be designed asap.

jabarros:
@jesmrec, rebase on master done

jesmrec:
Approved, nice work :)

@przybylski @jabarros A new rebase is needed, please.

jabarros:
Rebase done. Wating Travis.

Issue #1334, opened 2015-12-07
Sorting by recently modified should mix files and folders


jancborchardt:
Currently when sorting by recently modified first shows folders and then files. Rather we should mix them, whichever item was modified most recently. This is the same we do in the web interface.

Because right now the sorting is very strange for folders with a lot of items and folders – you need to scroll down a lot to get to the relevant items.

tobiasKaminsky:
But this is the way other file browser do this, oder?
(I can only verify this for dolphin [KDE file browser])

jancborchardt:
Desktop file managers might do that, but on mobile it’s impractical. :)

tobiasKaminsky:
Ok.
Then we do it :+1:

tobiasKaminsky:
Done in #1369

davivel:
Since there is a PR for this, let me remove the 'Junior Job' label.

abelgardep:
Done, will be included in 2.14

PR #1369, opened 2015-12-20
sort by date without obeying folder/files

tobiasKaminsky:
ref: #1334


tobiasKaminsky:
sort

tobiasKaminsky:
@AndyScherzinger @davivel @przybylski please review

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -184,15 +184,7 @@ public static RemoteFile fillRemoteFile(OCFile ocFile){
...
-                if (o1.isFolder() && o2.isFolder()) {
-                    Long obj1 = o1.getModificationTimestamp();
-                    return val * obj1.compareTo(o2.getModificationTimestamp());
-                }
-                else if (o1.isFolder()) {
-                    return -1;
-                } else if (o2.isFolder()) {
-                    return 1;
-                } else if (o1.getModificationTimestamp() == 0 || o2.getModificationTimestamp() == 0){
+                if (o1.getModificationTimestamp() == 0 || o2.getModificationTimestamp() == 0){
```
przybylski: I don't really like that if there is no timestamp we artificially making it equal with everything. This leads to unexpected behavior when we do have multiple zeros in collection.
I would rather not take any special care of this and boil down compare method to something like this:
```java
public int compare(OCFile o1, OCFile o2) {
    return val * (o1.getModificationTimestamp() - o2.getModificationTimestamp());
}
```

tobiasKaminsky: Yeah, but this is already the current behaviour.
So, we should discuss this in another issue?
(Although I do not know how often this problem might occur as the modification timestamp should always be set, or?

przybylski: In that case its ok for me to be merged

abelgardep:
To be done in #2710

Issue #788, opened 2014-12-08
Media player: Movie stop when turning device


LukeOwlclaw:
### Expected behaviour

When playing a movie and turning the device, movie (with sound) should continue playing nicely. (Just like VLC does it and it is done by oC app when playing audio.)
### Actual behaviour

Screen goes black and audio stops for a moment before it is resumed. Also the progress bar goes back to 0 for a moment.
### Steps to reproduce
1. Play movie (mp4 in my case)
2. Turn device (turning must be enabled of course)
### Environment data

Android version: 4.2.2.
Device model: Nexus 4
ownCloud app version: 1.6.2 (from master branch)

rishabh7m:
@LukeOwncloud @jabarros can I start on this if anyone is not working on this issue?

PR #1370, opened 2015-12-20
Media player: Movie stop when turning device

rishabh7m:
Fix for bug #788


przybylski:
This is not a correct way to handle this. Fixing this requires more sophisticated solution deep inside FileActivity

davivel:
As @przybylski , this is not the way. Including ```android:configChanges``` in ```AndroidManifest.xml``` is not enough to handle correctly configuration changes, and will have a lot of side effects leading to wrong behaviours in rotations.

What you try to do needs overriding also the method ```onConfigurationChanged(...)``` in the affected activities (see http://developer.android.com/intl/es/guide/topics/resources/runtime-changes.html).

Even in that case, that could not be the best strategy for our app, since there is already a lot of code in place to handle state in rotations taking into account that ```android:configChanges``` is **not** included.

Sorry, but this can't go in.

Issue #1358, opened 2015-12-17
Multiple files upload from Google Photos application not working


rawbertp:
Steps to reproduce:
1. Go to any folder in the ownCloud app
2. Click on Upload
3. Select "Content from other apps"
4. Select Google "Photos" app
5. Select multiple (e.g. 3) photos/files
6. Click "Done"

Actual outcome:
- Only one file is uploaded (instead of all 3)

Expected:
- All selected files get uploaded

rishabh7m:
@AndyScherzinger Hi,  I am working on this if anyone has not started yet.

AndyScherzinger:
Awesome! Go ahead, as far as I know no one else has started on this, looking forward to your fix :)

malkomich:
Uploads from external apps are currently being fixed by @jabarros in PR https://github.com/owncloud/android/pull/1643.

So maybe, when it gets approved by QA, we should close this issue and PR https://github.com/owncloud/android/pull/1373 too, since many things has changed in uploads code for the last release.

dpeger:
This actually still seems to be a thing. I this issue still WiP for more than 2 years?!

davigonz:
Hi @dpeger , I've just tested this feature with ownCloud 2.6.0 in a Samsung Galaxy S7 Edge with Android 7 and is working fine, could you give it a try? If it works for you, I will close this since is already implemented

dpeger:
@davigonz Oh, I kind of misread the issue, I thought it was about uploading images using the ownCloud web frontend, not the ownCloud app.

I'll delete my comment...

davigonz:
No problem @dpeger. @rawbertp , could you test it with 2.6.0 ownCloud app and tell us whether is still not working for you?

davigonz:
Closing this... if it happens again, please reopen it or open a new issue

PR #1373, opened 2015-12-20
Multiple files upload from Google Photos application not working

rishabh7m:
Fix for #1358


tobiasKaminsky:
//getClipData is only supported on api level 16+, Jelly Bean
is said in comment.
@davivel which is our minimum required api?

Besides that: There is already a for-loop in the lines above (619ff which can be used, if the api level is not the problem...?)

rishabh7m:
I think data.getData() will not be null if we select content from other apps. If we cancel the selection then resultCode will not be RESULT_OK. I have tested by not selecting anything, one selection and multiple selections from the other app.

davivel:
@rishabh7m , what Android version has the device you are using to test? As @tobiasKaminsky said, multiple selection is not possible through this workflow for Android lower than 4.1.

jesmrec:
Fixed in 2.11 with the native implementation

PR #1378, opened 2015-12-22
Force ant and android studio to use Java7

davivel:
Ready again for discussion

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,10 +53,13 @@ android {
         release.setRoot('build-types/release')
     }
 
-    android {
-        lintOptions {
-            abortOnError false
-        }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_7
+        targetCompatibility JavaVersion.VERSION_1_7
+    }
+
+    lintOptions {
+        abortOnError false
     }
 
     productFlavors {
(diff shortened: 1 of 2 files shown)


davivel:
@przybylski , do you what implications will have this change in terms of compatibility?

przybylski:
All enhancements introduced in Java 7 are backward compatible, and we are not using stuff which API has changed

tobiasKaminsky:
Are there android version out there that run with java 1.5? Then we cannot do the change as the byte code is not backward compatible...

przybylski:
Android is not using Java byte code. Java is just a medium which in the end is ran on dalvik or art

davivel:
Seen. The only problem I could fine, as described in http://tools.android.com/recent/androidstudio032released, is in using try-with-resources (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) . We shouldn't, since requires minSdkVersion >= 19, and we support >=14. 

Seems and assumable risk. Maybe we could specify a finer Lint configuration to fail with it, but I'm not sure the effort to do so is worth it.

Besides, this PR doesn't require QA, since affects just build options.

:+1:

davivel:
Please, notice I'm not approving my own stuff, real author is @przybylski . Check the commits.

PR #1438, opened 2016-01-18
Set username and password from Server URL if URL contains http basic auth

mnutt:
If user enters a server URL in the form `https://myuser:mypass@example.com`, on field blur, extract `myuser` and `mypass` and automatically fill them in the connection form.

### DIFF
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -790,12 +790,32 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 normalizeUrl(mHostUrlInput.getText().toString(), mServerInfo.mIsSslConn))) {
             // check server again only if the user changed something in the field
             checkOcServer();
+            extractCredentialsFromUrl();
         } else {
             mOkButton.setEnabled(mServerIsValid);
             showRefreshButton(!mServerIsValid);
         }
     }
 
+    private void extractCredentialsFromUrl() {
+        String uriInput = mHostUrlInput.getText().toString().trim();
+
+        if(uriInput.length() != 0) {
+            Uri uri = Uri.parse(uriInput);
+
+            String userInfo = uri.getUserInfo();
+            if(userInfo != null) {
+                String username = userInfo.split(":", 2)[0];
+                if(username.length() != 0) {
+                    mUsernameInput.setText(username);
+                }
+                String password = userInfo.split(":", 2)[1];
+                if(password.length() != 0) {
+                    mPasswordInput.setText(password);
+                }
+            }
+        }
+    }
 
     private void checkOcServer() {
         String uri = mHostUrlInput.getText().toString().trim();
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Code seems fine to me, but why should an user type such an url and not use the three input fields as they are described? 
I mean why should be catch this improper use of the url field?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -790,12 +790,32 @@ private void onUrlInputFocusLost() {
...
+            Uri uri = Uri.parse(uriInput);
+
+            String userInfo = uri.getUserInfo();
+            if(userInfo != null) {
+                String username = userInfo.split(":", 2)[0];
+                if(username.length() != 0) {
+                    mUsernameInput.setText(username);
+                }
+                String password = userInfo.split(":", 2)[1];
+                if(password.length() != 0) {
```
przybylski: NPE for: http://username@example.com

przybylski: Not NPE, but ArrayIndexOutOfBounds

mnutt:
It's a bit easier for me to copy-and-paste a single URL from desktop to phone than to manually copy and paste the URL, username, and password separately, switching back and forth each time. It's not really meant to replace the three input fields, just to act as a shortcut for people who combine them.

It's also a part of the HTTP URI spec and other clients (web browsers, command line, etc) typically tend to support it.

mnutt:
I fixed the ArrayIndexOutOfBounds by a slightly safer check for userInfo.

przybylski:
Seems fine, I re-wrote extract function, let me know what do you think in terms of readability.

```java
String urlText = mHostUrlInput.getText();
if (TextUtils.isEmpty(urlText)) return;

String userInfo = Uri.parse(urlText.trim()).getUserInfo();

if (TextUtils.isEmpty(userInfo)) return;

String[] infoParts = userInfo.split(":", 2);

if (infoParts.length > 0 && !infoParts[0].isEmpty()) {
	mUsernameInput.setText(userInfoParts[0]);
	if (infoParts.length > 1 && !infoParts[1].isEmpty())
		mPasswordInput.setText(userInfoParts[1]);
}

```

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -790,12 +790,31 @@ private void onUrlInputFocusLost() {
...
         } else {
             mOkButton.setEnabled(mServerIsValid);
             showRefreshButton(!mServerIsValid);
         }
     }
 
+    private void extractCredentialsFromUrl() {
+        String uriInput = mHostUrlInput.getText().toString().trim();
+
+        if(uriInput.length() != 0) {
```
przybylski: we hava a convention to put spaces after `if`s

rperezb:
For tracking purposes: https://github.com/owncloud/ios/pull/515

rperezb:
@mnutt it would be great if you may check whether the reported bugs on the iOS app works are not happening https://github.com/owncloud/ios/pull/515
thx!

PR #1453, opened 2016-01-27
Removed automatic back-browse; error codes from server do not disting…

davivel:
…uish between excess of permissions, removed share or removed file

STEPS TO REPRODUCE:
- As user1, share a folder with user2; then edit permissions on the share to disable 'can create'
- As user2, reshare the filder with user3; the edit permissions on the share to enable 'can create'

CURRENT BEHAVIOUR (before the fix)
- Error is shown **and the app browses back to the first view of sharing activity** automatically

EXPECTED BEHAVIOUR (with the fix)
- Error is shown and no automatic browse is done.


davivel:
cc @rperezb @jesmrec for QA; handle as a HOTFIX, please.

cc @masensio for code review.

masensio:
Code reviewed.
:+1:

rperezb:
:+1:

davivel:
Then in. Thanks.

Issue #1325, opened 2015-12-03
Clearing application data removes passcode


quincunx:
If the application data is cleared via app info (Settings > Apps > ownCloud > Storage > CLEAR DATA), then the PIN is removed, too. But as the account info is kept, starting the app again will give one access to the account without any PIN code.

Tested on Nexus 5 with Android 6.0 and ownCloud Android App 1.8.0.

przybylski:
Clearing data is a system setting and application data (SharedPreferences) cannot hide from it, since it is controlled by system. This is a security thread which doesn't have a simple way of handling.

tobiasKaminsky:
The only way is to also remove the account details when doing "clear data".

przybylski:
Since 'clear data' is a system feature I guess app cannot remove account (because it doesn't know when data was cleared).
The way out is instead of putting pin into shared preferences create a dummy file which would be encrypted by this PIN, and if decryption succeeds then we can unlock app.

tobiasKaminsky:
"clear data" also deletes the storage in /sdcard/owncloud/account-name or?
If this is true, then we would have an indicator: account folder is missing for existing account entry and then delete also the account entry.

przybylski:
No it is not removing data folder

davivel:
cc @rperezb @MTRichards @cmonteroluque 

To solve it we'll need:
1. Move all the information of the Account from current storage (AccountManager) to app database; (possible exception: account password; or maybe not)
2. Update all the code dependent of 1; not so bad as seems; specially important: code validating account and redirecting to log-in Activity when invalid.
3. Grant that the information of the accounts is correctly migrated for the existing users when the app is upgraded.

This way, when "clear data" is clicked not only the PIN will be removed, but also the account information, and the app will request log-in.

przybylski:
Please don't move Account information from AccountManager to app DB, the whole point was to have oc integrated with system, so android could schedule synchronizations of files (those "keep in sync" as well as file list), or maybe those functionalities were owned by some other app component?

przybylski:
And BTW why not use http://developer.android.com/reference/android/accounts/AccountManager.html#KEY_USERDATA to store PIN information?
This could lockup single account instead of whole app (but locking entire app is also possible).

davivel:
The point is keep the minimum information available in the AccountManager, this is:
  - accountName
  - password (maybe not)

And use the accountName to access the rest of the Account data in the app database.

[...]

PR #1461, opened 2016-02-02
Manage space activity

masensio:


src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+            appPrefsEditor.putString("PrefPinCode3", passCodeDigits[2]);
+            appPrefsEditor.putString("PrefPinCode4", passCodeDigits[3]);
+        }
+
+        appPrefsEditor.putBoolean("set_pincode", passCodeEnable);
+        appPrefsEditor.commit();
+
+
+        finish();
+        System.exit(0);
```
AndyScherzinger: Hi @masensio is it necessary to call System.exit(0) after you already called finish()?

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+
+    }
+
+    public void clearApplicationData() {
+        File cache = getCacheDir();
```
AndyScherzinger: cache is only referenced in the next line, so this could be shortened a bit :)

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+        File appDir = new File(cache.getParent());
+        if (appDir.exists()) {
+            String[] children = appDir.list();
+            for (String s : children) {
+                if (!s.equals("lib")) {
```
AndyScherzinger: should be implemented the other way round to be null safe. ```!"lib".equals(s)``` and I would also declare "lib" as a static final constant within the class.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+    private void clearData() {
+        // Save passcode from Share preferences
+        SharedPreferences appPrefs = PreferenceManager
+                .getDefaultSharedPreferences(getApplicationContext());
+
+        boolean passCodeEnable = appPrefs.getBoolean("set_pincode", false);
+
+        String passCodeDigits[] = new String[4];
+        if (passCodeEnable) {
+            passCodeDigits[0] = appPrefs.getString("PrefPinCode1", null);
```
AndyScherzinger: ```"PrefPinCode*"```should also be defined as static final strings in a central place which is also used in any other place these 4 Pin values are accessed/written within the application.

przybylski: In general I don't see a point in having PIN separated into different preferences fields. It's kinda confusing and makes comparison harder. You cannot set PIN in different length than 4, but separating this makes no sense to me

AndyScherzinger:
Hi @masensio 

I checked the code and as far as coding it looks good to me, just some minor comments. What I didn't understand is the logic because I see the following behavior:
* temporarily store the pin
* loop through the cache files and delete them
* set the pin in the preferences again

Does the looping through the cache empty the preferences? If not then the PIN doesn't need to be recovered since he doesn't get deleted. Besides that one thing I remember that would be missing if the code as of now doesn't purge the preferences is the lately merge #1099 since this implementation stores grid/list information per folder in preferences.

Hope my comments are of any help. I never implemented a manageSpaceActivity.

res/layout/activity_manage_space.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:text="@string/manage_space_description"
+        />
+
+    <android.support.v7.widget.AppCompatButton
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="@string/manage_space_clear_data"
+        android:id="@+id/clearDataButton"
+        android:layout_gravity="center_horizontal"
+        android:layout_margin="4dp"
```
przybylski: This should be taken from (or put to) dims.xml

res/layout/activity_manage_space.xml
```diff
@@ -0,0 +1,29 @@
...
+              android:layout_width="match_parent"
+              android:layout_height="match_parent">
+
+    <TextView
+        android:id="@+id/general_description"
+        android:layout_height="wrap_content"
+        android:layout_width="match_parent"
+        android:layout_gravity="center_vertical"
+        android:singleLine="false"
+        android:layout_margin="10dp"
```
przybylski: This should be taken from (or put to) dims.xml

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
+package com.owncloud.android.ui.activity;
```
przybylski: Copyright and author please

res/layout/activity_manage_space.xml
AndyScherzinger: we do have a value for that already called standard_margin (actually 16dp) and the same for padding (standard_padding) both are defined according to the material standards for margins and paddings.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
AndyScherzinger: I agree but this is imo how it is stroed today so the clear data work would be fine for now. And this should generally be changed but probably in a new PR, I think.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+                clearData();
+            }
+        });
+    }
+
+    /**
+     * Save passcode from Share preferences
+     * Clear the rest of data
+     */
+    private void clearData() {
```
przybylski: This method can hang the app and system will kill it. This should be handled by AsyncTask and maybe even provide some feedback to the user

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public void clearApplicationData() {
+        File cache = getCacheDir();
+        File appDir = new File(cache.getParent());
+        if (appDir.exists()) {
+            String[] children = appDir.list();
+            for (String s : children) {
+                if (!s.equals("lib")) {
+                    File fileToDelete = new File(appDir, s);
+                    deleteDir(fileToDelete);
```
przybylski: So lets, assume that deleteDir file will fail, and we won't do anything about it? That kinda sux

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        super.onOptionsItemSelected(item);
+        switch (item.getItemId()) {
+            case android.R.id.home:
+                finish();
+                break;
```
przybylski: You can early return here

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
AndyScherzinger: @przybylski I would then recommend to use a Loader since Android is then handling the life cycle issues in contrast to using a AsyncTask.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+                if (!s.equals("lib")) {
+                    File fileToDelete = new File(appDir, s);
+                    deleteDir(fileToDelete);
+                    Log_OC.d(TAG, "Clear Application Data, File: " + fileToDelete.getName()+ " DELETED *******");
+                }
+            }
+        }
+    }
+
+    public static boolean deleteDir(File dir) {
```
przybylski: This method can be moved to some different class, it will be more common scenario as app grows large (eg externalSD also removes dir structure).

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,130 @@
...
+        if (dir != null && dir.isDirectory()) {
+            String[] children = dir.list();
```
przybylski: you can call listFiles and do range based for loop.
also children can be null

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
przybylski: @AndyScherzinger but loader is used to provide data for future use, and this is not a case of data loading but real file deletion.

AndyScherzinger: @przybylski true, so this would be misleading. Thus no Loader in this case :)

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,131 @@
...
+                    deleteDir(fileToDelete);
+                    Log_OC.d(TAG, "Clear Application Data, File: " + fileToDelete.getName()+ " DELETED *******");
+                }
+            }
+        }
+    }
+
+    public boolean deleteDir(File dir) {
+        if (dir != null && dir.isDirectory()) {
+            String[] children = dir.list();
```
przybylski: you can call listFiles and do range based for loop.
also children can be null

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,131 @@
...
+                if (!s.equals("lib")) {
+                    File fileToDelete = new File(appDir, s);
+                    deleteDir(fileToDelete);
+                    Log_OC.d(TAG, "Clear Application Data, File: " + fileToDelete.getName()+ " DELETED *******");
+                }
+            }
+        }
+    }
+
+    public boolean deleteDir(File dir) {
```
przybylski: This method can be moved to some different class, it will be more common scenario as app grows large (eg externalSD also removes dir structure).

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: That's a try to enforce the VM is killed, besides of finishing the activity. The intention is that when the app is opened again, it's done from zero. Otherwise there are problems to recreate correctly the database and the app crashes.

But from your comment I had a quick look to the issue, and maybe ```System.exit(0)``` is doing nothing real. We'll check if it works fine also without it.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: :+1:

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: Would be better to have in a single value, but changing that now would require add the code to migrate it for current installations. Not the moment.

przybylski: Following the bad design will only make it worst. changing it into single value is not too hard to do, even in this PR

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
AndyScherzinger: I just meant this as a review comment. Can you take care of this @masensio? I woun't find the time to contribute anything this week :(

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: The scope of this PR is fixing #1325 ASAP. The storage scheme of the passcode will not be changed here.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: @AndyScherzinger , why the sadness? I was supporting your comment.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: :+1:

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: @przybylski's proposal here makes sense. Let's do it.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: Real chances that deletion here fails are really, really small. 

Nevertheless, let's be cautious, control the fail case and report the user.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: Given the size of the Activity, I wouldn't refactor this method to a different class "just for the future". Better refactoring when we have the common scenario than guessing we will have a common scenario.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: Null check needs to be done. @jesmrec , do we have a test case for using this button right after the app was installed and never opened? Should be interesting.

About range based loops, I discourage their use. Their performance is worse than index based loops, and the gain in readability (IMHO) is not worth it.

jesmrec: Right. This test case will be added.

przybylski: @davivel abor range based loop: http://stackoverflow.com/a/256861/969373

res/layout/activity_manage_space.xml
masensio: Done

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
masensio: Done

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
masensio: Now "PrefPinCode*" are static final string in PassCodeActivity.

IMHO, the passcode storage scheme must be changed in other issue, because that change needs different tests than #1325.

przybylski: "PrefPinCode*" as the name saids is a preference, and as such should be stored inside designated place, spreading them around the code will introduce chaos. IMO it should be placed in Prefrences.java

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
AndyScherzinger: @davivel sadness due to the fact that I can't contribute any coding work this week.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
masensio: We've checked the behaviour without System.exit(0) and it is necessary at least in Android 6. 
If we remove this line, we have a crash when trying to access to the app after removing the app data. The app isn't restarted properly, the database isn't recreated fine.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: Thanks, @przybylski . Pretty interesting, indeed.

My source of worry: https://www.youtube.com/watch?v=MZOf3pOAM6A ; results around time 3:00, they refer to range based loop as "for simple". Maybe the penalty is not strictly for Java VMs, but a problem in Android VM.

Please, let me clarify.  I will not ask anybody to avoid range base loops in contributions, unless it results in a real performance problem in the app. It's just that I will not recommend either replacing index based loops with range based loops, nor consider index based loops as a reason to block or reject PRs.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
AndyScherzinger: Thanks for making the effort. Good to know that it is necessary. :+1:

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: I understood later. We'll miss you, but we'll appreciate you the same, even more, when you come back :)

masensio:
Hi @all!

Code updated with your comments and suggestions ( I hope I have not missed anything).
Of course, you can review again if you want.
@jesmrec Ready for QA.

res/layout/activity_manage_space.xml
```diff
@@ -0,0 +1,46 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              android:orientation="vertical"
+              android:layout_width="match_parent"
+              android:layout_height="match_parent">
+
+    <TextView
```
AndyScherzinger: I would add the margin=standard_margin for both elements for the material-look.

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,177 @@
...
+            return  clearResult;
+        }
+
+        public boolean deleteDir(File dir) {
+            if (dir != null && dir.isDirectory()) {
+                String[] children = dir.list();
+                for (int i = 0; i < children.length; i++) {
+                    boolean success = deleteDir(new File(dir, children[i]));
+                    if (!success) {
+                        Log_OC.d(TAG, "File NOT deleted " + children[i]);
```
AndyScherzinger: I would use "w" level in case of unsuccessful deletion

AndyScherzinger:
Looks good to me @masensio. Just had two very minor comments which I missed in the first review. Sorry about that. Great work! :+1:

res/values/strings.xml
```diff
@@ -404,4 +404,9 @@
 
     <string name="action_switch_grid_view">Grid view</string>
     <string name="action_switch_list_view">List view</string>
+
+    <string name="manage_space_title">Manage space</string>
+    <string name="manage_space_description">Settings, database and server certificates from  %1$s\'s data will be deleted permanentlty. \n\nDownloaded files will be kept untouched.\n\nThis process can take some time.</string>
```
przybylski: There is a double space here

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,177 @@
...
+        actionBar.setTitle(R.string.manage_space_title);
+
+        TextView descriptionTextView = (TextView) findViewById(R.id.general_description);
+        descriptionTextView.setText(getString(R.string.manage_space_description, getString(R.string.app_name)));
+
+        Button clearDataButton = (Button) findViewById(R.id.clearDataButton);
+        clearDataButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                ClearDataAsynTask clearDataTask = new ClearDataAsynTask();
```
przybylski: There is no visual feedback to the user that something is going on

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                }
+            }
+            return  clearResult;
+        }
+
+        public boolean deleteDir(File dir) {
+            if (dir != null && dir.isDirectory()) {
+                String[] children = dir.list();
+                for (int i = 0; i < children.length; i++) {
```
przybylski: NPE

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+        }
+
+        public boolean clearApplicationData() {
+            boolean clearResult = true;
+            File appDir = new File(getCacheDir().getParent());
+            if (appDir.exists()) {
+                String[] children = appDir.list();
+                for (String s : children) {
```
przybylski: NPE

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: That can wait for further improvement of the activity.

przybylski: IMO this should be fixed properly, not just fixed.
As a user I would press this button several times if I wouldn't get any immediate response.

masensio:
Sorry, there is an error in the commit comment: "staring" should be "string"

src/com/owncloud/android/ui/activity/ManageSpaceActivity.java
davivel: This is not being just fixed. We are not just ignoring the comments in the review.

@jesmrec, let's add a test case to click several times on the button without waiting for response, and see what happens. If there's any problem, will add some visual feedback of progress.

davivel:
:+1: for me.

Please, @jesmrec , test it.

jesmrec:
@davivel ok, QA in process.

davivel:
In!

Issue #1401, opened 2016-01-04
[1.9.0] Can't copy to clipboard while renaming files


cdamken:
When we try to rename files with branded 1.9.0 the clipboard functions don't work as expected.

1.- Try to Double tap on the entry in the renaming dialogue (instead of touching a longer time) the function menu is darkened and the symbols doesn't work anymore.
2.- Trying to mark the text add selecting the scissors closes the dialogue.

Problem win Sony Xperia Z3 (Andriod 5.1.1)

1.9.0 Branded client.

I tried to reproduce it with onePlus 2 (5.1.1) 

The 1. was unabled to reproduce (I will ask for more detailed information)

The 2. was reproduced in 1.9.0 Branded and Unbranded client.

@MorrisJobke 
00004488

tobiasKaminsky:
I cannot reproduce any of it on Moto X Play with Android 6.
But this is maybe due to the fact that the copy&paste function is fresh implemented.

cdamken:
Now I can reproduce the first point (1.-) ;

The app is opened and I am long tapping an file-entry, so I'm getting the menu:
"Share, Open with, Synchronize, Rename....".
A short tap on /Rename/ opens
a bright dialogue box to enter a new filename. The background is getting darker.
Then I'm double tapping on a part of the filename. This part is marked and I'm getting a second menu between the dialogue box and the status bar. But this menu is also darkened and contains symbols for cut, copy and paste. When I'm tapping there to select a function like paste or cut, the dialogue box
closes.

@tobiasKaminsky Do you have a device with 5.1.1? I can repoduce with branden and unbraded versions on my phone.

tobiasKaminsky:
No, unfortunately I only have 6.x devices :/

MorrisJobke:
cc @rperezb @davivel

rperezb:
Just reproduced it, on Android 4.1.2 and 4.3 devices. 
The thing is that the edit menu appears below the rename view so when you select any edit action (copy/cut..) it's detected as cancel action (cancel the rename) so that the view is closed.
Android 6 works fine since the edit menu is above the rename view

AndyScherzinger:
@rperezb can you provide 1 or 2 screenshots? Also also only have Android 6 at the moment. Thanks! :)

rperezb:
Does this help?

![edit menu](https://cloud.githubusercontent.com/assets/1846927/12350696/f0ed91ba-bb78-11e5-8883-45f2bde7bc4c.png)

[...]

Issue #1343, opened 2015-12-13
Android 6: new permission system


tobiasKaminsky:
I have installed owncloud on a fresh android 6 system.
After installing I could not download any file because the "storage" permission was not granted.
This should, according to the new permission system, catched during runtime and the user should be asked if one wants to grant this permission.

AndyScherzinger:
We can implement this via the AppCompat lib so it works in a backwards compatible way :) What suprises me a little is the fact that the storage permission is missing since that would lead to the same problem on any Android version and Android 6 still supports the now legacy permission model?!

ContextCompat should be able to handle all this: https://developer.android.com/training/permissions/requesting.html But since I haven't played around with this I have no feeling how easy this will work and if using ApPCompat means it will work completely backward compatible... Using this then means #1220 has to be merged first since permissions require AppCompat >= 23.x

AndyScherzinger:
I deployed a permissions requsting implementation for WRITE_EXTERNAL_STORAGE in the Marshmallow branch and the beta version :grinning: 
This also means I implemented the new permissions API introduced with Android M in a backwards compatible (yet untested) way. Relevant commit can be checked out here: https://github.com/owncloud/android/commit/373235225780503555388bc4a099f3c517a8129a
Requesting the write permission leads to also getting the read permission when the write permission is granted by the user.

Looping in @jancborchardt @masensio @davivel @przybylski and also @LukasReschke since this is just a simple implementation to get the app working again for Marshmallow compile targeting (and the beta) but this has some impact on user interaction and is also a security/permissions topic in general. The implementation basically works like dropbox, that means whenever you open up the app the first time, add your first account and open up the list (after hitting the connect button) the app will ask you for the external storage access right. It will always ask for the right if it hasn't been granted since with compiling against v23 the app simple won't let you browse the storage, download or upload files - at least not on Marshmallow. 

In Marshmallow rights can be revoked even after installing the app and this is mainly why the check has to happen over and over again. In case the permission has been denied or revoked the app now shows a snack-bar with an explanation on why the user needs to grant the right (see screenshot 1). After the user clicks OK, the permission dialog pops up the give the user the possibility to grant the right (screenshot 2). 

Screenshot 1
![device-2015-12-29-005628](https://cloud.githubusercontent.com/assets/1315170/12027501/8f0abd60-adc7-11e5-8c0e-ab2e249648b3.png)

Screenshot 2
[...]

PR #1473, opened 2016-02-12
Move to compile target Marshmallow (Android 6 - v23) - rebased without FAB

davivel:
This Branch is *not* based on material_fab.

Works for gradle and ant.

What has been done so far:

* [x] Bumped compile target to 23
* [x] Bumped AppCompat to 23.1.1
* [x] Bumped build tools to 23.0.2 - same for the travis build
* [x] Fixed MediaService implementation
    * Moved on to NotificationCompat.Builder (as it is done in all aother placed within the app) since the Notification.setLatestEventInfo() has been removed with Android M, see http://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-notifications
    * added the material design "play button" icon for the notification/system bar
* [x] Added AppCompat Design lib 23.1.1
* [x] Fixed dependencies on deprecated Apache HTTP API in Android framework.
* [ ] ONGOING: v23 permission API calls for write access to external storage (necessary since compile target has been bumped to 23)
* [x] DONE- runtime permission check for CONTACTS group due to GET_ACCOUNTS permissions; check if this permission is strictly needed (MANAGE_ACCOUNTS is not in CONTACTS group, and maybe it's all that we need)


davivel:
How got here:
1.  ```git rebase --onto master material_fab target_mashmallow```; resolved some conflicts in the process, including removal of dependency on FAB library in ```build.gradle``` and ```setup_*``` files
2. ```git cherry-pick``` e494737
3. ```git cherry-pick``` a55e01c7c

davivel:
Checking updated documentation about GET_ACCOUNTS permissions, maybe we don't need to add the permission check for it: http://developer.android.com/intl/es/reference/android/Manifest.permission.html#GET_ACCOUNTS_PRIVILEGED

In any case, still needs to be there for Android <=5.1

davivel:
Tested: there is no need to check in runtime the if GET_ACCOUNTS is granted or not; since the app only uses its own accounts, the permissions is not really required from Android 6. If was really needed, the line https://github.com/owncloud/android/blob/master/src/com/owncloud/android/authentication/AccountUtils.java#L88 would be failing when the app has no permission on CONTACTS group, and that's not happening.

AndyScherzinger:
@davivel thanks for the rebase, great collaboration! Again! :+1: 
I am running the build on my phone (Android 6.0.1) since Saturday without any errors. The setup.md hat some merge conflicts which I resolved and also fixed the formatting of the unordered lists (https://github.com/owncloud/android/blob/target_marshmallow_rebased/SETUP.md), sorry for the amount of commits...I had to push to see if the formatting worked a couple of times.

davivel:
Thanks a lot, @AndyScherzinger . 

No sorry for the tiny commits, pal. Anyway, if you want a challenge, next time you don't like to have so many tiny commits in a row, you can try to squash them. See a couple of references: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html , https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/ .

About testing in real devices, we have what we need here to test Android 6. 

About the permissions handling, I agree that your solution should be enough for the moment, at least to the point that we can test and close #1401. This PR fixes it, as you commented there. 

I'll have a look to #1343. Adding some more checks of WRITE_EXTERNAL_STORAGE in FileDownloader and FileUploader could be interesting. In any case, that shouldn't block this PR.

davivel:
Oh, and I added some comments in AndroidManifest.xml about the permissions, if you are interested.

@jesmrec , I think testing on #1401 could be started.

@masensio , @malkomich , I made some extra changes by my own, other code review could be convenient. @AndyScherzinger , your comments, of course, are also welcome :)

AndyScherzinger:
Looks good to me :+1:

rperezb:
@malkomich how does this look to you?

masensio:
Code reviewed :+1:

PR #1484, opened 2016-02-22
Added further references in CONTRIBUTING.md documentation

AndyScherzinger:
* Added references to SETUP.md and Github issues tracker
* fixes some typos/formatting

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ If you have questions about how to use ownCloud, please direct these to the [mai
   - [iOS client](https://github.com/owncloud/ios-issues/issues)
   - [Desktop client](https://github.com/owncloud/mirall/issues)
   - [ownCloud apps](https://github.com/owncloud/apps/issues) (e.g. Calendar, Contacts...)
-* Search the existing issues first, it's likely that your issue was already reported.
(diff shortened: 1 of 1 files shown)


masensio:
Hi @AndyScherzinger,

thanks for updating CONTRIBUTING.md. Documentation updating is always welcome. 
Only one comment, I think is too much three commits for changing some characters and words in a text file. Please, take into account to reduce the number of commits for this kind of contributions.

:+1:  for the update.

AndyScherzinger:
Hi @masensio - will try to do that :) Also had this discussion with davivel. Any input on how to do this? I simply ended up both times with this amount of commits since I realized them during several readings, after the changes and @davivel suggested then squashing the commits. Reading the articles he mentioned said that you shouldn't sqash commits after they are available in the remote repo and squashing should be limited to the local repo. :(
Maybe it is okay to squash commits in the remote repo in case of branches where only one user is actually working with the branch. What do you think?

masensio:
Hi @AndyScherzinger,
I think if the branch has a lot of contributors the squash process can be difficult but if you are the only one you can squash without problems. After squashing and pushing your changes please add a comment to add a comment to warn that there has been a squash and should update their branch if you know there are more contributors with this branch (for code review or testing).

AndyScherzinger:
Ah okay, got it! :+1: Thanks for the feedback! :smiley:

masensio:
You're welcome!

PR #1485, opened 2016-02-24
Fix problems with Credentials error in uploads

masensio:


src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1025,6 +1027,10 @@ public void onRemoteOperationFinish(RemoteOperation operation, RemoteOperationRe
             onGetUserNameFinish(result);
         }
 
+        if (result.isSuccess() && mAction == ACTION_UPDATE_EXPIRED_TOKEN) {
+            setResult(FileActivity.RESULT_OK);
```
davivel: Could we use ```Activity.RESULT_OK```?  That way we don't need to add a dependency from ```AuthenticatorActivity``` to ```FileActivity```

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1419,6 +1425,9 @@ public void onAuthenticatorTaskCallback(RemoteOperationResult result) {
             }
 
             if (success) {
+                if (mAction == ACTION_UPDATE_EXPIRED_TOKEN) {
+                    setResult(FileActivity.RESULT_OK);
```
davivel: Same here

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -541,6 +543,17 @@ public void retryUpload(OCUpload upload) {
         }
     }
 
+    public void retryUploadsForAccount(Account account) {
```
davivel: Could we move this method to ```FileUploader```? As a static method ```retryUploadsForAccount(Account account, Context context)```, similar to the new static methods to ease the creation of upload requests. Thanks.

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -833,7 +834,7 @@ protected void requestCredentialsUpdate(Context context) {
                     AuthenticatorActivity.EXTRA_ACTION,
                     AuthenticatorActivity.ACTION_UPDATE_EXPIRED_TOKEN);
             updateAccountCredentials.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS);
-            startActivity(updateAccountCredentials);
+            startActivityForResult(updateAccountCredentials, UPDATE_CREDENTIALS_REQUEST_CODE);
```
davivel: This change is not needed anymore, right?

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -114,6 +113,8 @@
 
     protected static final long DELAY_TO_REQUEST_OPERATIONS_LATER = 200;
 
+    public static final int UPDATE_CREDENTIALS_REQUEST_CODE = 100;
```
davivel: I'd move this constant to ```UploadListActivity```, that's the Activity really needing it.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
masensio: Of course, we can use Activity.RESULT_OK but this change involves adding the import android.app.Activity too.

src/com/owncloud/android/files/FileOperationsHelper.java
masensio: Done

src/com/owncloud/android/ui/activity/FileActivity.java
masensio: Right.

masensio:
Close this pull request and create another new with relevant commits https://github.com/owncloud/android/commit/4d57de703245494c1ee228df927e08c2390f9a01 and https://github.com/owncloud/android/commit/ec709eca98a506fdea9520dd54adebcba0d71583

PR #1487, opened 2016-02-25
Use Upload End Time in Uploads View

masensio:


src/com/owncloud/android/ui/adapter/ExpandableUploadListAdapter.java
```diff
@@ -93,7 +93,11 @@ public String getGroupName() {
         public Comparator<OCUpload> comparator = new Comparator<OCUpload>() {
             @Override
             public int compare(OCUpload lhs, OCUpload rhs) {
-                return compareUploadId(lhs, rhs);
```
malkomich: Maybe the names of the parameters (lhs, rhs) are not enough explanatory

masensio: It is the name used by default in http://developer.android.com/intl/es/reference/java/util/Comparator.html.
Anyway I've changed it.

PR #1416, opened 2016-01-09
#1413 share view enhancement

AndyScherzinger:
For details and discussion please see #1413 

For reviewers, things that have been done in this PR:
* Moved from Switch zu AppCompatSwitch to have a material switch on all Android versions
* changed paddings/margins according to the introduced standard value fit
* added user/group icons to items to also visually distinguish between user and group line items

pinging @jancborchardt for visual review, open issue imho is now that I added icons for user/group (see screenshots) each line has an icon on the left (visual indication user or group item) and one on the right (delete action). Both are grey at the moment, but only the right icon is an action. My idea would be to color the delete icon in blue to indicate it is as an action. What do you think. We can also stay with it the way it is right now, whatever you choose :wink: 

**Detailed Screenshots (before/after)**

BEFORE: Share screen (on Android v15) with old Holo style switcher
device-2016-01-09-171558

AFTER: Share screen (on Android v15) with new Material style switcher, icons for user/group, Link items header being black and values being grey, switchers now top aligned to their corresponding headers
device-2016-01-09-163851

AFTER: Share screen (on Android v23) with new Material style switcher, icons for user/group, Link items header being black and values being grey, switchers now top aligned to their corresponding headers
device-2016-01-09-162507

AFTER: Share screen (on Android v23) being empty
device-2016-01-09-170355

BEFORE: User search (on Android v23)
device-2016-01-09-170930

AFTER: User search (on Android v15) (this has been styled like this before, since this is a standard UI component of Android)
device-2016-01-09-164649

AFTER: User search (on Android v23)
device-2016-01-09-163300

FIXED: Passwort dialog, buttons now colored correctly in the accent coloring and not in the Android platform color + fixed padding
device-2016-01-09-162523

PR #1492, opened 2016-03-03
Federated sharing

malkomich:
Enabled federate sharing, and autocompletion of users from well known servers. 
@davivel @rperezb

### DIFF
--- a/src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
+++ b/src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
@@ -60,7 +60,8 @@ public class CreateShareWithShareeOperation extends SyncOperation {
      *                      https://doc.owncloud.org/server/8.2/developer_manual/core/ocs-share-api.html .
      */
     public CreateShareWithShareeOperation(String path, String shareeName, ShareType shareType, int permissions) {
-        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType) && !ShareType.FEDERATED.equals(shareType)) {
+        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType)
+            && !ShareType.FEDERATED.equals(shareType)) {
             throw new IllegalArgumentException("Illegal share type " + shareType);
         }
         mPath = path;
--- a/src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
+++ b/src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
@@ -60,7 +60,7 @@ public class CreateShareWithShareeOperation extends SyncOperation {
      *                      https://doc.owncloud.org/server/8.2/developer_manual/core/ocs-share-api.html .
      */
     public CreateShareWithShareeOperation(String path, String shareeName, ShareType shareType, int permissions) {
-        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType)) {
+        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType) && !ShareType.FEDERATED.equals(shareType)) {
             throw new IllegalArgumentException("Illegal share type " + shareType);
         }
         mPath = path;
--- a/src/com/owncloud/android/operations/UnshareOperation.java
+++ b/src/com/owncloud/android/operations/UnshareOperation.java
@@ -78,7 +78,8 @@ public class UnshareOperation extends SyncOperation {
                 if (ShareType.PUBLIC_LINK.equals(mShareType)) {
                     file.setShareViaLink(false);
                     file.setPublicLink("");
-                } else if (ShareType.USER.equals(mShareType) || ShareType.GROUP.equals(mShareType) || ShareType.FEDERATED.equals(mShareType)){
+                } else if (ShareType.USER.equals(mShareType) || ShareType.GROUP.equals(mShareType)
+                    || ShareType.FEDERATED.equals(mShareType)){
                     // Check if it is the last share
                     ArrayList <OCShare> sharesWith = getStorageManager().
                             getSharesWithForAFile(mRemotePath,
--- a/src/com/owncloud/android/operations/UnshareOperation.java
+++ b/src/com/owncloud/android/operations/UnshareOperation.java
@@ -75,10 +75,10 @@ public class UnshareOperation extends SyncOperation {
             if (result.isSuccess()) {
                 Log_OC.d(TAG, "Share id = " + share.getRemoteId() + " deleted");
 
-                if (mShareType == ShareType.PUBLIC_LINK) {
+                if (ShareType.PUBLIC_LINK.equals(mShareType)) {
                     file.setShareViaLink(false);
                     file.setPublicLink("");
-                } else if (mShareType == ShareType.USER || mShareType == ShareType.GROUP){
+                } else if (ShareType.USER.equals(mShareType) || ShareType.GROUP.equals(mShareType) || ShareType.FEDERATED.equals(mShareType)){
                     // Check if it is the last share
                     ArrayList <OCShare> sharesWith = getStorageManager().
                             getSharesWithForAFile(mRemotePath,
--- a/src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
+++ b/src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
@@ -40,6 +40,7 @@ import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.shares.GetRemoteShareesOperation;
+import com.owncloud.android.lib.resources.shares.ShareType;
 import com.owncloud.android.utils.ErrorMessageAdapter;
 
 import org.json.JSONException;
@@ -70,9 +71,14 @@ public class UsersAndGroupsSearchProvider extends ContentProvider {
 
     public static final String AUTHORITY = UsersAndGroupsSearchProvider.class.getCanonicalName();
     public static final String ACTION_SHARE_WITH = AUTHORITY + ".action.SHARE_WITH";
-    public static final String DATA_USER = AUTHORITY + ".data.user";
-    public static final String DATA_GROUP = AUTHORITY + ".data.group";
-    public static final String DATA_REMOTE = AUTHORITY + ".data.remote";
+
+    public static final String USER = "user";
+    public static final String GROUP = "group";
+    public static final String REMOTE = "remote";
+
+    public static final String DATA_USER = AUTHORITY + ".data." + USER;
+    public static final String DATA_GROUP = AUTHORITY + ".data." + GROUP;
+    public static final String DATA_REMOTE = AUTHORITY + ".data." + REMOTE;
 
     private UriMatcher mUriMatcher;
 
--- a/src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
+++ b/src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
@@ -197,7 +197,12 @@ public class UsersAndGroupsSearchProvider extends ContentProvider {
                         displayName = getContext().getString(R.string.share_group_clarification, userName);
                         dataUri = Uri.withAppendedPath(groupBaseUri, shareWith);
                     } else if (ShareType.FEDERATED.getValue() == type && federatedShareAllowed) {
-                        displayName = getContext().getString(R.string.share_remote_clarification, shareWith);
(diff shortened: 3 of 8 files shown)


rperezb:
call for @jesmrec too!

AndyScherzinger:
Can we combine this with #1416 @davivel and @rperezb? :grin:

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -70,8 +72,14 @@
...
-    public static final String DATA_USER = AUTHORITY + ".data.user";
-    public static final String DATA_GROUP = AUTHORITY + ".data.group";
+
+    public static final String USER = "user";
+    public static final String GROUP = "group";
+    public static final String REMOTE = "remote";
+
+    public static final String DATA_USER = AUTHORITY + ".data." + USER;
+    public static final String DATA_GROUP = AUTHORITY + ".data." + GROUP;
+    public static final String DATA_REMOTE = AUTHORITY + ".data." + REMOTE;
```
AndyScherzinger: If we build the String this way then ".data." should also be declared as a constant :)

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -148,31 +156,45 @@ private Cursor searchForUsersOrGroups(Uri uri) {
...
             response = new MatrixCursor(COLUMNS);
             Iterator<JSONObject> namesIt = names.iterator();
-            int count = 0;
             JSONObject item;
             String displayName;
+            int count = 0;
             Uri dataUri;
             Uri userBaseUri = new Uri.Builder().scheme("content").authority(DATA_USER).build();
             Uri groupBaseUri = new Uri.Builder().scheme("content").authority(DATA_GROUP).build();
+            Uri remoteBaseUri = new Uri.Builder().scheme("content").authority(DATA_REMOTE).build();
```
AndyScherzinger: Same here, "content" should be a constant

src/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -116,25 +116,42 @@ protected void onNewIntent(Intent intent) {
         }
     }
 
-    private void doShareWith(String shareeName, boolean isGroup) {
+    private void doShareWith(String shareeName, String dataAuthority) {
+
+        int lastPointIndex = dataAuthority.lastIndexOf(".");
+        String type = dataAuthority.substring(lastPointIndex + 1);
+        ShareType shareType = null;
+        switch (type) {
```
AndyScherzinger: The switch could be substituted via a static final hashmap that has been statically prefilled during classloadinig (basically implementing a lookup table) but I leave this idea up for @davivel's vote :smiley_cat:

AndyScherzinger:
Code looks to to me :+1: 
Just some minor comments where the code can be (just slightly) improved.

davivel:
@AndyScherzinger , #1416 is planned for 1.9.2, we'll go with it in a week at latest :)

owncloud-android-library
```diff
@@ -1 +1 @@
-Subproject commit 1007a80534827479b0874ae942c8c7dbe87cac35
+Subproject commit 36da2a2336fdee71b544705b6a28b313d7f53e4d
```
davivel: @malkomich , you need to create a PR in the owncloud-library project too

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -148,31 +156,45 @@ private Cursor searchForUsersOrGroups(Uri uri) {
...
                     String shareWith = value.getString(GetRemoteShareesOperation.PROPERTY_SHARE_WITH);
-                    if (GetRemoteShareesOperation.GROUP_TYPE.equals(type)) {
+
+                    if (ShareType.GROUP.getValue() == type) {
                         displayName = getContext().getString(R.string.share_group_clarification, userName);
                         dataUri = Uri.withAppendedPath(groupBaseUri, shareWith);
+                    } else if (ShareType.FEDERATED.getValue() == type) {
+                        // check if the federated sharing is allowed
+                        FileDataStorageManager manager = new FileDataStorageManager(account, getContext().getContentResolver());
+                        boolean federatedShareAllowed = manager.getCapability(account.name).getFilesSharingFederationOutgoing().isTrue();
```
davivel: Checking if ```federatedShareAllowed``` should not be done in the ```while``` loop, but before entering it. There are zero chances of it changes between different iterations.

Moreover, probably it may go directly in ```onCreate``` , and ```federatedShareAllowed``` be converted in a member variable.

In both cases, we can rewrite the condition to ```} else if (ShareType.FEDERATED.getValue() = type && federatedShareAllowed)``` , so that we don't need to use a ```continue``` clause (that I'd rather to avoid as a general rule)

src/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -1,22 +1,21 @@
...
- *   but WITHOUT ANY WARRANTY; without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *   GNU General Public License for more details.
- *
- *   You should have received a copy of the GNU General Public License
- *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * ownCloud Android client application
  *
+ * @author masensio
+ * @author David A. Velasco
```
davivel: You may add yourself to the list of @authors if you wish to.

src/com/owncloud/android/ui/activity/ShareActivity.java
davivel: I'd prefer not, that's more hideous to read than a switch block, and this block is not so long.

src/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -116,25 +116,42 @@ protected void onNewIntent(Intent intent) {
...
-                (isGroup ? ShareType.GROUP : ShareType.USER),
-                getAppropiatePermissions()
+                shareType,
+                getAppropiatePermissions(shareType.equals(ShareType.FEDERATED))
         );
     }
 
 
-    private int getAppropiatePermissions() {
+    private int getAppropiatePermissions(boolean isFederated) {
```
davivel: This parameter seems too "cooked". I'd just pass ```shareType```.

src/com/owncloud/android/ui/fragment/EditShareFragment.java
```diff
@@ -147,9 +147,12 @@ private void refreshUiFromState(View editShareView) {
             setPermissionsListening(editShareView, false);
 
             int sharePermissions = mShare.getPermissions();
+            boolean isFederated = mShare.getShareType().equals(ShareType.FEDERATED);
             CompoundButton compound;
 
             compound = (CompoundButton) editShareView.findViewById(R.id.canShareSwitch);
+            if(isFederated)
```
davivel: Please, use curly brackets for single-line ```if```s too.

davivel:
Added some more comments.

Thanks, @AndyScherzinger :-)

src/com/owncloud/android/ui/activity/ShareActivity.java
davivel: I thought it twice. Your proposal, @AndyScherzinger, is indeed better.

The reason of using ```dataAuthority``` is that the contract of search providers does not allow return arbitrary parameters in a search. Otherwise we would return directly ```shareType```. We are simply "tricking" the search provider interface.

The responsibility to "translate the trick" would be better in ```UserAndGroupsSearchProvider``` than in its clients (in this case, ```ShareActivity```) . So, would be better to create the map you suggest in ```UserAndGroupsSearchProvider```, and a static public method that allow ```ShareActivity``` get the ```ShareType``` corresponding to the received URL.

malkomich:
I have just update the PR with the changes not included from the library and the changes from the code review @AndyScherzinger @davivel

malkomich:
Now the autocomplete feature matches the one in the web UI

malkomich:
Thanks @AndyScherzinger , changes commited ;)

davivel:
Code is :+1: to me, but needs to wait for https://github.com/owncloud/android-library/pull/111

davivel:
owncloud/android-library#111 is pending only of updating automatic tests; this is now ready to be tested.

src/com/owncloud/android/operations/CreateShareWithShareeOperation.java
```diff
@@ -60,7 +60,7 @@
      *                      https://doc.owncloud.org/server/8.2/developer_manual/core/ocs-share-api.html .
      */
     public CreateShareWithShareeOperation(String path, String shareeName, ShareType shareType, int permissions) {
-        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType)) {
+        if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType) && !ShareType.FEDERATED.equals(shareType)) {
```
davivel: Please, wrap this line shorter than 120 characters.

src/com/owncloud/android/operations/UnshareOperation.java
```diff
@@ -75,10 +75,10 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             if (result.isSuccess()) {
                 Log_OC.d(TAG, "Share id = " + share.getRemoteId() + " deleted");
 
-                if (mShareType == ShareType.PUBLIC_LINK) {
+                if (ShareType.PUBLIC_LINK.equals(mShareType)) {
                     file.setShareViaLink(false);
                     file.setPublicLink("");
-                } else if (mShareType == ShareType.USER || mShareType == ShareType.GROUP){
+                } else if (ShareType.USER.equals(mShareType) || ShareType.GROUP.equals(mShareType) || ShareType.FEDERATED.equals(mShareType)){
```
davivel: Please, wrap.

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -70,11 +73,32 @@
...
+    static {
+        sShareTypes = new HashMap<>();
+        sShareTypes.put(DATA_USER, ShareType.USER);
+        sShareTypes.put(DATA_GROUP, ShareType.GROUP);
+        sShareTypes.put(DATA_REMOTE, ShareType.FEDERATED);
+    }
+
+    public static ShareType getShareType(String authority) {
+
+        if (sShareTypes.containsKey(authority)) {
```
davivel: This method may be written in a single line; check the reference for ```HashMap#get```, it returns ```null``` if the key doesn't exist.

http://developer.android.com/intl/es/reference/java/util/HashMap.html#get(java.lang.Object)

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -148,31 +172,50 @@ private Cursor searchForUsersOrGroups(Uri uri) {
...
+                    if (ShareType.GROUP.getValue() == type) {
                         displayName = getContext().getString(R.string.share_group_clarification, userName);
                         dataUri = Uri.withAppendedPath(groupBaseUri, shareWith);
-                    } else {
+                    } else if (ShareType.FEDERATED.getValue() == type && federatedShareAllowed) {
+                        if (userName.equals(shareWith)) {
+                            displayName = getContext().getString(R.string.share_remote_clarification, userName);
+                        } else {
+                            displayName = getContext().getString(R.string.share_known_remote_clarification, userName,
+                                    shareWith.split("@")[1]);
```
davivel: Shouldn't use the element in position ```[1]```,  but the last element, not assuming that there will be only 2. In some servers "@" may be a valid character in the username.

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -148,31 +172,50 @@ private Cursor searchForUsersOrGroups(Uri uri) {
...
+                    } else if (ShareType.USER.getValue() == type) {
                         displayName = userName;
                         dataUri = Uri.withAppendedPath(userBaseUri, shareWith);
                     }
-                    response.newRow()
-                            .add(count++)             // BaseColumns._ID
-                            .add(displayName)         // SearchManager.SUGGEST_COLUMN_TEXT_1
-                            .add(dataUri);
+
+                    if (displayName != null && dataUri != null)
```
davivel: Please, use ```{ }``` to wrap conditional one-liners; it's easy to break something if the block is modified later.

davivel:
Well done, just a couple of minor things.

Please, remember to update the library reference also after the modifications inhttps://github.com/owncloud/android-library/pull/111

malkomich:
Fixed :)

davivel:
:+1:

I rebased to accomplish with branch protection. No conflict found.

PR #1510, opened 2016-03-15
Edit public share permission

malkomich:
As an Android user I want to decide whether edit is allowed or not when sharing via link a FOLDER so that I control the level of access

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -382,7 +382,7 @@
     <string name="share_via_link_expiration_date_label">Set expiration date</string>
     <string name="share_via_link_password_label">Password protect</string>
     <string name="share_via_link_password_title">Secured</string>
-    <string name="share_via_link_edit_permission_label">Allow edit</string>
+    <string name="share_via_link_edit_permission_label">Allow editing</string>
     <string name="share_get_public_link_button">Get link</string>
 
     <string name="share_search">Search</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -382,6 +382,7 @@
     <string name="share_via_link_expiration_date_label">Set expiration date</string>
     <string name="share_via_link_password_label">Password protect</string>
     <string name="share_via_link_password_title">Secured</string>
+    <string name="share_via_link_edit_permission_label">Allow edit</string>
     <string name="share_get_public_link_button">Get link</string>
 
     <string name="share_search">Search</string>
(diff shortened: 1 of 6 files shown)


res/values/strings.xml
```diff
@@ -382,6 +382,7 @@
     <string name="share_via_link_expiration_date_label">Set expiration date</string>
     <string name="share_via_link_password_label">Password protect</string>
     <string name="share_via_link_password_title">Secured</string>
+    <string name="share_via_link_edit_permission_label">Allow edit</string>
```
davivel: Should be ```Allow editing```

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -409,6 +410,25 @@ public void setPermissionsToShare(OCShare share, int permissions) {
         queueShareIntent(updateShareIntent);
     }
 
+    /**
+     * Updates a public share on a file to set its edit permissions.
+     * Starts a request to do it in {@link OperationsService}
+     *
+     * @param file                     File which public share will be constrained with an expiration date.
```
davivel: This comment is not correctly updated

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -432,6 +457,61 @@ public void onClick(View passwordView) {
...
+         * @param isChecked  New switch state.
+         */
+        @Override
+        public void onCheckedChanged(CompoundButton switchView, boolean isChecked) {
+            if (!isResumed()) {
+                // very important, setCheched(...) is called automatically during
+                // Fragment recreation on device rotations
+                return;
+            }
+            boolean permission;
```
davivel: The value of ```permission``` may be set in a single line

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -655,6 +740,25 @@ public void onClick(View v) {
                     mOnPasswordInteractionListener
             );
 
+            /// update state of the edit permission switch
+            Switch editPermissionSwitch = getEditPermissionSwitch();
+            if(getEditPermissionSection().getVisibility() == View.VISIBLE) {
```
davivel: Is this check really needed? Is there a way that the switch is clicked while being invisible?

davivel:
Code reviewed.

davivel:
:+1:

Issue #1383, opened 2015-12-24
Quick Account-Switching in navigation drawer


stonerl:
Currently when one has  multiple accounts in the app the only way to switch between these accounts is to open the settings and select the appropriate account.

Would it be possible to have a quicker way? e.g. from the sidebar?

tobiasKaminsky:
Yes, this is something that was discussed at #757 if I remember correctly.
But we should keep this issue open and change the topic to a more detailed one.
I will dig into my old code to see what I already have done and will create a PR soon for it.
After that it will be first in beta and after a while in the official version.

tobiasKaminsky:
> create a PR soon for it

:laughing: 
But now...#1537

AndyScherzinger:
current implementation is done in PR #1559

davivel:
Merged, will be released in version 2.2.0

PR #1537, opened 2016-03-18
Account switching in navigation drawer

tobiasKaminsky:
As wished in #1383

### DIFF
--- a/res/layout/drawer.xml
+++ b/res/layout/drawer.xml
@@ -28,33 +28,10 @@
         android:background="@color/background_color"
         >
 
-        <!--TODO re-enable when "Accounts" is available in Navigation Drawer-->
-        <!--<LinearLayout-->
-            <!--android:layout_width="match_parent"-->
-            <!--android:layout_height="wrap_content"-->
-            <!--android:layout_margin="5dp">-->
-
-            <!--<ImageView-->
-                <!--android:id="@+id/drawer_userIcon"-->
-                <!--android:layout_width="40dp"-->
-                <!--android:layout_height="40dp"-->
-                <!--android:src="@drawable/abc_ab_bottom_solid_dark_holo" />-->
-
-            <!--<TextView-->
-                <!--android:id="@+id/drawer_username"-->
-                <!--android:layout_width="wrap_content"-->
-                <!--android:layout_height="wrap_content"-->
-                <!--android:layout_gravity="center_vertical"-->
-                <!--android:layout_marginLeft="5dp"-->
-                <!--android:layout_marginStart="5dp"-->
-                <!--android:textAppearance="?android:attr/textAppearanceLarge" />-->
-
-        <!--</LinearLayout>-->
-
-        <!--<TextView-->
-            <!--android:layout_width="fill_parent"-->
-            <!--android:layout_height="2dip"-->
-            <!--android:background="@color/list_item_lastmod_and_filesize_text" />-->
+        <TextView
+            android:layout_width="fill_parent"
+            android:layout_height="2dip"
+            android:background="@color/list_item_lastmod_and_filesize_text" />
 
         <ListView
             android:id="@+id/drawer_list"
@@ -74,7 +51,7 @@
             android:background="@color/drawer_header_color">
 
             <ImageView
-                android:id="@+id/itemIcon"
+                android:id="@+id/drawer_userIcon"
                 android:layout_width="24dp"
                 android:layout_height="24dp"
                 android:layout_marginLeft="@dimen/standard_padding"
--- a/res/values-de/strings.xml
+++ b/res/values-de/strings.xml
@@ -19,8 +19,7 @@
   </string-array>
   <!--TODO re-enable when server-side folder size calculation is available   
     	<item>Biggest - Smallest</item>-->
-  <!--TODO re-enable when "Accounts" is available in Navigation Drawer-->
-  <!--<string name="drawer_item_accounts">Accounts</string>-->
+  <string name="drawer_item_accounts">Accounts</string>
   <string name="drawer_item_all_files">Alle Dateien</string>
   <!--TODO re-enable when "On Device" is available
     <string name="drawer_item_on_device">On device</string>-->
--- a/src/com/owncloud/android/ui/adapter/NavigationDrawerListAdapter.java
+++ b/src/com/owncloud/android/ui/adapter/NavigationDrawerListAdapter.java
@@ -24,21 +24,31 @@ package com.owncloud.android.ui.adapter;
 import android.accounts.Account;
 import android.accounts.AccountManager;
 import android.content.Context;
+import android.graphics.Color;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.BaseAdapter;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
+import android.widget.RadioButton;
+import android.widget.RadioGroup;
 import android.widget.TextView;
 
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.authentication.AccountUtils;
+import com.owncloud.android.lib.common.utils.Log_OC;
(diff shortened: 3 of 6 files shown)


tobiasKaminsky:
account


Ugly as hell...
@AndyScherzinger can you help me here with this? Pleeeease :) :)

przybylski:
I don't know if you want to add more dependencies but you migh wanna check this out https://github.com/mikepenz/MaterialDrawer

tobiasKaminsky:
Seems to be a very good lib, what do you think @AndyScherzinger @jancborchardt @davivel

AndyScherzinger:
@tobiasKaminsky I'll take a look and pimp it :) I also stumbled across the lib but with the more recent additions to the AppCompat Drawer implementation I am not sure it is worth the dependency. I'll see if we can add the functionality with the AppCompat lib. The neat thing about the Lib is everything comes out of the box (sticky footer, vector support, account switcher) on the other hand it also relies n the AppCompat lib and thus we would then always stay in sync with the versions imho which might be paintful in case we need a bug fix release from that lib which forces an AppCompat update.
I'll see what I can do without the library as a first try :)

jancborchardt:
Yep, it should definitely look like the switcher in the linked library. :)

Also, the last entry of the dropdown should then be »Add account« so we can completely remove this section from the settings.

AndyScherzinger:
In that case we probably should utilize the library with the mentioned benefits like:
* account switcher
* vector icons and thus coloring :)

Downside might be that besides a certain AppCompat lock (atm 23.2.1) due to the lib is that it adds some more AppCompat libs and lib internal stuff, see: https://github.com/mikepenz/MaterialDrawer/blob/develop/library/build.gradle

@davivel what is your opinion?

tobiasKaminsky:
> Also, the last entry of the dropdown should then be »Add account« so we can completely remove this section from the settings.

I remember that we discussed this and said that the drawer is just a switch for the accounts and should not function as add/remove. This would also keep the drawer clean and simple and focus on the tasks that are often used.

AndyScherzinger:
The account switcher would automatically bring such a feature, b/c the user switches swaps the list and the below menu/actions so this would be an option to use if we make use of the lib. Else we should keep it simple since the fully blown implementation like the one offered needs a lot of effort to implement by yourself.
Like I said the AppCompat dependency might be an issue. This would be fine with me in case we:
* wait until after the 1.9.2 release
* bump the AppCompat libs dependendy accordingly
* add the drawer lib

AndyScherzinger:
After a couple of hours of playing around with the actual implementation I couldn't get anywhere style-wise yet :disappointed:

AndyScherzinger:
I'm still working on it moving on to the actual ```NavigationView``` implementation, but might have to have a diskussion with @jancborchardt since that means some (auto) style changes since we will then be using Google's implementation for the drawer menu which has some implications (font, font-weight, etc.). Will keep you posted. In case Jan is fine with that I will also add the account switcher in the way it is known from other material designed apps :smile:

AndyScherzinger:
Screenshots of the drawer using the Google Support lib can now be found here: https://github.com/owncloud/android/pull/1552#issuecomment-199973134

tobiasKaminsky:
Just to understand: which lib do you mean?

AndyScherzinger:
As for the dependency discussion I am referring to https://github.com/mikepenz/MaterialDrawer as for the latest comments by me I am referring to the support lib from Google/Android. As for the https://github.com/mikepenz/MaterialDrawer lib we do get a fully implemented, abstract drawer lib at the cost of a competing appcompat lib dependency which makes things complicated. As for the later that means a bit more work now but then we are "just" depending on the google compatibility libs we are already depending on to support material design all the way back to 4.0

tobiasKaminsky:
Ok, I understand.
Is there a big (visual) difference between the screenshots in #1552 and the current beta version?

I would also rather keep the deps small and implement the account switching by ourselves.
You said, that you tried a couple of hours to design the account switching.
Where is the problem? Maybe I can help you..?

AndyScherzinger:
The problem I had was with the display of the accounts in the list especially with their size, disctances between radio button icon and circle-letter icon. Besides that I will anyways try to implement it the "standard" way if one can call it that as it is shown in many apps flipping the menu item/list for a accounts list via the account displayed in the drawer's header.

Question is if @jancborchardt is fine with the then changed visuals and if the overall decision trying to implement it ourselfs is also fine with @davivel (which I guess so, since we would not use a lib for that :). So right now I am simply waiting for Jan's comment on this :)

jancborchardt:
Looks good to me. Our implementation actually has nicer whitespace and the name text properly left aligned with the other list item text. :)

AndyScherzinger:
So moving to the "new" implementation would be fine? The alignment "only" matters until we move to the toolbar instead of the actionbar which I would have done already but had an unsolvable  problem with handling the icon state for drawer/up/back so with 1.8 I simply went back to the (old) action bar. Whenever we move to the actionbar which I would strongly suggest we do then the drawer will hover above the bar and thus the alignment will be fine. Also using the toolbar has some more benefits like customizing and animation/hiding. But in order to do that I might then be in need of some support from @davivel since to me it seemed some event propagation during folder navigation messed of my toolbar icon state.

AndyScherzinger:
@all please see #1559 which is in development introducing a completely new drawer implementation also incorporating a material style account switcher implementation.

res/layout/drawer.xml
```diff
@@ -28,33 +28,10 @@
...
-                <!--android:textAppearance="?android:attr/textAppearanceLarge" />-->
-
-        <!--</LinearLayout>-->
-
-        <!--<TextView-->
-            <!--android:layout_width="fill_parent"-->
-            <!--android:layout_height="2dip"-->
-            <!--android:background="@color/list_item_lastmod_and_filesize_text" />-->
+        <TextView
+            android:layout_width="fill_parent"
```
davivel: Please, as a general rule, use "match_parent"; "fill_parent" is obsolete.

res/values-de/strings.xml
```diff
@@ -19,8 +19,7 @@
   </string-array>
   <!--TODO re-enable when server-side folder size calculation is available   
     	<item>Biggest - Smallest</item>-->
-  <!--TODO re-enable when "Accounts" is available in Navigation Drawer-->
-  <!--<string name="drawer_item_accounts">Accounts</string>-->
+  <string name="drawer_item_accounts">Accounts</string>
```
davivel: We can't include changes in translated languages through git, or will break daily merge of translations. Please, undo this change, let the equivalent in res/values/strings.xml.

src/com/owncloud/android/ui/adapter/NavigationDrawerListAdapter.java
```diff
@@ -117,93 +127,91 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
+                    public void onCheckedChanged(RadioGroup group, int checkedId) {
+                        // checkedId is the RadioButton selected
+                        RadioButton rb = (RadioButton) view.findViewById(checkedId);
+
+                        AccountUtils.setCurrentOwnCloudAccount(mContext,rb.getText().toString());
+                        notifyDataSetChanged();
+                        mFileActivity.closeNavDrawer();
+
+                        // restart the main activity
+                        mFileActivity.restart();
```
davivel: Mmm, I will need to check if this is not too "heavy".

davivel:
@tobiasKaminsky , I made some comments just in case.

@AndyScherzinger , is this included in #1559, or makes it obsolete?

AndyScherzinger:
#1559 renders this PR obsolete, but we should keep it until 1559 is merged and out in the wild (already in beta since end of last week)

davivel:
Sorry, I don't understand. Why should be keep this open if it's obsolete?

AndyScherzinger:
We can close it but I would like to keep the branch itself for now just in case I need to peek for some of Tobias' code. This shouldn't happen, but you never know.

davivel:
Oh, that's no problem.

Closing as subsumed/merged by #1559. Thanks, @tobiasKaminsky & @AndyScherzinger

PR #1176, opened 2015-09-25
Multi select

tobiasKaminsky:
As discussed in #899 

TASKS:
- [x] Code review @jabarros
- [x] Changes after CR @jabarros
- [X] [QA] Create 'multiSelect' branch in QA Repo @jesmrec
- [X] [QA] Merge 'multiSelect' branch in QA Repo @jesmrec
- [X] [QA] Create test plan  @jesmrec
- [X] [QA] Validate test plan @jesmrec @mcastroSG

QA:
- [x] Check updated icons https://github.com/owncloud/android/pull/1424

Open bugs/improvements:
- [X] https://github.com/owncloud/android/issues/1347  [FIXED]
- [X] https://github.com/owncloud/android/issues/1411  [FIXED]  ~~#1714~~ #1715
- [X] https://github.com/owncloud/android/issues/1412 [FIXED]
- [X]  https://github.com/owncloud/android/issues/1605 [FIXED]
- [X] CRASH: https://github.com/owncloud/android/pull/1176#issuecomment-225573680 [FIXED] 
- [X] CRASH:https://github.com/owncloud/android/pull/1176#issuecomment-225602576 [FIXED]
- [x] "Move" option incorrect in multiselection https://github.com/owncloud/android/pull/1176#issuecomment-228726233 [FIXED]
- [X] Multiselection of av. offline https://github.com/owncloud/android/pull/1176#issuecomment-228735066 --- moved to new issue #1726 [FIXED] @jabarros
- [X] Select all option https://github.com/owncloud/android/pull/1176#issuecomment-229010605 --- moved to new issue #1727
- [X] Folder moved into itself https://github.com/owncloud/android/pull/1176#issuecomment-229029762 @davivel @jabarros [FIXED]
- [X] download file icon is the same as available offline https://github.com/owncloud/android/pull/1176#issuecomment-229891594 @davivel [FIXED]
- [X] av. offline icon https://github.com/owncloud/android/pull/1176#issuecomment-230223452 @davivel [FIXED]
- [X] copy/move target folders multiselection https://github.com/owncloud/android/pull/1176#issuecomment-231035343 @davivel [FIXED] 
- [X] Lack of 'three-dots' menu in tablets: https://github.com/owncloud/android/pull/1176#issuecomment-231965406 [WONT] @davivel  --> see #1738
- [X] 'Wait a moment' dialog has been removed without protecting files/folder being updated - should be recovered https://github.com/owncloud/android/pull/1176#issuecomment-230287774 @davivel [FIXED]

PR #1540, opened 2016-03-18
Send multiple files with resized image if available

tobiasKaminsky:
depends on #1176 and #1044 
with this PR one can select multiple images and send them e.g. to whatsapp.
This will soon be also in beta.


davigonz:
This is possible but without resizing the images, is not what the users expect

PR #1542, opened 2016-03-20
Move file via uploader

tobiasKaminsky:
When sharing a file with owncloud it is now possible to choose between copy and move, just like in the upload from within our app.
But in despite of this the default is always copy (as e.g. no one wants to move whatsapp images to owncloud as this might break the app).
Of course it can happen that a file is moved during sharing that cannot be moved (e.g. a cached image), then it will just copy it and fails silently.

@AndyScherzinger can you help me with the layout? I tried to do the same as in the upload_files_layout.xml, but it does not really work :/

wronglayout

### DIFF
--- a/res/layout/uploader_layout.xml
+++ b/res/layout/uploader_layout.xml
@@ -17,24 +17,17 @@
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
  -->
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-	android:layout_height="wrap_content" android:orientation="vertical"
-	android:layout_width="wrap_content" android:background="#fefefe"
-	android:gravity="center">
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	android:id="@+id/upload_files_layout"
+	android:layout_width="fill_parent"
+	android:layout_height="fill_parent"
+	android:background="@color/background_color"
+	android:orientation="vertical" >
 
-	<TextView android:layout_width="fill_parent"
-		android:text="@string/uploader_top_message"
-		android:layout_height="wrap_content"
-		android:id="@+id/drawer_username"
-		android:textColor="@android:color/black"
-		android:gravity="center_horizontal">
-	</TextView>
-
-	<FrameLayout android:layout_height="fill_parent"
-		android:layout_width="fill_parent"
-		android:id="@+id/frameLayout1"
-		android:layout_below="@+id/drawer_username"
-		android:layout_above="@+id/linearLayout1">
+	<FrameLayout
+		android:layout_height="388dp"
+		android:layout_width="match_parent"
+		android:id="@+id/frameLayout1">
 
 		<ListView android:id="@android:id/list"
 			android:layout_width="fill_parent"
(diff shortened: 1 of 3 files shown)


res/layout/uploader_layout.xml
```diff
@@ -17,24 +17,17 @@
...
-
-	<FrameLayout android:layout_height="fill_parent"
-		android:layout_width="fill_parent"
-		android:id="@+id/frameLayout1"
-		android:layout_below="@+id/drawer_username"
-		android:layout_above="@+id/linearLayout1">
+	<FrameLayout
+		android:layout_height="388dp"
+		android:layout_width="match_parent"
+		android:id="@+id/frameLayout1">
```
przybylski: If you are never accessing this layout in code then you should remove this id.

res/layout/uploader_layout.xml
```diff
@@ -43,33 +36,75 @@
 			android:dividerHeight="1dip">
 		</ListView>
 
+		<TextView android:layout_width="fill_parent"
+            android:text="@string/uploader_top_message"
+            android:layout_height="wrap_content"
+            android:id="@+id/drawer_username"
+            android:textColor="@android:color/black"
```
przybylski: Isn't there a color/owncloudDarkTextColor (or something similar) resource in the project?

res/layout/uploader_layout.xml
```diff
@@ -43,33 +36,75 @@
...
 	<LinearLayout
-	    android:id="@+id/linearLayout1"
-	    android:layout_width="fill_parent"
-	    android:layout_height="wrap_content"
-	    android:layout_alignParentBottom="true"
-	    android:orientation="horizontal" >
+		android:orientation="horizontal"
+		android:layout_width="match_parent"
+		android:layout_height="wrap_content"
+		android:paddingTop="8dp"
```
przybylski: There are standard padding in metrics files

res/layout/uploader_layout.xml
```diff
@@ -43,33 +36,75 @@
...
+		android:paddingLeft="16dp"
+		android:paddingRight="16dp">
+
+		<RadioGroup xmlns:android="http://schemas.android.com/apk/res/android"
```
przybylski: This namespace is already defined at the root node

src/com/owncloud/android/ui/activity/Uploader.java
```diff
@@ -424,6 +428,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
     private void populateDirectoryList() {
         setContentView(R.layout.uploader_layout);
+
+        mRadioBtnMoveFiles = (RadioButton) findViewById(R.id.uploader_radio_move);
+        mRadioBtnCopyFiles = (RadioButton) findViewById(R.id.uploader_radio_copy);
+        mRadioBtnCopyFiles.setChecked(true);
```
przybylski: Doesn't it override user decision when user de-selects this radio button and rotate the phone?

przybylski:
Review done

przybylski:
@tobiasKaminsky about your layout problems. You need to make the weight of file list 1 so it would stretch to the whole screen. 
As a design i would remove "choose upload folder:" view and move that to actionbar title.
Since this is triggered via system and chooser user already knows what app is he in, We don't need to take precious space and announce everywhere that he is in ownCloud (he knows that :) but that just my two ¢)

res/layout/uploader_layout.xml
tobiasKaminsky: There is textColor, which is also black. So I am using this.

res/layout/uploader_layout.xml
tobiasKaminsky: I am not that good with layouting. What do you mean?
Maybe also @AndyScherzinger can do this part.

src/com/owncloud/android/ui/activity/Uploader.java
tobiasKaminsky: No, but it will overwrite it again when changing a directory.
So I have to fix this.

tobiasKaminsky:
Thanks for the hint with weight=1.
Regarding the title and remaining layout: I left everything unchanged.
But it is a really good idea to change this.
Is this ok for you, @jancborchardt?

jancborchardt:
Yeah, @przybylski’s suggestion sounds good. :)

jesmrec:
Native uploader available. Closing this,

Issue #1432, opened 2016-01-12
Move Settings to bottom of sidebar, move Log to Advanced settings


jancborchardt:
First one to separate the settings from the actually daily useful filters. And then in the beta, the log entry should also be in the settings, advanced section. (Which I described in #1431)

AndyScherzinger:
@jancborchardt should we really move the log entry? The log entry is only shown:

* in the beta: so users can provide us with logs to support finding bugs 
* in the regular app when installing a debug build: so it is a dev feature only

both kind of make sense being accessed easily without having to navigate through the setting, wouldn't you agree? I am fine either way since being a developer means I use the IDE's console anyways. So if you think is simplifies the UI we can move it to the settings. No problem! :)

As for the settings item I'll have to check since the list is basically a menu rendered as a list. So it should be possible by removing it from the list and put in a separate bottom aligned item just for settings :)

AndyScherzinger:
Would it be finde with you to move to the new support lib navigation view? That is the view to render the menu items, example: http://antonioleiva.com/navigation-view/

jancborchardt:
Yeah, the reason I'm not happy with the log entry there is that the main interface elements of the beta should look exactly like the normal release. :)

When an actual error occurs, the log is presented anyway. And when you want to get the log, the extra step into settings is ok.

AndyScherzinger:
Oh, okay didn't know that. So yeah, when the log pops up anyways in case of an error than we can move it into the settings. :)

AndyScherzinger:
@jancborchardt just a general question wouldn't it make sense that if we move the menu entry for log to the settings screen to also make it available in the standard release? If so we might need some discussion with @davivel and @masensio on when to purge the log. For background processes like instant uploads etc. it might make sense to provide access to a log for reporting issues here. On the other hand but that again might be a privacy discussion would be automated error reporting to a central server like crashlytics which would be a whole other level of discussion, so I would postpone that one.

jancborchardt:
@AndyScherzinger sure! Then it could be in the »Advanced« section on the bottom, with PIN, Storage path and Cache size.

AndyScherzinger:
Logs has been moved to the settings screen, please see: #1552

AndyScherzinger:
Settings would be slightly seperated in the menu with the latest Drawer PR #1559

davivel:
> wouldn't it make sense that if we move the menu entry for log to the settings screen to also make it available in the standard release?

[...]

PR #1552, opened 2016-03-22
moved logs from drawer menu to the settings

AndyScherzinger:
As discussed in #1432 the Logs menu item from the drawer has been moved to the settings screen.

@jancborchardt I moved it to the "more" section instead of the "advanced" section - is this correct or should I move it to the "advanced section? 
device-2016-03-22-145222

### DIFF
--- a/res/values/setup.xml
+++ b/res/values/setup.xml
@@ -53,7 +53,7 @@
     <bool name="imprint_enabled">false</bool> 
     <bool name="recommend_enabled">true</bool>
     <bool name="feedback_enabled">true</bool>
-    <bool name="logger_enabled">true</bool>
+    <bool name="logger_enabled">false</bool>
     <string name="url_help">http://owncloud.com/mobile/help</string>
     <string name="url_imprint"></string>
     <string name="mail_recommend">"mailto:"</string>
--- a/res/values/setup.xml
+++ b/res/values/setup.xml
@@ -53,6 +53,7 @@
     <bool name="imprint_enabled">false</bool> 
     <bool name="recommend_enabled">true</bool>
     <bool name="feedback_enabled">true</bool>
+    <bool name="logger_enabled">true</bool>
     <string name="url_help">http://owncloud.com/mobile/help</string>
     <string name="url_imprint"></string>
     <string name="mail_recommend">"mailto:"</string>
--- a/res/values/drawer_resources.xml
+++ b/res/values/drawer_resources.xml
@@ -26,7 +26,6 @@
         <!--<item>@string/drawer_item_on_device</item>-->
         <item>@string/drawer_item_uploads_list</item>
         <item>@string/actionbar_settings</item>
-        <item>@string/actionbar_logger</item>
     </string-array>
 
     <!-- Nav Drawer Content Descriptions -->
@@ -37,7 +36,6 @@
         <!--<item>@string/drawer_item_on_device</item>-->
         <item>@string/drawer_item_uploads_list</item>
         <item>@string/drawer_item_settings</item>
-        <item>@string/drawer_item_logs</item>
     </string-array>
 
 </resources>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -27,7 +27,6 @@
     <string name="drawer_item_on_device">On device</string>-->
     <string name="drawer_item_settings">Settings</string>
     <string name="drawer_item_uploads_list">Uploads</string>
-    <string name="drawer_item_logs">Logs</string>
 	<string name="drawer_close">Close</string>
     <string name="drawer_open">Open</string>
     <string name="prefs_category_general">General</string>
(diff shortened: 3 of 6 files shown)


AndyScherzinger:
@jancborchardt I can also move the settings item in the drawer to the bottom, or separate it with a line like ather apps do it. Whatever floats your boat :wink:

AndyScherzinger:
Google apps just separate it with a line:
Gmail
device-2016-03-22-150703
Drive
device-2016-03-22-150808
Hangout
device-2016-03-22-150738

Dropbox - top right
device-2016-03-22-150846

Amaze file manager like Google bat as a sticky footer
device-2016-03-22-150908

AndyScherzinger:
@jancborchardt having done a local implementation using the Support Library for the drawer's content implementation the menu items would look like this...
device-2016-03-22-200828
with an item "active":
device-2016-03-22-200848
The difference is very, very subtile since the unactive grey is quite close to the primary color (dark blue).
The Drawer width has also increased to our actual one since this is the standard with in material...

What do you think? Moving on with this has it's benefits since checked/unchecked state, items via native menu.xml files/groups would then be available :)

jancborchardt:
I'd say either we stick it to the bottom or we leave it as it is. The line just increases the focus on settings more, which is not what we want. ;)

AndyScherzinger:
Okay :9 So when the reliable uploads are in place (on master) with the new drawer it would then look like this:
device-2016-03-22-224635

AndyScherzinger:
I removed the new drawer implementation here and started a new PR for the drawer over here: #1559 
So this PR is basically a move of the logs to the settings screen and a removal from the drawer so you can forget about the latest screenshots since this won't be part of this PR anymore.

AndyScherzinger:
@rperezb @jancborchardt @davivel @tobiasKaminsky This is DONE and can be reviewed...

tobiasKaminsky:
:+1: code is fine!

AndyScherzinger:
pinging @davivel and @malkomich minor change, freshly rebased, community reviewed, resolving  #1432 raised by @jancborchardt

jancborchardt:
cc @davivel @MTrichards @rperezb please review.

rperezb:
@jancborchardt @AndyScherzinger we will check this after the team week thx

jabarros:
:+1:

res/values/setup.xml
```diff
@@ -53,6 +53,7 @@
     <bool name="imprint_enabled">false</bool> 
     <bool name="recommend_enabled">true</bool>
     <bool name="feedback_enabled">true</bool>
+    <bool name="logger_enabled">true</bool>
```
davivel: Why is this here?

davivel:
:-1: ; please, see my question

res/values/setup.xml
AndyScherzinger: For the same reason the other are there ;) --> with this flag you can decide if the log part will be rendered in the settings screen or not. We can take it out if you want to. I just thought like with imprint, feedback etc. you would like to be able to have to feature toggle for logging also since I discussed the move to the settings screen with @jancborchardt and we agreed that after moving it to the settings screen it will also be visible in the official release.

AndyScherzinger:
:wink: please see my answer :smiley:

res/values/setup.xml
davivel: Thanks, I understand now. I totally oversaw the intention of making the menu available in the official release of the app, not only in debug mode. In this case, having this option here makes a lot of sense.

The point is, I'm not so sure we should have this is the official app. I will follow this up in #1432

davivel:
Thanks, @AndyScherzinger .

Blocked until a decision is taken about make the logs available in release mode or not.

cc @rperezb

jancborchardt:
Ok – let’s make it available in debug mode / beta only maybe for now. And just move the place of it. We’re doing 2 separate steps here (moving, and making it visible for normal users), that’s the problem.

davivel:
Thanks, @jancborchardt.

Proposal: let's keep the option in ```settings.xml```, but with default value to ```false```. Besides, let's modify the condition in ```Preferences.java``` so that the option is available if ```(loggerEnabled || BuildConfig.DEBUG)```. Seems to me more comfortable.

@AndyScherzinger , what do you think?

AndyScherzinger:
Yeah, sounds good to me! I can make the change during the next days! :)

AndyScherzinger:
@davivel added the changes as discussed :)

davivel:
Great job :)

:+1:

@jesmrec , your turn.

jesmrec:
@davivel @AndyScherzinger @jabarros 

Approved 👍

jabarros:
Cool, so it is prepared to be merged :)
@AndyScherzinger could you rebase it on master and merge it by yourself? Thanks!!

AndyScherzinger:
DONE! :+1:

rperezb:
Thanks!
Good to know that there is a new variable on the setup file ☺️

Issue #1556, opened 2016-03-22
Use toolbar instead of actionbar where applicable


AndyScherzinger:
Like the title says use the toolbar - Google's new, to be used implementation of the top bar instead of the action bar which can be considered legacy. 

**Motivation**
The actionbar is simply a bar which can have a menu, search widget and in newer AppCompat versions also a drawer implementation but isn't as flexible as the toolbar.

**Reasons and benefits**
The toolbar can be customized on many ways like but not limited to:
  * adding any kind of views to it, like the progressbar which would be nice especially when 
    * we integrate the multi select branch (no dark blue line below the then grey bar)
    * no dark blue line below the bar when the drawer is opened
  * animate it, like scrolling it away on a list/recycler view which seems nice on the file browsing views
  * the material drawer can be full height, like mentioned in #1210 because this is only possible for the toolbar not the action bar (it can be done with hacks though)

I would like to work on this in case we agree this is worth implementing which I strongly suggest to make the bar future proof and also to be able to implement the above mentioned gimmicks.

So pinging @davivel @jancborchardt  @malkomich @tobiasKaminsky and @przybylski for feedback on this matter.

davivel:
Great job here, and great presentation, @AndyScherzinger .

Just missed a section about "drawbacks" ;)

davivel:
Merged, will be released with version 2.2.0

AndyScherzinger:
:small_airplane:  **Awesome collab!**  :rocket:

PR #1493, opened 2016-03-03
Reliable uploads -  UPLOADS VIEW

davivel:
@owncloud/android-developers 

Time has come.

From my POV, this is finally ready to be tested by QA (cc @jesmrec).

Many hands contributed here, starting with the awesome effort of @LukeOwncloud. Sounds wise summoning many eyes to review it. So, any comment about the code will be welcome.

Though many things have been rewritten and considerably cleaned-up, sure you will still find many points that could (and should) be refactored. I can't promise that all of them may be attended in this PR, since we really, really, need to release this with next version 1.9.2, that should be very near in time to the release of OC 9. Evolution of testing is the leading criteria right now. 

This means: if we think we can refactor something in a safe manner without introducing an important risk for QA, we will do. If bugs appear and the code to fix is also subject of comments in the code review, we will apply the comments. 

But if change requests are big enough to require QA starts the testing all over again, we will have to wait for a better chance.

This one should not be too far in time. This PR lets aside an important topic that was originally addressed by @LukeOwncloud , the improvement of automatic retries. Right now instant uploads are retried when Wifi connection is established, in a similar way to the operation of the currently released version 1.9.1 [1]. But other convenient improvements will be addressed specifically in the short term [2], not in this PR. And that should be a good moment to improve the code of uploads further.

Come on Barbie, let's go party!

[1] You may notice some light improvements, but that's a side effect of the many changes done; was easier let it that way that adding artificial limits. Have fun finding them :)

[2] If you don't want to wait and prefer to implement them, go for it. Just mention me to let me know, and I'll try to help.

----

BUGS FOUND AND ENHANCEMENT:

- [x] Failed uploads can not be retried in another account #1497
- [x] Failed uploads are not retried by the button #1498
- [x] Deleting an account does not clear the current uploads #1499 
- [x] Killing the app leaves uploads in an inconsistent state #1500
- [x] Wording issues in uploads view #1501
- [x] Manual retries are "delayed #1502
- [x] Uploads in SAML requests authentication twice #1503
- [x] Current upload should be in the top of the uploads view #1511
- [x] Only current upload in progress naming as "Uploading" #1520
- [x] Retry failed uploads start uploading by the bottom #1521
- [x] Credentials requested on each failed upload #1522 
- [x] Uploaded files are possible to uploaded it again #1523
- [x] Title on top bar in upload view is shown sometimes owncloud instead of Uploads #1524 
- [x] Failed Uploads set as wrong status "Waiting for wifi connectivity"  #1525 
[...]

PR #1559, opened 2016-03-23
#1556 - Toolbar, new Drawer and new Account Manager

### LINKED ISSUE #1210: Full-height material drawer?
So I looked through the couple issue and PR threads pertaining to the Material redesign, and all I can tell is that at one point in the screenshots the drawer went at least up to the status bar, but then it started stopping at the app bar. Can't seem to find an issue pertaining to this, so...

### Expected behaviour
The drawer goes from the nav bar, or bottom of screen with hardware buttons, to the top of the screen, or at least to the status bar.

### Actual behaviour
The drawer stops at the app bar, almost certainly intentionally

### Steps to reproduce
1. Open app drawer, using hamburger button or edge swipe.

### Environment data
Android version: 5.1.1
Device model:  OnePlus One
Stock or customized system: Customized (CyanogenMod instead of Cyanogen OS)
ownCloud app version: 1.8.0
ownCloud server version: 8.1

Are logs relevant with intentional behavior?

Sorry if this is the wrong place, I wasn't quite sure where this goes.

### LINKED ISSUE #1383: Quick Account-Switching in navigation drawer

AndyScherzinger:
#This is the PR for #1556 - see todos and comments

At the moment my implementation is stable and available in the beta.

At the moment I also added the uploads menu item here as a dummy since this PR will be merged after the reliable uploads has been released so you all can already see how it will look like.

Material Design spec for side nav drawer: 
http://www.google.com/design/spec/layout/structure.html#structure-side-nav

----
What has been done already:
- [x] integrate with toolbar and get navigation right
- [x] integrate new drawer implementation where menu items do have a checked state
- [x] FIXES #1210
- [x] FIXES #1383 
- [x] Successor of  #1537
- [x] integrate material style account switcher
- [x] implement colored circle/letter user icon with the right diameter in alignment to the oC server's implementation (this has already been implemented by @tobiasKaminsky an the account switcher branch)
- [x] remove old drawer implementation.
- [x] propagate account manager changes to the drawer including active account swapping
- [x] check for necessary fix for #1572 after #1493 has been merged to master 
- [x] modify header design with larger user icon and menu item aligned switcher text and smaller overall drawer header height
- [x] implement account switching in header via avatar icons (always remember and show the last 3 accounts)
- [x] test what happens for really, really, really long account names with the menu items
- [X]  heavy testing for all Activities to make sure they all call the toolbar setup methods after they set the view content <<<< I did a lot of this during the code review; @jesmrec , not sure if you have more to do here. <<<< Present in test plan 

RELATED BUGS:
- [X] https://github.com/owncloud/android/issues/1657    [NOT REPRODUCIBLE]
- [X] Crash with no account: https://github.com/owncloud/android/pull/1559#issuecomment-251686981  [FIXED] 
- [X] Duplicated account in navigation drawer: https://github.com/owncloud/android/pull/1559#issuecomment-251689167 [FIXED] 
- [X] Inconsistent actions in options menu: https://github.com/owncloud/android/pull/1559#issuecomment-251692759 [OUT OF SCOPE]
- [X] Long names in manage accounts view https://github.com/owncloud/android/pull/1559#issuecomment-251729870 [FIXED]
- [x] Duplicated avatars in drawer https://github.com/owncloud/android/pull/1559#issuecomment-251933389 [FIXED] 
- [X] Avatars are not updated https://github.com/owncloud/android/pull/1559#issuecomment-251937443 [FIXED]
- [X] Drawer under the toolbar in multiselection https://github.com/owncloud/android/pull/1559#issuecomment-252194713 [FIXED] 
- [X] App closed if tap on back in login view https://github.com/owncloud/android/pull/1559#issuecomment-252231496 [FIXED]

POSSIBLE IMPROVEMENTS:
 - [X] Confirmation dialog to remove accounts: https://github.com/owncloud/android/pull/1559#issuecomment-251691507 [FIXED] 
 - [X] Displaying account info in the list using the same way as the current account https://github.com/owncloud/android/pull/1559#issuecomment-251735447 [WONT FIX]
 - [X] In drawer and manage accounts, either username or displayname, but both the same set up  https://github.com/owncloud/android/pull/1559#issuecomment-251900859 [FIXED]
 - [X] Deleted accounts "block" login account https://github.com/owncloud/android/pull/1559#issuecomment-251943423 [OUT OF SCOPE]
 - [X] Current account is not needed in the account switcher https://github.com/owncloud/android/pull/1559#issuecomment-251936645 [FIXED]
 - [X] Switch account by clicking on an account in "Manage Accounts"  https://github.com/owncloud/android/pull/1559#issuecomment-252567447 [FIXED]


AndyScherzinger:
Screenshot of the **initial** implementation showing the checked state. At the moment the drawer has the standard material width which is quite huge.
device-2016-03-23-162533

We can make that any size though :grin: 
device-2016-03-23-162847

Checked menu item state:
device-2016-03-23-163052


So looping in @jancborchardt for design, @davivel for possible discussions about integration at some point and @tobiasKaminsky for account switching (I'll ping yu again as soon as I have integrated the switcher implementation UI-wise).

AndyScherzinger:
small update, soon to be committed showing the account switch in the drawer header which is now 140dp fixed and will be 140dp and 164 for sw360dp, see also https://medium.com/android-news/navigation-drawer-styling-under-material-design-f0767882e692#.76h6eeqi9
device-2016-03-23-193334

tobiasKaminsky:
@AndyScherzinger account switching on the branch #1537 should be stable enough to merge it into this branch.
If you have problems, please let me know

AndyScherzinger:
@tobiasKaminsky thanks for the info! :+1: I'll ping you when I'll have the UI for it up and running and will need your help merging since I really don't know my way around git that much.

tobiasKaminsky:
You can also ping me and I do the merge. 
This should not be that hard ;)

AndyScherzinger:
Sounds perfect, great collaboration as always! :tada:

AndyScherzinger:
Screenshots of the actual progress.
* view can be switched between accounts and standard menu
* account add can be triggered but doesn't end up refreshing the account list (yet)
* manage account is just a placeholder atm.
* account switching is not implemented yet
* user-letter circles implemented with color coding, open issue: slightly misaligned, so whoever is good at math/canvas feel free to jump in. items are max sizes for the menu items, size can't be raised or at least I haven't found a way to do it...

Screenshot account list
device-2016-03-24-012519

Screenshot standard menu
device-2016-03-24-012536

AndyScherzinger:
Code reviews should probably wait until @jancborchardt provides feedback on the visuals :)

AndyScherzinger:
...enough for tonight... I now also implemented the account switching itself which has mostly been a move of all account related methods from the file activity to the drawer activity, plus some protected getter methods and binding the account clicks to the account switch.

AndyScherzinger:
**DONE**
* view can be switched between accounts and standard menu
* account switching is implemented
* drawer status is preserved during screen rotation

**OPEN**
* account add can be triggered but doesn't end up refreshing the account list
* manage account is just a placeholder atm.
* user-letter circles implemented with color coding, open issue: slightly misaligned, so whoever is good at math/canvas feel free to jump in. items are max sizes for the menu items, size can't be raised or at least I haven't found a way to do it...

tobiasKaminsky:
Cool stuff!
Can I therefore close my other PR #1537?

AndyScherzinger:
I guess so, but I think we should keep the branch until this PR is done just to make sure. Sorry that I didn't incorporate any changes from the other PR but down the road it seems easier/faster to just keep going on this branch.

tobiasKaminsky:
No problem :+1:

AndyScherzinger:
**DONE**
* account manager to change password and remove accounts
* account manager integrated
* accounts created via drawer menu get propagated back to the drawer incl. active account

**OPEN**
* added account via drawer doesn't swap the account for the file list
* account list changes via account manager don't get propagated back to the drawer / file list (need to call for results, send a dirty flag and update drawer and swap account in case active account changed)
* user-letter circles implemented with color coding, open issue: slightly misaligned, so whoever is good at math/canvas feel free to jump in. items are max sizes for the menu items, size can't be raised or at least I haven't found a way to do it...

Screenshot of the account manager
device-2016-03-25-010721

AndyScherzinger:
@jancborchardt any feedback on the visuals?

tobiasKaminsky:
You can remove all account related lines from Preferences as they are still in master.
If you need some code you can easily compare your file with branch master within androidStudio and in the resulting compare screen select those lines you want to have back.

AndyScherzinger:
:tada: The implementation is basically done and is ready for test, two probable issue still open, see below :tada:

**DONE**
* view can be switched between accounts and standard menu
* account switching is implemented
* drawer status is preserved during screen rotation
* account manager to change password and remove accounts
* account manager integrated
* accounts created via drawer menu get propagated back to the drawer incl. active account
* added account via drawer now swaps the account for the file list
* account list changes via account manager get propagated back to the drawer / file list
* house keeping - removed all account relevant code in Preferences
* account add can be triggered and refreshes the account list (in account manager)
* raise the size of the account switcher hitbox in the drawer for better usability

**OPEN**
* user-letter circles implemented with color coding, open issue: slightly misaligned, so whoever is good at math/canvas feel free to jump in. items are max sizes for the menu items, size can't be raised or at least I haven't found a way to do it...
* add account via drawer and then hitting the back button on the setup screen moves the app to the background. No idea why since this doesn't happen when you do the same thing in the account manager

AndyScherzinger:
Any code review and feedback is highly welcome! :smiley:

AndyScherzinger:
Upload file screen crashes...  Need to check that when I am back at my PC next week.

tobiasKaminsky:
@AndyScherzinger I have fixed this and merged it into beta :)

AndyScherzinger:
@tobiasKaminski Awesome! :smile: :tada:

AndyScherzinger:
Share screens also crash, will look into it the next days...

AndyScherzinger:
False alarm, this is already covered by your fix which I did not have on me phone yet. I will still habe a look since the share dialogs shouldn't inherit the drawer Activity imho.

AndyScherzinger:
@jancborchardt any preference on the positioning/size of the active/current account icon in the drawer?

Right now:
device-2016-03-29-105307

usually in other apps (I just resized this in paint, so please ignore the blurry icon...)
device-2016-03-29-105307 - kopie

AndyScherzinger:
I personally vote for the second version, since it is better to balance the alignments and support longer server urls version 1 looks a bit off.

AndyScherzinger:
@tobiasKaminsky did you push the merge to the beta branch on GH?

tobiasKaminsky:
@AndyScherzinger not now.
I tried, but it ended in a big mess ;)
The beta is so far away from master branch that it is really hard to include new branches as some classes have 4-5 branches that want to merge the same lines...

AndyScherzinger:
Allright, no worries @tobiasKaminsky we had to end up there at some point :( So pushing for the PRs to get merged and then probably restart beta releases.

@davivel just fyi with commit https://github.com/owncloud/android/pull/1559/commits/814ee2ff55927defa8ce6a9bd94af74f0f11003a the progress bar in the toolbar is now aligned with the color coding of the action bar, so the bar will have whatever the color that has been defined for the action bar :) This should make branding a bit more consistent for the toolbar :)

tobiasKaminsky:
@AndyScherzinger Of course you can try it ;) I think it can be easier for you as you know your changes better than me.
I will do the same tomorrow afternoon if you are not successful.

AndyScherzinger:
@tobiasKaminsky it gave it a try today and also failed since this change here is quite massive from a structural point of view but I think the activity hierarchy will benefit from it in the long run...

AndyScherzinger:
@davivel can you help me with this PR after the merge of #1493 or the release 1.9.2 respectively? I pulled in the PR #1493 and fixed everything so it should then just be a rebase on the master. But I still don't get rebases (right) so I would end up ruining this PR. As fort any conflicts with the relaible uploads changes since I did the local merge and push here already: This PRs changes always have to win + plus some possible fixes afterwards (I will take care of that), due to API changes by me.

AndyScherzinger:
@jancborchardt any input? :nail_care:

jancborchardt:
Here’s my proposal on the layout:
0061b5cc-f59e-11e5-9d75-cbd4c1b3a8ba

- Not going so crazy on the whitespace when we don’t have a background picture, or image-based switcher on the top right.
- Text properly left-aligned with the text of the navigation items.
- Slightly bigger profile image.

jancborchardt:
@AndyScherzinger so what do I see when I click on the name / dropdown? Have a small dropdown open for the switcher, with an »Add account« option? That would be good, instead of a full-screen window which rips you out of the flow.

AndyScherzinger:
You end up with this :grin: the switcher basically swaps the "menu" for an account list plus certain menu items für accounts in general (just like the GMail app does)
device-2016-03-30-164926

AndyScherzinger:
HAH! :tada: 
The rebase seems to have worked :smiley:

AndyScherzinger:
> 
- Not going so crazy on the whitespace when we don’t have a background picture, or image-based switcher on the top right.
- Text properly left-aligned with the text of the navigation items.
- Slightly bigger profile image.

Will do, even though looking through the Google screens http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-content the do it differently always offering the Account switch in the drawer's header. So probably as a follow up PR I will also implement the image-based switcher and would then go for the "standard" layout Google shows in its guide.
When doing so I would then also raise the header size again since the Google apps also have standard sizes, see https://medium.com/android-news/navigation-drawer-styling-under-material-design-f0767882e692#.2dcgj2v4t with hardcoded 140 dp / 164 dp height.

jancborchardt:
Ok, having the header size big is fine for me. The alignment of the text is important though.

Also, I would remove the »Manage accounts« option (because I guess it will redirect you to settings), completely remove the accounts stuff from settings, and then add a settings cog in that sidebar for each account to reach the account-specific connection settings.

jancborchardt:
And yeah, if we make the layout look exactly like this: http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-content with the profile pictures, that would be pretty perfect!

cc @DeepDiver1975 @rullzer for the API of profile pictures from the server.

AndyScherzinger:
It redirects you to a user management screen I created where you can create new users, delete users and change password they way Google apps do it and I removed the user management from settings since it doesn't belong there anyways. Please see screenshot in this thread for that. So the "Manage" is here to stay :P https://github.com/owncloud/android/pull/1559#issuecomment-201081043

AndyScherzinger:
Just to explain is in short: The drawer (besides the switching via circle avatars) including the account switcher works exactly like the one from GMail, Inbox, etc. (in case you have that app installed) :tada:

rullzer:
I know we all want avatars but currently there is no proper API for this and it is not scheduled for 9.1. So no promises. But yeah it will be awesome to have...

jancborchardt:
@AndyScherzinger the extra user mgmt screen is not really needed though – you can create new users via the »Add account« button. Then changing passwords and deleting users should be possible by clicking the settings button I proposed. :)

jancborchardt:
@AndyScherzinger cause that list already is there, in the dropdown. No need to have a separate view as it overcomplicates things / just duplicates the list mostly.

AndyScherzinger:
WHAT?!
If the user list is not in the settings screen anymore, how would you delete users and change passwords. You cannot do that within the drawer itself and it would not make any sense to me since the drawer is a (rather list based) menu... So there is no settings icon within user lines, the user line items are menu items so they are single actions as in change user context. One absolutely must navigate to a user management screen...REALLY!

AndyScherzinger:
And it is neither a dropdown, nor a list it is a menu ....

tobiasKaminsky:
Yeah, I remember also that we have discussed it this way:
Navigation drawer is just for switching accounts, editing/deleting will stay in settings as it is a not so often used function.

AndyScherzinger:
The only thing that would make any sense at all would be to have 
* an item for add account
* an item for remove account (the current one)
* an item for change password (for the current account)

And still this would totally differ from ANY Google app in the wild while you then would still and always have the Standard settings icon at the end of the list since that is defined in the material design guidelines for the drawer menu.

jancborchardt:
Well, there would be a settings cog (as said above) for the current account, next to the dropdown arrow. Clicking on that one would open a separate view which basically looks like the account connection view. (With additional deletion option.)

We only need one user list. And that’s the one directly in the sidebar. I can make quick mockups if needed. :)

AndyScherzinger:
Yes please, but still are you kidding me?! Besides the settings icon in the drawer header which is imho clutter you simplify what exactly? 
* It is not more intuitive 
* It is contrary to any other proper material design app
* It takes the same amount of clicks
* It even complicates things since I now would have to switch users in case I would want to change the passwords of several account. or would like to remove more than one account
* The only app that has a settings icon in the drawer header is dropbox and they navigate you to the settings not an account management interaction, which isn't according to material guidelines but due to the icon at least expected to some extend. So now we try to create a third variant ?????

jancborchardt:
:grin: ok, I will just create a mockup and maybe I’ll notice myself that it’s bullshit. ;) But at the moment it seems like it would vastly simplify the current state.

AndyScherzinger:
Ok :) I am happy to get tough otherwise!

tobiasKaminsky:
@AndyScherzinger relax :) :) 
I am sure that we come to a conclusion that we all like, as usual.

And I think you are right that we should take a look at other material designed apps.

AndyScherzinger:
I am fine :)  I just tend to get emotional when I strongly care about things and I really care about this one. As always hope to not have been offensive. 

The drawer implementation has been my 4 days vacation project, so I deeply care for thing one a bit more than I already care about my other PRs...  :(

AndyScherzinger:
Uhr, little typo,  I meant taught... Sry!

tobiasKaminsky:
Regarding the avatars: 
I have a code snippet that can load the avatar for the user, store them in the cache and show them in the account drawer list.
Only drawback is that there is no way to figure out that the avatar is up to date.
Maybe we could update the avatars when entering settings / editing one account.
I doubt that the avatars are changed very often.

When there is a proper API as @rullzer mentioned we could change it.
But I think it would be cool to show the avatars.
@jancborchardt @davivel it is your decision :) If yes, I will share the snippet.

AndyScherzinger:
Sounds good to me, even though it is a temporary hack until the Api is present. I could then implement the avatar based switch in the header :D

tobiasKaminsky:
@jancborchardt should we provide you an preview of this PR as a runnable app? (no beta, just to install it instead of the original one to test the behaviour?)

@AndyScherzinger I have tried it on an emulator. This is really awesome.
Just my opinion:
The icon on the right of the drawer (triangle up/down) is somewhat misleading. But I see that other apps doing the same, so I guess it is okay. My idea would be to change these icons to accounts / menu. (ref for the current implementation: https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0Bzhp5Z4wHba3WUpVTktSZWk1YjQ/patterns_navdrawer_behavior_temporary2.png)

Maybe remove the setting button in the account drawer view? As this view is solely about accounts and nothing more or less? Otherwise it is kind of duplicate?

tobiasKaminsky:
@AndyScherzinger my implementation of fetching the avatar is in branch "avatar".
As it is updated to master you can simply merge it into this PR and then delete the branch.

AndyScherzinger:
Sweet! 

Can you give some more details on changing the icon from accounts/menu? 

The actual implementation works like the one you referred to. So that has been my argument, simply that the drawer and switcher behave exactly like Google’s own apps.

tobiasKaminsky:
My idea was to show an account symbol when the normal drawer list is shown:
http://uxrepo.com/static/icon-sets/google-material/png32/256/000000/android-supervisor-account-256-000000.png
With this symbol it is clearer that you can change the accounts there.

On the account screen I just would show the normal hamburger symbol.

AndyScherzinger:
You mean the triangle which mimics a dropdown in the drawer? So if I click on the accounts icon you referenced what should then be displayed when the accounts list is shown in the drawer? The Hamburger? 
Imho the hamburger is associated with opening the drawer so I wouldn't use it for anything else. Still besides all the discussions I still don't understand why a standard look and behavior like in other proper material apps is now considered a bad UX thingy.

Can you do a mockup? I am probably just not getting the point here :crying_cat_face:

tobiasKaminsky:
hmm.. you are right, the hamburger is for showing the navigation drawer.
Then we can stay with the way you have done it!
Especially as you are right with the hint that other material apps do it the same way!
(I should just stay with coding and not with trying to enhance UX...)

AndyScherzinger:
Na, please keep going for UX too! Imho this is a very important discussion and the solution benefits from it. I am also no UX expert that is @jancborchardt, I am open minded and flexible here, but I am also having a strong opinion about it.

So I hope we end up having a nice and sweet solution for the navigation matter which makes sense to us all and it is more than just the "like all other apps" argument, which is valid but there should be better arguments than that.

Any news from you @jancborchardt? I hope I didn't out you off! :flushed:

AndyScherzinger:
I fixed the setting of the active menu item state and would like to get back to the drawer header discussion and on why why simply not implement it in alignment with the material design guidelines:
https://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-specs

AndyScherzinger:
So for local tests I checked out @jancborchardt's idea for the drawer header and the mine (imho the material design guideline):

*Jan:*
device-2016-04-04-161703

*Andy:*
device-2016-04-04-171029

The later basically had the standard 16dp whitespace around everything (also between avatar and text, which seems larger due to the fact that my username does not contain capital letters) and I replicates Googles guidelines to the max. The local implementation also slightly raises the drawer-header size and avatar size on tablet while keeping the whitespace in alignment with the material design style guide.

AndyScherzinger:
@jancborchardt one thing that came to my mind as an argument to keep the account management screen is that we could then incorporate more information there, for example displaying user quota in the account list screen (quota calls in the oC android lib to be merged first) and probably other stuff too :)

AndyScherzinger:
@tobiasKaminsky now I need your help :) I have integrated your avatar fetching code and changed it so it falls back to the 1-Letter username avatar generation in case no avatar is present and I also crop the avatars to be circular :) - Two thing left open I don't really know how to solve at the moment:

* Do the same thing for the account list: https://github.com/owncloud/android/pull/1559/files#diff-154af4ff2566e264dcca122072d20ae7R293
* dynamic sizing: I need to be able to generate the avatar in 3 different sizes
  * header cellphone
  * header tablet
  * account list

Work in Progress so far for you and @jancborchardt 
device-2016-04-05-110133

Next up will be the account switching via avatars in the drawer header as seen in the drawer screenshots from Google's Material Design guide and also mentioned by Jan: https://github.com/owncloud/android/pull/1559#issuecomment-203943273

Just to make sure we don't implement stuff that will be rejected in the end. Any feedback @davivel on the avatar loading mechnism implemented here by @tobiasKaminsky ?

AndyScherzinger:
Account switching via header avatars is now in place:
device-2016-04-05-162026

You can already click on the mini avatars which will trigger an account switch.

Open issue 
* as already mentioned before is that the generated avatars aren't dynamic regarding their size so the get cropped at the moment. The fetched ones from the server are "real" images and get resized.
* the avatars in the header are rather static in the way that they are chosen from the account list and in the order in which they are received from the account manager. Thus: The bis avatar is the current one chosen by the user, the small avatars to the right are always the first to from the account list that are not equal to the currently active one.

tobiasKaminsky:
Looks really great so  far.

I do not understand why the generated avatars (with the first letter) are not dynamic.
You have the function createAvatar with radiusInDp, so this should be enough, or?

The real avatars in the drawer list are just missing due to time, or do you have there problems?

Your second "issue": I do not understand the problem...

AndyScherzinger:
I kind of fixed the dynamic sizing with the latest commits but didn't get it quite right yet. It looks fine on xxhdpi but not on other resolutions yet.

The second problem is that I choose the avatars to be displayed from the account list, so they are always the same. For example if you have for accounts, the list always has the same order, so whichever is the current account the other two are always picked from the list and thus are always the same two, so the "fourth" account will never be one of the small avatars.

Another problem which I just found is that all the editText fields are messed up. for some reasons on 4.x versions of android the text is always white, some for the drawer menu items and their icons. So I will have to further investigate what is wrong there. I checked the master and the problem isn't present so it is something I introduced on this branch...

AndyScherzinger:
I made some progress...

* generic avatar tinting fixed for pre lollipop https://github.com/owncloud/android/pull/1559/commits/564b7b3df29f09241e96a0f9bf382f69d7f6ad10
* EditText fixed https://github.com/owncloud/android/pull/1559/commits/6436a45722a0743b88baeac3d69654718a125054
* generic avatar sizing fixed https://github.com/owncloud/android/pull/1559/commits/79776f202198bb725a411c13363e5b330e2ba314

AndyScherzinger:
@tobiasKaminsky maybe you can help me out with the avatars in the drawer menu itself and in the account list in the account manager screen - That would be awesome! :).

Code references:
* drawer menu: https://github.com/owncloud/android/pull/1559/files#diff-154af4ff2566e264dcca122072d20ae7R375
* account manager: https://github.com/owncloud/android/pull/1559/files#diff-ec86da1fce7563815b7cb2716757ebd3R90

I guess the setAvatar (https://github.com/owncloud/android/pull/1559/files#diff-154af4ff2566e264dcca122072d20ae7R446) would have to be moved to a helper class or something like that and the parameter would probably have to change or there need to be two different methods, since the menu item icon is a drawable and we can (or should) not work with layout ids for them. Where we can for the drawer header and for the list items in the account manager screen :)

AndyScherzinger:
Screenshot of the actual implementation:
device-2016-04-06-004916

AndyScherzinger:
FYI this is who the support lib handles waayyy to long menu titles:
device-2016-04-06-005636

AndyScherzinger:
So far tests have been done on 6.0.1 Nexus5X and 4.4.4 Neuxs7 (2012)

jancborchardt:
Very nice @AndyScherzinger! :)

Just one thing, the avatar placeholders should look more like the ones from core, ref this screenshot: https://github.com/owncloud/contacts/pull/261#issue-142581679
(Thinner font, a bit smaller and more centered.)

AndyScherzinger:
Sure thing! :) 

**DONE**
device-2016-04-06-163532

device-2016-04-06-163544

Is that the correct? 
* Size I simply calculated text size equals radius :D
* alignment is now properly calculated
* font isn't bold anymore

AndyScherzinger:
@tobiasKaminsky do you have the time to help me out with the avatr loading? :) --> https://github.com/owncloud/android/pull/1559#issuecomment-206023464 That would be awesome :)

jancborchardt:
Only 2 nitpicks:
- the superlong username is not ellipsized in that dropdown list
- the usernames in the dropdown list are not vertically aligned with the placeholder/avatar circle, they need to be moved a bit down (or rather the images need to be moved up)

tobiasKaminsky:
Maybe on the weekend or next work.
This week is rather busy :/

AndyScherzinger:
No worries @jancborchardt - I resized the avatars to Google's guideline (14sp) so they are now the same size as all the other icons in the menu and added the ellipsizing:
device-2016-04-06-212302

AndyScherzinger:
@tobiasKaminsky whenever you find the time :) I appreciate your help with this matter. BTW I just added commented out code for the on device menu item --> https://github.com/owncloud/android/pull/1559/commits/5451ca1cbb97bdc6891e433607780c3a0d87b58a which then just needs to be commented in whenever your PR gets merged :+1:

jancborchardt:
:+1: regarding design :)

AndyScherzinger:
Awesome :) :tada: 
I have finally also implemented the avatar loading for the menu items (cc @tobiasKaminsky):
device-2016-04-07-181319

anyone up for a code review? @tobiasKaminsky @przybylski @malkomich @davivel

tobiasKaminsky:
Great that you did it!
But it seems as the avatars from server are smaller than the generated one?

AndyScherzinger:
Argh, I missed that one (can really see the size difference on the phone, but on the screenshot). I pushed the fix :)

davivel:
Seems there is a great work here. 

Problem: there are **tons** of things to test. It's going to be hard.

I couldn't read all the issue, but regarding:

> It redirects you to a user management screen I created where you can create new users, delete users and change password they way Google apps do it and I removed the user management from settings since it doesn't belong there anyways. 

Is that in this PR also?

AndyScherzinger:
Yes, as soon as you click on manage accounts in the drawer's menu, kind of like when you do so in the GMail app. You can also run it in parallel using the beta branch which already includes this PR's code changes :)

AndyScherzinger:
@davivel @malkomich @tobiasKaminsky fancy doing a code review? :grinning:

MTRichards:
@davivel @rperezb maybe we can move this up?

AndyScherzinger:
That would be awesome @MTRichards! 
Beware though that this does not show the user's full name but (still) the username. Referring to the new issue opened #1639. At the moment the full name is not available. So pinging @davivel just in case the full name would be required to be included in this change because that would mean the app would have to start fetching this information and this would then be true for the iOS app as well.

AndyScherzinger:
@jancborchardt any news and progress on your idea on how to further improve account handling withing this PR? :smiley:

jancborchardt:
@AndyScherzinger I’d really like to have this merged before piling up more on it. ;)

rperezb:
@AndyScherzinger I edited your comment for better tracking 😉

AndyScherzinger:
@rperezb perfect! :+1: Way better!

davivel:
OK, time to have a look here. Much code to see, will take me a while.

AndroidManifest.xml
```diff
@@ -59,17 +59,19 @@
...
         android:icon="@drawable/icon"
         android:label="@string/app_name"
-        android:theme="@style/Theme.ownCloud"
+        android:theme="@style/Theme.ownCloud.Toolbar"
         android:manageSpaceActivity=".ui.activity.ManageSpaceActivity">
         <activity
             android:name=".ui.activity.FileDisplayActivity"
-            android:label="@string/app_name" >
+            android:label="@string/app_name"
+            android:theme="@style/Theme.ownCloud.Toolbar.Drawer">
```
davivel: Should be added also to ```.ui.activity.UploadListActivity``` ?

res/layout/account_action.xml
```diff
@@ -0,0 +1,46 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
```
davivel: ```match_parent``` better.

AndroidManifest.xml
AndyScherzinger: Yes :)

src/com/owncloud/android/utils/BitmapUtils.java
```diff
@@ -266,5 +274,41 @@ public static boolean isImage(File file) {
...
+
+    /**
+     * calculates the RGB value based on a given account name.
+     *
+     * @param accountName The account name
+     * @return corresponding RGB color
+     * @throws UnsupportedEncodingException if the charset is not supported
+     * @throws NoSuchAlgorithmException if the specified algorithm is not available
+     */
+    public static int[] calculateRGB(String accountName) throws UnsupportedEncodingException, NoSuchAlgorithmException {
```
davivel: Why is this here and not as a private method in ```BitmapUtils```? It's only used in that class. Is there any plan to use it in other places in the near future?

davivel: Forget it, it's not important. Maybe a more general name would be nice.

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -275,4 +284,52 @@ public static void colorSnackbar(Context context, Snackbar snackbar) {
...
+    /**
+     * fetches and sets the avatar of the current account in the drawer in case the drawer is available.
+     *
+     * @param account        the account to be set in the drawer
+     * @param userIcon       the image view to set the avatar on
+     * @param avatarRadius   the avatar radius
+     * @param resources      reference for density information
+     * @param storageManager reference for caching purposes
+     */
+    public static void setAvatar(Account account, ImageView userIcon, float avatarRadius, Resources resources,
```
davivel: Method ```DrawerActivity.setAvatar()``` and this one are very, very similar. We should try some refactoring there.

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -978,26 +524,24 @@ public void onServiceDisconnected(ComponentName component) {
...
     public FileUploaderBinder getFileUploaderBinder() {
         return mUploaderBinder;
     }
 
-
+    @Override
     public void restart(){
         Intent i = new Intent(this, FileDisplayActivity.class);
         i.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
-        i.addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP);
```
davivel: Maybe related to https://github.com/owncloud/android/issues/1657 ?

davivel:
@jesmrec , many things to test here. Let's talk tomorrow to define a good test plan.

@AndyScherzinger, @tobiasKaminsky , this is awesome material. Should I make a rebase on master, or can you do it?

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -82,21 +67,16 @@
...
-import java.util.ArrayList;
-
 
 /**
- * Activity with common behaviour for activities handling {@link OCFile}s in ownCloud
- * {@link Account}s .
+ * Activity with common behaviour for activities handling {@link OCFile}s in ownCloud {@link Account}s .
  */
-public class FileActivity extends AppCompatActivity
+public class FileActivity extends DrawerActivity
```
AndyScherzinger: Yes that was basically the idea and yes, composition would be preferable and yes that should be done in another PR

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -359,7 +358,223 @@ private Bitmap doFileInBackground() {
...
+
+                int px = getAvatarDimension();
+
+                // Download avatar from server
+                OwnCloudVersion serverOCVersion = AccountUtils.getServerVersion(mAccount);
+                if (mClient != null && serverOCVersion != null) {
+                    if (serverOCVersion.supportsRemoteThumbnails()) {
+                        GetMethod get = null;
+                        try {
+                            String uri = mClient.getBaseUri() + "" +
```
AndyScherzinger: Tricky... there is no official Avatar API... The implementation simply "knows" where to look for the avatar image and fetches it.

res/layout/account_action.xml
```diff
@@ -0,0 +1,46 @@
...
+    android:layout_height="72dp"
+    android:orientation="horizontal"
+    android:weightSum="1">
+
+    <ImageView
+        android:id="@+id/user_icon"
```
AndyScherzinger: correct yes, than it should probably be called "add_account_icon"

res/layout/account_action.xml
```diff
@@ -0,0 +1,46 @@
...
+    <TextView
+        android:id="@+id/user_name"
+        android:layout_width="0dp"
+        android:layout_height="fill_parent"
+        android:layout_marginLeft="28dp"
+        android:layout_weight="1"
+        android:gravity="center_vertical"
+        android:paddingRight="@dimen/standard_padding"
+        android:text="@string/placeholder_filename"
+        android:textColor="@color/primary"
```
AndyScherzinger: we should use "primary" as set here which is one of the oc blue colors.

AndroidManifest.xml
```diff
@@ -143,7 +145,7 @@
             android:name=".authentication.AuthenticatorActivity"
             android:exported="true"
             android:launchMode="singleTask"
-            android:theme="@style/Theme.ownCloud.noActionBar" >
+            android:theme="@style/Theme.ownCloud.Toolbar" >
```
AndyScherzinger: You got it right in the first place, There IS a toolbar now ;) (will have to check for toolbar on login screen though)

res/layout/drawer_header.xml
```diff
@@ -0,0 +1,123 @@
...
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_below="@+id/drawer_user_avatars"
+            >
+
+            <LinearLayout
```
AndyScherzinger: sure, go ahead!

res/layout/uploader_layout.xml
```diff
@@ -46,11 +50,23 @@
...
-	    android:layout_width="fill_parent"
-	    android:layout_height="wrap_content"
-	    android:layout_alignParentBottom="true"
-	    android:orientation="horizontal" >
+		android:layout_width="match_parent"
+		android:layout_height="wrap_content"
+		android:orientation="horizontal">
+
+		<ImageView
+			android:layout_width="match_parent"
```
AndyScherzinger: true!

src/com/owncloud/android/ui/TextDrawable.java
```diff
@@ -1,52 +1,126 @@
...
+import com.owncloud.android.utils.BitmapUtils;
+
+import java.io.UnsupportedEncodingException;
+import java.security.NoSuchAlgorithmException;
 
 /**
- * Created by tobi on 24.05.15.
+ * A Drawable object that draws text (1 character) on top of a circular/filled background.
  */
 public class TextDrawable extends Drawable {
```
AndyScherzinger: Any suggestion from your side? Go with ```SingleCharacterTextDrawable```?

src/com/owncloud/android/ui/activity/FileActivity.java
AndyScherzinger: I think so, yes. It's been a while.

src/com/owncloud/android/utils/BitmapUtils.java
AndyScherzinger: more general name, sure. I made it public since I saw no harm in it just in case someone might need such a function in the future somewhere.

src/com/owncloud/android/utils/DisplayUtils.java
AndyScherzinger: Yes I know :disappointed: but I didn't find an easy way on how to achieve that... :(

davivel:
Rebased on master.

AndyScherzinger:
:rocket: awesome! thanks @davivel :tada:

AndroidManifest.xml
davivel: That's not what I see in the style. 

In v21/styles.xml the style includes ```<item name="windowNoTitle">true</item>``` and ```<item name="windowActionBar">false</item>```; and in "general" styles.xml defines same values, and also extends them ```Theme.AppCompat.Light.NoActionBar```. 

Then, since ```AuthenticatorActivity``` does **not** extend ```ToolbarActivity```, there is no toolbar in it, what is right. And there is no styling info for toolbar or action bar or title, what is right too. So, perfect for this Activity :)

res/layout/toolbar_standard.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ownCloud Android client application
+
+  Copyright (C) 2016 ownCloud Inc.
```
davivel: To update to ownCloud Gmbh.

src/com/owncloud/android/ui/TextDrawable.java
davivel: That's better, but makes me wonder "why should I give a full name to print a single character" ? :D 

Maybe something like ```PrettyTextDrawable``` or ```PrettyRoundDrawable```, or like that?

davivel:
@AndyScherzinger , if you don't mind, I'll apply some of the requested changes by myself.

res/layout/accounts_layout.xml
```diff
@@ -15,13 +15,23 @@
...
-    android:layout_height="56dp"
-    android:gravity="center_vertical"
-    android:paddingLeft="@dimen/standard_padding"
-    android:paddingRight="@dimen/standard_padding"
-    android:textColor="@color/black"
-    android:textSize="18dp" />
\ No newline at end of file
+ -->
+<LinearLayout
+	android:id="@+id/upload_files_layout"
```
AndyScherzinger: I will fix it - quite confusing, yes

AndyScherzinger:
@davivel not at all! We are having a public holiday today, so I am rather afk today :)

res/layout/toolbar_standard.xml
AndyScherzinger: Done -> https://github.com/owncloud/android/pull/1559/commits/7a99787d9808c561dd35b31371f88b447d626b2c

src/com/owncloud/android/utils/BitmapUtils.java
AndyScherzinger: renamed https://github.com/owncloud/android/pull/1559/commits/99229cb6efe6d70397d469bad85290d6a8f98299 --> calculateAvatarBackgroundRGB

src/com/owncloud/android/ui/activity/BaseActivity.java
```diff
@@ -0,0 +1,250 @@
...
+     */
+    private Account mCurrentAccount;
+
+    /**
+     * Capabilites of the server where {@link #mCurrentAccount} lives.
+     */
+    private OCCapability mCapabilities;
+
+    /**
+     * Flag to signal that the activity will is finishing to enforce the creation of an ownCloud {@link Account}.
```
AndyScherzinger: fixed

AndyScherzinger:
Erm... where did my code changes (from yesterday) go?! 0_o

Like https://github.com/owncloud/android/pull/1559/commits/7a99787d9808c561dd35b31371f88b447d626b2c and

davivel:
Agh... sorry, I didn't notice you pushed changes... do you have it locally yet?

AndyScherzinger:
Unfortunately no, I do not have a local checkout anymore. But don't worry it was just the inc->gmbh and TextDrawable renaming. So not much work that got lost :smiley_cat:

davivel:
Uf, what a rest man.... I'm really, really sorry, I just made a ```push -f``` in the worst possible way, not even made ```git fetch``` before.

I made some changes to fix some visual problems, and to recover the ability to customize the toolbar color independently of Material Design guidelines (we need it!), sure you have noticed it. I planned to apply the rest of my comments this morning. Please, tell me if this is OK to you, or if you can apply them yourself instead.

AndyScherzinger:
No worries @davivel! I can apply them myself but won't find the time to do so before tonight. So if it is an urgent matter just go ahead! :+1:

res/layout/toolbar_standard.xml
```diff
@@ -0,0 +1,48 @@
...
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="?attr/actionBarSize"
+        android:background="?attr/colorPrimary"
+        app:layout_scrollFlags="scroll|enterAlways|snap"
+        app:popupTheme="@style/ThemeOverlay.AppCompat.Light"/>
+
+    <ProgressBar
```
davivel: I already took this into account. For the moment all the progress bars belong to fragments, instead of bar, though the other way around could be fine too.

src/com/owncloud/android/ui/TextDrawable.java
davivel: Sounds fine.

src/com/owncloud/android/utils/BitmapUtils.java
davivel: :+1:

res/layout/toolbar_standard.xml
AndyScherzinger: Fixed: https://github.com/owncloud/android/pull/1559/commits/65c5c376b227658c4a6b33e197d6329d28b01e68

res/layout/files.xml
```diff
@@ -18,9 +18,11 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
   -->
 <android.support.v4.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
+                                        xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@+id/drawer_layout"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    android:fitsSystemWindows="true"
```
davivel: Forget it. This is https://developer.android.com/reference/android/view/View.html#setFitsSystemWindows(boolean) , that's different; ```fitSystemWindows``` Vs ```fitsSystemWindows```. Come on, Google...

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
davivel: Let's keep it as it is right now and think about a separate refactoring (unless QA reveals hard bugs).

davivel:
You recovered it! Nice :)

I updated with the minor comments, the code review is OK to me. @AndyScherzinger , feel free to review my changes.

AndyScherzinger:
Well I just re-did the changes :wink: - Went through the latest changes, looks good to me :+1: 

One conceptual question, just to understand the reasons behind it, why did you move the horizontal progress bar to the fragment and not have it as part of the toolbar+accessing it via the activity?

davivel:
In tablets, landscape, the app shows two fragments, list and detail/preview of a file. With separate progress bars we are free to update both independently, each bound to the state of the fragment below.

It's not a big win, indeed. The progress bar in the details view is used with the sync-on-tap, when the user previews a file that was downloaded before but there is a new version in the server, that is downloaded automatically and then replaces the preview. The progress is ugly as hell, but we didn't want to show nothing nor expend a lot of time in redoing all the preview fragments.

I'm open to change it back again, just we need to be sure that for that use case we show something that makes sense.

AndyScherzinger:
Understood :) - No idea to be honest, I am not a UI/UX designer, I was just curious and asked :)

davivel:
@jesmrec , in general I think this is ready to QA. I will add some bugs I found on the road, please, tell me if you think they're blocking for the validation process.

davivel:
CRASH with no account:

-- STEPS TO REPRODUCE:
 1. With a single account in the app, open the drawer and tap on 'Manage accounts',
 2. Delete the single account using the trash-bin icon.
 3. Tap on BACK button or UP arrow in the Toolbar

-- EXPECTED BEHAVIOR:
Login view is shown to request the creation of an account

-- CURRENT BEHAVIOUR:
App crashes with NPE

davivel:
BUG: duplicated account in drawer:

-- STEPS TO REPRODUCE: 

  1. Clean installation of the app (not upgrading)
  2. Create first account
  3. Open navigation drawer
  4. Tap on triangle in the drawer header to show the submenu of accounts

-- EXPECTED BEHAVIOR:
Current account is shown in the list of accounts once

-- CURRENT BEHAVIOR:
Current account is shown in the list twice, one with the first letter of the username as the avatar, once with an user icon

EXTRA NOTES: do it "fast", the fact of not having the user avatar available could be a condition for the bug; after this is received, and the device rotated, the list of accounts is shown correctly.

davivel:
Needed improvement? . There is no confirmation dialog to delete an account. There wasn't before either, but the user had to long-tap an account, and then a second tap on 'Remove account' to get it deleted; now a single tap on the trash-bin icon is enough, seems too easy delete an account by accident.

Should we add a confirmation there?

davivel:
BUG? : content/order of options menu seems inconsistent, depending on the type of the current file . Specially remarkable, it's shown while choosing a target folder to move/copy a file in, and contains all the options of the main list of files (grid view, refresh account).

But I'm not sure if this is a bug here or in master. Needs to be checked.

davivel:
That's all I have, @jesmrec . Tell me if validation can start known these bugs exist.

I will not update this branch in the rest of the day.

jesmrec:
@davivel 

- The first two bugs do not block the validation, so they are reproduced with particular steps themselves.
- All destructive actions should be confirmed by the user to avoid that user's mistakes break the expected performance. I would add the confirmation, as i would add it when the user is going a delete a file/folder from server (out of the scope of this PR, only an example).
- I will check the order issue. 

QA process can start while these bugs and improvements are reviewed.

jesmrec:
Long user names/server url does not fit correctly in "manage accounts" view, with device in portrait orientation if it fills more that three rows 

For example:

screen shot 2016-10-05 at 18 37 00

a bigger padding or a smaller font could be solutions.

jesmrec:
(possible improvement)

Related to the very long names mentioned by @AndyScherzinger in https://github.com/owncloud/android/pull/1559#issuecomment-206025530

The place on the right of the avatar is 26 characters. This place can not be greater, but in many cases the string `username@server_uri(:port)` will be longer than 26 characters and

Would it be posible to set the accounts in the drawer in the same way as in the upper section? i mean, the username above the server uri, both on the right of the avatar.

src/com/owncloud/android/ui/activity/FileActivity.java
davivel: I'm checkin #1657; I can't reproduce it, but I'm reviewing the implementation of the switch in any case.

The deletion of this flag was in https://github.com/owncloud/android/pull/1559/commits/1ef9521ac83d2133b653e37156d498c32b03b38c , way before the bug #1657 was detected, so it wasn't removed to fix that bug.

I remember some strange behavior browsing the app that could be related to the absence of this flag. I will try to reproduce it and tell you.

jesmrec:
Related withe the previous message: the accounts are displayed in a different way in the drawer (with username) and in accounts manager (with displayname). All of them should have the same "format"

src/com/owncloud/android/ui/activity/FileActivity.java
davivel: OK, let's forget this. That flag is correctly removed, otherwise account switching doesn't work. I can't find the weird navigation behavior I remembered.

jesmrec:
### Bug

Steps:

1. Add an account. The user must have an avatar
2. Add more accounts of other server, which username is the same as in step 1, that do not have any avatar. For example:

user1@server1 <- has avatar
user1@server2 <- has no avatar
user1@server3 <- has no avatar

Actual behaviour

Some of the accounts use the same avatar than the first one

Expected behaviour

Each account, its avatar.

Tested with Huawei 6P, Android7

davivel:
About this:https://github.com/owncloud/android/pull/1559#issuecomment-251735447

Please, notice that during the rebase I took advantage of the ** display name ** in the drawer header. The display name was added to the app some weeks ago. 

So, in the drawer now we have the display name over the full account identifier (username@baseUrl) ; but in the list we are not showing the display name. This is exactly what Google does in its own apps.

Using the same approach in the list than in the header, the long identifier would be ellipsized exactly in the same point as now.

davivel:
I just noticed, in Google's account switcher the list below the header does not include the current account. Makes sense because, you know, who wants to change to the current account :D .

Should we remove it from that list too?

jesmrec:
@davivel, i agree with removing the current account from the list below.

jesmrec:
### Bug

Steps

1. In web, change the avatar, or the display name for another one which starts with a different letter (this change will change the avatar as well if the user does not have avatar)

Actual Behaviour

In app, avatar does not change 

Expected behaviour

Avatar change in the app as well

Tested with Nexus7 Android4.3, Huawei 6P, Android 7

davivel:
> In web, change the avatar
> ...
> In app, avatar does not change

Ladies and gentlemen, QA in da house.

:D

jesmrec:
The following bug is a regression. This is only for you to take in account and for asking your opinion to be checked in this issue or forgetting it to another moment:

1. In app, login in an account
2. In server, delete the account
3. In app, select the account in drawer
4. Redirection to login view

The account does not exist anymore, so credentials will never work (username is non-editable). If you tap on "Go Back" button, the files view or settings view (depending whether the app has more accounts or not) is shown for a little while, but you are redirected again to login view. It is a bit annoying to switch to another account or tap on "new account". 

Expected behaviour? Switch to another account or to login view if the app has no accounts. The question is if server side answers in the same way if the credentials have changed and if the account is removed to detect these two different cases.

AndyScherzinger:
currently active account isn't shown in the chooser in the drawer anymore. https://github.com/owncloud/android/pull/1559/commits/2f9890aecf2e2b3178c6cf3fdd536dd460003dd5

rperezb:
@hodyroff @pmaier1 fyi, this pr is on its way to be merged!

davivel:
> The question is if server side answers in the same way if the credentials have changed and if the account is removed to detect these two different cases.

Yes, the response is the same.

Options: modify the log-in view to allow the user to switch to other account from it. But I wouldn't do that in this PR.

jesmrec:
@davivel i agree then. This is absolutely out of scope. In future refactorings we could deal with this.

I will add it as a new issue for our backlog.

davivel:
I changed a bit the layout of accounts in the account manager to be more similar to the account header; I set 1 line for display name and 2 lines for full account name. Long display names still might be ellipsized "soon", but using four lines was too ugly in all the combinations I tried.

src/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -0,0 +1,702 @@
...
+                        BitmapUtils.bitmapToCircularBitmapDrawable(MainApp.getAppContext().getResources(), thumbnail)
+                );
+            } else {
+                // generate new avatar
+                if (ThumbnailsCacheManager.cancelPotentialAvatarWork(account.name, menuItem)) {
+                    final ThumbnailsCacheManager.AvatarGenerationTask task =
+                            new ThumbnailsCacheManager.AvatarGenerationTask(
+                                    menuItem, getStorageManager(), account
+                            );
+                    if (thumbnail == null) {
```
davivel: This block has some problems. This condition always will be ```true``` at this point, and the block inside should try to create the retrieve the avatar that doesn't exist yet, right? It doesn't, it just generates the colored letter. 

Please, @AndyScherzinger, tell me if I'm losing something.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -361,7 +360,223 @@ private Bitmap doFileInBackground() {
 
     }
 
-    public static boolean cancelPotentialWork(Object file, ImageView imageView) {
+    public static class AvatarGenerationTask extends AsyncTask<Object, Void, Bitmap> {
+        private final WeakReference<ImageView> mImageViewReference;
+        private final WeakReference<MenuItem> mMenuItemReference;
+        private static Account mAccount;
```
davivel: @tobiasKaminsky , what's the reason to make this ```static```? It's set in the construction of each instance, so, overwritten by the last one created. There will be problems if different avatars are created in parallel.

Could be related to https://github.com/owncloud/android/pull/1559#issuecomment-251933389

davivel:
@jesmrec , @AndyScherzinger , after my last comments about avatar generation, I wonder if we should disable it to merge this PR and fix the problems in a new one; fix the problems in this one; or merge it despite them and fix later.

jesmrec:
### Bug

Steps

1. Select multiple files/folders
2. Open the drawer

Actual Behaviour

Drawer is shown under the toolbar, so avatars are hidden

Expected behaviour

Drawer is shown over the toolbar

jesmrec:
@davivel i do not agree in merging this PR with these known bugs. This PR is too big and i understand not no to include the avatars if they need a special/different management. In this case, i would rather to disable avatars and handle them in other issue.

jesmrec:
Related to https://github.com/owncloud/android/pull/1559#issuecomment-251692759: the bug is in master. Up to you @davivel if it can be fixed in this PR or in another one

davivel:
> Related to #1559 (comment): the bug is in master. Up to you @davivel if it can be fixed in this PR or in another one

It's also in 2.1.2. I would deal with it as a separate issue.

davivel:
>@davivel i do not agree in merging this PR with these known bugs. This PR is too big and i understand not no to include the avatars if they need a special/different management. In this case, i would rather to disable avatars and handle them in other issue.

Let me have another look to see if I can find an easy solution.

jesmrec:
### Bug

Steps

1. Open the drawer
2. Tap in "Add account" (open login view)
3. Tap on Back button

Actual behaviour

App is closed

Expected behaviour

Go back to the view in which the drawer was opened.

Tested with Huawei 6P Android 7

davivel:
> Would it be posible to set the accounts in the drawer in the same way as in the upper section? i mean, the username above the server URI, both on the right of the avatar. 

It's not possible, unless we change all the implementation of the drawer (except the header). Right now it's based in a predefined component in the google support library, and all the entries we see in the bottom part are ```MenuItems```, not simple ```Views``` we can replace and draw as we prefer. In a nutshell, we can an icon to the left, and a title on the right, but not two lines per entry on the right.

I would leave it as it's right now. Full account names, although ellipsized, should be more informative than display names here. The change of the layout would be too much extra work for a small benefit. And others do in the same way, for example, Google apps, so we can trust that will not be strange to the user eyes.

src/com/owncloud/android/ui/activity/DrawerActivity.java
AndyScherzinger: I guess so... To be honest I just took the Avatar/Caching implementation done by @tobiasKaminsky and integrated it in this PR.

jesmrec:
I have noticed that in `Manage Accounts`view, the accounts are clickable with no actions. As possible enhacement, i would use the clickable property to switch the account, so this also something related to account management

davivel:
About https://github.com/owncloud/android/pull/1559#issuecomment-252194713, handling multiple selection when drawer is opened / closed.

I didn't find an easy way to keep the action mode bar under the navigation drawer.

I solved it like Gmail: when the drawer starts to open the action is disable; when the drawer is closed, the action mode is enabled again, and the items that were selected when it was disabled are checked again too. This works even if the device is rotated between drawer opening and closing.

davivel:
> I have noticed that in Manage Accountsview, the accounts are clickable with no actions. As possible enhacement, i would use the clickable property to switch the account, so this is also something related to account management

I missed it also at the beginning, though I guess that's becaused we are used to work that way. Google accounts manager doesn't allow changing the account there, though Google account manager doesn't below to a single app, it's an independent app to manage all Google accounts used by all the Google apps in the device, so maybe it's not a good example.

Any solution (change account on click or do nothing) works for me.

@AndyScherzinger , what do you think?

AndyScherzinger:
@davivel @jesmrec tricky one with the click action. I'd say that it should then probably launch the password change activity since that one is less "destructive" compared to the "remove account" action.

davivel:
@AndyScherzinger  , with 'changing the account' I meant 'switch to other account', no updating the password.

AndyScherzinger:
@davivel switching to other account would also be fine imho. :+1:

davivel:
Switch account on click in accounts manager added.

I also added a check mark on the current account. We had some internal discussion about how to do it, any feedback from @owncloud/android-developers or anybody else is more than welcome.

An example:

davivel:
By the way, the check mark is tinted with the same color as the toolbar.

AndyScherzinger:
Nice addition @davivel ! :+1: 
I went through Google's material design icons but couldn't find any other icon that would fit better.

davivel:
That last one was big.

Summary of changes:
 - the app is able to update user avatars after they are changed in the server; 
 - the avatar is refreshed whenever the root folder is refreshed (except as part of the full account refresh, that usually is in background);
 - the avatar is also refreshed when an account is added in the account manager;
 - the avatar is not downloaded again if didn't change (etag is used to prevent the transfer);
 - all the code involved in the load of the avatars or its fallbacks (generated avatar with colored letter or icon) has been moved to the AsyncTask in ```ThumbnailsCacheManager```, so that all of it runs in background; special mention about the method ```ThumbnailsCacheManager.getBitmapFromDiskCache(imageKey)```, that slows down the UI when called from the main thread (it's still called from the UI in different places, but not for avatars; I thought that refactoring also thumbnails would be too big and too risky).

Hope that's enough.

cc @AndyScherzinger , @jesmrec .

davivel:
I will rebase on master to get last translations (and to earn the last Github green check)

davivel:
And please, notice that now this depends also on owncloud/android-library#138

jesmrec:
The avatars changes will need extra checks. Working on it.

jesmrec:
Approved. 

These two issues have been created from the detected bugs as improvements:

https://github.com/owncloud/android/issues/1825
https://github.com/owncloud/android/issues/1826

davivel:
I updated the reference to the OC library after merging it's matching PR.

Thanks a lot, @AndyScherzinger . This contribution is pure LOVE.

Issue #1561, opened 2016-03-24
Tapping on share/link icon in the list (for items which already have been shared) should directly bring up share dialog


jancborchardt:
Currently this is just an icon. Why not make it do something? ;)

cc @tobiasKaminsky @AndyScherzinger @LukeOwncloud @stoyicker @Kernald

AndyScherzinger:
Sold ;) But can we then get some new icons for that? The share icon differs from the one on the web app and also has a weird color 0_o

Not sure how easy we can integrate itemClick, iconClick and longClickForMultiSelect though... this might be the tricky part here...

tobiasKaminsky:
But only on the list?
Then I can do this (and then try to get it on multiselect branch also working)

AndyScherzinger:
Yeah, imho just on the list :)

tobiasKaminsky:
I tried to implement that a click on the share icon shows the share dialog and a click on the link directly pops up the share link to app dialog.
But there seems to be a bug so the second part is not working at the moment.
@jancborchardt or should they both behave the same?

jancborchardt:
Maybe have them both behave the same in just opening the dialog for now.

And regarding the icons, use the ones from the core repo simply :)
https://github.com/owncloud/core/blob/master/core/img/actions/share.svg

tobiasKaminsky:
Then this is very simple to do :+1:

tobiasKaminsky:
Then we have a PR: #1577

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #1577, opened 2016-03-30
Click opens share dialog

tobiasKaminsky:
- reformatting the view generation
- using break for case, before that all cases were used (wonder that no unexpected behaviour occured)

@AndyScherzinger can you change the icon that @jancborchardt has mentioned in #1561 
https://github.com/owncloud/core/blob/master/core/img/actions/share.svg
I do not know how to get the right resolutions out of the svg file.
(maybe you can write how you do this)


jesmrec:
Current Ui/UX is optimal. Let's see in the future other improvements

PR #1599, opened 2016-04-09
Resized images

tobiasKaminsky:
--------- edited by @davivel

**BLOCKED** by https://github.com/owncloud/core/issues/16250

--------- edited by @jabarros
TASKs:
- [X] Changes after CR @jabarros
- [X] [QA] Create 'resizedImages' branch in QA Repo  @jesmrec
- [X] [QA] Merge 'resizedImages' branch in QA Repo
- [X] [QA] Create test plan @jesmrec 
- [ ] [QA] Validate test plan @jesmrec [WIP]

bugs and improvements:
- [ ] Unsupported image formats https://github.com/owncloud/android/pull/1599#issuecomment-229400874
- [ ] Bad preview of landscape images https://github.com/owncloud/android/pull/1599#issuecomment-229406324 @davivel [WIP]

---

This PR is now against master.

As this PR introduces a rather complex feature I will summarize it here.
- Clicking on an image loads a reduced version to the cache.
- The width and height is determinated by the screen.
- the benefit is a very small, but yet usable preview of the image.
- it can be used to send via whatsapp, or as a file to another app (like k9) with the correct filename.
- resized images are not treated as downloaded files. If the user wants to have the real file, he has to long-click and select download.
- clicking on an image shows first a square thumbnail (if available) until the resized image is loaded

**ToDo**
- Zooming a resized image with at least 2x (which can get blurry)
  - show snackbar "automatically download full image on zoom: yes"
  - Clicking "yes" after second time will offer option "always" to prevent showing the question. 
  - time out the snackbar the second time by just waiting will show "never download full image" -> "yes".
  - after both show second snackbar without action "default behaviour can be changed in settings"
- In setting: new entry "click on image downloads ..."
  - full image to storage
  - resized image to cache

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -154,7 +154,8 @@
 
         <provider
             android:name=".ui.adapter.DiskLruImageCacheFileProvider"
-            android:authorities="com.owncloud.imageCache.provider">
+            android:authorities="com.owncloud.imageCache.provider"
+            android:exported="true">
         </provider>
 
         <activity
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -152,6 +152,11 @@
             </intent-filter>
         </provider>
 
+        <provider
+            android:name=".ui.adapter.DiskLruImageCacheFileProvider"
+            android:authorities="com.owncloud.imageCache.provider">
+        </provider>
+
         <activity
             android:name=".authentication.AuthenticatorActivity"
             android:exported="true"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -131,7 +131,7 @@
             android:enabled="true"
             android:exported="true"
             android:label="@string/sync_string_files"
-            android:syncable="true" />
+            android:syncable="false" />
 
         <provider
             android:name=".providers.UsersAndGroupsSearchProvider"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -129,7 +129,7 @@
             android:name=".providers.FileContentProvider"
             android:authorities="@string/authority"
             android:enabled="true"
-            android:exported="true"
+            android:exported="false"
             android:label="@string/sync_string_files"
             android:syncable="false" />
 
--- a/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -29,10 +29,13 @@ import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.httpclient.methods.GetMethod;
 
 import android.accounts.Account;
+import android.accounts.AccountManager;
+import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Bitmap.CompressFormat;
 import android.graphics.BitmapFactory;
+import android.graphics.Point;
 import android.graphics.Canvas;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.ColorDrawable;
@@ -40,7 +43,11 @@ import android.graphics.drawable.Drawable;
 import android.media.ThumbnailUtils;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.view.Display;
+import android.view.View;
+import android.view.WindowManager;
 import android.widget.ImageView;
+import android.widget.ProgressBar;
 
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
@@ -58,27 +65,27 @@ import com.owncloud.android.utils.DisplayUtils;
  * Manager for concurrent access to thumbnails cache.
  */
 public class ThumbnailsCacheManager {
-    
+
     private static final String TAG = ThumbnailsCacheManager.class.getSimpleName();
-    
+
     private static final String CACHE_FOLDER = "thumbnailCache";
 
     private static final Object mThumbnailsDiskCacheLock = new Object();
     private static DiskLruImageCache mThumbnailCache = null;
     private static boolean mThumbnailCacheStarting = true;
-    
+
     private static final int DISK_CACHE_SIZE = 1024 * 1024 * 10; // 10MB
     private static final CompressFormat mCompressFormat = CompressFormat.JPEG;
     private static final int mCompressQuality = 70;
     private static OwnCloudClient mClient = null;
 
-    public static Bitmap mDefaultImg = 
+    public static Bitmap mDefaultImg =
             BitmapFactory.decodeResource(
                     MainApp.getAppContext().getResources(),
                     R.drawable.file_image
             );
 
-    
+
     public static class InitDiskCacheTask extends AsyncTask<File, Void, Void> {
 
         @Override
@@ -88,17 +95,17 @@ public class ThumbnailsCacheManager {
 
                 if (mThumbnailCache == null) {
                     try {
-                        // Check if media is mounted or storage is built-in, if so, 
+                        // Check if media is mounted or storage is built-in, if so,
                         // try and use external cache dir; otherwise use internal cache dir
-                        final String cachePath = 
-                                MainApp.getAppContext().getExternalCacheDir().getPath() + 
-                                File.separator + CACHE_FOLDER;
+                        final String cachePath =
+                                MainApp.getAppContext().getExternalCacheDir().getPath() +
+                                        File.separator + CACHE_FOLDER;
                         Log_OC.d(TAG, "create dir: " + cachePath);
                         final File diskCacheDir = new File(cachePath);
                         mThumbnailCache = new DiskLruImageCache(
-                                diskCacheDir, 
-                                DISK_CACHE_SIZE, 
-                                mCompressFormat, 
+                                diskCacheDir,
+                                DISK_CACHE_SIZE,
+                                mCompressFormat,
                                 mCompressQuality
                         );
                     } catch (Exception e) {
@@ -112,8 +119,8 @@ public class ThumbnailsCacheManager {
             return null;
         }
     }
-    
-    
+
+
     public static void addBitmapToCache(String key, Bitmap bitmap) {
         synchronized (mThumbnailsDiskCacheLock) {
             if (mThumbnailCache != null) {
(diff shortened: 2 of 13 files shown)


tobiasKaminsky:
Snackbar is only allowing one action.
So I have adapted the behaviour.
@jancborchardt @AndyScherzinger @davivel

tobiasKaminsky:
@jancborchardt @AndyScherzinger 
After thinking about it I think a snackbar (or similar) with two options would be better, or?

jancborchardt:
I thought we said zooming should always download the image? Without showing any snackbar or something like that?

tobiasKaminsky:
No, please read the other PR.
There me and other said that this against the expectation to save bandwidth and storage. So this should stay optional.
And with the snackbar approach every user can decide on its own how the default behaviour should be.

AndyScherzinger:
@tobiasKaminsky your approach works for me since you are right, the snackbar is defined with 0-1 action. For more than 1 action we would need to use dialogs or something else.

davivel:
Having a look now, just read the comments in the thread.

First ideas:
  - zoom behaviour should be in a separate PR to limit the scope of testing.
  - about sending preview image to other apps instead of the full size image without the using being aware of it: will this behaviour fit the expectations of most of the users?

AndroidManifest.xml
```diff
@@ -114,7 +114,7 @@
             android:name=".providers.FileContentProvider"
             android:authorities="@string/authority"
             android:enabled="true"
-            android:exported="false"
+            android:exported="true"
```
davivel: Why is this needed in this PR?

AndroidManifest.xml
```diff
@@ -125,6 +125,12 @@
             android:exported="false"
             android:label="@string/search_users_and_groups_hint" />
 
+        <provider
+            android:name=".ui.adapter.DiskLruImageCacheFileProvider"
```
davivel: Why is a new exported provider needed?

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -157,6 +165,12 @@ public ThumbnailGenerationTask(ImageView imageView, FileDataStorageManager stora
             mAccount = account;
         }
 
+        public ThumbnailGenerationTask(ImageView imageView, FileDataStorageManager storageManager,
+                                       Account account, ProgressBar progressWheel) {
+        this(imageView, storageManager, account);
```
davivel: Wrong indentention here.

tobiasKaminsky:
> zoom behaviour should be in a separate PR to limit the scope of testing.

This is a good idea.

> about sending preview image to other apps instead of the full size image without the user being aware of it: will this behaviour fit the expectations of most of the users?

If the image is not downloaded I assume that users know that the resized version is used.
The user has to explicit download a file to use this.
But maybe we can show this to the user by using "send resized image".
And use "send original size" when downloaded.

AndroidManifest.xml
tobiasKaminsky: This is for sending cached images to other apps.
Technically only a content:// is sent which points to this provider, as far as I remember.

AndroidManifest.xml
tobiasKaminsky: I have no idea :/
A quick test shows no problems if I revert this.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -281,18 +329,25 @@ private Bitmap doOCFileInBackground() {
                             try {
                                 String uri = mClient.getBaseUri() + "" +
                                         "/index.php/apps/files/api/v1/thumbnail/" +
```
jabarros: Please, set a constant for this

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -324,6 +334,26 @@ public void sendDownloadedFile(OCFile file) {
         }
     }
 
+    public void sendCachedImage(OCFile file) {
+        if (file != null) {
+            Intent sendIntent = new Intent(android.content.Intent.ACTION_SEND);
+            // set MimeType
+            sendIntent.setType(file.getMimetype());
+//            sendIntent.putExtra(Intent.EXTRA_STREAM, Uri.parse("content://" + DiskLruImageCacheFileProvider.AUTHORITY + "/#" + file.getRemoteId() + "#" + file.getFileName()));
```
jabarros: Remove this comment if it is code not needed

jabarros:
@tobiasKaminsky In a short term, QA team is going to validate this functionality. Could you take a look to the CR changes and do a rebase on master?

jesmrec:
The images in format .tiff are not previewed, because they are not supported (thumbnail is not shown). If you tap on a tiff image, a spinner appears forever on a black background. That can be confusing for the user, so that he/she does not know if there is a problem or not. If this image format (and another ones) are not supported, it would be better to show an alert or similar indicating that the image .tiff (or whatever) can not be previewed.

jesmrec:
if you tap on a "landscape" image, is not correctly resized. 

This is the downloaded image:

screen shot 2016-06-29 at 17 49 24

The previewed (not dowloaded) in portrait:

screen shot 2016-06-29 at 17 48 41

The previewed (nor downloaded) in landscape:

screen shot 2016-06-29 at 17 58 01

In this case, this is a "landscape" image (longer than widther), so the preview should fit with the device landscape orientation. The device in portrait orientation should show something like the first attached file. 

cc @tobiasKaminsky @owncloud/android-developers

jabarros:
@owncloud/android-developers 

due to high number of conflits detected during the rebase from master, I pushed the rebase with the fixes into a new one (resizedImages_fix_rebase) in order to avoid break this PR if some conflicts were not fixed okay.

So, if anyone will work here, please, better move to new resizedImages_fix_rebase branch.

davivel:
Will resume where @jabarros left it, in the branch resizedImages_fix_rebase, fixing the second bug.

AndroidManifest.xml
davivel: That is a different functionality requiring specific tests. I will move it to a separate branch.

Please, include in the description of the PR a summary of all the changes. Otherwise the test plans could not match the scope of the changes, since we don't keep an strict order for the tasks of reviewing code and defining tests - tasks that should be independent.

davivel: I'm sorry, you already commented it in the description: "it can be used to send via whatsapp, or as a file to another app".  

I apologize.

AndroidManifest.xml
```diff
@@ -129,9 +129,9 @@
             android:name=".providers.FileContentProvider"
             android:authorities="@string/authority"
             android:enabled="true"
-            android:exported="true"
+            android:exported="false"
```
davivel: Self-note: this was set to true with addition of Document Provider.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -336,24 +391,39 @@ private Bitmap handlePNG(Bitmap bitmap, int px){
             return resultBitmap;
         }
 
-        private Bitmap doFileInBackground() {
+        private Bitmap doFileInBackground(Boolean mIsThumbnail) {
```
davivel: Self note: why not boolean?

AndroidManifest.xml
```diff
@@ -152,6 +152,12 @@
             </intent-filter>
         </provider>
 
+        <provider
+            android:name=".ui.adapter.DiskLruImageCacheFileProvider"
+            android:authorities="com.owncloud.imageCache.provider"
```
davivel: Self-note: check if should be in setup.xml

davivel:
Rebased double checked and pushed here; no need to go to other branch.

davivel:
> The images in format .tiff are not previewed, because they are not supported (thumbnail is not shown). If you tap on a tiff image, a spinner appears forever on a black background. That can be confusing for the user, so that he/she does not know if there is a problem or not. If this image format (and another ones) are not supported, it would be better to show an alert or similar indicating that the image .tiff (or whatever) can not be previewed.

Even more: if a resized preview can't be get, I'd say the file should be downloaded instead.

davivel:
> In this case, this is a "landscape" image (longer than widther), so the preview should fit with the device landscape orientation. The device in portrait orientation should show something like the first attached file.

Previews should use the same rules to adapt the size of the image to the screen that those used to render the full image. In your sample the preview is resulting in a clip of the image, but it should be a rescaled version.

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -199,10 +205,56 @@ public void onSaveInstanceState(Bundle outState) {
...
         if (getFile() != null) {
-            mLoadBitmapTask = new LoadBitmapTask(mImageView, mMessageView, mProgressWheel);
-            //mLoadBitmapTask.execute(new String[]{getFile().getStoragePath()});
-//            mLoadBitmapTask.execute(getFile().getStoragePath());
-            mLoadBitmapTask.execute(getFile());
+            mImageView.setTag(getFile().getFileId());
+
+            if (mShowResizedImage){
+                Bitmap resizedImage = ThumbnailsCacheManager.getBitmapFromDiskCache(
+                        String.valueOf("r" + getFile().getRemoteId()));
```
davivel: Prefix "r" in the identifier to request a resized preview, and prefix "t" in the identifier to request a thumbnail? Is not too tricky? Seems that ```ThumbnailsCacheManager``` needs different methods for both cases.

davivel:
> Previews should use the same rules to adapt the size of the image to the screen that those used to render the full image. In your sample the preview is resulting in a clip of the image, but it should be a rescaled version.

And in the very end, the problem is what @tobiasKaminsky already stated in the very beginning:

> Remaining problem:
owncloud/core#16250

IMHO, this is **blocking**. 'Til that issue is not addressed, this shouldn't be merged.

zgmnkv:
Hi guys, any update on this one?
Downloading all the images that I viewed consumes a lot of memory on my device.

jesmrec:
not for the moment... you know, we are opened to contributions!!! :)

jesmrec:
Will check with next core generation.

PR #1633, opened 2016-04-25
return false if remove failed

LukeOwncloud:
replaces https://github.com/owncloud/android/pull/1028

if OCFile is null --> cannot delete ==> return false.

Note: if `removeDBData` and `removeLocalCopy` are false, nothing will be deleted, but function still returns `true`. This might be intended, though.

### DIFF
--- a/src/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/src/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -484,6 +484,8 @@ public class FileDataStorageManager {
                     }
                 }
             }
+        } else {
+            success = false;
         }
         return success;
     }


davivel:
Thanks, @LukeOwncloud . 

> Note: if removeDBData and removeLocalCopy are false, nothing will be deleted, but function still returns true. This might be intended, though.

Might be, I don't remember, but seems reasonable.

:+1: 

Do you remember a use case where a NULL value gets to the ```OCFile``` parameter?

LukeOwncloud:
No, not really. I remember that while debugging it happened but I do recall when, why, or how. It could be that it was my fault after all. But I remember that this function returning true caused me some headaches.

davivel:
Puf, checking the code, I can't imagine any valid use case to reproduce it from the UI. Maybe deleting a file in the Android app that was previously deleted in the server while the Android app is refreshing the folder. But the deletion should be requested before the database is refreshed, and then be performed after the database is refreshed. Too tricky.

@jesmrec , @rperezb , no QA needed but regression tests in next release. 

@LukeOwncloud , could you rebase the PR on master and merge then by yourself? Thanks!

LukeOwncloud:
@AndyScherzinger Could you assist? I did quite grasp the whole concept of rebasing. I believe you did!?! Could you do me a favor and rebase the PR on master and merge then? And please copy the git commands you used and post them here. I'd really like to understand how it works! Thanks a lot!

AndyScherzinger:
@LukeOwncloud unfortunately I can't because it is not a owncloud repo but your personal fork and thus can't do a rebase. The commands I use would then be (you need to have an up to date master though) would then be:
```
git checkout master
git pull
git checkout patch-remove-failed3
git rebase master
(git rebase --continue)
git push -f origin patch-remove-failed3
```

davivel:
Thanks, @LukeOwncloud !

LukeOwncloud:
Thank you, @AndyScherzinger! That worked great!

AndyScherzinger:
Great! Glad I could help! :)

Issue #827, opened 2014-12-31
Problem with instant upload and multiple accounts


jancborchardt:
If you have multiple accounts, the instant upload causes a taken photo to get uploaded to whichever account was active at the time. It’s not clear that this is the case, and we should probably specify it in some way.

(I’m pretty sure I opened an issue about it but can’t find it anymore.) cc @rperezb @davivel @tobiasKaminsky 

___

Info by @lefherz 

### Actual behaviour
In https://central.owncloud.org/t/android-instant-upload-with-multiple-accounts-select-target-account/7427 a user reports that in a multi-account setup they can not select the account to which photos shall be instantly uploaded.

The photos are currently only going to the last account which was used. This means that they end up in the wrong folder sometimes. In nextcloud it seems to show the same behavior: https://github.com/nextcloud/android/issues/440

### Expected behaviour
Users should be able to select to which account they want to upload, either after taking a photo or in general.
 
### Steps to reproduce
1. Take a photo
2. Go to the ownCloud app and select another account
3. Take another photo
4. look where they have been uploaded to

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Android version: Android 7.1

Device model: Samsung Galaxy S3 Mini (i8190)

Stock or customized system: Lineage OS 14.1

ownCloud app version: 2.7.0

ownCloud server version: one of them 10.0.7, the other 10.0.6

tobiasKaminsky:
Or another would be:
All settings are account specific.
If you switch, all settings will be remembered.
Then there could be one account with instant upload and one without it.
And only one instant upload can be active in the same time.

floriangosse:
@tobiasKaminsky I think good solution would be global settings where I can choose the account for instant upload and account specific settings where I can define the path for instant upload.

jancborchardt:
Right … when there is more than one account there should be a dropdown in the instant upload section like: »Account to upload to«:
* Active account
* `jan@firstaccount.com` (default)
* `jan@secondaccount.com`
* …

The reason the first account instead of the active account should be the default since that’s the big source of confusion in the current approach. What do you think @tobiasKaminsky @AndyScherzinger @Kernald @Stoyicker @davivel?

jancborchardt:
(Btw @floriangosse just saw you’re in Berlin – you coming to https://owncloud.org/conf ?)

davivel:
[...]

PR #1637, opened 2016-04-26
configure account for instant upload

LukeOwncloud:
As discussed in https://github.com/owncloud/android/issues/827, instant uploads should always use the same account (rather than the currently active one).

edited by @jabarros

- [x] Create 'instant_upload_multi_account' branch from master in android repository @jabarros
- [ ] Merge 'instant_upload_multi_account' branch into master
- [ ] Download LukeOwncloud:instant_upload_multi_account branch changes @jabarros (WIP)
- [ ] Create test plan @jesmrec
- [ ] Validate test plan @jesmrec

### DIFF
--- a/src/com/owncloud/android/utils/FileStorageUtils.java
+++ b/src/com/owncloud/android/utils/FileStorageUtils.java
@@ -29,6 +29,7 @@ import third_parties.daveKoeller.AlphanumComparator;
 
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.lib.resources.files.RemoteFile;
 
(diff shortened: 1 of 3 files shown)


src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -126,6 +127,30 @@ public static String getInstantUploadFilePath(Context context, String fileName)
...
+    }
+
+    private static Account getAccount(Context context, String prefName) {
+        SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
+        String accountName = pref.getString(prefName, null);
+        Account account = AccountUtils.getOwnCloudAccountByName(MainApp.getAppContext(),
+                accountName);
+        return account;
+    }
+
```
malkomich: I think it makes more sense to add these new methods to `AccountUtils` class, dont you think?

malkomich:
I guess a new setting options is needed in order to set the account chosen for the instant uploads, because currently the destination account keeps changing when you change the active one.

davivel:
Hi, @LukeOwncloud. I made a fast test, and there is an important problem here. If the user enables the instant upload and doesn't change the default folder, no account is saved as target account and the upload is not done. The broadcast receiver logs:

'''07-13 11:07:55.544 14302-14302/com.owncloud.android W/com.owncloud.android.files.InstantUploadBroadcastReceiver: No account found for instant upload, aborting```

Could you have a look to it?

danimo:
@LukeOwncloud hey, any chance you can take a look at this again?

jesmrec:
I agree with the comments of @malkomich and @davivel. It is not much intuitive to switch the default account until the user selects one.

Thinking in posible options to select a default account for the instant uploads:

1) In the `Upload Path`section, show a pop-up to select the account to instantly upload to. After that, browsing in that account to select the folder. The advantage is: possibility of selecting different accounts for photos and videos and ease to use. The problem in this case is the upgrade from previous version. Which user account should be the default from scratch if the user had the instant uploads enabled?

2) In the below section of instant uploads, where the Camera folder and the Copy-Move options are placed, set another option to select the default account. This one has the advantage of giving users backwards compatibility: by not selecting any account, the behaviour would be upload to the current one, as right now is performed. By default the account is empty, so that upgrade from previous version is easier.

What do you mean? more possibilities?

cc @jabarros @davivel @malkomich @rperezb @owncloud/android-developers

davivel:
@jesmrec , I just pushed a work-in-progress branch with something similar to your option 2) . It's done from the recent changes to update instant uploads for Android 7, in branch https://github.com/owncloud/android/tree/instant_upload_set_account . 

Let's move the focus there, since @LukeOwncloud seems not available for this anymore.

Thanks

LukeOwncloud:
I am sorry, guys! Indeed I am not available these days. But I truly hope I will be eventually.

PR #1643, opened 2016-04-28
Uploads from content uris

jabarros:


src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -765,27 +766,40 @@ private void requestSimpleUpload(Intent data, int resultCode) {
...
+                    // Call to copy and then upload the selected file
+                    copyThenUpload(uris, remotePaths);
+
+                    cursor.close();
+                    return;
+
+                }
+                // and what happens in case of error?; wrong target name for the upload
+            } catch (Exception e) {
+                Log_OC.e(TAG, "Error while trying to copy and upload a content schema type file ", e);
```
malkomich: schema type **content**, in this case ;)

malkomich:
All code seems OK, just that log detail ^^ :+1:

malkomich:
Cool, we can ping @jesmrec to test it when he gets less busy ;)

jesmrec:
approved 👍

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -765,27 +766,39 @@ private void requestSimpleUpload(Intent data, int resultCode) {
...
+                    // Call to copy and then upload the selected file
+                    copyThenUpload(uris, remotePaths);
+
+                    cursor.close();
+                    return;
+
+                }
+                // and what happens in case of error?; wrong target name for the upload
+            } catch (Exception e) {
+                Log_OC.e(TAG, "Error while trying to copy and upload a schema type content file ", e);
```
davivel: This will end in a silent error; we should show the ```Toast``` that was introduced in version 2.0.0, previous line number 786

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -765,27 +766,39 @@ private void requestSimpleUpload(Intent data, int resultCode) {
...
+                    Uri[] uris = new Uri[]{selectedFileUri};
+                    String[] remotePaths = new String[]{remotePath};;
+
+                    // Call to copy and then upload the selected file
+                    copyThenUpload(uris, remotePaths);
+
+                    cursor.close();
+                    return;
+
+                }
```
davivel: The ```else``` case corresponding to the ```if``` block ending here should also lead to showing an error message via a ```Toast```. Let's compare it with the error messages shown in activity ```Uploader```

davivel:
@jabarros , let's refactor a bit.

Code in ```Uploader#uploadFiles()``` is better handling both ```file://``` and ```content://``` URIs pointing to files to upload. Let's put it in a common place so that ```FileDisplayActivity#requestSimpleUpload()``` takes advantage of it, and avoid this way duplicated code that evolves differently. We've suffered a lot of bugs in the past due to this.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -726,84 +726,29 @@ private void requestMultipleUpload(Intent data, int resultCode) {
...
-                Log_OC.e(TAG, "Couldn't resolve path to file");
-                Toast t = Toast.makeText(
-                        this, getString(R.string.filedisplay_unexpected_bad_get_content),
-                        Toast.LENGTH_LONG
-                );
-                t.show();
-                return;
-            }
-        }
+        ArrayList<Parcelable> mStreamsToUpload = new ArrayList<Parcelable>() {{
```
davivel: Local variable shouldn't have ```m``` prefix, you, copy-paster :P

src/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -177,7 +177,7 @@ public static String convertIdn(String url, boolean toASCII) {
      * @param filepath: Content Uri converted to string format
      * @return String: fileExtension (type '.pdf'). Empty if no extension
      */
-    public static String getComposedFileExtension(String filepath) {
+    public static String getFileExtension(String filepath) {
```
davivel: This method is not used anymore; should be removed.

src/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
```diff
@@ -238,7 +238,7 @@ protected void onPostExecute(ResultCode result) {
             listener.onTmpFilesCopied(result);
 
         } else {
-            Log_OC.i(TAG, "User left Uploader activity before the temporal copies were finished ");
+            Log_OC.i(TAG, "User left ReceiveExternalFilesActivity activity before the temporal copies were finished ");
```
davivel: This log should not include the name of the left activity now that could be any; just something like "the caller activity" or "the parent activity"

src/com/owncloud/android/ui/helpers/UriUploader.java
```diff
@@ -0,0 +1,188 @@
...
+            mCode = UriUploaderResultCode.ERROR_READ_PERMISSION_NOT_GRANTED;
+            Log_OC.e(TAG, "Permissions fail", e);
+
+        } catch (Exception e) {
+            mCode = UriUploaderResultCode.ERROR_UNKNOWN;
+            Log_OC.e(TAG, "Unexpted error", e);
+
+        } finally {
+            // Save the path to shared preferences; even if upload is not possible, user chose the folder
+            PreferenceManager.setLastUploadPath(mUploadPath, mActivity);
```
davivel: This should be kept only in ```ReceiveExternalFilesActivity```, was created just for that entry point.

src/com/owncloud/android/ui/helpers/UriUploader.java
```diff
@@ -0,0 +1,188 @@
...
+
+            }
+
+        } catch (SecurityException e) {
+            mCode = UriUploaderResultCode.ERROR_READ_PERMISSION_NOT_GRANTED;
+            Log_OC.e(TAG, "Permissions fail", e);
+
+        } catch (Exception e) {
+            mCode = UriUploaderResultCode.ERROR_UNKNOWN;
+            Log_OC.e(TAG, "Unexpted error", e);
```
davivel: Good moment to fix the typo in the log message :D

src/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -476,124 +471,40 @@ private boolean somethingToUpload() {
...
-     * @param sourceUris        Array of content:// URIs to the files to upload
-     * @param remotePaths       Array of absolute paths to set to the uploaded files
-     */
-    private void copyThenUpload(Uri[] sourceUris, String[] remotePaths) {
-        showWaitingCopyDialog();
-
-        CopyAndUploadContentUrisTask copyTask = new CopyAndUploadContentUrisTask(this, this);
-        FragmentManager fm = getSupportFragmentManager();
-        TaskRetainerFragment taskRetainerFragment =
-            (TaskRetainerFragment) fm.findFragmentByTag(FTAG_TASK_RETAINER_FRAGMENT);
```
davivel: This code is being lost; this is important, we can have problems in rotations. Need to have it also in ```UriUploader```

src/com/owncloud/android/ui/helpers/UriUploader.java
```diff
@@ -0,0 +1,188 @@
...
+     * @param remotePaths       Array of absolute paths to set to the uploaded files
+     */
+    private void copyThenUpload(Uri[] sourceUris, String[] remotePaths) {
+        if (mShowWaitingDialog) {
+            mActivity.showLoadingDialog(mActivity.getResources().
+                    getString(R.string.wait_for_tmp_copy_from_private_storage));
+        }
+
+        CopyAndUploadContentUrisTask copyTask = new CopyAndUploadContentUrisTask
+                ((CopyAndUploadContentUrisTask.OnCopyTmpFilesTaskListener)mActivity, mActivity);
```
davivel: The casting of the first parameter in the last minute is slightly dangerous. Let's avoid it adding a separated parameter in the constructor to receive an ```OnCopyTmpFilesTaskListener``` instance.

Besides, NULL should be a valid value for it. ```FileDisplayActivity``` is doing nothing in ```onTmpFilesCopied(...)```, could pass a NULL value for the listener and wouldn't need to implement ```OnCopyTmpFilesTaskListener```.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
davivel: Besides, we should move this code out of the method, or change it to receive directly an ``ArrayList<Parcelable>``` object instead of an ```Intent data```. See the caller code; in some situations ```requestSimpleUpload``` is called in a loop, what will result in a several ```UriUploader``` objects created and, more problematic, in several ```CopyAndUploadContentUrisTask```

davivel:
@jabarros , reviewed. Great job you did it there, let's give it the final shot.

davivel:
@jabarros , I made some cosmetic changes. IMHO this is :+1:

Take a look to my final commit and tell us what you think about.

jabarros:
Great last changes!
We have done a good improvement with this refactor.
I will do a rebase and ping QA to test it! :)

jesmrec:
Approved (again) 👍

davivel:
Seen last changes by @jabarros . :+1:

PR #1646, opened 2016-04-29
make readme/repository more attractive for contributors

jancborchardt:
Same as for iOS https://github.com/owncloud/ios/pull/636 – @tobiasKaminsky @AndyScherzinger @davivel please review. :)

@DeepDiver1975 can you also add https://play.google.com/store/apps/details?id=com.owncloud.android to the URL field of the repository? And change the description to:

> :iphone: Android app for ownCloud

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,15 @@
-#This is the Android client for [ownCloud][0]
+# [ownCloud](https://owncloud.org) Android app
+[![](https://owncloud.org/wp-content/themes/owncloudorgnew/assets/img/clients/buttons/googleplay.png)](https://play.google.com/store/apps/details?id=com.owncloud.android)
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Looks good to me :+1:

jancborchardt:
Ah I just remembered – maybe we should add the beta apk link for testing? @tobiasKaminsky which one should I best use? https://github.com/owncloud/android/blob/beta/apks/latest if that works?

AndyScherzinger:
latest should work just fine :) which ATM it is the 20160427 release which works best.

jancborchardt:
Problem is that I can not see the 0427 apk in that folder :/

tobiasKaminsky:
Great idea to link the beta.
I will make sure that there is a latest.apk that always link to the latest apk.

rperezb:
👍

davivel:
No QA needed. In

PR #1652, opened 2016-05-03
Change the default text color of login view to black

malkomich:


davivel:
Code :+1:

davivel:
Tested :+1:

PR #1655, opened 2016-05-04
Remote folder size

tobiasKaminsky:
Folder size is now shown by remote computed folder size.


jabarros:
LGTM

tobiasKaminsky:
@LukeOwncloud @AndyScherzinger do you want to do a code review?

AndyScherzinger:
Looks good to me :+1:

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -154,7 +154,11 @@ public static String getParentPath(String remotePath) {
     public static OCFile fillOCFile(RemoteFile remote) {
         OCFile file = new OCFile(remote.getRemotePath());
         file.setCreationTimestamp(remote.getCreationTimestamp());
-        file.setFileLength(remote.getLength());
+        if (remote.getMimeType().equalsIgnoreCase("DIR")){
+            file.setFileLength(remote.getSize());
```
davivel: Ah, I almost missed this part, that's why I thought it couldn't be working.

@PVince81 , do you recall what's the minimum server version including the property ```oc:size``` ?

PVince81: Not sure, could be OC 6 or OC 7. That property has been there forever... But it's only on folders as far as I remember, because files have "getcontentlength" but folders don't have that one officially.

davivel: Yes, you are right, that's what we need, the folders. Thanks!

tobiasKaminsky: I am always trying to hide the important things ;) ;)

davivel: LOL 

Was checking the library, we explicitly request the ```oc:size``` property from https://github.com/owncloud/android-library/pull/55 . We should make some minimum test with server 8.0.

Also would be wise to check what minimum OC server we grant the app works with. Do you recall it, @rperezb ?

davivel:
With OC server 8.0.12 works fine.

With OC server 7.0.14 all folders show 0B as size. Server 7 is ending it's life cycle, but we should support it.

Ideas to handle it:
 - in the android-library repo, change the default value of WebDavEntry#mSize to -1, instead of 0;
 - in this PR, when size is < 0, hide the size in the list of files;
 - in this PR, when server version is < 8.0, hide the option to sort by size; OR just leave it there, and let the folders are sorted according to the -1, maybe not so bad

tobiasKaminsky:
I think option 3 is ok as the user still can sort all files by size.

src/com/owncloud/android/utils/FileStorageUtils.java
rperezb: @davivel that's a good question!
For sure, I recall that we don't support oc4, I remember that the entrypoint changed that time (LOL), having said that, I will go with what we support as platform, we are about to stop supporting oc7 https://github.com/owncloud/core/wiki/Maintenance-and-Release-Schedule
@MTRichards is the one

rperezb:
Having in mind that oc7 is not being supported any more, I agree with @tobiasKaminsky

davivel:
> Having in mind that oc7 is not being supported any more

If that's the case, maybe we don't need the special cases. I'm not sure about server 8.0.0, but 8.0.12 is working correctly. Is it worth that we add special handling for possible differences between minor versions?

davivel:
Yes, checking https://github.com/owncloud/core/wiki/Maintenance-and-Release-Schedule seems that 7.0.15 dies this current month

rperezb:
@davivel 3 days...

tobiasKaminsky:
If OC7 is not supported any longer then we can omit every check for the version, which makes the code less complicated :+1:

davivel:
Recovering focus on this. 

Works with all the maintained versions of server, so :+1: from code perspective.

Unfortunately, there's too much already for 2.1.0, so needs to go to 2.2.0.

@jesmrec , ping us before testing to rebase.

jesmrec:
@davivel, i'm going to start to work with this PR. If you can rebase it...

davivel:
Sure, in a minute.

davivel:
@jesmrec , rebased.

davivel:
All is ready -> in

PR #1659, opened 2016-05-05
Automatic unit tests

davivel:
Added first local unit test, runnable in Gradle/Android Studio environment


davivel:
cc @jesmrec , @jabarros

jabarros:
👍

PR #1665, opened 2016-05-09
Add instantupload behaviour copy

zmatsuo:
I was set options as follow:
  Original file will be..  = "kept in original folder"
  "Upload pictures via wifi only" = off

When I took a picture without wifi, Owncloud app uploads the picture with 3G.
And I want to look the picture with Owncloud app(I love grid view),
Owncloud app downloads the picture with _3G_.

So I think that need "copied to app folder" option.
This is the reason that I made this PR.


tobiasKaminsky:
This was the previous behaviour prior to my PR.
And in all discussions there was no need to have an option to keep it the previous way.
But, for me, this would be ok.
@jancborchardt @davivel ?

res/values-ja-rJP/strings.xml
```diff
@@ -16,9 +16,9 @@
   <string-array name="actionbar_sortby">
     <item>A-Z</item>
     <item>最新 - 最古</item>
+    <!--TODO re-enable when server-side folder size calculation is available   
+    <item>Biggest - Smallest</item>-->
```
tobiasKaminsky: Please do not modify any language specific files as they are all handled by Transifex.

res/values/attrs.xml
```diff
@@ -14,10 +14,12 @@
...
 	</string-array>
 
 	<string-array name="pref_behaviour_entryValues">
 		<item>NOTHING</item>
 		<item>MOVE</item>
+		<item>COPY</item>
 	</string-array>
-</resources>
\ No newline at end of file
+</resources>
```
tobiasKaminsky: To better and easier reviews: please try to avoid such changes as they are not changing real code but show up as changes.

zmatsuo:
I reflected a review. Thank you @tobiasKaminsky .

PR #1671, opened 2016-05-10
Added first instrumented tests

davivel:


davivel:
@jabarros , @jesmrec , should be just burocracy

jesmrec:
👍 all right

davivel:
@AndyScherzinger , that duplication was on purpose. I will add a comment. But please, we need to complete branch merging first.

We'll create a final PR from automatic_tests to master and wait for comments / review of everybody.

Thanks for your patience.

AndyScherzinger:
Allright, sorry but any trouble I might have caused!

PR #1681, opened 2016-05-18
Translation of de-AT, Typo

theskystalker:
I'm using German Android with country settings Austria, so the App is merging android/res/values-de/strings.xml with android/res/values-de-rAT/strings.xml because values-de-rAT is currently not complete. This file should only contain country specific strings, but now it contains some strings, and some not, and many strings with the same meaning in AT and DE are different with no qualification.

### DIFF
--- a/res/values-de-rAT/strings.xml
+++ b/res/values-de-rAT/strings.xml
@@ -21,7 +21,7 @@
     <string name="drawer_item_on_device">On device</string>-->
   <string name="prefs_category_general">Allgemein</string>
   <string name="prefs_accounts">Konten</string>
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Hi @theskystalker all translations are handled through Transifex https://www.transifex.com/owncloud-org/owncloud/language/de_AT/ and then automatically pushed to this repository. So in order to change any translations you would need to register on transifex, request access on the owncloud project and then do the translation. Your changes would then by incorporated overnight.

theskystalker:
Hi, thanks i will try it over transifex, this request can be closed.
BR
Stefan

Issue #1136, opened 2015-09-04
Automated synchronization of folders (scope: inside 'owncloud' folder)


MTRichards:
# As an ownCloud user, I want to be able to sync a folder from my Android mobile device to the server so that I can access and edit files easily across devices without thinking about it

Acceptance Criteria:
- Two-way syncing, from the folder set, to all its descendants
- For folders existing in an OC account (syncing local folders only existing in the device is out of the scope of this issue).
- The current Favorites star is to be renamed ~~Sync~~ "Available offline"
- The Sync star icon is to be replaced with  ~~a circular static sync icon~~ an arrow pointing downwards, place over an horizontal line; see more about icons in #1218 
- ~~Sync icon is to be the same color as the star and found in the same location on the iOS device (yellow)~~ (iOS)
- ~~Sync icon will automatically replace existing favorite directories on files currently set as favorites~~
- ~~Sync for iOS can only occur when the app is open, as there is no service mechanism for initiating sync in the background. As a result, every time the app is opened, the local cache is refreshed, and any new files are synced with the server.~~ (iOS)
- Once the app is opened, sync of individual directories happens when the user navigates into the directory
- When a user is browsing files and pulls down to refresh the directory, any files and folders in that directory are synced automatically (including folder children).
- When a directory is syncing, if ~~the app~~any folder is blocked, a circular icon shows that ~~the app~~ the folder is working on the sync

---

@davivel updated after May 16 Design Days

---

edited by @jabarros
- [x] Create 'available_offline_folders' branch from master in android repository @jabarros
- [x] [DEV] Allow set a folder as favorite (menu) @jabarros
- [x] [DEV] Save the favorite folder state on db and unset as favorite all children files that were favorites @jabarros
- [x] [DEV] Modify places where favorite folders are being syncing to sync also folders when file is folder: Case of folders syncing (pull down or enter in) @jabarros
- [x] [DEV] Modify places where favorite folders are being synced to sync also folders when file is folder: Case of full account sync and refresh account (in menu) @jabarros
- [x] [DEV] Check service that observes changes (FileObserveService) to take into account changes inside sync folders @malkomich
- [x] [DEV] Get folder data correctly, to assure a favorite keeps being favorite while a refresh operation is done @malkomich
- [x] [DEV] Create method to check if file is favorite or belongs to a favorite folder @jabarros
- [x] [DEV] Check on unfavorite request if the file is favorite by inheritance or not to allow execute the operation @malkomich
[...]

PR #1698, opened 2016-06-02
Available offline folders

jabarros:
Issue #1136


davivel:
Reviewing...

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -309,7 +294,7 @@ public void saveFolder(
             cv.put(ProviderTableMeta.FILE_ACCOUNT_OWNER, mAccount.name);
             cv.put(ProviderTableMeta.FILE_LAST_SYNC_DATE, file.getLastSyncDateForProperties());
             cv.put(ProviderTableMeta.FILE_LAST_SYNC_DATE_FOR_DATA, file.getLastSyncDateForData());
-            cv.put(ProviderTableMeta.FILE_KEEP_IN_SYNC, file.isFavorite() ? 1 : 0);
+            //cv.put(ProviderTableMeta.FILE_KEEP_IN_SYNC, file.getFavoriteStatus());
```
davivel: The comment should be removed.

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -387,7 +372,7 @@ public void saveFolder(
         cv.put(ProviderTableMeta.FILE_ACCOUNT_OWNER, mAccount.name);
         cv.put(ProviderTableMeta.FILE_LAST_SYNC_DATE, folder.getLastSyncDateForProperties());
         cv.put(ProviderTableMeta.FILE_LAST_SYNC_DATE_FOR_DATA, folder.getLastSyncDateForData());
-        cv.put(ProviderTableMeta.FILE_KEEP_IN_SYNC, folder.isFavorite() ? 1 : 0);
+        //cv.put(ProviderTableMeta.FILE_KEEP_IN_SYNC, folder.getFavoriteStatus());
```
davivel: Same here.

src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -54,6 +54,47 @@ public OCFile createFromParcel(Parcel source) {
     public static final String PATH_SEPARATOR = "/";
     public static final String ROOT_PATH = PATH_SEPARATOR;
 
+    public enum FavoriteStatus {
```
davivel: This is a great opportunity to stop calling it "Favorite" also internally, and start calling it "AvailableOffline".

src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -68,7 +109,7 @@ public OCFile createFromParcel(Parcel source) {
     private boolean mNeedsUpdating;
     private long mLastSyncDateForProperties;
     private long mLastSyncDateForData;
-    private boolean mFavorite;
+    private int mFavoriteStatus;
```
davivel: Why ``ìnt```member, instead of using ```FavoriteStatus```?

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -214,14 +214,14 @@ private void filter(List<Integer> toShow, List <Integer> toHide) {
         }
 
         // FAVORITES
-        if (mFile == null || synchronizing || mFile.isFolder() || mFile.isFavorite()) {
+        if (mFile == null || synchronizing || mFile.getFavoriteStatus() == OCFile.FavoriteStatus.FAVORITE.getValue()) {
```
davivel: If ```getFavoriteStatus``` returned a ```FavoriteStatus``` value instead of ```int```, ```.getValue()``` wouldn't be needed.

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -465,10 +470,23 @@ public void toggleFavorite(OCFile file, boolean isFavorite) {
...
-        if (file.isFavorite()) {
+        if (file.getFavoriteStatus() == OCFile.FavoriteStatus.FAVORITE.getValue()) {
             syncFile(file);
         }
     }
+
+    private void toggleAvailableOfflineFilesInFolder(OCFile file, boolean isAvailableOffline) {
+        int favoriteStatus = isAvailableOffline ?
+                OCFile.FavoriteStatus.FAVORITE.getValue() : OCFile.FavoriteStatus.NO_FAVORITE.getValue();
+        Vector<OCFile> filesInFolder = mFileActivity.getStorageManager().getFolderContent(file);
```
davivel: This might be hard in performance if the subtree under ```file``` is really big. We should try to change the property for any descendant minimizing database access.

src/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -349,11 +349,6 @@ private void synchronizeData(ArrayList<Object> folderAndFiles, OwnCloudClient cl
         // get 'fresh data' from the database
         mLocalFolder = mStorageManager.getFileByPath(mLocalFolder.getRemotePath());
 
-        // parse data from remote folder 
-        OCFile remoteFolder = FileStorageUtils.fillOCFile((RemoteFile) folderAndFiles.get(0));
```
davivel: Sorry, I'm confused. Why are these three lines removed?

davivel:
Seen, @jabarros see my comments, please.

src/com/owncloud/android/operations/RefreshFolderOperation.java
malkomich: Because it was creating a new OCFile from the remote file data, without enough information like the favourite state.

I replaced it with an operation from the storage manager (line 440) which gets all the needed data including favorite state, since the folders to sync here are always previously created, to save correctly the folder without losing its data.

davivel: The ```OCFile``` instance you are retrieving in line 440 is the same you have in the field ```mLocalFolder```; that's why its local properties are copied into ```remoteFolder``` right in the code where you removed it. The idea is not doing an extra call to the database, since we consider at this moment that ```mLocalFolder``` is "fresh enough".

If you miss a local property, such as availableOfflineStatus (favorite before), it could just be copied in the same way.

davivel: > If you miss a local property, such as availableOfflineStatus (favorite before), it could just be copied in the same way.

Even better, apply the change @jabarros made later, modifying ```saveFolder``` so that it doesn't change ```availableOfflineStatus``` at all. We should change at some moment ```FileDataStorageManager.saveFolder``` and ```saveFile``` so that only properties that might really have changed are updated.

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -276,16 +276,27 @@ private void addObservedFile(OCFile file, Account account) {
...
     private void addObservedFile(String localPath, Account account) {
         File file = new File(localPath);
-        String parentPath = file.getParent();
-        FolderObserver observer = mFolderObserversMap.get(parentPath);
-        if (observer == null) {
-            observer = new FolderObserver(parentPath, account, getApplicationContext());
-            mFolderObserversMap.put(parentPath, observer);
-            Log_OC.d(TAG, "Observer added for parent folder " + parentPath + "/");
+
+        if(file.isDirectory()) {
```
davivel: This will not work if no file of the folder was downloaded before, the folder will not exist in the local file system.

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -276,16 +276,27 @@ private void addObservedFile(OCFile file, Account account) {
...
+            File[] children = file.listFiles();
```
davivel: And similarly, this will only add to observe the files that were downloaded before.

We need to extend ```FolderObserver.java``` so that can attend any file inside.

davivel: Might seem odd, since we only observe changes in files downloaded and we enforce a synchronization when a folder/file is 

But think in the next case:
1. Folder is set as available offline, and all its contents are correctly sync'ed.
2. A file is added in the server side into the the folder (via web UI or desktop client)
3. The user browses in the folder via the Android app, or a full account sync is triggered; the new file in the server side is downloaded
-- from this moment, the file should be monitored for local changes; but it's not, because it didn't exist when the folder was set as available offline.

davivel:
Rebased on master. Let's give it another shot. Users need it!

davivel:
Resuming. Let's finish this.

davivel:
Rebased on master, let's move on.

davivel:
Tracking pending tasks in #1136

src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -60,6 +60,47 @@ public OCFile createFromParcel(Parcel source) {
...
+         */
+        NOT_AVAILABLE_OFFLINE(0),
+
+        /**
+         * File is available offline
+         */
+        AVAILABLE_OFFLINE(1),
+
+        /**
+         * File belongs to any available offline folder
```
davivel: Typo in comment - 'an'

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -790,23 +867,45 @@ private boolean copyFile(File src, File target) {
...
 
-        if (c.moveToFirst()) {
-            do {
-                OCFile child = createFileInstance(c);
-                // TODO Enable when "On Device" is recovered ?
-                // if (child.isFolder() || !onlyOnDevice || onlyOnDevice && child.isDown()){
-                ret.add(child);
-                // }
-            } while (c.moveToNext());
+        if (c!= null) {
```
davivel: This "patch" is also needed for methods that return individual files.

davivel: No, this is not a good idea. We are already updating the property ```FILE_KEEP_IN_SYNC``` for all the descendents when its value is changed by the user for a folder.

For descendent files/folders that are created or discovered (from the server) after the user sets a folder as available offline, we should write the appropriate inherited value then - better increase time when writing the database than when reading it. 

So let's remove this from here and store ```AVAILABLE_OFFLINE_PARENT``` correctly instead of inferring it on reads.

davivel: DONE.

davivel:
@jesmrec , ready to test from my POV.

@nasli , @owncloud/android-developers , other reviews are welcome.

davivel:
Rebased, and all known bugs fixed or ready to be tested.

jesmrec:
Approved

src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -60,6 +60,47 @@ public OCFile createFromParcel(Parcel source) {
...
+
+        AvailableOfflineStatus(int value) {
+            this.value = value;
+        }
+
+        public int getValue() {
+            return value;
+        }
+
+        public static AvailableOfflineStatus fromValue(int value) {
```
jrodgut-sg: you could use existing .ordinal() method and values()[index] to get them so that you wouldn't need to have a value attribute (though maybe less control as you'd need to take care to add new enum values at the end just to preserve existing indexes)

davivel: Sure, though is it a big win? Keeping an ordinal relationship between the elements is not important for our case, and the attribute is not too heavy. I'd rather to have control about the values than about the order.

Is there any other benefit I'm losing?

jrodgut-sg: Nah,  I don't see disadvantages on having more attributes, I just meant that the benefit would be that you'd make sure that the same index is not used in 2 elements by mistake as it is assigned automatically

davivel: Good point. There it goes :)

jrodgut-sg: 👍

jrodgut-sg:
👍

PR #1699, opened 2016-06-03
Allow changing APK names

davivel:


davivel:
@jabarros , @malkomich , have a look please.

No change in the app itself.

jabarros:
Reviewed! :+1:

PR #1700, opened 2016-06-03
Load release signing config from environment variables

davivel:


davivel:
cc @jabarros @malkomich 

No change in the app itself, no QA needed.

jabarros:
:+1: !

PR #1753, opened 2016-07-27
Changed delay of the runnable in startSyncFolderOperation adding 350ms.

atsag:
Added a further 350ms delay before the execution of the runnable in the startSyncFolderOperation method, so as to ensure that a refresh of the folder data will happen. The 350ms delay is a pure product of experimentation, and so may be adjusted according to the needs.


davivel:
Thanks, @atsag .

Seems you inadvertently changed the reference to the ownCloud-library project. Could you set it back to previous value, and push your branch again?

atsag:
@davivel   Yes, but i could use some help with the necessary steps.

davivel:
Sure! Let's see if we can find the right commands. Starting in the root folder of your project, type next commands:

```
git checkout refresh_folders_startup
cd owncloud-android-library
git checkout master
cd ..
cd user_manual/ocdoc
git checkout master
cd ..
git add .
git commit -m "Fix reference to subprojects"
git push origin refresh_folders_startup
```

That should make it.

Let me know if anything goes wrong.

atsag:
I am sorry @davivel but I have a problem with my owncloud-android-library submodule, and so cannot complete the build.

davivel:
I guess that the commands in owncloud-android-library are failing...

Try this other thing instead:

```
git checkout master
git pull upstream
git submodule update
git checkout refresh_folders_startup
git add .
git commit -m "Fix reference to subprojects"
git push origin refresh_folders_startup
```

PR #1764, opened 2016-08-11
Automatic tests appium

jesmrec:
Appium tests updated:
- Login suite


davivel:
This is deprecated in favor of #1835

PR #1765, opened 2016-08-11
Changed delay of the runnable in startSyncFolderOperation adding 350ms.

atsag:
@davivel 
David thank you for all your help. This is the same as https://github.com/owncloud/android/pull/1753 , but now the code and the commits are sorted in a better way.


davivel:
Thanks @atsag.

:+1:

PR #1770, opened 2016-08-17
Adapt to new OC_Log.startLogging(...) with path parameter

davivel:
@jesmrec , could receive minimal testing to check that logs are OK. I wouldn't create a test plan.

Needed for owncloud/android-library#96


davivel:
Uf, this has been killed by the rebase of the library on master.

We need to #1771 before this one.

PR #1744, opened 2016-07-20
Multi-Window support added for android N

GurpreetSK95:
device-2016-07-20-155248

PR #1771, opened 2016-08-17
Update of build target to Android Nougat, with multiwindow

davivel:
Replaces #1744, after rebase and update of reference to library

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ before_install:
   - chmod +x ./wait_for_emulator.sh
   - ./wait_for_emulator.sh
 script:
+# force init of adb way before any Gradle task that uses ADB; prevents a timeout error that skips emulators of android-24
+  - adb devices
 # build app and assemble APK, in debug mode
   - ./gradlew assembleDebug
 # run all the local unit tests of app module
@@ -33,6 +35,7 @@ script:
   - adb shell am instrument -w -e debug false -e class com.owncloud.android.datamodel.OCFileUnitTest com.owncloud.android.test/android.support.test.runner.AndroidJUnitRunner
 env:
   global:
+  - ADB_INSTALL_TIMEOUT=10 # minutes (4 minutes by default)
   - secure: h4Y7ZvgbvOj5T71ubRcw3Fy3KXF8qHugRFLGK3q2R9YuRsDAf8XH+Y/UiXyH8sac2QSj7Zlny1kA1DEJgwhTXs9wsAVKVJCQNTJGEvhm/4uQgPeNMzMv07Lqe8V+KUBsFH5qhfPJO357ERW0k2f2qljoLSHtHStclt7iGvFdynA=
   - secure: o9L6lXWpXowhQSdiUSmajliBUkQ6n7NrBUqhC09lqe7yXSGhEsgGRXqHoT3q2B4uIqGSiLCa9HQbW0dfDQCs+pADmzHIl3zbTViR88TSaIhOiTrqMUUl5iaO++pneZ2TzgU9bbGHbl6Ixjc6iALH2+F7P+RUM6vLTNPcfnCJa3g=
   - secure: ydxZrS7+1ht3p1tC6DE9W9bjLQGjMkwFBwyhNmcdEM6538kN8ZCBQe7NxSsCrC1nSDBLQ17Cziv0XJHl/pCfPrbkrPPgVFpjDfPeqC2zcGcCNcFQUEylXCvQ4uRU2hKL6dPqCsOQ57Pv3qwpPoprl/usoN5Wh8V7BKplU88ZaYM=
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,6 @@ script:
   - adb shell am instrument -w -e debug false -e class com.owncloud.android.datamodel.OCFileUnitTest com.owncloud.android.test/android.support.test.runner.AndroidJUnitRunner
 env:
   global:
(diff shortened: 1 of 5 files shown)


davivel:
OK, license/cla test now passes because @GurpreetSK95 already signed in https://github.com/owncloud/android-library/pull/126 .

:+1: ; please, note I'm not approving my one code, I just rebased code by @GurpreetSK95 from #1744 .

@jesmrec , this is basically a technical PR, but if maybe you want to test multi-window mode a big. Shouldn't be a big deal, I don't think it requires a test plan; it's a feature of the OS, besides, not of the app.

AndyScherzinger:
@davivel this PR isn't working for travis! In order to get this working it needs to:
* target v24 but run the tests on v23
* not use JACK since that seems to break the travis builds too

davivel:
Thanks, @AndyScherzinger . I forgot to commit the update of the reference to the library, that's why appear v23 builds.

Let's see if that's enough. What problems do you know about JACK in Travis?

AndyScherzinger:
I never got that working (yet), for some reasons the builds keep failing. But lets see if it works now :)

davivel:
There it goes, our beloved workaround app...

davivel:
@AndyScherzinger , you were right, something is not working between Jack & Travis.

Disabled Jack -> back to Java 7

AndyScherzinger:
Pushed a fix, let's see if travis is happy with it :)

davivel:
The idea is reduce dependencies -> get rid of android-23

AndyScherzinger:
@davivel I know, but when I was playing around with travis a lately builds kept failing with using the emulator with a v24 image...

AndyScherzinger:
both builds are running. Let's see who wins :wink:

davivel:
:D

davivel:
You did it

```test(AVD) - 7.0' for 'android:debug': Unknown API Level```

stupid emulator...

AndyScherzinger:
I missed updating the android-lib submodule... so my build broke due to the 24.0.0 build tool version from the old submodule link :(

AndyScherzinger:
I'll try another push...

davivel:
Let's see. 

Anyway, the problem of the emulator is installing the APK. The tests in the library are working fine.

Might be a timeout problem (something like http://stackoverflow.com/questions/32952413/gradle-commands-fail-on-api-23-google-api-emulator-image-armeabi-v7a). In that case, might be prevented installing without gradle.

AndyScherzinger:
The timeout could be fixed like this I guess: https://github.com/googlemaps/android-maps-utils/commit/807f515f535a4cea06ed8b4299548ed74350a126
```  - ADB_INSTALL_TIMEOUT=20 # minutes (2 minutes by default)```

davivel:
Forget it, doesn't make sense. That job took just 31.429 secs to fail. Can't be that. 

We'll have to go with the android-23 emulator for the moment.

AndyScherzinger:
@davivel build works :smiley:

davivel:
OK, I'll have a look locally to the issue of the emulator. If I don't find out something in a while, let's just move on with this.

:+1: for you, @AndyScherzinger  :)

jesmrec:
I have checked the multiWindow in emulator and real device with Android N and works properly. The apk is also correctly built in this branch ;)

Approved. 

cc @davivel @AndyScherzinger

davivel:
This should make the trick to work with emulator of Android 24. I think the problem is this other timeout: https://android.googlesource.com/platform/tools/base/+/master/ddmlib/src/main/java/com/android/ddmlib/Device.java#90 

That is not configurable. Seems the ```installDebug``` task can't get the API level of the emulator fast enough because it's the first task working with ```adb``` and the call to get the API level finishes before is completely up with wrong value. That would explain the ```test(AVD) - 7.0' for 'android:debug': Unknown API Level``` message.

AndyScherzinger:
@davivel ```test(AVD) - 7.0' for 'android:debug': Unknown API Level``` issue remains :( Maybe we still need to add the the ```- ADB_INSTALL_TIMEOUT=20 # minutes (2 minutes by default)``` or something similar as some kind of a "wait"?

davivel:
Dammit. It worked perfectly in the test branch.

I'll make a last try with ```- ADB_INSTALL_TIMEOUT=20``` and moving the ```adb devices``` earlier. If that fails, let's go with android-23 emulator.

davivel:
Enough. It's not stable, and even when it works, the next test fails.

Rolling back to your proposal.

davivel:
Great. So, the source of the problem is not the version of the emulator itself. Probably the upgrade of gradle tools or the build tools. Thinking...

davivel:
Seems that ```adbOptions``` in ```build.gradle``` helped; or maybe it's just a coincidence, but two greens in a row is much better than what we had before. Let's go with this, and if breaks again, will disable instrumented tests until Google or Travis improve this mess.

davivel:
Come on, rebase on master :S

AndyScherzinger:
Downside is now that the CI burns 30 minutes to do this :(

davivel:
Well, CI it's not supposed to be fast... but yes, I share the feeling.

AndyScherzinger:
It'll die again... https://travis-ci.org/owncloud/android/builds/153250705#L2093

davivel:
Disabling.

davivel:
@AndyScherzinger , are you fine with this?

AndyScherzinger:
@davivel absolutely! :+1:

davivel:
Then all in.

Thanks both @GurpreetSK95 and @AndyScherzinger for your work here.

PR #1783, opened 2016-08-25
Instant uploads based on watch for new files in camera folder

### LINKED ISSUE #6: Instant Upload based on folder observance instead of broadcasted events
We discussed and debugged the issue already on IRC some time ago. It seems that the camera App from HTC does something different which doesn't trigger the instant upload. Tested with HTC Desire S (Android 2.3.5, HTC Sense 3.0) and the ownCloud Android Client 1.3.8

EDITED by @davivel
SUMMARY on 26/02/2016

The problem: third party camera apps often fail to broadcast the ```Intent```s with action ```android.hardware.action.NEW_PICTURE``` and ```android.hardware.action.NEW_VIDEO``` that the OC app listens for to know when a new picture or video was taken.

The probable solution: keep alive a FolderObserver on the official folder where pictures / videos are stored to know when a new picture or video was taken.

Main references on this: 
https://github.com/owncloud/android/issues/6#issuecomment-22494946
https://github.com/owncloud/android/issues/6#issuecomment-60381095
#1059 

To keep in mind:
- There are third party camera apps that don't use the official DCIM folder to store their captures; let's ignore them first. 
- Android Services are not granted to be there forever; this could result in problems to keep the FolderObserver alive; risk is low, since that's the approach already used to observe kept-in-sync / favourite files, and our experience in this point has been fine.
- The app keeps too many Services. We should tend to merge them in a different architecture, with a more uniform synchronization model. Internally, the concept of 'instant upload' should evolve to 'automatic addition, later sync'.

davivel:
Fixes #6, and means instant uploads... :
  - ... work in Android 7;
  - ... work with more camera apps.

New requisites for camera apps will be:
  - store pictures and videos in public storage (there's not much sense in doing other way);
  - store pictures and videos in same folder (this can be extended in the future if needed)

____

BUGS & IMPROVEMENTS

- [X] Camera MX does not upload videos https://github.com/owncloud/android/pull/1783#issuecomment-242351587  [FIXED]
- [X] tmp file in video uploads "only wifi" with no connection https://github.com/owncloud/android/pull/1783#issuecomment-242382128 [FIXED]
- [X] Changing from 4G to only wifi, the last file is not retried https://github.com/owncloud/android/pull/1783#issuecomment-242670409 [FIXED]
- [X] Instant uploads with default camera do not work in Samsung Note 4 with Android 6 https://github.com/owncloud/android/pull/1783#issuecomment-242675221 [WONT]. Moved to https://github.com/owncloud/android/issues/1786


davivel:
@jesmrec , ready to test.

@nasli and @owncloud/android-developers , code reviews are welcome.

src/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -35,25 +39,79 @@
...
+
+    private static final String PREF__INSTANT_PICTURE_ENABLED = "instant_uploading";
+    private static final String PREF__INSTANT_VIDEO_ENABLED = "instant_video_uploading";
+    private static final String PREF__INSTANT_PICTURE_WIFI_ONLY = "instant_upload_on_wifi";
+    private static final String PREF__INSTANT_VIDEO_WIFI_ONLY = "instant_video_upload_on_wifi";
+    private static final String PREF__INSTANT_UPLOAD_ACCOUNT_NAME = "instant_upload_account_name";  // NEW - not saved yet
+    private static final String PREF__INSTANT_PICTURE_UPLOAD_PATH = "instant_upload_path";
+    private static final String PREF__INSTANT_VIDEO_UPLOAD_PATH = "instant_video_upload_path";
+    private static final String PREF__INSTANT_UPLOAD_BEHAVIOUR = "prefs_instant_behaviour";
+    private static final String PREF__INSTANT_UPLOAD_SOURCE = "instant_upload_source_path";   // NEW - not saved yet
```
davivel: Self note: ```instant_upload_source_path``` is saved, the comment is wrong.

src/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -35,25 +39,79 @@
...
+        result.setWifiOnlyForPictures(
+            prefs.getBoolean(PREF__INSTANT_PICTURE_WIFI_ONLY, false)
+        );
+        result.setWifiOnlyForVideos(
+            prefs.getBoolean(PREF__INSTANT_VIDEO_WIFI_ONLY, false)
+        );
+        result.setUploadAccountName(
+            prefs.getString(
+                PREF__INSTANT_UPLOAD_ACCOUNT_NAME,
+                AccountUtils.getCurrentOwnCloudAccount(context).name
```
davivel: Self-note: This will produce a ```NPE``` when all accounts are deleted. NEEDs to be fixed.

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -338,7 +382,46 @@ private void removeObservedFile(String localPath) {
...
+            mInstantUploadsObserver = new InstantUploadsObserver(config, getApplicationContext());
+            mInstantUploadsObserver.startWatching();
+
+        } else if ( mInstantUploadsObserver != null) {
+            // observer exists and can handle changes in configuration -> update observer
+            mInstantUploadsObserver.updateConfiguration(config);
+
+        } else {
+            Log_OC.w(TAG, "No current observer for instant uploads, nothing to observe in configuration");
+            // this should not happen, but it's not dangerous
```
davivel: Comment is wrong, this will happen when instant uploads are disabled at any time that the service is initialized (for instance: in BOOT time or when the OC app is started).

Log should be ```info```, not level.

src/com/owncloud/android/services/observer/InstantUploadsObserver.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.Set;
+
+/**
+ * Observer watching a folder to request the upload of new pictures or videos inside it.
+ */
+public class InstantUploadsObserver extends FileObserver {
+
+    private static String TAG = InstantUploadsObserver.class.getSimpleName();
+
+    // TODO calculate CORRECT mask
```
davivel: It's already correct, AFAIK.

src/com/owncloud/android/services/observer/InstantUploadsObserver.java
```diff
@@ -0,0 +1,251 @@
...
+     * @param context           Used to start an operation to upload a file, when needed.
+     */
+    public InstantUploadsObserver(InstantUploadsConfiguration configuration, Context context) {
+        super(configuration.getSourcePath(), UPDATE_MASK);
+
+        if (context == null) {
+            throw new IllegalArgumentException("NULL context argument received");
+        }
+
+        // TODO - work if camera folder doesn't exist, but is created later?
```
davivel: Works, remove.

src/com/owncloud/android/ui/activity/LocalFolderPickerActivity.java
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        // inflate and set the layout view
+        setContentView(R.layout.files_folder_picker);   // beware - inflated in other activities too
+        if (savedInstanceState == null) {
+            createFragments();
+       }
+
+        // set input controllers
+        //mCancelBtn = (Button) findViewById(R.id.upload_files_btn_cancel);
```
nasli: Removed if not needed

nasli:
Wow, awesome refactor and fast-timing for Android7 ! Let's release it! 👍

davivel:
For the record: Travis fails getting build dependencies from Google, code is OK. Let's see with a retry,

jesmrec:
@davivel i have noticed this:

The "Camera MX" app does not upload the videos taken into its folder `...DCIM/CameraMX' . The following log is print at the moment the video starts recording:

`08-25 13:00:37.256 11051-11068/com.owncloud.android W/InstantUploadsObserver: Camera app saved an empty or temporary file, ignoring VIDEO_20160825_130036.mp4`

With this app the photos works correctly. I have tested other apps and works properly with videos and photos. 

Perhaps is a corner case that only happens with this app, but having a look can be a good idea :)

davivel:
I'll have a look to see if we can do something fast. Video recording might become a problem. Seems that every camera app handles file its-very-own-way. I tested with Google Camera and OpenCamera, both save the video in different ways (OpenCamera keeps it simple). 

Let me debug a bit and cross your fingers.

Any case involving video uploads should be blocked meanwhile.

davivel:
MOG. The camera is "writing&closing" the same file twice after creating. Probably it's doing:
 - start recording: create new file, then close it empty;
 - stop recording: open same file, save video and close file;

We are not tracking that workflow. And it's a crappy one, since it's indistinguishable of updating a video file, what we didn't want to track.

I need to think a bit about this. Ideas are welcome.

Instant video was working for this app in Android 6 with ownCloud 2.1.0 ?

jesmrec:
i'm not pretty sure but i think they did. Checking...

jesmrec:
@davivel checked with 2.0.1 + Camera MX and works correctly. Videos & Photos are uploaded without error.

davivel:
OK, seems I have a solution.

davivel:
@jesmrec, should work now.

@nasli , you should give another look to code updated. All the changes are in file ```InstantUploadsObserver```

jesmrec:
With Android5 and Android4 (backwards compatibility)

Steps:

1. Enable instant uploads for videos with "only wifi"
2. Disable wifi
3. Take a video

Actual behaviour:

In uploads view, the video appears twice as pending to upload

 Video_1234.mp4
 Video_1234.mp4.tmp

Expected behaviour:

In uploads view, only one file related to the video appears in the queue

Nexus5 v5.0.1

davivel:
Really? That .tmp never should have got here.

What camera app are you using in those devices?

jesmrec:
i used default camera.

davivel:
OK, that was stupid from my side.

Fixed now also, @jesmrec

jesmrec:
### Bug (low severity)

### Steps:

1. In a device with a SIM card, enable the instant uploads with only wifi
2. Switch the wifi off (4G is up)
3. Take several photos/videos -> pending of wifi connection
4. Switch the wifi on

### Actual behaviour

All the photos/videos are uploaded but the last one, that have to be tapped manually to upload

### Expected behaviour

All the photos are uploaded


From plane mode to wifi mode -> works fine

Tested with Android 7. Lower Android versions, work correctly

davivel:
Taking a look.

jesmrec:
### Bug (only in samsung note 4 Android 6)

1. Enable instant uploads 
2. Set the default camera folder with the default camera
3. Take some photos and videos

### Actual behaviour

Nothing is uploaded

### Expected behaviour

Content is uploaded

davivel:
That's worse. Checking.

davivel:
As discussed offline: this last bug is very, very specific of the OS version + camera app + camera app configuration, maybe also model, and there is no way we can fix this with a limited effort. Let's open another issue for it.

jesmrec:
@davivel working fine. Approved.

Now we should be aware of posible differences related to device/software vendors that can cause exceptions like https://github.com/owncloud/android/pull/1783#issuecomment-242675221. 

Great job ;)

davivel:
Rebasing.

Great job you :D

PR #1808, opened 2016-09-14
dian

EricHehe:
CustomPopup.this.dismiss();?????????????
i'm playing a joke


davivel:
AFAIK that class is dead code. Better delete it than fix it.

>i'm playing a joke

Then I guess we can close the PR :)

Issue #1818, opened 2016-10-01
resource leak: database cursor not closed


andrew659:
### Actual behaviour
-database cursor may be left open

### Expected behaviour
-cursors even empty ones (moveToFirst returns false) should be closed
 
The database cursor used by the delete() method in FileContentProvider class may be left unclosed if cursor.moveToFirst() returns false:

https://github.com/owncloud/android/blob/master/src/com/owncloud/android/providers/FileContentProvider.java

```
if (c != null && c.moveToFirst()) {
                    remoteId = c.getString(c.getColumnIndex(ProviderTableMeta.FILE_REMOTE_ID));
                    //ThumbnailsCacheManager.removeFileFromCache(remoteId);
                    c.close();
                }
```

Is this a bug?

GurpreetSK95:
@andrew659 The code should be put inside a try block and the cursor should be closed in the finally block. I'm working on it RN

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #1819, opened 2016-10-03
Resource leak db

GurpreetSK95:
Issue #1818 
Changes made to FileContentProvider.java


src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -96,20 +96,26 @@ private int delete(SQLiteDatabase db, Uri uri, String where, String[] whereArgs)
...
+                    Log_OC.d(TAG, "Removing FILE " + remoteId);
+
+                    count = db.delete(ProviderTableMeta.FILE_TABLE_NAME,
+                            ProviderTableMeta._ID
+                                    + "="
+                                    + uri.getPathSegments().get(1)
+                                    + (!TextUtils.isEmpty(where) ? " AND (" + where
+                                    + ")" : ""), whereArgs);
+                }catch (Exception e){
+                    e.printStackTrace();
```
AndyScherzinger: Hi @GurpreetSK95 please use a log statement instead of print stack trace :)

src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -96,20 +96,26 @@ private int delete(SQLiteDatabase db, Uri uri, String where, String[] whereArgs)
...
-                String remoteId = "";
-                if (c != null && c.moveToFirst()) {
-                    remoteId = c.getString(c.getColumnIndex(ProviderTableMeta.FILE_REMOTE_ID));
-                    //ThumbnailsCacheManager.removeFileFromCache(remoteId);
+                try {
+                    String remoteId = "";
+                    if (c != null && c.moveToFirst()) {
+                        remoteId = c.getString(c.getColumnIndex(ProviderTableMeta.FILE_REMOTE_ID));
+                        //ThumbnailsCacheManager.removeFileFromCache(remoteId);
+                        c.close();
```
AndyScherzinger: ```c.close(); ``` can be removed here since your finally block will take care of this anyways :)

AndyScherzinger:
Hi @GurpreetSK95 I found another (slight) thing, sorry for the delay.

src/com/owncloud/android/providers/FileContentProvider.java
GurpreetSK95: Sorry for this. I should have seen this earlier

AndyScherzinger:
@davivel code looks good imho :+1:

src/com/owncloud/android/providers/FileContentProvider.java
AndyScherzinger: No worries! Happy to help with code reviews :smiley:

davivel:
Checking.

davivel:
@GurpreetSK95 , could you explain a bit more? What's the ```Exception``` that needs to be caught in that block to prevent a leak of the ```Cursor```? Did you find a concrete situation where a ```Cursor``` leak occurred?

Thanks.

AndyScherzinger:
@davivel imho the code has been missing a finally { //close cursor } in general and didn't close the cursor at all before this code change. Thus you always have a resource leak since the cursor isn't closed. :wink:

davivel:
The ```close()``` was there, though now I see one case were it could be not called even without exceptions: when ```c!=null```  and ```c.moveToFirst()``` returns ```false```.

src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -96,20 +96,25 @@ private int delete(SQLiteDatabase db, Uri uri, String where, String[] whereArgs)
         switch (mUriMatcher.match(uri)) {
             case SINGLE_FILE:
                 Cursor c = query(db, uri, null, where, whereArgs, null);
```
davivel: If we are going to add a ```catch``` block for runtime exceptions, then we should include this line in the ```try``` block, separating the declaration of ```c``` from its assignment, like:

```
Cursor c= null;
try {
   c= query(db, uri, null, where, whereArgs, null);
```

src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -96,20 +96,25 @@ private int delete(SQLiteDatabase db, Uri uri, String where, String[] whereArgs)
...
+                    count = db.delete(ProviderTableMeta.FILE_TABLE_NAME,
+                            ProviderTableMeta._ID
+                                    + "="
+                                    + uri.getPathSegments().get(1)
+                                    + (!TextUtils.isEmpty(where) ? " AND (" + where
+                                    + ")" : ""), whereArgs);
+                }catch (Exception e){
+                    Log_OC.d(TAG, "Datebase Error " + e);
+                }finally {
                     c.close();
```
davivel: And then modify the finally block to:

```
} finally {
  if (c != null) {
     c.close()
  }
}
```

davivel:
Please, @GurpreetSK95 , could you also rebase the branch on master?

davivel:
Ah, I found where did this come from. Sorry, I missed it.

jesmrec:
Already done

PR #1835, opened 2016-10-27
Login Test Suite Espresso

jesmrec:
UI testing suite on the ´AuthenticatorActivity´ activity to check the login feature.

Included:
- Positive and negative test cases. 
- Checks in both orientations 
- Accounts Manager: to add/remove ownCloud accounts from device.

To do:
- SAML Servers

### DIFF
--- a/androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
+++ b/androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
@@ -24,6 +24,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
 import android.os.Bundle;
+import android.os.Build;
 import android.os.RemoteException;
 import android.os.SystemClock;
 import android.support.test.InstrumentationRegistry;
@@ -38,7 +39,6 @@ import static android.support.test.espresso.action.ViewActions.replaceText;
 import static android.support.test.espresso.action.ViewActions.scrollTo;
 import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
 import static android.support.test.espresso.matcher.ViewMatchers.withContentDescription;
-import static android.support.test.espresso.matcher.ViewMatchers.withParent;
 import static android.support.test.espresso.matcher.ViewMatchers.withText;
 import static org.hamcrest.Matchers.allOf;
 import static org.junit.Assert.assertTrue;
@@ -52,7 +52,6 @@ import org.junit.After;
 import org.junit.FixMethodOrder;
 import org.junit.Rule;
 import org.junit.Test;
-import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.junit.runners.MethodSorters;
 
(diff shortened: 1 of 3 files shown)


androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
```diff
@@ -19,59 +20,88 @@
...
+    private Context targetContext = null;
 
+    private String testUser = null;
+    private String testUser2 = null;
+    private String testPassword = null;
+    private String testPassword2 = null;
+    private String testServerURL = null;
+    private int trusted = -1;
+
+    /*
```
davivel: This is begging for an enum ;)

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
```diff
@@ -81,59 +111,460 @@ protected Intent getActivityIntent() {
...
+
+            // Check that login button keeps on being disabled
+            onView(withId(R.id.buttonOK))
+                    .check(matches(not(isEnabled())));
+
+            // Check that SSL server is not trusted
+            onView(withId(R.id.server_status_text))
+                    .check(matches(withText(R.string.ssl_certificate_not_trusted)));
+
+            Log_OC.i(LOG_TAG, "Test not accept not secure passed");
```
davivel: No assert in this test?

jesmrec: yes, line 179 contains the `matches` assertion. This is a negative test case, so only is checked the error message.

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
jesmrec: you mean, set the `trusted` as an enum with the values in the comment.

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
davivel: :+1:

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
davivel: Yes, defining an enum type with those values an use the constants of the enum to make it easier to read all over the code.

Example:https://github.com/owncloud/android/blob/master/src/com/owncloud/android/datamodel/UploadsStorageManager.java#L49

jesmrec: @davivel done!! please review

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
```diff
@@ -543,20 +581,20 @@ private void setFields (String connectionString, String username, String passwor
      */
     private void checkStatusMessage(){
 
-        switch (trusted){
-            case (-1):
+        switch (servertype.getStatus()){
```
davivel: An enum can be used directly in a ```switch``` statement. You don't need to get its internal value, just write

```switch (servertype)```

and then the cases as

```case HTTP:```,  etc...

File: androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java [hunk 4]
```diff
@@ -81,59 +138,471 @@ protected Intent getActivityIntent() {
...
+    public void init() {
+        Bundle arguments = InstrumentationRegistry.getArguments();
+
+        testUser = arguments.getString("TEST_USER");
+        testUser2 = arguments.getString("TEST_USER2");
+        testPassword = arguments.getString("TEST_PASSWORD");
+        testPassword2 = arguments.getString("TEST_PASSWORD2");
+        testServerURL = arguments.getString("TEST_SERVER_URL");
+        trusted = Integer.parseInt(arguments.getString("TRUSTED"));
+        switch (trusted) {
```
davivel: I would move this logic to get an enum from an int to a static method in the enum, similar to the method ```fromValue``` here: https://github.com/owncloud/android/blob/master/src/com/owncloud/android/datamodel/UploadsStorageManager.java#L76

That way you could read the servertype here, or at any other place, directly as

```servertype = ServerType.fromValue(Integer.parseInt(arguments.getString("TRUSTED")));```

and the field ```trusted``` wouldn't be needed anymore.

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
jesmrec: done!

androidTest/java/com/owncloud/android/authentication/AuthenticatorActivityTest.java
jesmrec: done, thanks!

davivel:
In.!

PR #1844, opened 2016-12-04
Issue 1562 fix

LuisVilarBarbosa:
To solve the issue 1562, we analyzed, in the beginning, the situation using the user interface and, then, the code. At the interface level, we verified that the problem occurred both with files as with folders and that it also happened if, as soon as we shared it, we asked for the link twice; the first worked well but, from the second time, no longer.
At the code level, we found that, sometimes, the ArrayList returned by the “getData” method in “ShareActivity” didn't always have only one element, having more than one, i.e. it had an “User” link and a “Public” link.
Then, we analyzed the code of the ‘RemoteOperationResult’ class, which is used by diverse classes, and found that it had a method called “setData”. After this, we identified, with the “Find Usages” capability of the “Android Studio”, all the locations where “setData” was used and found that only the classes “ShareToRemoteOperationResultParser” and “ReadRemoteFolderOperation” could create an ArrayList with more than one element.
After analyzing what those two classes added to them ArrayList, before applying the method “setData”, we understood that the problem was not in “ReadRemoteFolderOperation”, because that class deals with the data from a remote folder and her child-files. So, we verified if there were different types of links and found in “ShareType” that that was true, being really dangerous to only accept “Public” links in ”ShareToRemoteOperationResultParser” (now we know that it really would give problems), so we decided to go with a solution that is looks like the presented in the issue, but that we think that is better than the workaround. We think that this is the best solution to the problem, not bringing negative effects to the application running behavior.
Later, we confirmed that, as expected, the links were added by creation order, so, if the public link was the first to be generated (or was not generated), the problem would never occur, otherwise, the problem would occur.


LuisVilarBarbosa:
Some commits from our next iteration are present here, because I didn't know that I should close the pull request. The commits that were to belong to this pull request end at 2016/12/04. The other commits are present in another pull request that I have done some minutes ago.

PR #1845, opened 2016-12-05
Handle strings.xml as unformatted, unless otherwise needed

davivel:


davivel:
cc  @owncloud/android-developers

davivel:
@jesmrec , FYI

jesmrec:
The problem here happens when two encoding characters are set in a branding option, but not consecutively

- http://own%20_%20cloud.com/mobile/help -> does not work
- http://own%20%20cloud.com/mobile/help -> work

The fix solves correctly the bug

CC @davivel

davivel:
cc @owncloud/android-developers

PR #1851, opened 2016-12-15
Keep a single session with the server if its version is >= 9.1

### LINKED ISSUE #1841: use session cookies for android app
### Actual behaviour
android app ignores session cookies.

### Expected behaviour
android should check for the session cookies so the access can be revoked fromm the server side.
 
### Steps to reproduce
1. create user test on server
2. login with user on server
3. login from android.
4. on personal page on server look for active sessions
5. android client is not shown in sessions overview.

### Environment data
Android version: android 6 

Device model: 
moto g3 
Stock or customized system:
cyanogenmod
ownCloud app version:
2.1.2
ownCloud server version:
9.1.2 expected to work from 9.1.0

davivel:
Fixes #1841 

Requires owncloud/android-library#141


davigonz:
:+1:

PR #1859, opened 2016-12-20
Update version code in JB workaround app

davivel:


davivel:
By the power of hotfix... self :+1:

PR #1864, opened 2016-12-23
Bug 1722 maintenance mode

### LINKED ISSUE #1722: Client does not recognize maintenance mode
### Steps to reproduce
1. Switch ownCloud server into maintenance mode
### Expected behaviour

The Android client should inform the user that the server currently in maintenance mode or unreachable.
### Actual behaviour

The client shows no message and displays folders as empty
### Affected software versions

All

 
### Tell us what could be improved:

Have the client display a notification when the server is unavailable (HTTP 503)

____

BUGS & IMPROVEMENTS

- [X] Error in uploads view https://github.com/owncloud/android/issues/1722#issuecomment-269292313 @davigonz [FIXED] @jesmrec
- [X] Share with users does not show error https://github.com/owncloud/android/issues/1722#issuecomment-269295600 @davigonz [FIXED] @jesmrec
- [X] Shibboleth server does not recognize maintenance mode https://github.com/owncloud/android/issues/1722#issuecomment-269300730 @davigonz [FIXED] @jesmrec
- [X] Unsharing in maintenance mode removes the file https://github.com/owncloud/android/issues/1722#issuecomment-269603786 @davigonz [FIXED] @jesmrec

davigonz:
Fix #1722 

Requires: https://github.com/owncloud/android-library/pull/144

### DIFF
--- a/res/values-b+crp/strings.xml
+++ b/res/values-b+crp/strings.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_name">ownCloud</string>
+</resources>
--- a/res/values-b+crp/strings.xml
+++ b/res/values-b+crp/strings.xml
@@ -1,4 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<resources>
-    <string name="app_name">ownCloud</string>
-</resources>
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1280,7 +1280,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mServerStatusText = R.string.auth_redirect_non_secure_connection_title;
                 break;
             case MAINTENANCE_MODE:
-                mServerStatusText = R.string.auth_maintenance_mode;
+                mServerStatusText = R.string.maintenance_mode;
                 break;
             default:
                 mServerStatusText = 0;
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1280,7 +1280,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mServerStatusText = R.string.auth_redirect_non_secure_connection_title;
                 break;
             case MAINTENANCE_MODE:
-                mServerStatusText = R.string.maintenance_mode;
+                mServerStatusText = R.string.auth_maintenance_mode;
                 break;
             default:
                 mServerStatusText = 0;
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1214,74 +1214,77 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
         mServerStatusIcon = R.drawable.common_error;    // the most common case in the switch below
 
         switch (result.getCode()) {
-        case OK_SSL:
-            mServerStatusIcon = R.drawable.ic_lock;
-            mServerStatusText = R.string.auth_secure_connection;
-            break;
+            case OK_SSL:
+                mServerStatusIcon = R.drawable.ic_lock;
+                mServerStatusText = R.string.auth_secure_connection;
+                break;
 
-        case OK_NO_SSL:
-        case OK:
-            if (mHostUrlInput.getText().toString().trim().toLowerCase().startsWith("http://") ) {
-                mServerStatusText = R.string.auth_connection_established;
-                mServerStatusIcon = R.drawable.ic_ok;
-            } else {
-                mServerStatusText = R.string.auth_nossl_plain_ok_title;
+            case OK_NO_SSL:
+            case OK:
+                if (mHostUrlInput.getText().toString().trim().toLowerCase().startsWith("http://")) {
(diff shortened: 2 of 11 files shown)


davigonz:
@davivel , @jesmrec , bug fixed, pending to review and pass QA.

res/values-b+crp/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
davivel: Where does this file come from? ```app_name``` should be only on file res/values/setup.xml

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1092,7 +1092,8 @@ private void onGetUserNameFinish(RemoteOperationResult result) {
             if (success)
                 finish();
         } else {
-            updateStatusIconFailUserName();
+            int statusText = result.getCode() == ResultCode.MAINTENANCE_MODE ? R.string.maintenance_mode : R.string.auth_fail_get_user_name;
```
davivel: Line too long, please wrap it on 120 characters.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1214,74 +1215,77 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
...
-            break;
-
-        case OK_NO_SSL:
-        case OK:
-            if (mHostUrlInput.getText().toString().trim().toLowerCase().startsWith("http://") ) {
-                mServerStatusText = R.string.auth_connection_established;
-                mServerStatusIcon = R.drawable.ic_ok;
-            } else {
-                mServerStatusText = R.string.auth_nossl_plain_ok_title;
+            case OK_SSL:
```
davivel: This update of indentation, was on purpose or triggered by automatic formatting in your AS? Please, **disable** it if that's the case.

davivel: But don't revert the change. It's just to prevent it happens in the future.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1369,9 +1373,9 @@ private void updateAuthStatusIconAndText(RemoteOperationResult result) {
     }
 
 
-    private void updateStatusIconFailUserName(){
+    private void updateStatusIconFailUserName(int statusText){
```
davivel: Please, change the name of the method or add a separate method for maintainance mode and leave this one without parameters. In this state, the name of the method doesn't match semantically what it does.

src/com/owncloud/android/providers/UsersAndGroupsSearchProvider.java
```diff
@@ -288,15 +288,19 @@ public void run() {
...
-                        getContext().getApplicationContext(),
-                        ErrorMessageAdapter.getErrorCauseMessage(
-                                result,
-                                null,
-                                getContext().getResources()
-                        ),
-                        Toast.LENGTH_SHORT
-                ).show();
+
+                // Edited: this toast message has no sense. If operation is being passed as null to getErrorCauseMessage(),
```
davivel: That's true, except when the ```result``` is an error included in ```ErrorMessageAdapter#isNetworkError(...)``` . In that case the value of ```operation``` is not considered at all in ```ErrorMessageAdapter#getErrorCauseMessage(...)``` .

src/com/owncloud/android/utils/ErrorMessageAdapter.java
```diff
@@ -92,6 +93,9 @@ public static String getErrorCauseMessage(RemoteOperationResult result,
                 } else if (result.getCode() == ResultCode.INVALID_CHARACTER_DETECT_IN_SERVER) {
                     message = res.getString(R.string.filename_forbidden_charaters_from_server);
 
+                } else if (result.getCode() == ResultCode.MAINTENANCE_MODE) {
```
davivel: Instead of adding the case of ```ResultCode.MAINTENANCE_MODE``` to every operation block, I would just add it to ```getErrorMessage()```  and ```isNetworkError()```. Maybe the methods of these two methods could be changed to ```getCommonErrorMessage()```  and ```isCommonError()``` to show better their intention.

res/values-b+crp/strings.xml
davigonz: Sure, I will delete it.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
davigonz: I've used the Reformat code option. This is how it looked before the indentation:
```javascript
switch (x) {
case 1: if (something) {
        }
case 2: if (other) {
        }
}
```
In my opinion, this indentation makes the contents of the switch stand out better. :wink:

src/com/owncloud/android/authentication/AuthenticatorActivity.java
davigonz: Sure, it's such a good refactoring approach. 👏 👏

davigonz:
I've just added all the requested changes. Please @davivel , check all the modifications and @jesmrec , pass QA again. Thanks!

jesmrec:
Right, all this changes in review stage contain significative changes that make necessary a new QA review.

jesmrec:
Approved again 👍 

CC @davivel @davigonz

src/com/owncloud/android/authentication/AuthenticatorActivity.java
davivel: :+1:

davivel:
Fino filipino :+1:

PR #1869, opened 2016-12-30
Encode properly the password

### LINKED ISSUE #1842: OCS Share API. password parameter not x-www-form-encoded
### Actual behaviour
Example
When I put _Hello%20World_ as password the client doesn't encode the value and sends it without any change to the server.
So, when I navigate to the url returned by the server and introduce _Hello%20World_ as password I get password authentication error... The password that I should introduce to get access is _Hello World_

Other example
If I choose _Hello&World_ as password -> The server takes _Hello_ as password 

### Expected behaviour
The password should be exactly what I typed
 
### Steps to reproduce

The password parameter in https://doc.owncloud.org/server/9.1/developer_manual/core/ocs-share-api.html#update-share is sent by the android client to the Server without x-www-form-encoding 
The windows desktop client is sending the password in the right way , so the server doesn't get messed up

Don't know

### Environment data
Android version:
7.1.1

Device model: 
Nexus 5x

Stock or customized system:
Stock

ownCloud app version:
2.1.2
ownCloud server version:
9.1.1

davigonz:
Fix #1842

### DIFF
--- a/src/com/owncloud/android/ui/dialog/SharePasswordDialogFragment.java
+++ b/src/com/owncloud/android/ui/dialog/SharePasswordDialogFragment.java
@@ -1,20 +1,20 @@
 /**
- *   ownCloud Android client application
- *   @author masensio
- *   Copyright (C) 2016 ownCloud GmbH.
+ * ownCloud Android client application
  *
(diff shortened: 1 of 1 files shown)


jesmrec:
Approved.

PR #1304, opened 2015-11-23
Remove >>Details<< action from file menu. #1275

DougJod:
@jancborchardt @AndyScherzinger @LukeOwncloud 
The redundant 'Details' menu item has been removed from the long press action, but the FileDetailFragment is still in use.

Issue #1275, opened 2015-11-09
Remove »Details« action from file menu


jancborchardt:
The details screen serves no real purpose anymore. It only duplicates the info which is available elsewhere:
- name
- size
- date
- file type (it even wrongly states SVG files are of type JPEG)
- ability to favorite

So I would say we should get rid of that screen and remove the action in the long-press file menu. This also serves as reducing the actions in the file menu, see https://github.com/owncloud/android/issues/1144

cc @tobiasKaminsky @AndyScherzinger @davivel @masensio @MTRichards

AndyScherzinger:
How can this info be reached? Files else open with an app that support the mime type, so if it is a pdf for example I can only see the "details" via the menu item on the long click action, right?

LukeOwncloud:
@AndyScherzinger name, size, date are shown in folder view. ability to favorite is in menu.

@jancborchardt What is not available is the exact, absolute file date (including seconds). But I think it is not very important, same as the file type. So I like the idea to remove the details view.

jancborchardt:
@AndyScherzinger yeah, that screen can only be reached by the details item in the long click action. Otherwise it’s displayed for filetypes which can’t be opened by ownCloud. Which is ok – but it shouldn’t be duplicated as action in the long press menu.

In addition, the duplicate info there about »type, size, date« can be removed, and that whole screen should be beautified a bit more. Kind of like the empty folder message in the web interface of ownCloud:
![capture du 2015-11-09 15-44-51](https://cloud.githubusercontent.com/assets/925062/11036435/d7e67d1c-86f8-11e5-9878-7e333da3070a.png)

AndyScherzinger:
That isn't happening (anymore), at least not for me on the branches. It just displays a Toast saying no app found for the file type. So at least on the branches (will have to check the master), the long click really is the only way to get to the details screen...

I get @LukeOwncloud's comment and yes, atm I would also suggest to drop the screen. So not sure if the Toas is fine or the screen makes it look a bit nicer :)

jancborchardt:
Nah, the toast is fine. Even better, because it makes no sense redirecting people to a separate screen when the file can’t be opened anyway. However I can’t see the toast. For me for example, a .gpx file brings up the details screen.

AndyScherzinger:
Than it works as designed already ... on master :) 

(probably even in the release 1.8.0; not sure though)

AndyScherzinger:
[...]

PR #1873, opened 2016-12-31
Remove >>Details<< action from file menu #1275

AkshitKrNagpal:
#1304 and #1275

### DIFF
--- a/res/menu/file_actions_menu.xml
+++ b/res/menu/file_actions_menu.xml
@@ -87,12 +87,6 @@
         app:showAsAction="never"
         android:orderInCategory="1" />
     <item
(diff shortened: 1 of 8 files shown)


jesmrec:
Removing <<Details>> from main menu is not a good idea IMHO. Taking in account that the files view has already enough information, maybe some extra info about the files could be set in Details, for example, tagging info.

davivel:
Sorry, @AkshitKrNagpal , this will not go in. I agree with @jesmrec , see my explanation in https://github.com/owncloud/android/issues/1275#issuecomment-286499303

My apologies, I should have addressed this issue long time ago.

PR #1880, opened 2017-01-12
Show message when the server is incompatible with sharing files with …

### LINKED ISSUE #1308: Share: unspecific error message
Hi,

I wanted to use "share with users", but all I got was:
Sorry, your server version does not allow share with users within clients.
Please contact your administrator

Although I am the admin I was clueless ;)
After digging in the code the problem is that my server is 8.1.x and not 8.2.x.

So I suggest to change the error message to:
Sorry, your server version is 8.1.2 and sharing with users within clients is supported by at least 8.2.0. Please upgrade the server (or similar).

davigonz:
…users

Fix #1308

### DIFF
--- a/res/layout/share_file_layout.xml
+++ b/res/layout/share_file_layout.xml
@@ -94,31 +94,7 @@
                 android:layout_gravity="start"
                 android:text="@string/share_with_user_section_title"
                 android:textColor="@color/color_accent" />
-
-            <LinearLayout
-                android:layout_width="match_parent"
-                android:layout_height="match_parent"
-                android:orientation="vertical">
-
-                <TextView
-                    android:id="@+id/shareNoUsers"
-                    android:layout_width="match_parent"
-                    android:layout_height="wrap_content"
-                    android:padding="@dimen/standard_half_padding"
-                    android:text="@string/share_no_users"
-                    android:textSize="12sp" />
-
-                <android.support.v7.widget.AppCompatButton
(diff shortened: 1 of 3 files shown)


davigonz:
@davivel, @jesmrec  , this is fixed, pending to review and pass QA.

jesmrec:
Approved

res/layout/share_file_layout.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
```
davivel: I don't like your formatter :P

res/layout/share_file_layout.xml
```diff
@@ -99,7 +93,31 @@
                 android:layout_height="wrap_content"
                 android:layout_gravity="start"
                 android:text="@string/share_with_user_section_title"
-                android:textColor="@color/color_accent"/>
+                android:textColor="@color/color_accent" />
+
+            <LinearLayout
```
davivel: @davigonz , with this change the 'add user or groups' button is shown over the list of shared users, instead of below. Was this on purpose?

res/layout/share_file_layout.xml
davigonz: Is not mine, is Android Studio code formatter 😄

PR #1885, opened 2017-01-17
Update the Android manual to reflect version 2.2.0

settermjd:
This commit is part of owncloud/documentation/issues/2811. It doesn't
necessarily change a lot. But there were notable areas which were out of
date. For what it's worth, the manual isn't conclusive. I'll endeavour
to make it so, as time allows.


user_manual/android_app.rst
```diff
@@ -2,202 +2,305 @@
...
-The ownCloud menu at the top left shows the name of the logged-in user, has a 
-shortcut to your files, the **Uploads** button, and the **Settings** button.
+Change User Password
+^^^^^^^^^^^^^^^^^^^^
+
+To change a user's password, click the key icon, next to the user's details. 
+This will display the user details page, with the ownCloud server URI and user account pre-filled.
+Enter a new password, and click "Connect", and the password will be updated.
+
+.. NOTE::
```
davivel: I'm not sure this is a good place to talk about the PIN. Right now the option is quite separated from the 'Manage accounts' view.

user_manual/android_app.rst
```diff
@@ -2,202 +2,305 @@
...
-The Uploads page displays the status of files currently uploading, a list of 
-your recently-uploaded files, and a Retry option for any failed uploads. If 
-credentials to access the file have changed, you'll see a credentials error. 
-Tap the file to retry, and you'll get a login screen to enter the new 
-credentials. 
+To remove an account, click the rubbish bin icon, next to the key icon. 
+This will display a confirmation dialog, asking if you want to remove the account.
+If you click "Yes", the account will be removed. 
+This action logs both you out of the server and removes your account settings **and**
+files from your Android device.
```
davivel: This is not true. The database with the list of files is deleted, but the files that were downloaded in the device at the moment of the removal will still be there after deleting the account, in the public partition. Deletion here is pending to do from some time now.

user_manual/android_app.rst
davivel: Probably a better place is the ** Application Settings ** section. There is where the PIN code option (now its called 'Passcode lock') is.

settermjd:
@davivel sorry for not getting back to this sooner.

user_manual/android_app.rst
settermjd: Thanks for the clarification. I'm not sure if I took this information from somewhere else, and modified it, or assumed it based on my experimentation with the Android client. Either way, I'll updated it.

user_manual/android_app.rst
settermjd: How about moving the documentation there, but still referencing it here as well?

settermjd:
@davivel do you have time for a final review?

user_manual/android_app.rst
davivel: Works for me.

user_manual/android_app.rst
davivel: Thanks a lot.

PR #1896, opened 2017-02-06
Show error when there is a external storage error

### LINKED ISSUE #1004: External Storage Errors Passed from Server to Mobile (Android)
As an ownCloud administrator, I would like a mechanism to pass external storage credential errors to the mobile clients so that users on their devices can address any problems with external storage, and continue to use ownCloud even after an error (in conjunction with: owncloud/core#16306 )

Acceptance Criteria:
- The error needs to be able to identify a problem with external storage and pass an error that does not cause the client to delete the directory, rather a temporary error that notifies the user that the directory is not syncing
- All errors are assumed to be temporary network or credential issues, rather than permanent failures
  - If external storage is permanently affected, the admin will have to remove it (or the user, if it is a personal mount) to resolve the errors
- The error needs to specifically identify the kind of problem (such as a credentials error), server not found error, etc
- The error also needs to specify a path so that the error can be addressed by the clients (such as opening a browser to the appropriate URL) and then allowing the user to resolve the issue. The error cannot be assumed to be in the root directory, as external storage can be moved around inside owncloud

iOS Version: https://github.com/owncloud/ios/issues/382

Desktop Version: https://github.com/owncloud/client/issues/3308

LONGER TERM:
The desktop clients would provide a communications path for passing messages to and from ownCloud clients (such as a credentials error AND a built in login screen) - so a general class of error that can be reused would be a good concept here. Related: owncloud/core#16772

____

BUGS & IMPROVEMENTS

- [X] Incorrect error message uploads https://github.com/owncloud/android/issues/1004#issuecomment-278648662 [FIXED]
- [X] Incorrect error message downloads  https://github.com/owncloud/android/issues/1004#issuecomment-278649569  [FIXED]

davigonz:
Requires https://github.com/owncloud/android-library/pull/153

Fix #1004


davigonz:
@jesmrec , @jabarros This bug is fixed, pending to review and pass QA

PR #1897, opened 2017-02-08
Login error messages

### LINKED ISSUE #878: If login fails, android app should show the error message of status.php
If there is a login problem, the android app should display the error message of https://example.com/status.php

I guess this would avoid a lot of confusion and false bug reports.

davigonz:
Fix #878


davigonz:
@jabarros , @jesmrec this bug is fixed, pending to review and pass QA

jesmrec:
If the login is forbidden by a firewall rule, the displayed message is `Forbidden`. The handling errors in firewall used to set a longer message @davigonz

jesmrec:
The error when the firewall does not allow the login is solved. It is important to know, that the tests with firewall were performed with a server v8.2, due to the problem in server side related woth firewall in versions 9 and later.

All the remaining errors are correctly displayed.

Approved.

jabarros:
👍

Issue #592, opened 2014-08-22
Video files to be stream instead of downloading and than playing


supportreq:
Hi,

Is it possible for the video files [mp4 h264] which are stored on owncloud, to be streamed directly, instead of them being downloaded and than played.

Using Android 4.4.4 Nexus 5 android app 1.5.8

____

PENDING TASKS:

- [X] [GIT] Create branch in Android repo
- [X] Use ExoPlayer library instead of VideoView because it was not possible to set headers for the authentication in < Lollipop devices.
- [X] Create custom http data source for setting authentication headers (ExoPlayer doesn't implement it).
- [X] Add full screen mode manually  (ExoPlayer doesn't implement it). 
* **Important**: This full screen mode is different to the one we previously had in the app. To go into full screen mode there is a new button in the playback controls and to exit the full screen mode, there is other button in the same place when full screen mode is enabled.
- [X] Handle full screen modes. Two full screen modes to implement depending on the Android versions:
* **< 4.4 KitKat:** Full screen with a limitation, the least user interaction will cause navigation controls to reappear immediately. Once the navigation controls have reappeared, they are not hidden again, even if the user clicks in the view.
* **\>= 4.4 KitKat:** App goes truly full screen using **immersive mode**, system bars are automatically hidden. User swipes to display the system bars.
- [X] Options in menu. Remove options not necessary to stream 
- [X] If the file is being downloaded you can not stream it, and if the user taps over the file a progress bar will be shown 
- [X] If new version is available it should works as always. Update the current file while play it. 
- [X] Detect SSL certificate to choose between download/stream. In this case, when the certificate is untrusted, an specific error message is shown to the user and after that, the streaming view is closed and the download starts automatically. 
- [X] Improve full screen button behaviour, avoiding show it till the player is ready to reproduce the video
- [X] Hide unnecessary options to stream in tablets [FIXED] @davigonz , @jesmrec 
- [X] Video playback state and autoplay are missed while rotating the screen in preview mode, not in full screen mode [FIXED] @davigonz @jesmrec 
- [X] Video playback state and autoplay are missed while changing from preview mode to full screen mode and vice versa [FIXED] @davigonz @jesmrec 
- [X] [QA] Design test plan @jesmrec
- [x] [QA] Validate test plan [WIP] @jesmrec
- [x] [GIT] Merge branch in Android repo

BUGS & IMPROVEMENTS

[...]

PR #1907, opened 2017-02-24
Video streaming

davigonz:
Implements the feature of the issue #592


davigonz:
@jesmrec , @jabarros This new feature is completed, pending to review and pass QA

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1305,16 +1317,36 @@ public void onServiceConnected(ComponentName component, IBinder service) {
...
+
+                    // If the file is being downloaded, assure that the fragment to show is details
+                    // fragment, not the streaming video fragment which has been previously
+                    // set in chooseInitialSecondFragment method
+
+                    FileFragment secondFragment = getSecondFragment();
+                    if (secondFragment != null && secondFragment instanceof PreviewVideoFragment) {
+                        cleanSecondFragment();
+
+                        Fragment detailFragment = FileDetailFragment.newInstance(getFile(), getAccount());
```
davivel: Maybe we could call ```showDetails(getFile())``` here instead of these 4 duplicated lines.

src/com/owncloud/android/ui/preview/CustomHttpDataSourceFactory.java
```diff
@@ -0,0 +1,94 @@
...
+
+import com.google.android.exoplayer2.upstream.DataSource;
+import com.google.android.exoplayer2.upstream.DefaultHttpDataSource;
+import com.google.android.exoplayer2.upstream.HttpDataSource.BaseFactory;
+import com.google.android.exoplayer2.upstream.HttpDataSource.Factory;
+import com.google.android.exoplayer2.upstream.TransferListener;
+
+import java.util.Map;
+
+/** A {@link Factory} that produces {@link DefaultHttpDataSource} instances. */
```
davivel: Comment to update.

src/com/owncloud/android/ui/preview/CustomHttpDataSourceFactory.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private final String userAgent;
+    private final TransferListener<? super DataSource> listener;
+    private final int connectTimeoutMillis;
+    private final int readTimeoutMillis;
+    private final boolean allowCrossProtocolRedirects;
+    private final Map<String, String> headers;
+
+    /**
+     * Constructs a DefaultHttpDataSourceFactory. Sets {@link
```
davivel: Comment to update

src/com/owncloud/android/ui/preview/CustomHttpDataSourceFactory.java
```diff
@@ -0,0 +1,94 @@
...
+        this.listener = listener;
+        this.connectTimeoutMillis = connectTimeoutMillis;
+        this.readTimeoutMillis = readTimeoutMillis;
+        this.allowCrossProtocolRedirects = allowCrossProtocolRedirects;
+        this.headers = params;
+    }
+
+    @Override
+    protected DefaultHttpDataSource createDataSourceInternal() {
+        DefaultHttpDataSource defaultHttpDataSource = new DefaultHttpDataSource(userAgent, null, listener, connectTimeoutMillis,
```
davivel: Please, keep lines under 120 characters.

src/com/owncloud/android/ui/preview/CustomHttpDataSourceFactory.java
davivel: Or maybe not?

src/com/owncloud/android/ui/preview/PreviewUtils.java
```diff
@@ -0,0 +1,149 @@
...
+import com.owncloud.android.lib.common.accounts.AccountUtils;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+
+/**
+ * Created by davidgonzalez on 22/2/17.
```
davivel: Automated comment?

src/com/owncloud/android/ui/preview/PreviewUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+        } else {
+
+            try {
+
+                //Get account credentials asynchronously
+                final GetCredentialsTask task = new GetCredentialsTask();
+                task.execute(account);
+
+                OwnCloudCredentials credentials = task.get();
```
davivel: This will not work fine.

You run a task as an ```AsyncTask``` because you want it occurs in another thread, because the task is too long and don't want the current thread (probably the main thread) uses all that time in this.

But then you call ```task.get()```, and that **waits** in the current thread until the task is finished.

So, in the end, this works just exactly as if you do not use an ```AsyncTask```, but with more code. Not a good deal.

src/com/owncloud/android/ui/preview/PreviewUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Task for getting account credentials asynchronously
+     */
+    private static class GetCredentialsTask extends AsyncTask<Object, Void, OwnCloudCredentials> {
+        @Override
+        protected OwnCloudCredentials doInBackground(Object... params) {
+            Object account = params[0];
+            try {
+                OwnCloudCredentials ocCredentials = AccountUtils.getCredentialsForAccount(MainApp.getAppContext(), (Account) account);
```
davivel: Line too long.

src/com/owncloud/android/ui/preview/PreviewVideoActivity.java
```diff
@@ -1,235 +1,356 @@
...
-     *  Called when the activity is first created.
-     *  
-     *  Searches for an {@link OCFile} and ownCloud {@link Account} holding it in the starting {@link Intent}.
-     *  
-     *  The {@link Account} is unnecessary if the file is downloaded; else, the {@link Account} is used to 
-     *  try to stream the remote file - TODO get the streaming works
-     * 
-     *  {@inheritDoc}
-     */
+    private static final DefaultBandwidthMeter BANDWIDTH_METER = new DefaultBandwidthMeter();
```
davivel: Why static?

src/com/owncloud/android/ui/preview/PreviewVideoActivity.java
```diff
@@ -1,235 +1,356 @@
...
-        Intent i = new Intent();
-        i.putExtra(EXTRA_AUTOPLAY, mVideoPlayer.isPlaying());
-        i.putExtra(EXTRA_START_POSITION, mVideoPlayer.getCurrentPosition());
-        setResult(RESULT_OK, i);
-        mVideoPlayer.stopPlayback();
-        super.onBackPressed();
+    public void onResume() {
+        super.onResume();
+        Log_OC.v(TAG, "onResume");
+        if ((Util.SDK_INT <= 23 || player == null)) {
```
davivel: Why the difference depending on API level? Just curious.

res/layout/preview_video_fragment.xml
```diff
@@ -33,27 +33,15 @@
         android:id="@+id/visual_area"
         android:layout_width="match_parent"
         android:layout_height="0dp"
-        android:layout_above="@+id/media_controller"
         android:layout_alignParentTop="true"
         >
 
-        <VideoView
-            android:id="@+id/video_preview"
+        <com.google.android.exoplayer2.ui.SimpleExoPlayerView android:id="@+id/video_preview_fragment"
```
davivel: Mmmm, this file is *not* being used anymore. Both the ```PreviewVideoFragment``` and ```PreviewActivity``` use ```res/layout/video_player.xml```. Please, remove one of them

build.gradle
```diff
@@ -28,13 +28,20 @@ repositories {
     }
 }
 
+configurations.all {
+    resolutionStrategy {
+        force "com.android.support:support-annotations:${supportLibraryVersion}"
```
davigonz: To fix this conflict which has been triggered while adding the Exoplayer library dependency:

Conflict with dependency 'com.android.support:support-annotations' in project 'android'. Resolved versions for app (25.0.1) and test app (24.2.1) differ.

res/layout/video_preview.xml
```diff
@@ -0,0 +1,33 @@
...
+    android:keepScreenOn="true"
+    android:background="@color/black">
+
+    <com.google.android.exoplayer2.ui.SimpleExoPlayerView android:id="@+id/video_player"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        app:controller_layout_id="@layout/custom_playback_control"/>
+
+    <!-- last to draw, will be top in Z axis -->
+    <include layout="@layout/top_progress_bar"/>
```
davigonz: Yes, it is used when the certificate is untrusted and the video file starts to download automatically

res/values/strings.xml
```diff
@@ -469,12 +469,18 @@
...
     <string name="failed_upload_quota_exceeded_text">Not enough space in the account</string>
     <string name="auth_forbidden">Forbidden authentication</string>
-
+    
     <string name="actionbar_privacy_policy">Privacy policy</string>
     <string name="prefs_privacy_policy">Privacy policy</string>
     <string name="privacy_policy_error">"An error has occurred: "</string>
     <string name="service_unavailable">Server is unavailable</string>
 
+    <string name="uploads_view_upload_status_failed_storage_not_available_error">Storage not available</string>
```
davigonz: Done

res/values/strings.xml
```diff
@@ -469,12 +469,18 @@
...
+    <string name="streaming_certificate_error">Streaming cannot be initialized because your server certificate is untrusted. File download will start automatically</string>
+    <string name="streaming_unrecognized_input">This video format is not supported</string>
+    <string name="streaming_file_not_found_error">This video file is no longer available in the server</string>
+    <string name="streaming_common_error">This video cannot be reproduced</string>
```
davigonz: After trying to identify the specific error, this is the message we show because we are not sure if the error is due to the streaming or the local playing itself.

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -134,30 +137,34 @@ private void hideAll(Menu menu) {
      * @param toHide            List to save the options that must be shown in the menu.
      */
     private void filter(List<Integer> toShow, List <Integer> toHide) {
+
         boolean synchronizing = anyFileSynchronizing();
 
+        boolean videoPreviewing = anyFileVideoPreviewing();
```
davigonz: Not really, this variable is used to check if the file is being previewed, with streaming or not. In other part of this file I use the code below to check if the file is being streamed:

`!anyFileDown() && anyFileVideoPreviewing()`

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -289,6 +297,20 @@ private boolean anyFileUploading(FileUploaderBinder uploaderBinder) {
         return uploading;
     }
 
+    private boolean anyFileVideoPreviewing() {
+        final HookActivity activity = (HookActivity) mContext;
```
davigonz: Updated

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -289,6 +297,20 @@ private boolean anyFileUploading(FileUploaderBinder uploaderBinder) {
...
+        Fragment secondFragment = activity.getSupportFragmentManager().findFragmentByTag(
+                TAG_SECOND_FRAGMENT);
+        boolean videoPreviewing = false;
+        if (secondFragment instanceof PreviewVideoFragment) {
```
davigonz: Updated

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -29,6 +29,7 @@
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.design.widget.Snackbar;
+import android.support.v4.app.Fragment;
```
davigonz: Removed

src/com/owncloud/android/ui/preview/PreviewAudioFragment.java
```diff
@@ -508,7 +508,7 @@ private void openFile() {
      * @return 'True' if the file can be handled by the fragment.
      */
     public static boolean canBePreviewed(OCFile file) {
-        return (file != null && file.isDown() && file.isAudio());
+        return (file != null && file.isAudio());
```
davigonz: Ups, added again

src/com/owncloud/android/ui/preview/CustomHttpDataSourceFactory.java
davigonz: Yes, updated

src/com/owncloud/android/ui/preview/PreviewVideoActivity.java
davigonz: Not needed, changed

src/com/owncloud/android/ui/preview/PreviewUtils.java
davigonz: Updated, adding the preparing player code to an AsyncTask.

src/com/owncloud/android/ui/preview/PreviewVideoActivity.java
davigonz: I'm not sure about this, I've gotten this from the Exoplayer demo example. At first glance , it has no sense but I would have to check it.

davigonz: I've definitely deleted it, not needed.

res/values/strings.xml
davivel: Well, should be possible to know... checking if the file is downloaded ;)

src/com/owncloud/android/files/FileMenuFilter.java
davivel: Then we do not allow anymore to rename or copy the file in the preview mode. But we allowed it before. Why not now?

res/layout/video_preview.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
davivel: Maybe we should rename this as ```preview_video_fragment```, for consistency with the other layouts.

res/layout/video_preview.xml
davivel: :+1:

res/values/strings.xml
```diff
@@ -469,12 +469,19 @@
...
     <string name="prefs_privacy_policy">Privacy policy</string>
     <string name="privacy_policy_error">"An error has occurred: "</string>
     <string name="service_unavailable">Server is unavailable</string>
 
+    <string name="streaming_certificate_error">Streaming cannot be initialized because your server certificate is untrusted. File download will start automatically</string>
+    <string name="streaming_unrecognized_input">This video format is not supported</string>
+    <string name="streaming_file_not_found_error">This video file is no longer available in the server</string>
+    <string name="previewing_video_common_error">This video cannot be reproduced</string>
+
+    <string name="previewing_video_not_supported">Video previewing is not supported in your Android version. File download will start automatically and you will be able to open it with an external player</string>
```
davivel: This error message was the same in the previous review? Seems too long. Do you think that explaining that file download will be done is needed? Should be quite obvious.

src/com/owncloud/android/files/FileMenuFilter.java
davivel: OK, they were not available either before, this was an inconsistency of multiple-selection mode. Works for me.

src/com/owncloud/android/ui/preview/PrepareVideoPlayerAsyncTask.java
```diff
@@ -0,0 +1,170 @@
...
+    private Account mAccount;
+    private Handler mHandler;
+
+    private static final DefaultBandwidthMeter BANDWIDTH_METER = new DefaultBandwidthMeter();
+
+
+    public PrepareVideoPlayerAsyncTask(Activity activity, OCFile file, Account account,
+                                       Handler mainHandler) {
+        mContext = activity.getApplicationContext();
+        mListener = new WeakReference<>((OnPrepareVideoPlayerTaskListener)activity);
```
davivel: I'd rather use ```OnPrepareVideoPlayerTaskListener``` as the parameter type, that would prevent the need of a casting.

src/com/owncloud/android/ui/preview/PrepareVideoPlayerAsyncTask.java
```diff
@@ -0,0 +1,170 @@
...
+
+        if (file.isDown()) {
+
+            return new DefaultHttpDataSourceFactory(MainApp.getUserAgent(), bandwidthMeter);
+
+        } else {
+
+            try {
+
+                OwnCloudCredentials credentials = AccountUtils.getCredentialsForAccount(MainApp.getAppContext(), account);
```
davivel: Line too long.

res/values/strings.xml
davigonz: This error message is not the same as the previous review, the one you mean is just above it (previewing_video_common_error). This (previewing_video_not_supported) is particularly shown while using ICE CREAM SANDWICH devices.

davivel:
@davigonz , just a couple of minors, overall is OK.

src/com/owncloud/android/ui/preview/PrepareVideoPlayerAsyncTask.java
davigonz: Casting from activity to OnPrepareVideoPlayerTaskListener would be needed in this case even though the mListener be OnPrepareVideoPlayerTaskListener.

res/values/strings.xml
davivel: OK, thought I would use "Video streaming is not supported in your Android version" 

Moreover, I would not even show an error. Just would download the file. That's what already happens in the released app, the expectation does not have to change.

res/layout/video_preview.xml
davigonz: Renaming it as preview_video_fragment could be confusing because this layout is currently being used in the preview video fragment and also in the full screen mode, when any fragment is shown.

res/values/strings.xml
davivel: Also, you considered ICE CREAM SANDWICH edge case in ```PreviewMediaFragment```, but not in ```PreviewVideoActivity```.

src/com/owncloud/android/ui/preview/PrepareVideoPlayerAsyncTask.java
davivel: Better in the client of the call than in the implementation. In this method, you are requesting your future callers for a parameter with a concrete interface, but you never use it according to that interface - you use the object with another interface totally unrelated. A client not knowing the method implementation could pass just any Activity, not implementing ```OnPrepareVideoPlayerTaskListener``` , and get the task fail. What the method really needs is an ```OnPrepareVideoPlayerTaskListener``` that is what should request.

res/layout/video_preview.xml
davivel: Yes, I know. OK, not really important.

res/values/strings.xml
davigonz: Yes, I prefer your error message by I don't agree with you about not showing the error message. It's the current behaviour in the released app but we should give feedback to the user as much as possible. If the user see that the download starts automatically, he will wonder why. 

ICE CREAM SANDWICH is only considered in PreviewVideoFragment because there is no way to go into full screen mode if the video player is not still prepared (full screen button is hidden till the video player is prepared)

src/com/owncloud/android/ui/preview/PrepareVideoPlayerAsyncTask.java
davigonz: Yes, you're right, but I've followed the same implementation as AuthenticatorAsyncTask, which constructor is doing the casting in the same way. 

Here it is: https://github.com/owncloud/android/blob/master/src/com/owncloud/android/authentication/AuthenticatorAsyncTask.java

davivel: Yes, but ```AuthenticatorAsyncTask``` is using the ```Activity``` as an ```Activity```: 

```mContext = activity.getApplicationContext()```

This other ```AsyncTask``` is not doing so.

davivel: WT... this ```AsyncTask``` **is** doing so... Sorry, I don't know how I missed that.

In any case, would be wiser to include two different parameters in the interface, one for the ```Activity``` and one for the listener. If the client passes a single object for them is up to it.

davigonz: This AsyncTask is doing so actually, I've changed the casting to the callers anyway

davivel:
Ready to go; @jesmrec , your call.

jesmrec:
Approved

CC @davigonz @davivel

PR #1914, opened 2017-03-08
Delete unneeded permissions added by Activity template

davivel:


davivel:
Mandatory for release in progress. cc @davigonz 

Not QA needed.

PR #1927, opened 2017-03-27
Retry uploads after lost network

### LINKED ISSUE #1684: Adapt to Doze
Doze (1) can disable network access at any moment, once the screen is off. This could result in pending uploads or downloads being abruptly interrupted (and not automatically retried, at this moment).

We need to study (experiment) the real impact on our app and adapt to it.

(1) https://developer.android.com/training/monitoring-device-state/doze-standby.html

UPDATE:

Doze (from Android 6) and App Standby (from Android 7) disable access to network when some conditions are met. These conditions never apply when apps are open, but the ownCloud Android app does a lot of work in background. 

Main problems we detected are about pending uploads interrupted, and never retried again - simply left as failed in the uploads view, and leaving the task of retrying them for the user. Downloads can be easily interrupted too, but the feedback we have from users is focused in uploads, probably because uploading multiple files at once is a use case more common than downloading multiple files at once.

Another operation that might be interrupted is automated sync of full account, but this is retried by the system automatically, shouldn't be a problem. Besides, we have no feedback saying otherwise.

Solution applied is scheduling retries of uploads and downloads failed due to lack of network, using the JobScheduler to do it whenever Wifi network is available. We "hardcoded" the requirement of Wifi to make it available sooner, but we expect to make this detail configurable by the user in the scope of #41. 

It is worth to say that retries will not be done as soon as the Wifi is back. Android system will decide when exactly the retry is done, granting that Wifi network is available. This is a constrain imposed by JobScheduler, and a trend that Google is introducing step by step. More constrains in this sense are coming with Android O, and we'll need to assume that we'll have less and less control over background operation.

____

BUGS & IMPROVEMENTS

 Downloads

- [X]  Crash when tapping on notification https://github.com/owncloud/android/issues/1684#issuecomment-291542941 [FIXED] @davigonz , @jesmrec 
- [X] Switching the device off does not resume the downloads https://github.com/owncloud/android/issues/1684#issuecomment-292171604 [FIXED] @davigonz @jesmrec 
- [X] With no connection loss, restarting the device does not resume https://github.com/owncloud/android/issues/1684#issuecomment-294431219
- [X] SSL initialization failed https://github.com/owncloud/android/issues/1684#issuecomment-292456651 [FIXED]
- [X] Not all downloads resumed https://github.com/owncloud/android/issues/1684#issuecomment-292469463  [FIXED]
- [X] Android 6: downloads stucked https://github.com/owncloud/android/issues/1684#issuecomment-294718015 [FIXED]
- [X] Android 5: Download interrupted is never resumed https://github.com/owncloud/android/issues/1684#issuecomment-295275970  [FIXED]
- [X] Android 5: Instant uploads get freezed after a connection loss https://github.com/owncloud/android/issues/1684#issuecomment-295306015 [FIXED] @davivel

 Uploads

- [X] Several uploads pending, not all are resumed https://github.com/owncloud/android/issues/1684#issuecomment-291869388
https://github.com/owncloud/android/issues/1684#issuecomment-294438205
- [X] Instant uploads https://github.com/owncloud/android/issues/1684#issuecomment-291920392  
- [X] Crashes after some uploads https://github.com/owncloud/android/issues/1684#issuecomment-291926676 https://github.com/owncloud/android/issues/1684#issuecomment-293214748 [FIXED]

davigonz:
Fix #1684 

Requires https://github.com/owncloud/android-library/pull/159

### DIFF
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -87,6 +87,7 @@ public class FileDownloader extends Service
     public static final String ACCOUNT_NAME = "ACCOUNT_NAME";
 
     public static final String KEY_FILE_REMOTE_PATH = "FILE_REMOTE_PATH";
+    public static final String KEY_ACCOUNT_NAME = "ACCOUNT_NAME";
 
 
     private static final String TAG = FileDownloader.class.getSimpleName();
@@ -472,6 +473,8 @@ public class FileDownloader extends Service
                                 PersistableBundle extras = new PersistableBundle();
                                 extras.putString(KEY_FILE_REMOTE_PATH, mCurrentDownload.getFile()
                                         .getRemotePath());
+
+                                extras.putString(KEY_ACCOUNT_NAME, mCurrentAccount.name);
 
                                 builder.setExtras(extras);
                                 JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -482,7 +482,7 @@ public class FileDownloader extends Service
                                 extras.putString(EXTRA_ACCOUNT_NAME, mCurrentAccount.name);
                                 builder.setExtras(extras);
 
-                                JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
+                                JobScheduler jobScheduler = (JobScheduler) this.
                                         getSystemService(Context.JOB_SCHEDULER_SERVICE);
                                 jobScheduler.schedule(builder.build());
                             }
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -473,16 +473,15 @@ public class FileDownloader extends Service
                                 builder = new JobInfo.Builder(jobId, mServiceComponent);
 
                                 // require unmetered network
-                                builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
+                                builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_UNMETERED);
 
                                 // Extra data
                                 PersistableBundle extras = new PersistableBundle();
                                 extras.putString(EXTRA_REMOTE_PATH, mCurrentDownload.getFile()
                                         .getRemotePath());
-
                                 extras.putString(EXTRA_ACCOUNT_NAME, mCurrentAccount.name);
-
                                 builder.setExtras(extras);
+
                                 JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
                                         getSystemService(Context.JOB_SCHEDULER_SERVICE);
                                 jobScheduler.schedule(builder.build());
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -470,6 +470,9 @@ public class FileDownloader extends Service
                                 // require unmetered network
                                 builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_UNMETERED);
 
+                                // Persist job and prevent it from being deleted after a device restart
+                                builder.setPersisted(true);
+
                                 // Extra data
                                 PersistableBundle extras = new PersistableBundle();
                                 extras.putString(EXTRA_REMOTE_PATH, mCurrentDownload.getRemotePath());
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -26,6 +26,10 @@ import android.accounts.OnAccountsUpdateListener;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.Service;
+import android.app.job.JobInfo;
+import android.app.job.JobScheduler;
+import android.content.ComponentName;
+import android.content.Context;
 import android.content.Intent;
 import android.os.Binder;
 import android.os.Handler;
@@ -33,6 +37,7 @@ import android.os.HandlerThread;
 import android.os.IBinder;
 import android.os.Looper;
 import android.os.Message;
+import android.os.PersistableBundle;
 import android.os.Process;
 import android.support.v4.app.NotificationCompat;
 import android.util.Pair;
@@ -50,13 +55,14 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.files.FileUtils;
-import com.owncloud.android.ui.notifications.NotificationUtils;
 import com.owncloud.android.operations.DownloadFileOperation;
 import com.owncloud.android.ui.activity.FileActivity;
 import com.owncloud.android.ui.activity.FileDisplayActivity;
+import com.owncloud.android.ui.errorhandling.ErrorMessageAdapter;
+import com.owncloud.android.ui.notifications.NotificationUtils;
 import com.owncloud.android.ui.preview.PreviewImageActivity;
 import com.owncloud.android.ui.preview.PreviewImageFragment;
-import com.owncloud.android.ui.errorhandling.ErrorMessageAdapter;
+import com.owncloud.android.utils.ConnectivityUtils;
 
 import java.io.File;
 import java.lang.ref.WeakReference;
@@ -80,6 +86,9 @@ public class FileDownloader extends Service
     public static final String EXTRA_LINKED_TO_PATH = "LINKED_TO";
     public static final String ACCOUNT_NAME = "ACCOUNT_NAME";
 
+    public static final String KEY_FILE_REMOTE_PATH = "FILE_REMOTE_PATH";
+
+
     private static final String TAG = FileDownloader.class.getSimpleName();
 
     private Looper mServiceLooper;
@@ -97,6 +106,7 @@ public class FileDownloader extends Service
     private NotificationCompat.Builder mNotificationBuilder;
     private int mLastPercent;
 
+    private static int jobId = 0;
 
     public static String getDownloadAddedMessage() {
         return FileDownloader.class.getName() + DOWNLOAD_ADDED_MESSAGE;
@@ -443,7 +453,31 @@ public class FileDownloader extends Service
 
                     if (!downloadResult.isSuccess() && downloadResult.getException() != null) {
 
+                        if (!ConnectivityUtils.isNetworkActive(getApplicationContext())) {
 
+                            ComponentName mServiceComponent = new ComponentName(this,
+
+                                    RetryDownloadJobService.class);
+
+                            JobInfo.Builder builder;
+
+                            // Schedule future retry of failed download due to network error from Android 5
+                            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+                                builder = new JobInfo.Builder(jobId, mServiceComponent);
+                                // require unmetered network
+                                builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
+
+                                // Extra data
+                                PersistableBundle extras = new PersistableBundle();
+                                extras.putString(KEY_FILE_REMOTE_PATH, mCurrentDownload.getFile()
+                                        .getRemotePath());
+
+                                builder.setExtras(extras);
+                                JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
+                                        getSystemService(Context.JOB_SCHEDULER_SERVICE);
+                                jobScheduler.schedule(builder.build());
+                            }
+                        }
                     }
 
                     /// notify result
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -477,8 +477,7 @@ public class FileDownloader extends Service
 
                                 // Extra data
                                 PersistableBundle extras = new PersistableBundle();
-                                extras.putString(EXTRA_REMOTE_PATH, mCurrentDownload.getFile()
-                                        .getRemotePath());
+                                extras.putString(EXTRA_REMOTE_PATH, mCurrentDownload.getRemotePath());
                                 extras.putString(EXTRA_ACCOUNT_NAME, mCurrentAccount.name);
                                 builder.setExtras(extras);
 
--- a/src/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/com/owncloud/android/files/services/FileDownloader.java
@@ -84,10 +84,7 @@ public class FileDownloader extends Service
     public static final String EXTRA_FILE_PATH = "FILE_PATH";
     public static final String EXTRA_REMOTE_PATH = "REMOTE_PATH";
     public static final String EXTRA_LINKED_TO_PATH = "LINKED_TO";
-    public static final String ACCOUNT_NAME = "ACCOUNT_NAME";
-
-    public static final String KEY_FILE_REMOTE_PATH = "FILE_REMOTE_PATH";
-    public static final String KEY_ACCOUNT_NAME = "ACCOUNT_NAME";
+    public static final String EXTRA_ACCOUNT_NAME = "EXTRA_ACCOUNT_NAME";
 
 
     private static final String TAG = FileDownloader.class.getSimpleName();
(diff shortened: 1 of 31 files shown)


davigonz:
@davivel , @jesmrec this issue is fixed, pending to review and pass QA

src/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -80,6 +86,10 @@
     public static final String EXTRA_LINKED_TO_PATH = "LINKED_TO";
     public static final String ACCOUNT_NAME = "ACCOUNT_NAME";
 
+    public static final String KEY_FILE_REMOTE_PATH = "FILE_REMOTE_PATH";
```
davivel: I would use ```EXTRA_REMOTE_PATH```  instead. It's already there, and means the same.

src/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -80,6 +86,10 @@
...
+    public static final String KEY_ACCOUNT_NAME = "ACCOUNT_NAME";
```
davivel: I'd rather call it ```EXTRA_ACCOUNT_NAME``` , to follow the same convention as other keys used for ```extras```  in ```Bundle``` .

Besides, since ```FileDownloader```  does not really need this string for itself, I would declare it in ```RetryDownloadJobService``` instead of here.

src/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -97,6 +107,7 @@
     private NotificationCompat.Builder mNotificationBuilder;
     private int mLastPercent;
 
+    private static int jobId = 0;
```
davivel: What's this for?

src/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -441,6 +452,39 @@ private void downloadFile(String downloadKey) {
...
+
+                                ComponentName mServiceComponent = new ComponentName(this,
+
+                                        RetryDownloadJobService.class);
+
+                                JobInfo.Builder builder;
+                                
+                                builder = new JobInfo.Builder(jobId, mServiceComponent);
+                                // require unmetered network
+                                builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
```
davivel: Comment says ```require unmetered network```, and so it should be; but parameter does not fit the comment; please, use ```JobInfo.NETWORK_TYPE_UNMETERED```.

src/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -441,6 +452,39 @@ private void downloadFile(String downloadKey) {
...
+
+                                // Extra data
+                                PersistableBundle extras = new PersistableBundle();
+                                extras.putString(KEY_FILE_REMOTE_PATH, mCurrentDownload.getFile()
+                                        .getRemotePath());
+
+                                extras.putString(KEY_ACCOUNT_NAME, mCurrentAccount.name);
+
+                                builder.setExtras(extras);
+                                JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
```
davivel: Use ```this```  instead of ```getApplicationContext()```. ```FileDownloader``` is a ```Context``` , and we only need a context here to access a system service, not to bind anything to it, so the current context is good enough.

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -941,6 +949,34 @@ public void uploadFile(String uploadKey) {
...
+
+                            ComponentName mServiceComponent = new ComponentName(this,
+
+                                    RetryUploadJobService.class);
+
+                            JobInfo.Builder builder;
+
+                            builder = new JobInfo.Builder(jobId, mServiceComponent);
+                            // require unmetered network
+                            builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
```
davivel: Again, type should be ```JobInfo.NETWORK_TYPE_UNMETERED```

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -941,6 +949,34 @@ public void uploadFile(String uploadKey) {
...
+                            PersistableBundle extras = new PersistableBundle();
+                            extras.putString(KEY_FILE, mCurrentUpload.getRemotePath());
```
davivel: The keys for the extras should be the same for downloads and uploads. They mean the same -  account **name** and file **remote path** .

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -941,6 +949,34 @@ public void uploadFile(String uploadKey) {
...
+                            extras.putString(KEY_ACCOUNT, mCurrentAccount.name);
+                            builder.setExtras(extras);
+                            JobScheduler jobScheduler = (JobScheduler) getApplicationContext().
```
davivel: Same about the context.

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -169,6 +175,8 @@
     private NotificationCompat.Builder mNotificationBuilder;
     private int mLastPercent;
 
+    private static int jobId = 0;
```
davivel: Same here, what's this for.

src/com/owncloud/android/files/services/RetryUploadJobService.java
```diff
@@ -0,0 +1,40 @@
...
+import com.owncloud.android.datamodel.UploadsStorageManager;
+import com.owncloud.android.db.OCUpload;
+
+@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+public class RetryUploadJobService extends JobService {
+
+    @Override
+    public boolean onStartJob(JobParameters jobParameters) {
+
+        String file = jobParameters.getExtras().getString(FileUploader.KEY_FILE);
```
davivel: Please, rename variable ```file```  to something more accurate, such as ```fileRemotePath```  or ```remotePath```.

src/com/owncloud/android/utils/ConnectivityUtils.java
```diff
@@ -44,4 +44,22 @@ public static boolean isAppConnected(Context context) {
...
+
+        boolean isNetworkActive = true;
+
+        ConnectivityManager cm = (ConnectivityManager)context.
+                getSystemService(Context.CONNECTIVITY_SERVICE);
+
+        NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
+
+        // Check network availabality (no info object implies no connectivity)
+        if (activeNetwork == null || (activeNetwork!= null &&
```
davivel: I would simplify to

```return (activeNetwork != null && activeNetwork.isConnectedOrConnecting())``` .

No ```isActiveNetwork``` variable declared nor assigned, easier condition.

File: src/com/owncloud/android/files/services/FileDownloader.java (hunk 2 above)
Comment (davigonz) [inline src/com/owncloud/android/files/services/FileDownloader.java]: Sure, I will change the name to EXTRA_ACCOUNT_NAME but about declaring it in RetryDownloadJobService, there are other final variables such as EXTRA_REMOTE_PATH which is declared in FileDownloader but it's used where download broadcast is received.

Comment (davivel) [inline src/com/owncloud/android/files/services/FileDownloader.java]: The constants are public, so they might be used in many places.

But, who should be the responsible of its declaration? Who should be the owner? These constants are, in practice, part of the public interface of the services, they are needed to any client that want to send and Intent to the service, they are part of the contract. And the contract should be defined by the server part, so that different clients can access to the server.

File: src/com/owncloud/android/files/services/FileDownloader.java (hunk 3 above)
Comment (davigonz) [inline src/com/owncloud/android/files/services/FileDownloader.java]: Here is the official info about it:

captura de pantalla 2017-03-30 a las 9 53 39

With this in mind, the approach I've used is not the proper one, because using the same id for different jobs implies overwriting other jobs, and therefore, not taking into account some of the failed uploads or downloads.

Other alternative may be using the download or upload file id, but this file id (if I'm not wrong) can be the same for different OC accounts so we have the same problem, we would overwrite jobs of other accounts.

Comment (davivel) [inline src/com/owncloud/android/files/services/FileDownloader.java]: fileId is unique; the problem is that uploads of new files have no fileId, it's set when the upload is succeeded . A big legacy mistake.

Comment (davigonz) [inline src/com/owncloud/android/files/services/FileDownloader.java]: Fixed

File: src/com/owncloud/android/files/services/FileUploader.java (hunk 9 above)
Comment (davigonz) [inline src/com/owncloud/android/files/services/FileUploader.java]: As I commented above, it's already fixed.

File: src/com/owncloud/android/utils/ConnectivityUtils.java (hunk 11 above)
Comment (davigonz) [inline src/com/owncloud/android/utils/ConnectivityUtils.java]: I can't delete the activeNetwork == null because as the comment above the if says, no info object implies no connectivity.

Comment (davivel) [inline src/com/owncloud/android/utils/ConnectivityUtils.java]: When there is no info object (```activeNetwork == null```), the condition I wrote is evaluated as

```(null != null && activeNetwork.isConnectedOrConnecting())```

```(false && activeNetwork.isConnectedOrConnecting())```

```false````

Comment (davigonz) [inline src/com/owncloud/android/utils/ConnectivityUtils.java]: OMG, you're right, silly mistake 🤓

File: src/com/owncloud/android/files/services/FileDownloader.java [hunk 12]
```diff
@@ -441,6 +447,47 @@ private void downloadFile(String downloadKey) {
...
+                            if (android.os.Build.VERSION.SDK_INT >=
+                                    android.os.Build.VERSION_CODES.LOLLIPOP) {
+
+                                ComponentName mServiceComponent = new ComponentName(this,
+
+                                        RetryDownloadJobService.class);
+
+                                JobInfo.Builder builder;
+
+                                IndexedForest<FileDownloader> fileDownloaderIndexedForest =
```
davivel: We don't need to create a new ```IndexedForest``` object. ```mPendingUploads``` is an ```IndexedForest```, it's perfect to call ```buildKey```.

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -941,6 +947,46 @@ public void uploadFile(String uploadKey) {
...
+                        // Schedule future retry of failed upload due to network error from Android 5
+                        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+
+                            ComponentName mServiceComponent = new ComponentName(this,
+
+                                    RetryUploadJobService.class);
+
+                            JobInfo.Builder builder;
+
+                            IndexedForest<FileDownloader> fileDownloaderIndexedForest =
```
davivel: Same here.

src/com/owncloud/android/files/services/RetryUploadJobService.java
```diff
@@ -0,0 +1,40 @@
...
+public class RetryUploadJobService extends JobService {
+
+    @Override
+    public boolean onStartJob(JobParameters jobParameters) {
+
+        UploadsStorageManager uploadsStorageManager = new UploadsStorageManager(getContentResolver());
+
+        String fileRemotePath = jobParameters.getExtras().getString(FileUploader.EXTRA_REMOTE_PATH);
+
+        String account = jobParameters.getExtras().getString(FileUploader.EXTRA_ACCOUNT_NAME);
```
davivel: Please, rename variable to ```acountName```

src/com/owncloud/android/files/services/FileDownloader.java
davivel: :+1:

davivel:
Works for me; @jesmrec , in your hands now.

jesmrec:
Approved

src/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -939,6 +798,36 @@ public void uploadFile(String uploadKey) {
...
+                            TAG,
+                            String.format(
+                                "Exception in upload, network is OK, no retry scheduled for %1s in %2s",
+                                mCurrentUpload.getRemotePath(),
+                                mCurrentAccount.name
+                            )
+                        );
+                    }
+                }
+
```
davigonz: You could add here an else case and log "Success OR fail without exception for...", just as you've done in FileDownloader

src/com/owncloud/android/files/services/RetryDownloadJobService.java
```diff
@@ -0,0 +1,80 @@
...
+                TAG,
+                String.format(
+                    "Account %1s was deleted, no retry will be done",
+                    accountName
+                )
+            );
+        }
+
+        jobFinished(jobParameters, false);  // done here, real job was delegated to another castle
+        return true;    // TODO or false? what is the real effect, Google!?!?!?!?
```
davigonz: Google, we love you ❤️

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -0,0 +1,317 @@
...
+ */
+
+public class TransferRequester {
+
+    private static final String TAG = TransferRequester.class.getName();
+
+    /**
+     * Call to upload several new files
+     */
+    public void uploadNewFile(
```
davigonz: I would rename this method as uploadNewFiles since it's used for uploading several files.

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -0,0 +1,317 @@
...
+            behaviour,
+            createRemoteFile,
+            createdBy
+        );
+    }
+
+    /**
+     * Call to update multiple files already uploaded
+     */
+    private void uploadUpdate(Context context, Account account, OCFile[] existingFiles, Integer behaviour,
```
davigonz: Same here, rename as uploadsUpdate?

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -0,0 +1,317 @@
...
+                    remotePath,
+                    accountName
+                )
+            );
+        }
+    }
+
+
+    /**
+     * Schedule a future trasnfer of an upload, to be done when a connection via an unmetered network (free Wifi)
```
davigonz: Typo here, transfer instead of trasnfer

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -0,0 +1,317 @@
...
+                             Boolean forceOverwrite) {
+
+        uploadUpdate(context, account, new OCFile[]{existingFile}, behaviour, forceOverwrite);
+    }
+
+
+    /**
+     * Call to retry upload identified by remotePath
+     */
+    public void retry (Context context, OCUpload upload) {
```
davigonz: Rename as retryUpload? I know that we are not retrying the downloads currently, but it could be a future feature, and we would have to add here a retryDownload method. :wink:

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /**
+     * Private implementation of retry.
+     *
+     * @param context           Caller {@link Context}
+     * @param account           OC account where the upload will be retried.
+     * @param upload            Persisted upload to retry.
+     */
+    private void retry(Context context, Account account, OCUpload upload) {
```
davigonz: Same here, retryUpload?

davivel:
@davigonz , take a look to the changes I added and review the code, please.

src/com/owncloud/android/files/services/TransferRequester.java
davivel: I wouldn't go that way. I'd prefer a future where the interface is more like ```retrySync```, or just ```sync``` , that keep duplicating methods for uploads and downloads when most of them are, essentially, the same.

Besides, the upload is already in the type of the parameters.

src/com/owncloud/android/files/services/TransferRequester.java
davivel: Same here :)

src/com/owncloud/android/files/services/FileUploader.java
davivel: Done

src/com/owncloud/android/files/services/RetryDownloadJobService.java
davivel: Love is in the air :D

src/com/owncloud/android/files/services/TransferRequester.java
davivel: Done

src/com/owncloud/android/files/services/TransferRequester.java
davivel: Done

src/com/owncloud/android/files/services/TransferRequester.java
davivel: Fixed

davigonz:
👍

PR #1944, opened 2017-04-17
Support hidden version in status.php taking version from capabilities…

davivel:
… API as preferred


davivel:
This is stupid. 

@DeepDiver1975 , please, take a look to repo configuration. We can't merge recent PRs. Same problem in the iOS repo.

PR #1951, opened 2017-04-27
Connect 'trash' and 'link' buttons of public shares to actions

davivel:
Requires https://github.com/owncloud/android-library/pull/160

### DIFF
--- a/src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
+++ b/src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
@@ -35,10 +35,10 @@ import com.owncloud.android.operations.CreateShareWithShareeOperation;
 import com.owncloud.android.operations.DownloadFileOperation;
 import com.owncloud.android.operations.MoveFileOperation;
 import com.owncloud.android.operations.RemoveFileOperation;
+import com.owncloud.android.operations.RemoveShareOperation;
 import com.owncloud.android.operations.RenameFileOperation;
 import com.owncloud.android.operations.SynchronizeFileOperation;
 import com.owncloud.android.operations.SynchronizeFolderOperation;
-import com.owncloud.android.operations.UnshareOperation;
 import com.owncloud.android.operations.UpdateSharePermissionsOperation;
 import com.owncloud.android.operations.UpdateShareViaLinkOperation;
 import com.owncloud.android.operations.UploadFileOperation;
@@ -230,7 +230,7 @@ public class ErrorMessageAdapter {
 
             }
 
-        } else if (operation instanceof UnshareOperation) {
+        } else if (operation instanceof RemoveShareOperation) {
 
             if (result.getData() != null && result.getData().size() > 0) {
                 message = (String) result.getData().get(0);     // share API sends its own error messages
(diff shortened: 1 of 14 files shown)


src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
```diff
@@ -443,7 +443,7 @@ private static String getGenericErrorMessageForOperation(RemoteOperation operati
                 ) {
             message = res.getString(R.string.share_link_file_error);
 
-        } else if (operation instanceof UnshareOperation) {
+        } else if (operation instanceof RemoveShareOperation) {
             message = res.getString(R.string.unshare_link_file_error);
```
davigonz: You could rename unshare_link_file_error to remove_link_file_error in order to be consistent with the name of the operation

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -280,13 +280,16 @@ private void initShareViaLinkListener(View shareView) {
...
+        //GetLink from the server and show ShareLinkToDialog
+        mListener.copyOrSendPublicLink(share);
     }
 
     @Override
-    public void deletePublicLink(OCShare share) {
-
+    public void removePublicLink(OCShare share) {
+        // Unshare
+        Log_OC.d(TAG, "Unsharing " + share.getName());
```
davigonz: Removing instead of Unsharing....

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -280,13 +280,16 @@ private void initShareViaLinkListener(View shareView) {
...
+    public void copyOrSendPublicLink(OCShare share) {
+        //GetLink from the server and show ShareLinkToDialog
+        mListener.copyOrSendPublicLink(share);
     }
 
     @Override
-    public void deletePublicLink(OCShare share) {
-
+    public void removePublicLink(OCShare share) {
+        // Unshare
```
davigonz: Remove instead of Unshare

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -490,7 +491,7 @@ private void updateListOfUserGroups() {
     public void unshareButtonPressed(OCShare share) {
         // Unshare
         Log_OC.d(TAG, "Unsharing " + share.getSharedWithDisplayName());
```
davigonz: Removing....

src/com/owncloud/android/ui/fragment/SearchShareesFragment.java
```diff
@@ -229,7 +229,7 @@ public void onDetach() {
     @Override
     public void unshareButtonPressed(OCShare share) {
         // Unshare
-        mListener.unshareWith(share);
+        mListener.removeShare(share);
         Log_OC.d(TAG, "Unshare - " + share.getSharedWithDisplayName());
```
davigonz: Remove?

src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
davivel: We prefer not renaming strings. When we do that, they are handled by Transifex as if a string was removed and other was added, so that we lose the translations and the same string needs to be translated again.

davivel:
@davigonz , updated. Please, check again.

davigonz:
👍

PR #1954, opened 2017-04-27
Connect angle icon with dialog to update public share

davivel:


src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -282,7 +282,7 @@ public boolean onPreferenceClick(Preference preference) {
                             Intent intent = new Intent(Intent.ACTION_VIEW, uriUrl);
                             startActivity(intent);
                         }
-                        //ImprintDialog.newInstance(true).show(preference.get, "IMPRINT_DIALOG");
+                        //ImprintDialog.newInstanceToCreate(true).show(preference.get, "IMPRINT_DIALOG");
```
davigonz: You've changed this line but why is still commented? Is not going to be used?

PR #1959, opened 2017-05-05
Show warning about public shares is enabled by admin in server side

davivel:
Requires https://github.com/owncloud/android-library/pull/162

### DIFF
--- a/src/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/com/owncloud/android/providers/FileContentProvider.java
@@ -50,7 +50,6 @@ import com.owncloud.android.db.ProviderMeta;
 import com.owncloud.android.db.ProviderMeta.ProviderTableMeta;
 import com.owncloud.android.lib.common.accounts.AccountUtils;
 import com.owncloud.android.lib.common.utils.Log_OC;
-import com.owncloud.android.lib.resources.shares.OCShare;
 import com.owncloud.android.lib.resources.shares.ShareType;
 import com.owncloud.android.utils.FileStorageUtils;
 
@@ -809,7 +808,7 @@ public class FileContentProvider extends ContentProvider {
                     db.setTransactionSuccessful();
 
                     // SQLite does not allow to drop a columns; ftm, we'll not recreate
-                    // the files table without the column, just forget about
+                    // the files table without the column FILE_PUBLIC_LINK, just forget about
 
                 } finally {
                     db.endTransaction();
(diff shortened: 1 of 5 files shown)


src/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -839,12 +835,28 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                 try {
                     db.execSQL("ALTER TABLE " + ProviderTableMeta.CAPABILITIES_TABLE_NAME +
                             " ADD COLUMN " + ProviderTableMeta.CAPABILITIES_SHARING_PUBLIC_MULTIPLE
-                            + " INTEGER " + " DEFAULT 0");
+                            + " INTEGER " + " DEFAULT -1");
```
davigonz: Was 0 value not being used as FALSE and 1 as TRUE?

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -410,6 +410,15 @@ private boolean isPublicShareDisabled() {
      */
     private void updateListOfPublicLinks() {
 
+        // Hide warning about public links if not enabled in server
```
davigonz: According to the variable name, it would be Show instead of Hide, wouldn't it?

src/com/owncloud/android/providers/FileContentProvider.java
davivel: I updated it, since these are ```CapabilityBooleanType``` instances, that hold three values: FALSE (0), TRUE (1) and UNKNOWN (-1). UNKNOWN seems more appropriate as default.

src/com/owncloud/android/ui/fragment/ShareFileFragment.java
davivel: The comment is for the next "paragraph" of code, no just the variable. The code executed conditionally is on ```(!showDisplayWarning)```, negated. And what it does is setting the visibility of the message to ```GONE```, so it's effectively hiding it (or doing nothing).

davigonz: 👍

src/com/owncloud/android/providers/FileContentProvider.java
davigonz: 👍

PR #1965, opened 2017-05-18
Load privacy policy zoomed out for better support of non-responsive p…

davivel:
…ages

Fixes owncloud/enterprise#1964


davivel:
@jesmrec , let's test this as part of the regression tests in the release. It's an easy fix.

davivel:
Targeting to release branch instead...

davivel:
Will add it directly

PR #1967, opened 2017-05-18
added community to make releases more prominent in public

Kawohl:
It would be great to have the information prior to the release so we can publish this stuff better.


davivel:
:+1:, thanks @Kawohl

PR #1975, opened 2017-05-29
Reword change log for version 2.4

davivel:


davivel:
No functional change, no QA involved

cc @jesmrec

PR #1979, opened 2017-06-04
Preference to set if notifications should be shown (second PR)

AigerimBazarbayeva:
Instant upload notifications for pictures and videos were not optional. Now, after changes in code it is possible to enable and disable feature of instant notifications.


davigonz:
Just trying to revive the stuck tests.

jesmrec:
instant uploads were re-designed, so the logic here is outdated. Will create a new issue to implement inside the new architecture scope.

PR #1986, opened 2017-06-14
[Espresso] Adapt login suite to lookup servers

jesmrec:
Lookup servers (with a dispatcher to several server instances) are now totally supported by the login suite with basic auth. Before, they were already supported, but the test case related with login in an already attached account (negative test). This test case will be skipped for lookup servers. It is not worthwhile to parameterize a test matrix [userX, instanceX] just to check an error message, also supported for other kind of servers.

@davivel @davigonz


davigonz:
LGTM @jesmrec , please rebase with master, the branch is out-of-date.

jesmrec:
@davigonz Thanks for the review. Branch rebased. Pending CI.

jesmrec:
All passed. Merging...

PR #1988, opened 2017-06-16
Write only share

### LINKED ISSUE #1971: Write-only share link for folders
AC:
 - [x] Rename "allow uploads" option in public shares of folders as "allow editing"
 - [x] Add checkbox "show file listing" checked by default if '"allow editing" is checked
 - [x] Disable checkbox if "allow editing" is unckecked
- [x] Show option only if capability 'supports_upload_only' is set 
- [x] Create share in server according to new show file listing switch state
- [x] Update share in server according to new show file listing switch state
- [x] Design test plan @jesmrec
- [x] Validate test plan @jesmrec

davigonz:
Fixes #1971 
Requires https://github.com/owncloud/android-library/pull/168

### DIFF
--- a/src/com/owncloud/android/operations/UpdateShareViaLinkOperation.java
+++ b/src/com/owncloud/android/operations/UpdateShareViaLinkOperation.java
@@ -43,6 +43,7 @@ public class UpdateShareViaLinkOperation extends SyncOperation {
     private String mPassword;
     private Boolean mPublicUpload;
     private long mExpirationDateInMillis;
+    private int mPermissions;
 
     /**
      * Constructor
@@ -56,6 +57,7 @@ public class UpdateShareViaLinkOperation extends SyncOperation {
         mPassword = null;
         mExpirationDateInMillis = 0;
         mPublicUpload = null;
+        mPermissions = OCShare.DEFAULT_PERMISSION;
     }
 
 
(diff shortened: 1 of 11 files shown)


davigonz:
@davivel @jesmrec Write only share is ready to review and pass QA

src/com/owncloud/android/operations/UpdateShareViaLinkOperation.java
```diff
@@ -56,6 +57,7 @@ public UpdateShareViaLinkOperation(long shareId) {
         mPassword = null;
         mExpirationDateInMillis = 0;
         mPublicUpload = null;
+        mPermissions = OCShare.DEFAULT_PERMISSION;
```
davivel: Why not 0? What if we don't want to change the current permission?

davigonz: There's no way to keep the current permission right now. If we don't send the permission we want in each update, it will be set to READ, have a look at https://github.com/owncloud/core/blob/7b7dfbcaf7360c88125d6ce5365146623d6c7195/apps/files_sharing/lib/API/Share20OCS.php#L370

davivel: :+1:

Issue #1870, opened 2016-12-30
Instant upload never started - ANDROID 6


M0ses:
### Actual behaviour
When taking a picture its not getting uploaded

### Expected behaviour
pictures should be uploaded, at least when connecting WIFI next time
 
### Steps to reproduce
1.  Configure InstantUpload
2.  Take picture
3. Check files on server

### Environment data
Android version:
multiple Versions

Device model: 
BQ Aquaris e4.5
BQ Aquaris X5 Plus
Huawei P8 Lite

Stock or customized system:
Stock rom's / mostly rooted

ownCloud app version:
2.1.1

ownCloud server version:
9.1.0
(problem exists since a year with multiple versions

### Logs
No relevant logs on server side ATM. Please let me know what to provide exactly.

Noticed this messages in 'adb logcat'

```
12-30 21:48:31.886  1871  3900 I ActivityManager: Killing 5166:com.owncloud.android/u0a95 (adj 15): empty #17
12-30 21:48:43.908  1871  1884 I ActivityManager: Start proc 6979:com.owncloud.android/u0a95 for broadcast com.owncloud.android/.files.BootupBroadcastReceiver
12-30 21:48:59.192  1871  1898 I ActivityManager: Waited long enough for: ServiceRecord{7a8fd19 u0 com.owncloud.android/.services.observer.FileObserverService}
```

M0ses:
I could fix it on my BQ Aquaris X5 Plus

It seems to be a permissions problem, but I could not fix it by revoking and granting permissions again. 
I had to deinstall and install the owncloud app again.

I will do some further investigation on the other devices

jesmrec:
@M0ses is Android 7 your OS?

houseofbugs:
I am having issues where nothing is triggering on my Android app running 2.2.0. I have the latest OwnCloud server installed (just installed today too).

M0ses:
@jesmrec 

Sorry for delay. To give one example:

* BQ Aquaris X5 Plus
* Android Version:: 6.0.1
* App Version: 2.1.1
  * Upgrade to 2.2.0 did not help
  * Deinstall/Install app again did the trick
* Config Pictures/Videos: 
  * Path: /SofortUpload
  * Wifi only
  * Camera directory: /storage/emulated/0/DCIM/Camera
  * keep original
* AFAICR : App was upgraded (from Version unknown)
  * App taken from F-Droid
[...]

Issue #1829, opened 2016-10-23
Instant Uploads never started - ANDROID 7


kigero:
### Actual behaviour

With Instant Upload turned on, no pictures or videos are automatically uploaded, with or without the "wifi only" toggle on.  After reading the related issues, it seems like generally Instant Upload works on Android 7, but some specific phones are having issues.
### Expected behaviour

Pictures and videos should automatically upload.
### Steps to reproduce
1. Enable instant upload in the settings.  Set the appropriate folder for pictures (tried both internal and external SD card).
2. Take a picture.
3. The picture is not uploaded - and does not show in the Uploads section.

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes
### Environment data

Android version: 7

Device model: LG v20

Stock or customized system: Stock

ownCloud app version: 2.1.2 (built from master branch on Oct 22).

ownCloud server version: 8.1
#### Web server error log

No relevant logs found.
#### ownCloud log (data/owncloud.log)

No logs found.

davivel:
Hi, @kigero .

What camera app are you using? 

What are the exact paths you are setting for the camera folder?

Was the device originally shipped with Android 7, or was it upgraded from Android 6?

Thanks in advance.

kigero:
I"m using the default camera app (com.lge.camera).

The path that I am currently using for the SD card is: /storage/6264-3033/DCIM/Camera.  The path that is used by the internal card - and which Owncloud originally picked up automatically - is: /sdcard/DCIM/Camera.  I've tried both with no success.

It originally shipped with Android 7.

Thanks!

piperfect:
I am having similar results with the LG V20 without using an external SD card.  The default camera path after install is /storage/emulated/0/DCIM/Camera

davivel:
Thank you both. We need to fix this, but version 2.2.0 needs to go before the end of the year and it's currently too loaded. I'm sorry, will have to wait to the next one.

kigero:
I did a bit of digging this morning and found a few interesting notes that might be useful when it comes time to take a look at this:

1. The InstantUploadsObserver  just doesn't seem to be triggering at all on that folder, but it does seem to be configured and started properly - I'm not seeing any errors in the log.
[...]

PR #1990, opened 2017-06-16
Detect instant uploads with different observer implementations depend…

davivel:
…ing on device manufacturer.

First improvement to address #1829 and #1870 to test in Beta app.

This address the first and easiest to detect problem: detection via ```android.os.FileObserver``` doesn't work at all in devices of certain manufacturers, specially in Android 6 and Android 7. This issue has been filed in Android repo, but not attended. For instance, here:

https://issuetracker.google.com/issues/37131428

Some other times, it has been responded with ```GOTO(manufacturer)```, such as here: 

https://issuetracker.google.com/issues/37120851

This solution is based in active polling. The polling time has been set to 5 minutes, so, **uploads are not instant anymore**, might be a delay of **up to 5 minutes** from the moment they are taken to the moment they are uploaded.

### DIFF
--- a/src/com/owncloud/android/services/observer/FileObserverService.java
+++ b/src/com/owncloud/android/services/observer/FileObserverService.java
@@ -3,7 +3,7 @@
  *
  *   @author David A. Velasco
  *   Copyright (C) 2012 Bartek Przybylski
- *   Copyright (C) 2016 ownCloud GmbH.
+ *   Copyright (C) 2017 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License version 2,
@@ -182,7 +182,7 @@ public class FileObserverService extends Service {
         mAvailableOfflineObserversMap = null;
 
         if (mInstantUploadsObserver != null) {
-            mInstantUploadsObserver.stopWatching();
+            mInstantUploadsObserver.stopObserving();
             mInstantUploadsObserver = null;
         }
 
@@ -416,27 +416,31 @@ public class FileObserverService extends Service {
                 mInstantUploadsObserver == null
             )  {
             // no current observer -> create it
-            mInstantUploadsObserver = new InstantUploadsObserver(config, getApplicationContext());
-            mInstantUploadsObserver.startWatching();
+            mInstantUploadsObserver = InstantUploadsObserverFactory.newObserver(
+                config,
+                getApplicationContext()
+            );
+            mInstantUploadsObserver.startObserving();
 
         } else if ( !config.isEnabledForPictures() && !config.isEnabledForVideos() &&
                     mInstantUploadsObserver != null
             ) {
             // nothing ot observe -> stop current observer
-            mInstantUploadsObserver.stopWatching();
+            mInstantUploadsObserver.stopObserving();
             mInstantUploadsObserver = null;
-
-        } else if ( mInstantUploadsObserver != null &&
-                    !mInstantUploadsObserver.getSourcePath().equals(config.getSourcePath())
-            ) {
-            // source path to watch was changed -> stop current observer, create a new one
-            mInstantUploadsObserver.stopWatching();
-            mInstantUploadsObserver = new InstantUploadsObserver(config, getApplicationContext());
-            mInstantUploadsObserver.startWatching();
 
         } else if ( mInstantUploadsObserver != null) {
             // observer exists and can handle changes in configuration -> update observer
-            mInstantUploadsObserver.updateConfiguration(config);
+            boolean configurationUpdated = mInstantUploadsObserver.updateConfiguration(config);
+            if (!configurationUpdated) {
+                // current observe cannot handle this configuration change -> replace with a new one
+                mInstantUploadsObserver.stopObserving();
+                mInstantUploadsObserver = InstantUploadsObserverFactory.newObserver(
+                    config,
+                    getApplicationContext()
+                );
+                mInstantUploadsObserver.startObserving();
+            }
 
         } else {
             Log_OC.i(TAG, "Instant uploads are disabled, no current observer -> nothing to do");
--- a/src/com/owncloud/android/services/observer/FileObserverService.java
+++ b/src/com/owncloud/android/services/observer/FileObserverService.java
@@ -412,7 +412,7 @@ public class FileObserverService extends Service {
         PreferenceManager.InstantUploadsConfiguration config =
             PreferenceManager.getInstantUploadsConfiguration(this);
 
-        if ((   config.isEnabledForPictures() || config.isEnabledForVideos()) &&
+        if ((config.isEnabledForPictures() || config.isEnabledForVideos()) &&
                 mInstantUploadsObserver == null
             )  {
             // no current observer -> create it
@@ -422,14 +422,14 @@ public class FileObserverService extends Service {
             );
             mInstantUploadsObserver.startObserving();
 
-        } else if ( !config.isEnabledForPictures() && !config.isEnabledForVideos() &&
+        } else if (!config.isEnabledForPictures() && !config.isEnabledForVideos() &&
                     mInstantUploadsObserver != null
             ) {
             // nothing ot observe -> stop current observer
             mInstantUploadsObserver.stopObserving();
             mInstantUploadsObserver = null;
 
-        } else if ( mInstantUploadsObserver != null) {
+        } else if (mInstantUploadsObserver != null) {
             // observer exists and can handle changes in configuration -> update observer
             boolean configurationUpdated = mInstantUploadsObserver.updateConfiguration(config);
             if (!configurationUpdated) {
--- a/src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnCommonsIOFileMonitor.java
+++ b/src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnCommonsIOFileMonitor.java
@@ -1,0 +1,203 @@
+/**
+ *   ownCloud Android client application
+ *
+ *   @author David A. Velasco
+ *   Copyright (C) 2017 ownCloud GmbH.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License version 2,
+ *   as published by the Free Software Foundation.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.owncloud.android.services.observer;
+
+import android.content.Context;
+import android.os.Handler;
+
+import com.owncloud.android.db.PreferenceManager.InstantUploadsConfiguration;
+import com.owncloud.android.lib.common.utils.Log_OC;
+
+import org.apache.commons.io.monitor.FileAlterationListener;
+import org.apache.commons.io.monitor.FileAlterationMonitor;
+import org.apache.commons.io.monitor.FileAlterationObserver;
+
+import java.io.File;
+
+/**
+ * Observer watching a folder to request the upload of new pictures or videos inside it.
+ *
+ * Implementation based in {@link FileAlterationObserver}, from package {@link org.apache.commons.io.monitor},
+ * a component for monitoring file system events included inf Commons IO utilities by Apache Software Foundation.
+ *
+ * {@see https://commons.apache.org/proper/commons-io/}
+ */
+public class InstantUploadsObserverBasedOnCommonsIOFileMonitor
+    extends FileAlterationObserver implements InstantUploadsObserver {
+
+    private static final String TAG = InstantUploadsObserverBasedOnCommonsIOFileMonitor.class.getName();
+
+    private static final long serialVersionUID = 4821269467392712079L;
+
+    private static final int POLL_PERIOD_IN_MS = 1000 * 60 * 5;     // 5 minutes
+
+    private InstantUploadsConfiguration mConfiguration;
+    private Context mContext;
+    private InstantUploadsHandler mInstantUploadsHandler;
+
+    private FileAlterationMonitor mMonitor;
+
+    private final Object mLock = new Object();  // to sync mConfiguration, mainly
+
+    public InstantUploadsObserverBasedOnCommonsIOFileMonitor(
+        InstantUploadsConfiguration configuration,
+        Context context) {
+
+        super(configuration.getSourcePath());
+
+        if (context == null) {
+            throw new IllegalArgumentException("NULL context argument received");
+        }
+
+        mConfiguration = configuration;
+        mContext = context;
+        mInstantUploadsHandler = new InstantUploadsHandler();
+
+        mMonitor = new FileAlterationMonitor(POLL_PERIOD_IN_MS); // TODO some reasonable period
+    }
+
+    /**
+     * Updates the configuration for instant uploads with the one received.
+     *
+     * Source path of both the new and the current configurations must be the same.
+     *
+     * @param configuration     New configuration for instant uploads to replace the current one.
+     * @return                  'True' if the new configuration could be handled by the observer,
+     *                          'false' otherwise.
+     */
+    @Override
+    public boolean updateConfiguration(InstantUploadsConfiguration configuration) {
+        if (configuration == null) {
+            return false;
+        }
+        synchronized (mLock) {
+            if (mConfiguration.getSourcePath().equals(configuration.getSourcePath())) {
+                mConfiguration = configuration;
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public void startObserving() {
+        try {
+            initialize();
+            addListener(new CameraFolderAlterationListener());
+            mMonitor.addObserver(this);
+            mMonitor.start();
+
+        } catch (Exception e) {
+            Log_OC.e(
+                TAG,
+                "Exception starting to watch camera folder, instant uploads will not work",
+                e
+            );
+        }
+    }
+
+    @Override
+    public void stopObserving() {
+        try {
+            mMonitor.stop();
+            mMonitor.removeObserver(this);
+            checkAndNotify();
+            destroy();
+
+        } catch (Exception e) {
+            Log_OC.w(
+                TAG,
+                "Exception stopping to watch camera folder: " + e.getMessage()
+            );
+        }
+
+    }
+
+    private class CameraFolderAlterationListener implements FileAlterationListener {
+
+        private static final int HANDLE_DELAY_IN_MS = 200;
+
+        private Handler mHandler = new Handler();
+
+        @Override
+        public void onStart(FileAlterationObserver observer) {
+            Log_OC.v(TAG, "onStart called");
+        }
+
+        @Override
+        public void onDirectoryCreate(File directory) {
+            Log_OC.i(TAG, "onDirectoryCreate called for " + directory.getAbsolutePath());
+        }
+
+        @Override
+        public void onDirectoryChange(File directory) {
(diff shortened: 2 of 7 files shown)


src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnCommonsIOFileMonitor.java
```diff
@@ -0,0 +1,203 @@
...
+        public void onFileCreate(final File file) {
+            if (file != null) {
+                Log_OC.i(TAG, "onFileCreate called for " + file.getAbsolutePath());
+                synchronized (mLock) {
+                    final String fileName = file.getAbsolutePath().substring(
+                        mConfiguration.getSourcePath().length() + 1
+                    );
+
+                    /// delay a bit the execution to deal with possible renames of files (for instance: Google Camera)
+                    mHandler.postDelayed(
```
davigonz: I don't fully understand the delay that is being used here just to handle file renaming. Does the Google Camera change the file name after creating it? If so, shouldn't this case be managed in onFileChange method?

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
```diff
@@ -0,0 +1,194 @@
...
+public class InstantUploadsObserverBasedOnINotify extends FileObserver implements InstantUploadsObserver {
+
+    private static final String TAG = InstantUploadsObserverBasedOnINotify.class.getSimpleName();
+
+    private static final int CREATE_MASK = (
+            FileObserver.CREATE | FileObserver.MODIFY | FileObserver.CLOSE_WRITE |
+            FileObserver.MOVED_TO
+    );
+
+    private static final int ALL_EVENTS_EVEN_THOSE_NOT_DOCUMENTED = 0x7fffffff;   // NEVER use 0xffffffff
```
davigonz: Why not use 0xffffffff? If you suggest not to use it, a comment explaining the reason could be interesting.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
```diff
@@ -0,0 +1,194 @@
...
+
+        if (path != null && path.length() > 0) {
+            synchronized (mLock) {
+                Log_OC.d(TAG, "Observed camera folder is " + mConfiguration.getSourcePath());
+
+                if ((event & FileObserver.CREATE) != 0) {
+                    // new file created, let's watch it; false -> not modified yet
+                    mObservedChildren.put(path, false);
+                }
+                if ((   (event & FileObserver.MODIFY) != 0) &&
```
davigonz: Too much space here, before **(event...**

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
```diff
@@ -0,0 +1,194 @@
...
+                        mObservedChildren.containsKey(path) &&
+                        !mObservedChildren.get(path)
+                    ) {
+                    // watched file was written for the first time after creation
+                    mObservedChildren.put(path, true);
+                }
+                if (   (event & FileObserver.CLOSE_WRITE) != 0 &&
```
davigonz: Same here

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
```diff
@@ -0,0 +1,194 @@
...
+                    // for instance, Google Camera does so right after
+                    // saving a video recording
+                    handleNewFile(path);
+                }
+            }
+        }
+
+        if ((event & IN_IGNORE) != 0 &&
+            (path == null || path.length() == 0)) {
+            Log_OC.d(TAG, "Stopping the observance on " + mConfiguration.getSourcePath());
```
davigonz: Here is a log which says that the **observance is being stopped**, but where is the code that implements this stopping?

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Request the upload of a file just created if matches the criteria of the current
+     * configuration for instant uploads.
+     *
+     * @param fileName      Name of the file just created
+     */
+    private void handleNewFile(final String fileName) {
+
+        /// delay a bit the execution to deal with possible renames of files (for instance: Google Camera)
+        mHandler.postDelayed(
```
davigonz: I'm not sure if the delay here is needed, because if the file is renamed (by Google Camera for instance), this method will be called from **event & FileObserver.MOVED_TO**, after the file renaming, as you comment in https://github.com/owncloud/android/pull/1990/files#diff-14969411cf4e79636d20fff685a98c1dR129

src/com/owncloud/android/services/observer/InstantUploadsObserverFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Builds new instances of {@link InstantUploadsObserver}, using the appropritate implementation.
+ */
+class InstantUploadsObserverFactory {
+
+    private static final String TAG = InstantUploadsObserverFactory.class.getName();
+
+    public static InstantUploadsObserver newObserver(InstantUploadsConfiguration config, Context context) {
+        if (Build.MODEL.toLowerCase().contains("nexus") ||
```
davigonz: I guess that this is a temporal solution because it's not kind of scalable to check the different device models to use an InstantUploadsObserver or the other one. 

Besides, it is possible that people with Nexus or Pixel devices use cameras different to the Google one and find problems with instant uploads if we use **InstantUploadsObserverBasedOnINotify** instead of **InstantUploadsObserverBasedOnCommonsIOFileMonitor**.

 If I'm not wrong, it's more a matter of the camera app than of the device models.

davivel: It is a "fast-&-dirty" check. But maybe it's good enough for production. Or maybe we go in the end with the method based on commons.io for all the devices. In the end, the idea that "current implementation works in Google devices but not in others" is not exact, it is a rough estimation based on the feedback we have from users and their devices. We do not know the real reason why ```FileObserver``` works in some devices and not in others. 

Moreover, for this concrete problem - the fail of ```FileObserver``` to observer, there is nothing that points to camera apps as responsible of the problem. 

Remember: at the moment we have two methods of detection of new pictures working at the same time in the app: ```FileObserver``` on the camera folder, and listening for ```NEW_PICTURE``` broadcasts. The second one is the one we know is affected by camera apps, not the first one.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: Fixed

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: Fixed

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: Fixed.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnCommonsIOFileMonitor.java
davivel: This is just following the same process that was done before in the implementation based on ```FileObserver```. Let's discuss it down below.

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -411,27 +411,31 @@ private void updateInstantUploadsObservers() {
                 mInstantUploadsObserver == null
```
davivel: Fixed.

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -411,27 +411,31 @@ private void updateInstantUploadsObservers() {
...
             // no current observer -> create it
-            mInstantUploadsObserver = new InstantUploadsObserver(config, getApplicationContext());
-            mInstantUploadsObserver.startWatching();
+            mInstantUploadsObserver = InstantUploadsObserverFactory.newObserver(
+                config,
+                getApplicationContext()
+            );
+            mInstantUploadsObserver.startObserving();
 
         } else if ( !config.isEnabledForPictures() && !config.isEnabledForVideos() &&
```
davivel: Fixed

src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -411,27 +411,31 @@ private void updateInstantUploadsObservers() {
...
 
-        } else if ( mInstantUploadsObserver != null &&
-                    !mInstantUploadsObserver.getSourcePath().equals(config.getSourcePath())
-            ) {
-            // source path to watch was changed -> stop current observer, create a new one
-            mInstantUploadsObserver.stopWatching();
-            mInstantUploadsObserver = new InstantUploadsObserver(config, getApplicationContext());
-            mInstantUploadsObserver.startWatching();
-
         } else if ( mInstantUploadsObserver != null) {
```
davivel: Fixed

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: I'll extend the comment.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: The event is received *after* the stop occurs, not before. The system decided to stop sending events about changes in the file, or our observer was unsuscribed of the events before.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: If the delay is removed, the file will be uploaded twice: the first one, when the file is created; the second one, when the file is renamed; both actions would be understood as a different picture, though it's the same. Two requests to ```FileUploader``` would be done. The first one would probably fail because the file wouldn't have the original name when the service really tarts the transfer. The second on would succeed. Both would appear in the uploads view.

The trick here is completed in ```InstantUploadsHandler``` . Before requesting the upload, the handler checks that the file really exists in the path. Without the delay, the check will succeed for the file before the rename. With the delay, when the handler checks the existence the rename already happened, and the first detection will be silently discarded.

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: I will refactor the delay into ```InstantUploadsHandler```

src/com/owncloud/android/services/observer/InstantUploadsObserverBasedOnINotify.java
davivel: I will refactor the delay into ```InstantUploadsHandler```

davigonz: Got it, thank you for the explanation

src/com/owncloud/android/services/observer/InstantUploadsObserverFactory.java
davigonz: 👍

davivel:
Applied changes in CR

davigonz:
@davivel LGTM, we can continue with QA

jesmrec:
After some exploratory testing (not deep), the feature seems to work as expected. Ready for beta-testing.

@davivel @davigonz

davivel:
Going there!

src/com/owncloud/android/services/observer/InstantUploadsObserverFactory.java
mario: Sure we do - we do know why FileObserver works on some devices and doesn't on others :P You'll also notice polling doesn't work on a bunch of devices too :P But yea... scroll through all the commits first ;)

mario:
Some interesting inspirations are taken from Nc, right? :D

I look forward to your next implementation to see how "similar" it is ... :)

michaelstingl:
@mario You can follow the next steps here:
https://github.com/owncloud/android/milestone/27

jesmrec:
forking inspirations :)

mario:
@jesmrec nothing against you, but if it means what I think it means I'd be careful what you "fork", especially if it's my inspirations or code. I'm all for collaborating, but credit where credit is due.

Luckily you're not part of ownCloud, or I'd think you are very dangerous to the project health....

jesmrec:
@mario No bad intentions in my comment, only kidding. Open source works in this way and i know it. For me there is no problem that anyone inspires in other's code. And... unluckily i'm member 💃  (kidding again). Regards.

davivel:
@mario, 

> Some interesting inspirations are taken from Nc, right? :D

Yes, as I said here: https://github.com/owncloud/android/issues/1870#issuecomment-315038682

> I look forward to your next implementation to see how "similar" it is ... :)
> @jesmrec nothing against you, but if it means what I think it means I'd be careful what you "fork", especially if it's my inspirations or code. I'm all for collaborating, but credit where credit is due.

We don't grab or copy code of others without proper credit. If you find any code that fails in this, please, open an issue or comment it and the topic will be seriously handled. 

The next rewrite of instant uploads doesn't depend on me anymore, but you can be sure the code will not be blindly copy-pasted from NEXTcloud nor from any other project.

> Luckily you're not part of ownCloud, or I'd think you are very dangerous to the project health....

ownCloud is very lucky for having @jesmrec in its ranks. His skills and effort are crucial for the good health of both mobile apps. You will notice if you stay over here time enough, instead of judging anybody from a single comment.

mario:
@davivel right. I would have however at least appreciated notes in the source code, given that there is an open recognition from your side that you have no idea why the pre-existing code didn't work, and that there are clear traces of inspiration in your implementation. 

Also, the code in Nc had tons of workarounds for shortcomings of the external library which you "unfortunately" didn't copy, so you get further issues from that which is bad for customers, along with the already existing issues. Plus you added your own issues :P

So next time you'd like to do something like this, especially if it's my code, I'd appreciate if you'd consult me. Who knows, I might even be able to help you rather than finding of this in a bad, bad way.

I look forward to seeing how your NEW instant uploads will look like.

As for @jesmrec, I saw he's part of Solid GEAR, which in my limited understanding, is an agency that works for ownCloud.

Cheers,
Mario

jancborchardt:
> We don't grab or copy code from others without proper credit nor ignoring licenses. If you find any code that fails in this, please, open an issue or comment it and the topic will be seriously handled.

@davivel It already happened a year ago with https://github.com/owncloud/core/pull/25940#issuecomment-244061191 (also by a SolidGear employee) where code was copied verbatim without any attribution whatsoever and it wasn’t handled or reacted to at all.

davivel:
> I would have however at least appreciated notes in the source code, given that there is an open recognition from your side that you have no idea why the pre-existing code didn't work, and that there are clear traces of inspiration in your implementation.

You are right, adding some comments in code about why we use two different implementations of obrservers and the knowledge we have or I lack about the library / framework they are based on, would be handy for further development. Something like what I commented in the description of this PR, or what I answered in the code review, would be useful in the code and not just in Github. I will take note of that for the future.

About adding notes related to the "clear traces of inspiration", I'm sorry, it just didn't occurred to me. This is the first time I meet somebody that would like to be credited for inspiring about the existence of a third-party library. It's something I never felt the need for myself, so I guess that's why it didn't come to my head.

> Also, the code in Nc had tons of workarounds for shortcomings of the external library which you "unfortunately" didn't copy, so you get further issues from that which is bad for customers, along with the already existing issues. Plus you added your own issues :P

Thanks for caring about our customers, that's very kind from your part. Let me tell it: we do care a lot about all the users of the Android app, no matter if customers or not. We've always done our best to handle their issues, and we are aware we don't always respond as they would like, but we'll not give up in our intention of doing the best mobile apps in their kind.

About the shortcomings we didn't copy, it was not something "unfortunate", it was totally intended. We just didn't copy your code, shortcomings included. We don't copy code.

> So next time you'd like to do something like this, especially if it's my code, I'd appreciate if you'd consult me. Who knows, I might even be able to help you rather than finding of this in a bad, bad way.

If I'm not wrong, with "to do something like this", you mean to the process of 1) reading the code you wrote and released in a public FOSS repository ; 2) learn something about it ; 3) get my own conclusions and write my own code using what I learnt . Did I catch it correctly?

Personally, I will respect your desires. I will not read and learn from your code without consulting you in the process. @davigonz , @jesmrec , let me humbly recommend you to do the same.

@mario, let me suggest you to add a warning in that sense in the README.md file of your project. My experience in FOSS projects is limited, but I'd say that consulting the author of the project before reading and learning from their code is not a common expectation. An explicit warning could help you to prevent this to happen again in the future.

> I look forward to seeing how your NEW instant uploads will look like.

I will be very clear here. We know you announced a new implementation of your auto uploads included in the last beta version of the Nextcloud Android app. I didn't read a single line of code of it. AFAIK, nobody in the ownCloud mobile team did it. We were busy with our own stuff.

I'm pretty sure you made a great work there. Our plans for the NEW instant uploads included read your code and learn about it. But seems pretty clear that's something you don't like, so I would recommend abort that part and just don't read your implementation.

Fortunately, our plans included more stuff. A very high-level summary of them can be found in https://github.com/owncloud/android/issues/2028 .

Also, feel free to read our implementation and learn about it, if you think it's valuable for you, when it's done. You don't need to consult us first.

> As for @jesmrec, I saw he's part of Solid GEAR, which in my limited understanding, is an agency that works for ownCloud.

That's correct, but that doesn't disqualify him, or any of us, as an actual member of ownCloud. That's the way we have felt and behaved for over five years now. That doesn't justify either that you suggest he is dangerous for ownCloud in any way.

davivel:
@jancborchardt ,

> @davivel It already happened a year ago with owncloud/core#25940 (comment) (also by a SolidGear employee) where code was copied verbatim without any attribution whatsoever and it wasn’t handled or reacted to at all.

That saddens me. I remember that problem, and I though it was addressed in that moment.

As you know, my knowledge about the details of what happens in the ```core``` repository is very limited, and more limited is my ability to "move stuff" there. Nevertheless, I will try to move a fix forward.

Meanwhile, I'd suggest you to add another comment in that pull request to grab some attention from our mates in ```core```, or even opening a new issue in the repository about it. Otherwise, your interest in the topic could be missinterpreted as low.

jancborchardt:
> Meanwhile, I'd suggest you to add another comment in that pull request to grab some attention from our mates in core, or even opening a new issue in the repository about it. Otherwise, your interest in the topic could be missinterpreted as low.

Attribution should be a given, not only after nagging. My interest that this happens is high, however I’m not inclined to spend my time running after people and reminding them. It is something which needs to happen inside the organization and people not giving the proper attribution & credit. So I suggest you communicating this matter on an internal Owncloud company mailing list for example.

davivel:
> Attribution should be a given, not only after nagging.

Totally agree.

> however I’m not inclined to spend my time running after people and reminding them.

However, you have no problem in spending your time running after people and reminding them if the target is blaming about mistakes or contradictions, as you just did here. What makes me wonder if you really want to get the problem fixed or are just enjoying a little bit of trolling.

jancborchardt:
There has been code taken without attribution, I point it out and request you to bring it up in your organization, and that makes me a troll? Pointless to argue here.

mario:
I like how you guys copy&paste code without understanding what it means for your business licence-wise, and how you "inspire" yourself without thinking attribution is needed. I've tried to do my best here and get in touch with some oC team members who have made absolutely no effort to fix this and have proved rather harsh in my efforts to fix this situation.

Given I choose how I spend my free time, my free time will not be spent on helping your project or fixing things that YOUR team members asked me to fix for you. And please, I beg of you, just try to take ONE LINE OF CODE from me directly, then it'll get interesting ... :)

Oh, and just keep liking your own posts, that's so awesome!

PR #1994, opened 2017-06-23
Show beta version in name of beta APK generated with gradle

davivel:
With command ```./gradlew assembleBeta``` , before this PR the file generated is called ```android_null-beta.apk```; with this PR it becomes ```android_2.5.0-beta.1-beta.apk```

Last minute fix.


jesmrec:
👍

Issue #1993, opened 2017-06-23
"Show file listing" option in server v10.0.0


jesmrec:
### Steps to reproduce
 **Server 10.0.0**: Folder -> Public Share  -> Create ("+" button)

### Current behaviour
The option "Show File Listing" is shown. 

### Expected behaviour
WebUI in servers 10.0.0 does not handle this option, which is available from 10.0.1

Related with #1971

PR #1999, opened 2017-06-28
Log enabled in beta

davigonz:
Related to #1993 

When we designed the beta app, the beta buildType included debug mode to log different events in the app. 

Since FDroid doesn't allow to upload the app with debug mode and besides, it makes no sense because beta app needs to be signed to be published, I've enabled the logs in the beta buildType, so enabling debug is not longer needed to include logs in the beta app.

@jesmrec

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -149,7 +149,6 @@ android {
 
     lintOptions {
         abortOnError false
-        checkReleaseBuilds false
     }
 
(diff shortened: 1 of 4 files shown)


davigonz:
@jabarros Could you please review the code here? Thank you!

jabarros:
LGTM

Issue #2004, opened 2017-07-12
Forbidden reshare not reflected in the UI 


davivel:
#I have disabled resharing in the oC server settings, but if I create a new share in the Android UI, it still shows the option "Can share enabled". In the web UI or desktop UI this option isn't shown in the same case.

- Web UI
<img width="726" alt="oc_sharing" src="https://user-images.githubusercontent.com/214010/27550141-dd8e5872-5a9e-11e7-8987-e0b73055f8b6.png">

- Android UI

<img width="726" alt="oc_sharing" src="https://user-images.githubusercontent.com/15261363/29159589-6365ca88-7daf-11e7-8eb4-8c86750b45d5.png">

### Steps to reproduce
1. Disable "Allow reshare" in Admin settings
2. Create share in iOS app

### Expected behaviour
"Can share" should be hidden if resharing has been disabled

### Actual behaviour
"Can share" is visible

### Server configuration

**ownCloud version:**
9.1.6, 10.0.2

### Client
**Android version:**
6.0.1

**ownCloud app version:**
2.4.0

**Device model:**
Samsung Galaxy Note 4

Sibling: https://github.com/owncloud/ios/issues/915

jesmrec:
QA blocked, pending to check federated shares with SAML servers. Caused by:

https://github.com/owncloud/user_shibboleth/issues/195

jesmrec:
Let's move on this issue. In spite of having the bug with SAML, i don't expect a bad behaviour in Android app. In the worse case, the behaviour will be isolated and will be faced easily.

@davigonz @davivel

PR #2019, opened 2017-08-24
Forbidden reshare

davigonz:
Fix https://github.com/owncloud/android/issues/2004

### DIFF
--- a/src/com/owncloud/android/files/FileMenuFilter.java
+++ b/src/com/owncloud/android/files/FileMenuFilter.java
@@ -215,8 +215,12 @@ public class FileMenuFilter {
                 (capability.getFilesSharingApiEnabled().isTrue() ||
                         capability.getFilesSharingApiEnabled().isUnknown()
                 );
+
+        boolean allowResharing = isSingleSelection() && anyFileSharedWithMe() &&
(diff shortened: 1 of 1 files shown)


davigonz:
@davivel @jesmrec this bug is fixed, pending to review and pass QA

src/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -215,8 +215,12 @@ private void filter(List<Integer> toShow, List <Integer> toHide) {
                 (capability.getFilesSharingApiEnabled().isTrue() ||
                         capability.getFilesSharingApiEnabled().isUnknown()
                 );
+
+        boolean notAllowResharing = isSingleSelection() && anyFileSharedWithMe() &&
```
davivel: @davigonz , is the ```isSingleSelection()``` part really needed in this condition? Would appropriate to allow resharing multiple files if resharing is disabled by admin?

davigonz: No, it's not needed. This condition, indeed, it's being checked in the if just below and the share button will be hidden when is not a single selection.

PR #2023, opened 2017-08-31
Oauth2 support bugfixing

davivel:


src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1018,11 +1028,14 @@ private void startSamlBasedFederatedSingleSignOnAuthorization() {
...
-                + AccountUtils.getWebdavPath(mServerInfo.mVersion, mAuthTokenType);
+        ArrayList<String> targetUrls = new ArrayList<>();
+        targetUrls.add(
+            mServerInfo.mBaseUrl
+                + AccountUtils.getWebdavPath(mServerInfo.mVersion, mAuthTokenType)
+        );
         WebViewDialog dialog = WebViewDialog.newInstance(
-            targetUrl,
-            targetUrl,
+            targetUrls.get(0),
```
davigonz: You could pass targetUrls variable just one time and access to targetUrls.get(0) from the WebViewDialog.newInstance method

src/com/owncloud/android/ui/dialog/WebViewDialog.java
```diff
@@ -213,16 +211,6 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
     }
 
     @Override
-    public void onSaveInstanceState(Bundle outState) {
```
davigonz: Do not we need onSaveInstanceState any more? Are we handling the screen rotation from other point?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
davivel: True, but that would make the interface of ```WebViewDialog.newInstance```  more difficult to understand.

Now it is:  ```WebViewDialog.newInstance(String initialUrl, List<String> targetUrls, ...)```

The two parameters are separated because they have an important meaning to differentiate: URL loaded when the webview is shown VS list of URLs that will make the webview to close if they are get. 

In this case, the initial URL is the same as one the URLs that should close the dialog. But that's a coincidence, if you check the changes in ```AuthenticatorActivity```, the initial URL doesn't match any target URL.

So, to unify both parameters in a single one we should have an interface like:

```WebViewDialog.newInstance(List<String> urls, ....)``` 

where the first item in urls is the initial URL, and the rest of items are the targets URLs. This meaning is not understandable just watching the method signature, and could easily lead to mistakes in the caller.

davigonz: Understood, it makes sense to separate them in two different parameters, thanks

davigonz:
LGTM 👍

PR #2035, opened 2017-09-22
Enable private-link sharing when property is available via WebDAV

davivel:


davivel:
Sorry, not for 2.5.0 release

PR #2037, opened 2017-09-27
Set minimum of 4 characters to search for sharees in server, and maxi…

davivel:
…mum of 30 sharees per response.


davivel:
@jesmrec , @davigonz, last minute addition.

Fix #enterprise/2189

PR #2040, opened 2017-10-03
Release 2.5.0

davigonz:


oc_jb_workaround/build.gradle
```diff
@@ -8,6 +8,11 @@ android {
     compileSdkVersion 24
     buildToolsVersion '25.0.2'
 
+    defaultConfig {
+        versionCode = 100036
+        versionName = "1.0.36"
+    }
```
jesmrec: why did you move version form manifest to build.gradle in workaround?

davigonz: This change is not mine but I guess that has been done just to follow the same behaviour that we have included in the app manifest and build.gradle, where we are also using defaultConfig block to encapsulate default settings for each build variant (release, debug, beta), although In oc_jb_workaround build gradle we only have defined the release build variant.

PR #2043, opened 2017-10-10
Fix Espresso tests after UI changes in login view

jesmrec:
Right now, login view only shows URL field from scratch and not username and password. These changes have been developed in automatic tests to adapt new login view:

- Check username and password are hidden, as precondition
- Check status message after URL connection, by clicking on scroll and not in username field as before


davigonz:
@jesmrec LGTM :+1:

jesmrec:
Thanks @davigonz. Let's merge then.

PR #2049, opened 2017-10-19
Update issue_template.md

davigonz:
The test/test credentials are not working anymore for the demo instance, we will use other instead.


jesmrec:
Code is OK for me.

Are these credential changes in demo instance going to be kept? any reason to grant admin permissions to everyone who wants to test something?

davigonz:
@jesmrec 

> Are these credential changes in demo instance going to be kept? 

I hope so

> any reason to grant admin permissions to everyone who wants to test something?

No idea, I didn't notice this change till I have tested the old credentials. Maybe the demo instance responsible should know about it.

jesmrec:
@davigonz thanks, in any case it is better to keep correct credentials in templates.

davigonz:
Not needed finally

PR #2053, opened 2017-10-24
Fix #2052

### LINKED ISSUE #2052: CookieSyncManager not needed from 21
Hi, I found CookieSyncManager is instantiated in [WebviewDialog](https://github.com/owncloud/android/blob/f58114698b971365d4be0f9632cb58a0666b15c5/src/com/owncloud/android/ui/dialog/WebViewDialog.java#L154). I understand it was instantiated to avoid crashes when instantiating CookieManager. But from API 21, it is no longer needed and deprecated. CookieManager can work well by itself. (According to the [doc](https://developer.android.com/reference/android/webkit/CookieSyncManager.html))

frankFic:
I created a pull request to fix #2052 I reported

### DIFF
--- a/src/com/owncloud/android/ui/dialog/WebViewDialog.java
+++ b/src/com/owncloud/android/ui/dialog/WebViewDialog.java
@@ -26,12 +26,13 @@ import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Dialog;
 import android.content.DialogInterface;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.support.annotation.NonNull;
 import android.support.v4.app.DialogFragment;
-import android.support.v4.app.FragmentTransaction;
 import android.support.v4.app.FragmentManager;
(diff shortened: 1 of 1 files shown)


davigonz:
Hi @frankFic , thank you very much for this contribution, I've just approved the changes.

Next step, a bit of QA by @jesmrec

jesmrec:
Code changes only affects those devices with API >= 21, so such version will be the target in QA. 

Checked both auth methods that involve webviews to get credentials form the user in case of first authentication and credential changes as well.

All correct, approved.

jesmrec:
@frankFic the branch is outdated with master, please rebase and it will be ready to merge.

frankFic:
@jesmrec Thanks, I updated the branch. Is that OK?

jesmrec:
@frankFic travis is passing, when it finishes i think we are allowed to merge. CC @davigonz

jesmrec:
Merged, thanks again!!

Issue #2028, opened 2017-09-05
Drop instant uploads in favor of automatic (non-instant) camera uploads


davivel:
#From the feedback we got from https://github.com/owncloud/android/issues/1829 and https://github.com/owncloud/android/issues/1860, we know that we can't trust in the current implementation of instant uploads to grant that all the picture we'll be eventually uploaded.

At this moment we have tried all the things possible, in our knowledge, to upload the files "instantly", i.e: as soon as the pictures are taken.

We need to forget the idea of immediately and focus in reliability. Most of the feedback we have is from people that are much more concerned about having all there pictures effectively pushed to the server side than about the time needed to get it.

The problem of the current implementation is that the file observers we register for changes in the camera folder "live" in a background service that we try to keep alive forever. If that service is killed for any reason, the observers "die" and pictures are not detected anymore until the service is restarted. The restart is done when the app is opened again, or when the device is rebooted.

In the past, the death of the service didn't happen often. But from Android 4.4, Google is silently killing "immortal" services more and more often with each Android version. With Android 8 Google finally is documenting that services in background will not be allowed to live more than a limited time.

We need to go for a total rewrite of the detection of pictures, that will be based in the next key points:

1. Schedule a job that syncs the camera folder periodically if auto uploads are enabled.
2. Every time the job is triggered it should check the contents of the camera folder to find all the pictures that are not uploaded to the server. The job itself needs to compare the pictures in the local folder with the pictures that are already uploaded -- it cannot trust in being notified from the system per picture.
3. The job will upload new pictures as new files to the server; if there are pictures already know but the app that were not successfully uploaded, the job will resume / retry those uploads too.

There are many other details that be improved on the road, but these points address the reliability problem we have in the current implementation.

_____
**TASKS:** 

 - [x] Dive inside the current code to understand it better
 - [x] Create new job to: 
    - Check content of local camera folder. 
    - Check content of upload folder.
    - Compare folders content and upload those images that don't exist in the server.
    - Retry previous unsuccessful uploads.
- [x] Upload only new pictures/videos (if user deletes a picture/video from the remote folder, it shouldn't be uploaded again) 
- [x] Handle only-wifi option
- [x] Clean up previous implementation

_____
**BUGS & IMPROVEMENTS**
 - [X] (1) Crash in **Android 8** when the ownCloud app is in background and tries to enqueue an upload with the camera file. [FIXED] 
[...]

PR #2062, opened 2017-11-06
Drop instant uploads in favour of camera uploads [WIP]

davigonz:
Implements #2028

___ 

BUGS & IMPROVEMENTS

- [X] (1) [WORDING] Label for paths [FIXED] 
- [X] (2) [Bug] Pics with no loading files view https://github.com/owncloud/android/pull/2062#issuecomment-347603181 [FIXED]  
- [X] (3) [Bug] Pics never uploaded https://github.com/owncloud/android/pull/2062#issuecomment-347837777  [FIXED, NEW EXPECTED BEHAVIOUR]
- [X] (4) [Bug] Pics with feature disabled are uploaded https://github.com/owncloud/android/pull/2062#issuecomment-347838543  [FIXED- new implementation in https://github.com/owncloud/android/pull/2062#issuecomment-354288647 
- [X] (5) [Bug] Crash when local camera folder doesn't exist (DCIM/Camera by default) https://github.com/owncloud/android/pull/2062#issuecomment-349980510 [FIXED] 
- [X] (6) [Bug] Pics are uploaded after disabling option and executing job https://github.com/owncloud/android/pull/2062#issuecomment-351343526 [FIXED] - new implementation https://github.com/owncloud/android/pull/2062#issuecomment-354288647 @jesmrec 
- [X] (7) [Bug] Job checks wrong camera folder https://github.com/owncloud/android/pull/2062#issuecomment-351432415 [NOT REPRODUCIBLE]
- [x] (9) Update from latest version https://github.com/owncloud/android/pull/2062#issuecomment-356580698 [WON'T BE FIXED]


davigonz:
@jesmrec Can you have a look at this PR? It will be included in the coming 2.5.1 beta release

jcmontesmartos:
LGTM

jesmrec:
### Wording (1)

Current:

`Upload picture path` 
`Upload video path`

Suggested:

`Picture upload path`
`Video upload path`

jesmrec:
### Bug (2)

Steps:

1. Enable instant uploads for pics or videos
2. Tap on circle button to vanish the app and go to camera app
3. Take some pics

Current

Pics are never uploaded

Expected:

Pics are uploaded after some time (less than 15 min)

Tested with Galaxy Note 4 v6, Nexus 6P v7.1

jesmrec:
@davigonz: If i take some pics, and i disable the feature before these pics are uploaded, which is the expected behaviour?

davigonz:
> If i take some pics, and i disable the feature before these pics are uploaded, which is the expected behaviour?

@jesmrec The periodic job is automatically cancelled and the app won't upload those pics. See: https://github.com/owncloud/android/pull/2062/files#diff-257db3cd78d8a1eab3328301bfcaaab1R86

jesmrec:
### Bug (3)

Steps:

1. Enable instant uploads for pics or videos
2. Take same pics
3. Before the pics are uploaded, disable instant uploads

Current

Pics are never uploaded

Expected:

Pics are uploaded. 

Tested with Galaxy Note 4 v6, Nexus 6P v7.1

jesmrec:
### Bug (4)

Steps:

1. Enable instant uploads 
2. Take same pics
3. Before the pics are uploaded, disable instant uploads
4. Take more pics
5. After some time (more than 15 minutes), enable camera uploads

Current

All pics are uploaded (steps 2. and 4.)

Expected:

Only pics at 2. should be uploaded 

Tested with Galaxy Note 4 v6, Nexus 6P v7.1

davigonz:
### Bug (5)

Steps:

1. Enable camera uploads.
2. Select a specific folder.
3. Delete that folder. 
4. Take some pics.

Current:

App crashes when camera uploads synchronization starts. 

Expected:

App doesn't crash and camera uploads synchronization is properly performed. 

Tested with Nexus 5X v8.0.0

jesmrec:
Bugs (1), (2) and (5), fixed

Bugs (3) and (4) are still reproducible (Nexus 9 Android 7, Nexus 5 Android 6)

@davigonz

jesmrec:
@### Bug (6)

Steps:

1. Enable instant uploads 
2. Take some pics
3. Before the pics are uploaded, disable instant uploads and take more pics
(after a bit of time, pics in step 1. are uploaded correctly and 3. are ignored)
4. After that, take more pics
5. After some time (more than 15 minutes), enable camera uploads

Current

Pics of 4. are uploaded

Expected:

Pics of 4. are not uploaded

Tested with Samsung Galaxy S7 v7

jesmrec:
i have noticed this behaviour:

if you switch the Camera folder, the next time the job is executed, it will be uploaded all the pics in such folder from the last job execution (not from the point the Camera folder was switched).

So, each time the job is executed, it is uploaded the content related with the set-up at this time. if you change four times the camera folder during the slot between job executions, it will be taken in account the one in the moment of the job execution.

I think, we can take this as expected behaviour.

jesmrec:
### Bug (7)

Steps:

Not easy to reproduce, but these steps can be performed:

1.- Switch camera folder to another one (for example, Camera MX)
2.- Take pics with the new camera (they are uploaded, correct behaviour)
3.- Switch again to the original camera folder

Current

After some time, checking logs: there are jobs executing over the Camera MX folder

Expected:

Only checks over the current camera folder are executed.

Tested with Samsung Galaxy S7 v7

davigonz:
### Bug (8)

**Steps:**

1. Enable camera uploads for pictures
2. Take some pics (1 and 2)
3. Wait till job is executed, uploading pictures 1 and 2 properly
4. Take a new pic (3) before the job in step 3 is executed again (<15 minutes)
5. Disable camera uploads for pictures before the job in step 3 is executed again (<15 minutes)
6. Enable camera uploads for pictures before the job in step 3 is executed again (<15 minutes). A new job will be schedule here.
7. After a while, the job in step 6 will be executed.

Scheme:

img_20171228_124842

**Current:**

App doesn't upload picture 3.

**Expected:**

Picture 3 should be uploaded.

**Tested with Samsung Galaxy S7 Edge v7.0**

davigonz:
Taking into account the [bug 8](https://github.com/owncloud/android/pull/2062#issuecomment-354283023), and after thinking during a while how to fix it (a third date needed?) I would say in this point that we are including too complexity here, specially added when trying to upload pending pictures once camera uploads feature has been disabled. Let's see an example with some times:

_10:00:_ user enables camera uploads for pictures and the job to check pictures and videos from camera folder is automatically scheduled.
_10:05:_ user takes picture number 1.
_10:10:_ user takes picture number 2.
_10:12:_ user disables camera uploads for pictures.
_10:15_: **job is executed, but should it upload pictures 1 and 2?** The initial implementation for new camera uploads didn't take into account this and didn't upload pictures 1 and 2 while current implementation does.

To face that issue, we included:
- A new finish date that saves the moment in which the feature has been disabled, to fix [bug 4](https://github.com/owncloud/android/pull/2062#issuecomment-347838543).
- Separate jobs for pictures and videos to upload pending pictures when camera uploads for pictures has been recently disabled but camera uploads for videos is enabled.

So, we have been questioning **what the user expects once camera uploads has been disabled**:
1. Pictures/videos taken when the feature was enabled are uploaded, even after disabling the feature.
2. Pictures/videos taken when the feature was enabled aren't uploaded once the user disables the feature, although he has taken the pictures/videos while the feature was enabled. 

Since it is being difficult to fit the moment when the job to upload pictures/videos is executed and the moment when the user enables or disables the feature and it seems that we are adding complexity that might trigger pictures not to be uploaded, I would go back to the behaviour defined in step 2 above, showing an additional dialog to the user with a message like:

**"Are you sure you want to disable this feature? The pending pictures and videos won't be uploaded"**

CC @jesmrec @michaelstingl @nasli

michaelstingl:
> "Are you sure you want to disable this feature? The pending pictures and videos won't be uploaded"

Very clear, easy to understand 👍

nasli:
Agree @davigonz, we should go with 2 aproach, for me when I disabled the feature I will expect that pending photos are not going to be uploaded and with the warning the user will know, and also same behavior than in iOS 👌

jesmrec:
It's ok for me. We can afford the lost uploads after disabling by awarding the user of the behaviour in this point.

jesmrec:
At this point, bugs (4) and (6) make no sense anymore, so the disabling of the option discards the possibility of uploading pics/videos taken before this action.

The disabling/enabling of the feature will be tested again, and in case of finding bugs, they will be a different ones.

jesmrec:
### Bug (9)

1.- Install previous version (2.5.0)
2.- Enable instant uploads and set some config related (folder...)
3.- Upgrade to current

Current behaviour: all previous set-up was reset, all is disabled
Expected: previous set-up is not reset

Tested with Nexus 5X v8

jesmrec:
Bug (7) checked again and not reproducible

davigonz:
@jesmrec I think there's nothing we can do related to [Bug 9](https://github.com/owncloud/android/pull/2062#issuecomment-356580698) since we have updated the name of the preferences from instant upload whatever to camera uploads whatever (see https://github.com/owncloud/android/pull/2062/files#diff-7757df0546633dccb162e75d58b8d019R25) and are like new preferences for the Android system.

jesmrec:
In this case, we have to find out a way to let users know that they have to re-enable the feature after upgrading. 

CC @michaelstingl

michaelstingl:
Detect the old setting was present and fire up a dialogue? (only once, of course)

davigonz:
> In this case, we have to find out a way to let users know that they have to re-enable the feature after upgrading

> Detect the old setting was present and fire up a dialogue? (only once, of course)

@michaelstingl @jesmrec 

Unfortunately, all the references to the previous instant uploads feature are replaced with new ones and there's not an easy way to detect if the old setting was enabled without using and maintaining a new table in database, shared preference or similar, which from my point of view, is not a good solution.

Besides, I'd say that it's not a good idea to warn the user about "re-enabling" the feature since the feature to be added (camera uploads) is not the same as the previous feature (instant uploads) and users have to be aware of that to avoid misunderstandings about the moment in which pictures/videos will be uploaded.

So, I think that we can solve this by explaining the feature properly to the user. The fact is that, in this moment, we only have the changelog, blog posts and central topic to inform the users about the new features we add but I've just proposed to include a [new wizard](https://github.com/owncloud/android/issues/2090) that appears directly in the app, in which we can include this kind of information, more visible for the users.

jesmrec:
Many apps are able to show a hint/alert with the new features after upgrading (for example, Telegram). The wizard is a good idea, but it should be a first guide for new users (like iOS does). I think it should not be triggered after each upgrade.

davigonz:
@jesmrec it is possible to move forward this so that we can start pushing the 2.6.0 release?

jesmrec:
@davigonz final checks are being performed right now. Very soon i expect to approve this to be included in 2.6.0

jesmrec:
Approved.

PR #2065, opened 2017-11-17
Add touchImageView source code

davigonz:
Necessary to include future beta versions in F-Droid since .jar files are no longer allowed. 

See this line in [F-Droid inclusion policy](https://f-droid.org/docs/Inclusion_Policy/?title=Inclusion_Policy): 
> Binary dependencies such as JAR files have to be replaced by source-built versions or used from a trusted repository (see manual).

Discussion opened: https://gitlab.com/fdroid/fdroiddata/merge_requests/2657


davigonz:
@jesmrec Can you have a look at it? I have basically moved the TouchImageView library classes to a new module. You do not need to review the code in the library classes but the dependencies in AndroidManifest.xml and build.gradle files.

jesmrec:
@davigonz will the library code be included in the app also for release?

davigonz:
> will the library code be included in the app also for release?

@jesmrec I think so, as long as we do not come up with another idea, because in the way we were releasing the last beta versions (from master branch directly) , maintaining the library code in a .jar file for release versions and in source code for beta version could be a mess since we should have to do the changes every time we release a new version (release or beta).

jesmrec:
For mi it is ok :)

davigonz:
> For mi it is ok :)

Great, let's merge then

PR #2085, opened 2018-01-03
Go back camera uploads [WIP] 

davigonz:
With this PR we are going to revert the last changes applied for camera uploads feature, but in a safe way, without rewriting the commits history, just in case we need some of that implementation in the future.

Taking into account [this comment](https://github.com/owncloud/android/pull/2062#issuecomment-354288647), I have gone from [Pics never uploaded commit](https://github.com/owncloud/android/pull/2062/commits/ed25e084f89a9f2613630eee17e1fbf7a26a795d) and included a new dialog to ask the user to confirm camera uploads disabling.


davigonz:
I will merge this branch, the code will be reviewed in https://github.com/owncloud/android/issues/2028

PR #2086, opened 2018-01-03
Revert camera uploads

davigonz:
I'm facing some problems to include these changes into drop_instant_uploads branch, let's get it another try


davigonz:
Not needed finally, commits included directly in https://github.com/owncloud/android/pull/2062

PR #2088, opened 2018-01-10
Include beta signing configuration

davigonz:
This PR includes the signing configuration to compile the beta app automatically


davigonz:
@jesmrec thanks, merging...

Issue #2027, opened 2017-09-04
Feature Request - Camera Upload


gpullen:
Under the 'Upload' button,  it would be great if we can immediately take a photo or series of photos that are saved directly to the cloud.

Many thanks

jesmrec:
Thanks for your suggestion!!

sd1998:
Is someone working on this issue?

sd1998:
@davigonz Thanks I will be working in this Issue

davigonz:
Great @sd1998 ! If you have any question, do not hesitate in asking me

sd1998:
@davigonz Thanks :)

sd1998:
@davigonz Should I also add the functionality to directly upload a video taken from the camera?

gpullen:
@sd1998 Yes if possible, this would really help us!

Many thanks for working on this!

sd1998:
@gpullen I will be sending the PR for the same by tomorrow

sd1998:
@gpullen 
I am having this error while running this app on my device

<img width="409" alt="screen shot 2017-10-27 at 12 24 39 am" src="https://user-images.githubusercontent.com/22047421/32071627-79d4ea2a-baad-11e7-831f-e26f72ec8f56.png">

<img width="1340" alt="screen shot 2017-10-27 at 12 24 56 am" src="https://user-images.githubusercontent.com/22047421/32071633-7b71d910-baad-11e7-85d2-7a38559fb554.png">

Please help me with this so that I can test the initial changes that I have done.

gpullen:
No idea, I'm afraid I'm not an android coder... Just an end user requesting a feature!  Good luck!

sd1998:
@davigonz Can you help me with the same?

davigonz:
> I am having this error while running this app on my device

> Can you help me with the same?

Follow up: https://github.com/owncloud/android/issues/2060

davigonz:
> Should I also add the functionality to directly upload a video taken from the camera?

@sd1998 sorry I missed this question, that issue is being addressing in instant uploads feature https://github.com/owncloud/android/issues/2028. Do not worry about that.

davigonz:
@gpullen @sd1998 I include here a mockup with a new button with the text "Take picture" that appears when pressing the floating action button, as @jesmrec suggests in https://github.com/owncloud/android/pull/2059#issuecomment-341375182

[...]

PR #2097, opened 2018-01-24
Pictures can be upload directly clicking them from the device's camera.

sd1998:
Fix to Issue #2027 

_____

BUGS & IMPROVEMENTS

- [X] (1) File name of pics https://github.com/owncloud/android/pull/2097#issuecomment-362556980 [FIXED]
- [X] (2) Option switched from copy to move https://github.com/owncloud/android/pull/2097#issuecomment-362571538 [FIXED]
- [X] (3) Wording issues [FIXED]


davigonz:
Hi @sd1998 , thank you for opening this again, using a new branch for a new feature is always the best approach 👍 

I've seen that the commit of the android-library (7c7cad7e990705c180f4fe2934191c317a3f72fc) is not the proper one, please update it to the latest one (0e7fd8ee8fb1ad082acba8911412571b206ee1f0). 

From your Android app directory:

`cd owncloud-android-library`
`git checkout 0e7fd8ee8fb1ad082acba8911412571b206ee1f0`
`cd ..`
`git add owncloud-android-library`
`git commit ...`
`git push ...`

sd1998:
@davigonz Is it the latest one now? (I have just changed it)

davigonz:
>  Is it the latest one now? (I have just changed it)

@sd1998 Yes, the library change has just disappeared from the **Files changed** tab, thanks 👏

sd1998:
@davigonz Has the QA started?

res/values/strings.xml
```diff
@@ -82,6 +82,10 @@
     <string name="file_list_file">file</string>
     <string name="file_list_files">files</string>
     <string name="filedetails_select_file">Tap on a file to display additional information.</string>
+    <string name="upload_owncloud"> Upload to OwnCloud </string>
+    <string name="upload_bottom_sheet_displayed"> Upload Bottom Sheet Displayed </string>
```
davigonz: This resource is not being used, please delete it

res/values/strings.xml
```diff
@@ -82,6 +82,10 @@
...
+    <string name="upload_bottom_sheet_collapsed"> Upload Bottom Sheet Collapsed </string>
```
davigonz: Same here, delete it

res/values/strings.xml
```diff
@@ -176,6 +180,7 @@
     <string name="foreign_files_remote_text">"Remote: %1$s"</string>
     <string name="upload_query_move_foreign_files">There is not enough space to copy the selected files into the %1$s folder. Would you like to move them instead? </string>
     <string name="pass_code_enter_pass_code">Please insert your passcode</string>
+    <string name="bottom_sheets_layout_behaviour"> android.support.design.widget.BottomSheetBehavior </string>
```
davigonz: Not used, can be deleted

res/values/strings.xml
```diff
@@ -205,6 +210,9 @@
     <string name="media_rewind_description">Rewind button</string>
     <string name="media_play_pause_description">Play or pause button</string>
     <string name="media_forward_description">Fast forward button</string>
+    <string name="button_text_camera"> Camera </string>
```
davigonz: Why are you using spaces before and after the value of this resource? If you want to add a margin or padding, please include it in the xml file instead of here.

res/values/strings.xml
```diff
@@ -205,6 +210,9 @@
...
+    <string name="upload_from_camera_title"> Upload From Camera </string>
```
davigonz: Same here, there are several ones

res/values/strings.xml
```diff
@@ -205,6 +210,9 @@
...
+    <string name="capture_button_text"> Capture </string>
```
davigonz: Not used?

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -376,26 +454,36 @@ protected Boolean doInBackground(Void... params) {
...
+                } else {
+                    data.putExtra(EXTRA_CHOSEN_FILES, mFileListFragment.getCheckedFilePaths());
+                }
 
                 SharedPreferences.Editor appPreferencesEditor = PreferenceManager
                         .getDefaultSharedPreferences(getApplicationContext()).edit();
 
-
-                if (mRadioBtnMoveFiles.isChecked()){
+                if(requestCode == FileDisplayActivity.REQUEST_CODE__UPLOAD_FROM_CAMERA){
```
davigonz: You are using the condition `if(requestCode == FileDisplayActivity.REQUEST_CODE__UPLOAD_FROM_CAMERA){` twice here to set several params, is not possible to use just one and unify all the code?

davigonz:
> Has the QA started?

@sd1998 No, I've just passed what I think is going to be last code review. After that, @jesmrec will start QA.

sd1998:
@davigonz I will work on this I remember using some of the string resources in Log_OC.d() I will check them again. If I am not using them then I will remove them.
Regarding the last one I think one way to unify the code will be to create another class that extends AsyncTask and execute it when the user selects the upload from camera option from the bottom sheet. But I did not do that in the first place because the new AsyncTask will be very much similar to CheckAvailableSpaceTask class (the part in doInBackground() and onPostExecute()) which will result in much more code duplication hence I used the current approach. Should I add another AsyncTask and not use the if statements?

davigonz:
> Regarding the last one I think one way to unify the code will be to create another class that extends AsyncTask and execute it when the user selects the upload from camera option from the bottom sheet. But I did not do that in the first place because the new AsyncTask will be very much similar to CheckAvailableSpaceTask class (the part in doInBackground() and onPostExecute()) which will result in much more code duplication hence I used the current approach. Should I add another AsyncTask and not use the if statements?

@sd1998 What I meant is something easier, like this: 

screen_shot_2018-01-25_at_15_23_51

sd1998:
@davigonz Ok I will work on this.

res/values/strings.xml
sd1998: Actually I am using upload_owncloud

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
sd1998: Yes I have done that

sd1998:
@davigonz Are there any more changes required?

res/values/strings.xml
davigonz: I meant upload_bottom_sheet_displayed, which you've already deleted. 

There're still some space at the beginning and end of *Upload to ownCloud* in upload_owncloud resource. Do you really need it? If so, please use margin or padding in the xml layout.

res/values/strings.xml
davigonz: Are you using button_text_camera resource?

res/values/strings.xml
davigonz: capture_button_text seems not to be used either.

davigonz:
> Are there any more changes required?

@sd1998 thanks for applying the changes, have a look at the next three comments:

- https://github.com/owncloud/android/pull/2097#discussion_r164354395

- https://github.com/owncloud/android/pull/2097#discussion_r164354765

- https://github.com/owncloud/android/pull/2097#discussion_r164354983

res/values/strings.xml
sd1998: Yes I just saw that. I used it when I was not using bottom sheets and instead using a FAB.

sd1998:
@davizon I have done all the changes.

davigonz:
@sd1998 It looks good to me, changes approved, ready to pass QA @jesmrec

jesmrec:
Starting QA here...

jesmrec:
**IMPROVEMENT** (1)

When a pic is taken, the name it is upload with is the epoch time of the moment. This is not incorrect, of course, but maybe is more clear to set the calendar date as filename, for example: `2018-02-02-1159.JPG` or similar, like the style used for most of the cameras.

what do you think?

sd1998:
@jesmrec I think epoch time will be better as it is possible to take two picture in a minute and upload them. Thus resulting in naming conflicts. For ex. If two pictures are taken at 11:59 then both of them will be named 2018-02-02-1159.

jesmrec:
it can be set always the second fraction, i mean `2018-02-02-115934.JPG` what do you think @davigonz ?

davigonz:
> it can be set always the second fraction, i mean 2018-02-02-115934.JPG what do you think @davigonz ?

@jesmrec @sd1998 I would use something like IMG_20180131_185750.jpg , which is the current format used by the default Android camera app to save the taken pictures and the users are used to it.

jesmrec:
**BUG (2)**

Steps to reproduce:

1. Install the app
2. Try to upload any file (not form camera). The option copy/move (bottom of the view) is set to copy by default
3. Go back to files view, open FAB and upload file from camera.
4. Take a pic and accept it -> it is uploaded correctly
5. Try to upload again a file (not from camera)

Current: Option has switched from copy to move
Expected: Option remains as copy

Tested with Nexus5 v6

sd1998:
@davigonz @jesmrec This is the current naming format in which the pictures are getting uploaded.

sd1998:
@jesmrec [Bug](https://github.com/owncloud/android/pull/2097#issuecomment-362571538) is beacuse I am moving the files that are uploaded directly from the camera. I will fix this now.

jesmrec:
@sd1998 is there no other option to name the pics?

sd1998:
@jesmrec No. But should I add one. Naming the picture as in the user should be given an option to name the picture everytime?

jesmrec:
No, i do not mean that the user set a name by typing it everytime he/she takes a pic. This would be bothering.

if you'd have the choice to set a name to the pic file to upload before queuing it, i would set something like @davigonz suggested here https://github.com/owncloud/android/pull/2097#issuecomment-362571250

sd1998:
@jesmrec So I just have to change the name of each image taken to the naming scheme suggested by @davigonz Before queuing it for upload?

jesmrec:
yes, this would be the scenario:

1. User taps on FAB, select update, select update from camera
2. Camera starts, user takes a pic, user accepts the pic
3. Pic is queued with name `IMG_20180131_185750.jpg` (for example, with a more clear timestamp)
4. Pic is uploaded

any question related, do not hesitate to ask us

sd1998:
@jesmrec I have fixed both the bugs.

jesmrec:
@sd1998 i will check, thanks!

sd1998:
@jesmrec There was a small bug (If the picture was taken in February then the image name would contain the month value as 2 instead of 02) with the new image naming scheme (better Timestamp) I . have fixed it now.

jesmrec:
both bugs are fixed. Let's continue with QA

davigonz:
@sd1998 I didn't notice you are using "Upload From Camera" with the second and third word starting with capital letter, it would be better "Upload from camera", wouldn't it? @jesmrec

res/values/strings.xml
```diff
@@ -82,6 +82,8 @@
     <string name="file_list_file">file</string>
     <string name="file_list_files">files</string>
     <string name="filedetails_select_file">Tap on a file to display additional information.</string>
+    <string name="upload_owncloud">Upload to ownCloud</string>
```
jesmrec: the word "ownCloud" should be parameterized with the value in `setup.xml` file "app_name"

jesmrec:
@davigonz right, the correct default format should be "Upload from camera", only capitals in the first word

sd1998:
@davigonz Sorryfor that I will change it. Similarly for the word "ownCloud" should I use it from setup.xml?

sd1998:
@jesmrec @davigonz I have fixed the issue.

jesmrec:
All issues fixed, all working fine. Approved.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -22,14 +22,20 @@
...
+import android.graphics.Bitmap;
+import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Environment;
 import android.preference.PreferenceManager;
+import android.provider.MediaStore;
 import android.support.v4.app.DialogFragment;
 import android.support.v7.app.ActionBar;
+import android.util.Log;
```
davigonz: Import not needed, you can use the option Code => Optimize Imports once you have finished your code or from time to time to delete automatically the unused imports

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -39,6 +45,7 @@
 import android.widget.RadioButton;
 import android.widget.TextView;
 
+import com.google.android.exoplayer2.C;
```
davigonz: Same here

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -48,7 +55,10 @@
 import com.owncloud.android.ui.fragment.LocalFileListFragment;
 import com.owncloud.android.utils.FileStorageUtils;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
```
davigonz: Same here

davigonz:
@sd1998 just some unused imports, delete them and we will merge, no extra QA needed @jesmrec

Thank you so much again @sd1998 for implementing this interesting feature

davigonz:
@sd1998 

@jesmrec tested with different devices and the bottom sheet was completely opened, but I have noticed this behaviour with Nexus 6P

upload_from_camera

Since the handling of this menu in different devices is likely difficult, there's a constant in the `setPeekHeight` function you are using [here](https://github.com/owncloud/android/pull/2097/files#diff-092d6016acf68fd310127b29d371a852R296) to set auto height.

If you want to see what I mean, right click in `setPeekHeight` of `uploadBottomSheetBehavior.setPeekHeight(400)` and Go to => Declaration . There's a constant called `PEEK_HEIGHT_AUTO`, which value is -1 and is used to set auto height.

Thus, please use -1 in uploadBottomSheetBehavior.setPeekHeight(-1) since 400 can be a good height value for some devices but a bad one for other ones.

sd1998:
@davigonz The bottom sheet not being completely visible on some devices that thought did cross my mind but I still used 400 (static values) as I thought that 400 was greater that the height taken by the layout and hence should on any device it should fit the whole bottom sheet. I did not take in account the concept of density independent pixels. I have changed it to -1 now.

davigonz:
Sorry @sd1998 for these requested changes after code review but there are some of them that you don't see till you test the app in a specific device. 

In the GIF attached in my previous comment, you will notice also that the camera icon is smaller than the files one, moving the **Upload from camera** text a little to the left. 

I've been checking the resources folders and in the drawable-xxxhdpi folder, the camera icon has a resolution of 72x72 while the file icon resolution is 96x96, so the camera icon resolution in that case should be 96x96 as well, same as the other xxxhdpi resources.

jesmrec:
@davigonz @sd1998 the fix related with `PEEK_HEIGHT_AUTO`works fine when the device is in portrait, but if you set the device in landscape the view is partially hidden

feb-06-2018 12-45-39

sd1998:
@davigonz @jesmrec I have fixed [this](https://github.com/owncloud/android/pull/2097#issuecomment-363395222) and regarding the bottom sheet should I implement a different layout for upload bottom sheet when the device is in landscape configuration? In this new layout we can make the files and the upload from camera option to be side by side rather than being one below another.

davigonz:
> In this new layout we can make the files and the upload from camera option to be side by side rather than being one below another.

Even with the layout your propose, If we use **PEEK_HEIGHT_AUTO**, we would continue having the problem described in https://github.com/owncloud/android/pull/2097#issuecomment-363398512 and if we do not use **PEEK_HEIGHT_AUTO**, it would reappear the problem described in https://github.com/owncloud/android/pull/2097#issuecomment-363387082.

There is likely to be a common solution to show all the options both in landscape and portrait modes.

The problem with bottom sheets in landscape mode seems to be known, have a look at https://stackoverflow.com/questions/41591733/bottom-sheet-landscape-issue , it may useful for you

jesmrec:
@sd1998 do you think is feasible to fix that?

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -100,6 +104,11 @@
     private OCFile mFile = null;
     private FileListListAdapter mAdapter;
 
+    private LinearLayout uploadFilesLinearLayout;
```
davigonz: These four variables do not need to be declared globally, you can declare it directly in registerFabUploadListeners method, where are going to be used.

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -257,8 +266,35 @@ private void registerFabUploadListeners() {
...
+                    public boolean onTouch(View v, MotionEvent event) {
+                        UploadFilesActivity.startUploadActivityForResult(getActivity(),((FileActivity) getActivity()).getAccount(),
+                                FileDisplayActivity.REQUEST_CODE__UPLOAD_FROM_CAMERA);
+                        dialog.hide();
+                        return false;
+                    }
+                });
+                uploadToTextView.setText(String.format(getResources().getString(R.string.upload_to),getResources().getString(R.string.app_name)));
+                uploadBottomSheetBehavior = BottomSheetBehavior.from((View) uploadBottomSheet.getParent());
+                uploadBottomSheetBehavior.setPeekHeight(-1);
```
davigonz: I've managed to fix https://github.com/owncloud/android/pull/2097#issuecomment-363387082 and  https://github.com/owncloud/android/pull/2097#issuecomment-363398512 to give you a hand, follow the next steps:

1. Delete line `uploadBottomSheetBehavior.setPeekHeight(-1);`
2. Include the next code:

```
dialog.setOnShowListener(new DialogInterface.OnShowListener() {
    @Override
    public void onShow(DialogInterface dialog) {
        uploadBottomSheetBehavior.setPeekHeight(uploadBottomSheet.getMeasuredHeight());
    }
});
```
With this code, we wait till the dialog is shown, measure its height and use it to set the height of the bottom sheet. I've tested it with different devices (portrait & landscape) and seems to be working properly.

davigonz:
Hi @sd1998, I've managed to fix the problem with the bottom sheet height to give you a hand, please have a look at the changes I requested above.

Besides, please update the camera icon for xxxhdpi as I described in https://github.com/owncloud/android/pull/2097#issuecomment-363395222.

After that, @jesmrec will perform the last tests, we will merge and continue with other interesting features.

Thanks

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
sd1998: @davigonz Thanks for the help. I trying to find the value of the height of bottom sheet using ViewTreeObserver class but it did not work out. I will implement the changes.

sd1998:
@davigonz I have made all the changes and added the xxxhdpi icon

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -292,8 +288,14 @@ public boolean onTouch(View v, MotionEvent event) {
                     }
                 });
                 uploadToTextView.setText(String.format(getResources().getString(R.string.upload_to),getResources().getString(R.string.app_name)));
-                uploadBottomSheetBehavior = BottomSheetBehavior.from((View) uploadBottomSheet.getParent());
-                uploadBottomSheetBehavior.setPeekHeight(-1);
+                final BottomSheetBehavior uploadBottomSheetBehavior = BottomSheetBehavior.from((View) uploadBottomSheet.getParent());
+                uploadBottomSheetBehavior.setState(BottomSheetBehavior.STATE_EXPANDED);
```
davigonz: I think you don't need this line

sd1998: @davigonz Yes I forgot to remove it. Sorry I will remove it now.

davigonz:
Thanks @sd1998 , let's start with the last QA round @jesmrec

jesmrec:
The icon size is correct and now the bottom sheet appears and vanish correctly (checked with three devices in different screen sizes).

Finally, we can approve this

davigonz:
Thanks for the QA @jesmrec , I will merge when @sd1998 rebases the branch with master

sd1998:
@davigonz I have done the rebase.

davigonz:
@sd1998 @jesmrec Merged, this feature will be included in the next 2.7.0 release, great work 👍

sd1998:
@davigonz @jesmrec Thanks.

Issue #2009, opened 2017-07-21
App is reopened from login view after double tap on "Back" button


jesmrec:
### Steps to reproduce
In login view, tap twice in "Back" button (triangle button)

### Current behaviour
App is closed

### Expected behaviour
App is closed and reopened

yongjiexu:
I think this should be a feature, not a bug.

jesmrec:
@yongjiexu maybe... depending on the expected behaviour.

sd1998:
@jesmrec I would like to work on this issue.

jesmrec:
@sd1998 ok! first of all, try to reproduce the problem by yourself, so you can check what is really happening. Please, open a new branch to develop the fix and tell us if you need help.

sd1998:
@jesmrec I tried to reproduce the problem. The login layout is account_setup.xml in res/layout right? What I have seen is that when the activity(AuthenticatorActivity.java) launches the on screen keyboard opens and then if I press back two times the app is closed. To fix this issue when the user presses the back button the app should not close right? Or the app should close and relaunch again. And this should not happen when the user is adding a second account because in that case on pressing back button twice the app returns to the FileDisplayActivity.java.

jesmrec:
Two taps on the back button are performed without the screen keyboard, and no accounts attached. I mean, first tap on the back button to vanish the keyboard, and then, double tap to reproduce the problem. The expected behaviour is that the app is closed as well as with one tap. Maybe it is not easy to reproduce (for that reason the low severity).

sd1998:
@jesmrec So after the keyboard disappears if the user taps the back button two times the app should not close?

jesmrec:
It should close. The wrong behaviour is that it does not. It closes and reopens.

For that reason, i pointed you to try to reproduce it before start developing :)

sd1998:
@jesmrec  I am trying to reproduce it but it happens sometimes (very rarely) and most of the times it just closes.

sd1998:
@jesmrec I think I just have to eliminate the very rarely possibility of the activity to relaunch.

jesmrec:
Not reproducible anymore

PR #2099, opened 2018-02-01
The app closes if the back button is pressed twice

sd1998:
Fix issue #2009

### DIFF
--- a/src/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -195,6 +195,8 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
 
     private String mAuthStatusText = "";
 
+    private boolean backPressed = false;
+
     private int mAuthStatusIcon = 0;
     
     private String mAuthToken = "";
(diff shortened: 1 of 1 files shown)


sd1998:
@jesmrec I think now the issue has been fixed.

jesmrec:
Ok, let's move to ready to test and will pass code review and QA

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -308,6 +310,16 @@ public void onClick(View v) {
         //Log_OC.e(TAG,  "onCreate end");
     }
 
+    @Override
+    public void onBackPressed(){
+        if(backPressed){
+            super.onBackPressed();
```
davigonz: I'm not able to stop the debugger here, the **backPressed** variable is set to true below, but in which situation it is possible enter this method with backPressed true?

sd1998: @davigonz The bug was quite tough to reproduce because it is quite rare(because if back button is clicked once the activity closes) I just added backPressed condition as a backup that in case if the back button is pressed twice the activity should be closed.

jesmrec:
The bug is not easy to reproduce, but the code developed can avoid the relaunch on the second press.  Not QA for this with the difficulties to reproduce. From my side, we can merge this fix

davigonz:
> The bug is not easy to reproduce, but the code developed can avoid the relaunch on the second press. Not QA for this with the difficulties to reproduce. From my side, we can merge this fix

@jesmrec @sd1998 sorry but I wouldn't merge it if we are not sure whether this code really fix the problem. I mean, it might have no effect in the app and we are including extra code.

sd1998:
@davigonz I don't think the code is unnecessary as the problem is the app is not closing if the back button is pressed twice this can occur if the activity is not able to exit on single backpress (which can be one of the reasons) hence if that does not happen then when the user presses the back button next time the activity will surely close because of the code that I have added.

Issue #2071, opened 2017-12-07
Include new dialog to encourage users to add a new app review in PlayStore


davigonz:
It would be great to increase the number of app reviews in Play Store so that we can receive more feedback from our users.

The idea is to create a new dialog that would appear from time to time (we will think about the exact frequency), asking the users for a new review and redirect them to the PlayStore.

Mockup:

![rate_dialog](https://user-images.githubusercontent.com/15261363/33764566-2db0c1d0-dc15-11e7-9a40-820ee074fcd6.png)

davigonz:
Rate dialog mockup: 
![rate_dialog](https://user-images.githubusercontent.com/15261363/33764566-2db0c1d0-dc15-11e7-9a40-820ee074fcd6.png)

harshitbansal05:
@davigonz can i work on this issue?

harshitbansal05:
@davigonz should I use a library for this purpose(there are some present)? Also, the design should be the same as in the screenshot?

harshitbansal05:
@davigonz @PVince81 do i need to set up a server or is there a demo server url so I can login into app.

michaelstingl:
https://demo.owncloud.org/
user: demo
pw: demo:

davigonz:
> should I use a library for this purpose(there are some present)? Also, the design should be the same as in the screenshot?

Hi @harshitbansal05 , you don't need a new library, dialog fragments can be used for that purpose, we already have many of them implemented in our code, see [this](https://github.com/owncloud/android/tree/master/src/com/owncloud/android/ui/dialog). 

If you need more information. here you have the official documentation about Android dialogs: https://developer.android.com/guide/topics/ui/dialogs.html?hl=en

davigonz:
@harshitbansal05 

> Also, the design should be the same as in the screenshot?

That was a initial mockup but you can propose a new one if you want.

> do i need to set up a server or is there a demo server url so I can login into app.

You can use our demo site: https://demo.owncloud.org/login with the next credentials: 

- admin/admin
- test/test

jesmrec:
Ready to be released in 2.7.0

michaelstingl:
Related blog post:
https://owncloud.org/news/owncloud-app-in-review-heaven/

PR #2102, opened 2018-02-03
add rate-app dialog fragment

harshitbansal05:
This pr solves issue #2071. The current period for the first dialog to be shown is 2 days, while the min. number of launches are also 2. Clicking on the neutral button further prevents the dialog from opening for another day.

### DIFF
--- a/res/layout/rate_me_dialog.xml
+++ b/res/layout/rate_me_dialog.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ownCloud Android client application
+
+  Copyright (C) 2016 ownCloud GmbH.
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License version 2,
+  as published by the Free Software Foundation.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:padding="@dimen/standard_padding"
+    android:filterTouchesWhenObscured="true">
+
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:drawableRight="@drawable/ic_star_black_24dp"
+        android:drawableEnd="@drawable/ic_star_black_24dp"
+        android:drawablePadding="16dp"
+        android:textColor="@color/color_accent"
+        android:textSize="20sp"
+        android:text="@string/rate_dialog_title"/>
+
+    <View
+        android:layout_width="match_parent"
+        android:layout_height="0.5dp"
+        android:layout_marginTop="16dp"
+        android:layout_marginBottom="16dp"
+        android:background="@color/color_accent"/>
+
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:gravity="center"
+        android:textSize="16sp"
+        android:textColor="@color/color_accent"
+        android:layout_marginStart="16dp"
+        android:layout_marginEnd="16dp"
+        android:layout_marginBottom="16dp"
+        android:text="@string/rate_dialog_description" />
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="16dp"
+        android:layout_marginBottom="16dp"
+        android:weightSum="3"
+        android:gravity="center">
+
+        <Button
+            android:id="@+id/button_no_thanks"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:background="@color/color_accent"
+            android:textColor="@color/white"
+            android:textSize="12sp"
+            android:text="@string/button_no_thanks_name"/>
+
+        <Button
+            android:id="@+id/button_later"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:layout_marginStart="8dp"
+            android:layout_marginEnd="8dp"
+            android:background="@color/color_accent"
+            android:textColor="@color/white"
+            android:textSize="12sp"
+            android:text="@string/button_later_name"/>
+
+        <Button
+            android:id="@+id/button_rate_now"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:background="@color/color_accent"
+            android:textColor="@color/white"
+            android:textSize="12sp"
+            android:text="@string/button_rate_now_name"/>
+    </LinearLayout>
+</LinearLayout>
--- a/res/layout/rate_me_dialog.xml
+++ b/res/layout/rate_me_dialog.xml
@@ -2,7 +2,7 @@
 <!--
   ownCloud Android client application
 
-  Copyright (C) 2016 ownCloud GmbH.
+  Copyright (C) 2018 ownCloud GmbH.
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License version 2,
@@ -27,7 +27,8 @@
     <TextView
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_gravity="center"
+        android:layout_marginStart="8dp"
+        android:layout_marginEnd="8dp"
         android:drawableRight="@drawable/ic_star_black_24dp"
         android:drawableEnd="@drawable/ic_star_black_24dp"
         android:drawablePadding="16dp"
@@ -46,11 +47,10 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_gravity="center"
-        android:gravity="center"
         android:textSize="16sp"
         android:textColor="@color/color_accent"
-        android:layout_marginStart="16dp"
-        android:layout_marginEnd="16dp"
+        android:layout_marginStart="8dp"
+        android:layout_marginEnd="8dp"
         android:layout_marginBottom="16dp"
         android:text="@string/rate_dialog_description" />
 
@@ -58,7 +58,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="16dp"
-        android:layout_marginBottom="16dp"
+        android:layout_marginBottom="8dp"
         android:weightSum="3"
         android:gravity="center">
 
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -520,6 +520,12 @@
     <string name="media_service_notification_channel_description">See music player</string>
     <string name="file_sync_notification_channel_name">File sync</string>
     <string name="file_sync_notification_channel_description">See file sync result</string>
+
+    <string name="rate_dialog_title">Rate ownCloud app!</string>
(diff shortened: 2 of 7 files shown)


harshitbansal05:
@jesmrec can you please review this?

davigonz:
Hi @harshitbansal05 , thank you for this contribution, we will take care of it

src/com/owncloud/android/ui/dialog/RateMeDialog.java
```diff
@@ -0,0 +1,165 @@
...
+
+        Button rateNowButton = (Button) v.findViewById(R.id.button_rate_now);
+        Button laterButton = (Button) v.findViewById(R.id.button_later);
+        Button noThanksButton = (Button) v.findViewById(R.id.button_no_thanks);
+
+        rateNowButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                String packageName = getArguments().getString(APP_PACKAGE_NAME);
+                Uri uri = Uri.parse("market://details?id=" + packageName);
```
davigonz: I would create a new constant in the top of the class to include `"market://details?id="`

src/com/owncloud/android/ui/dialog/RateMeDialog.java
```diff
@@ -0,0 +1,165 @@
...
+                {
+                    flags |= Intent.FLAG_ACTIVITY_CLEAR_WHEN_TASK_RESET;
+                }
+                goToMarket.addFlags(flags);
+
+                try {
+                    startActivity(goToMarket);
+                } catch (ActivityNotFoundException e) {
+                    getActivity().startActivity(new Intent(Intent.ACTION_VIEW,
+                            Uri.parse("http://play.google.com/store/apps/details?id=" + packageName)));
```
davigonz: Same here for `"http://play.google.com/store/apps/details?id="`

src/com/owncloud/android/ui/dialog/RateMeDialog.java
```diff
@@ -0,0 +1,165 @@
...
+                editor.putLong("date_neutral", System.currentTimeMillis());
+                editor.apply();
+                dialog.dismiss();
+            }
+        });
+
+        noThanksButton.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                SharedPreferences preferences = getActivity().getSharedPreferences("app_rater", 0);
```
davigonz: From this class you could access to the preferences names constants declared in AppRater

src/com/owncloud/android/AppRater.java
```diff
@@ -0,0 +1,77 @@
...
+public class AppRater {
+
+    private static final String DIALOG_RATE_ME_TAG = "DIALOG_RATE_ME";
+
+    private final static int DAYS_UNTIL_PROMPT = 2;
+    private final static int LAUNCHES_UNTIL_PROMPT = 2;
+    private final static int DAYS_UNTIL_NEUTRAL_CLICK = 1;
+
+    public static void appLaunched(Context mContext, String packageName) {
+        SharedPreferences prefs = mContext.getSharedPreferences("app_rater", 0);
```
davigonz: The preferences names (app_rater, date_neutral, date_first_launch, don't_show_again) should go in constants in the top of the class

res/layout/rate_me_dialog.xml
```diff
@@ -0,0 +1,97 @@
...
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<LinearLayout
```
davigonz: The dialog design is pretty good but I would align the content to the left, since default Android UI tends to have the headers and text in that position while iOS usually show the content more centered in the middle, like you have done in your dialog. I think something like the next dialog would be better

res/layout/rate_me_dialog.xml
```diff
@@ -0,0 +1,97 @@
...
+        android:layout_marginStart="16dp"
+        android:layout_marginEnd="16dp"
+        android:layout_marginBottom="16dp"
+        android:text="@string/rate_dialog_description" />
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="16dp"
+        android:layout_marginBottom="16dp"
```
davigonz: 8dp would be enough as bottom margin

src/com/owncloud/android/ui/dialog/RateMeDialog.java
```diff
@@ -0,0 +1,165 @@
...
+            public void onClick(View v) {
+                String packageName = getArguments().getString(APP_PACKAGE_NAME);
+                Uri uri = Uri.parse("market://details?id=" + packageName);
+                Intent goToMarket = new Intent(Intent.ACTION_VIEW, uri);
+
+                /// To count with Play market back stack, After pressing back button,
+                /// to taken back to our application, we need to add following flags to intent.
+                int flags = Intent.FLAG_ACTIVITY_NO_HISTORY |
+                        Intent.FLAG_ACTIVITY_MULTIPLE_TASK;
+                if (Build.VERSION.SDK_INT >= 21)
```
davigonz: Instead of 21, please use Build.VERSION_CODES.LOLLIPOP

davigonz:
Hi @harshitbansal05 , I've just requested some changes, if there's any question ping me, thanks

harshitbansal05:
@davigonz i have done the requested changes...

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -199,6 +200,8 @@ protected void onCreate(Bundle savedInstanceState) {
         }   // else, Fragment already created and retained across configuration change
 
         Log_OC.v(TAG, "onCreate() end");
+
+        AppRater.appLaunched(this, getApplicationContext().getPackageName());
```
davigonz: I think you don't need here the application context since you can use the activity context, so just `getPackageName()` should work

davigonz:
> i have done the requested changes...

@harshitbansal05 Thanks for applying the changes, I suggested a new one

harshitbansal05:
@davigonz done..

davigonz:
Changes approved, ready to pass QA @jesmrec

jesmrec:
I have tried to rate the app with two different devices. The dialog is OK, then i select "Rate now" but i get an "Item not found" from Play Store (gmail account attached)

@harshitbansal05

jesmrec:
fixed by installing a signed release app.

jesmrec:
I miss here a brandable option in `setup.xml`for the case of users with their own distribution method. Default should be with the option enabled

davigonz:
> I miss here a brandable option in setup.xmlfor the case of users with their own distribution method. Default should be with the option enabled

I don't understand exactly what you mean here, are you talking about building the link to PlayStore to own distributions? If I'm not wrong, the code is getting the package name to build the link to PlayStore, so it should be enough.

jesmrec:
I mean, only a parameter in `setup.xml` to enable/disable the feature. What do you think about @harshitbansal05 ?

michaelstingl:
> I mean, only a parameter in setup.xml to enable/disable the feature.

Yes, good point. This is a requirement for users of a branded version.

harshitbansal05:
Yes, this can be done @jesmrec...

jesmrec:
ok @harshitbansal05, please, let us know anything you need

davigonz:
Hi @harshitbansal05 , could you include the flag to enable/disable the feature as @jesmrec suggested, so we can continue with QA here? Thanks

davigonz:
@harshitbansal05 thank you for including the flag to enable/disable the feature, let's proceed with some QA @jesmrec

res/values/strings.xml
```diff
@@ -520,6 +520,12 @@
     <string name="media_service_notification_channel_description">See music player</string>
     <string name="file_sync_notification_channel_name">File sync</string>
     <string name="file_sync_notification_channel_description">See file sync result</string>
+
+    <string name="rate_dialog_title">Rate ownCloud app!</string>
```
jesmrec: @harshitbansal05 ownCloud should not appear as name, in case of branding (thinking a bit in the future, so the option will be disabled). It should be used the `app_name`parameter.

jesmrec:
>This pr solves issue #2071. The current period for the first dialog to be shown is 2 days, while the >min. number of launches are also 2. Clicking on the neutral button further prevents the dialog from >opening for another day.

If i understood correctly:

- By default, the dialog will never raise before 2 days after installing. (managed by `DAYS_UNTIL_PROMPT`)
- After the second day, the dialog will raise if the number of launches is, at least, 2. If not, after the second launch, whenever it happens (managed by `LAUNCHES_UNTIL_PROMPT`)
- With neutral button (`Later` option), the dialog will raise the following day (managed by `DAYS_UNTIL_NEUTRAL_CLICK`). From here, in a loop after user clicks on `No thanks`or `Rate now`. 

@harshitbansal05 @davigonz

davigonz:
> If i understood correctly:
> 
> - By default, the dialog will never raise before 2 days after installing. (managed by DAYS_UNTIL_PROMPT)
> - After the second day, the dialog will raise if the number of launches is, at least, 2. If not, after the second launch, whenever it happens (managed by LAUNCHES_UNTIL_PROMPT)
> - With neutral button (Later option), the dialog will raise the following day (managed by DAYS_UNTIL_NEUTRAL_CLICK). From here, in a loop after user clicks on No thanks or Rate now.

Yes, think so, @harshitbansal05 is that correct?

harshitbansal05:
yes, @jesmrec @davigonz...

jesmrec:
thanks @harshitbansal05 i will take in account these tips for the test plan

jesmrec:
Here i'm having some problems to reproduce some scenarios. The AppRater is called in the onCreate() method of FileDisplayActivity. This method is called when the app is opened for the first time and then, after log in to show the list of files. I was not able to get the AppRater code from this moment. Killing the app, reopening it... all cases the file list is shown and onCreate() and AppRater are not invoked. 

Any ideas? @davigonz @harshitbansal05 @theScrabi

jesmrec:
First QA round finished and all worked fine!! i will review again and if all is OK, ready to move forward.

The only issue to think about is the expected behaviour for beta versions. Do we expect beta testers to get this dialog? maybe it should be enabled only in `release`builds.

jesmrec:
@harshitbansal05 if you can have a look to https://github.com/owncloud/android/pull/2102#pullrequestreview-100779423

we will have the feature finished, once we have decided including/excluding this in betas.

davigonz:
> we will have the feature finished, once we have decided including/excluding this in betas.

@jesmrec 

Disabling the feature in beta versions is the best approach in my opinion since we won't have a separate beta app from now on and therefore, no more beta reviews. 

What about including that check? @harshitbansal05

davigonz:
Although thinking a bit more about it, beta testers might be interested on reviewing the release app as well, what's your opinion? @jesmrec @michaelstingl @harshitbansal05

jesmrec:
A beta version has testing purposes, and bugs/improvements can be suggested by beta testers to improve the release app. It is not the aim of a beta to rate the release build. The release app is the one to be rated, i mean.

michaelstingl:
@harshitbansal05 @davigonz Will this be shown to F-Droid users? They maybe don't have a Play Store account at all…

theScrabi:
> Will this be shown to F-Droid users? They maybe don't have a Play Store account at all…

Some have, and I've seen several apps on fdroid doing this, so we don't necessarily remove it for fdroid builds.

michaelstingl:
@theScrabi okay, thanks for clarification.

jesmrec:
In this case, this is ready to rebase + merge

davigonz:
Hi @harshitbansal05, I've seen that you are agree with the previous comment, does it mean that you are working on this feature? We would like to include it in 2.7.0 release, if you are busy and cannot move this feature forward, we can take care of it. 

Indeed, we have already created a new PR to include your contribution: https://github.com/owncloud/android/pull/2156 , do you agree or do you want us to wait a bit for it?

davigonz:
@harshitbansal05 thanks for this contribution, your code will be merged via https://github.com/owncloud/android/pull/2156 and included in 2.7.0

PR #2110, opened 2018-02-06
FileObserverService: use startForegroundService on Oreo and later

invisiblek:
* In API 26 and higher, startForegroundService needs to be used. This
  fixes the following crash upon device boot:

AndroidRuntime: java.lang.RuntimeException: Unable to start receiver com.owncloud.android.broadcastreceivers.BootupBroadcastReceiver: java.lang.IllegalStateException: Not allowed to start service Intent
{ act=com.owncloud.android.services.observer.FileObserverService.action.START_OBSERVATION cmp=com.owncloud.android/.services.observer.FileObserverService }:
app is in background uid UidRecord{783d156 u0a93 RCVR idle change:uncached procs:1 seq(0,0,0)}

* Reference: https://developer.android.com/about/versions/oreo/background.html#migration

### DIFF
--- a/src/com/owncloud/android/services/observer/FileObserverService.java
+++ b/src/com/owncloud/android/services/observer/FileObserverService.java
@@ -33,6 +33,7 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.os.Build;
 import android.os.IBinder;
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v4.util.Pair;
@@ -99,7 +100,11 @@ public class FileObserverService extends Service {
     public static void initialize(Context context) {
         Intent i = new Intent(context, FileObserverService.class);
         i.setAction(ACTION_START_OBSERVE);
-        context.startService(i);
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.N_MR1) {
+            context.startForegroundService(i);
+        } else {
+            context.startService(i);
+        }
     }
 
     /**


src/com/owncloud/android/services/observer/FileObserverService.java
```diff
@@ -99,7 +100,11 @@
     public static void initialize(Context context) {
         Intent i = new Intent(context, FileObserverService.class);
         i.setAction(ACTION_START_OBSERVE);
-        context.startService(i);
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.N_MR1) {
```
davigonz: The problem you report appears in In API 26 and higher, but N_MR1 is API 25, it should be O, which is API 26

davigonz:
Thanks @invisiblek for reporting this bug, I've just requested some changes

src/com/owncloud/android/services/observer/FileObserverService.java
davigonz: Ups, forget my comment, just seen you are using > instead of >=

davigonz:
@invisiblek Please rebase with master and @jesmrec will start with a bit of QA

invisiblek:
Closing this PR for now, this isn't the right fix. Per the documentation I linked above, I think we want to use a job scheduler for this. Using startForegroundService requires you to call Service.startForeground() or you'll hit an ANR. More info: https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground

The problem with that, is you need to show a notification for Service.startForeground(), which is probably not something that makes sense for this.

davigonz:
Ok @invisiblek , no problem, we will wait to see your final fix. Meanwhile, you could create a new issue in our repo, where we can discuss about it, and we will connect the new PR with it when is ready.

PR #2117, opened 2018-02-12
Use proper name for the Upload from camera feature

davigonz:
According to the final design of new [Upload from camera](https://github.com/owncloud/android/issues/2027) feature, the name to be included in the bottom sheet should be "Picture from camera" instead of "Upload from camera" since the _upload_ word already appears in the bottom sheet title.

upload bottom sheets


davigonz:
@jesmrec can you please review this PR? It's a string resource change, no extra QA needed.

jesmrec:
Review approved. Good point.

Issue #702, opened 2014-11-02
Support animated GiF-Images


tobiasKaminsky:
Add functionality to show animated gif files in slideshow (image preview)
___

TASKS

- [X] [GIT] Create branch gif_support.
- [x] [RESEARCH] Search a gif library. 
- [x] [DEV] Include library dependency, adding the required attributions.
- [x] [DEV] Implement GIF support.
- [X] [QA] Design test plan @jesmrec
- [X] [QA] Execute test plan @jesmrec 
- [ ] [GIT] Merge branch gif_support into master.

tobiasKaminsky:
see #1111 "Schnappszahl " ;)

jancborchardt:
Animated GIF support is quite a high priority! :D (Not joking)

@tobiasKaminsky @Kernald @LukeOwncloud @AndyScherzinger anyone of you want to take a stab at this, or know someone who wants to join development and would like to take this on? :)

Kernald:
Doesn't #1111 fixes it already?

jancborchardt:
Haha, how did I miss that? :D In that case can you folks please review #1111?

sd1998:
@davigonz  Is this issue fixed? If not I would like to work on it.

jesmrec:
To be included in 2.7.0

PR #2123, opened 2018-02-22
Gif support

davigonz:
Implements #702


davigonz:
@jesmrec I've replaced the library to make zoom in the pictures with a new one (Photoview) and use also a new library (Glide) to reproduce the gifs and show the images, I tell you in case you need to include more tests.

jesmrec:
Approved

build.gradle
```diff
@@ -40,7 +44,6 @@ configurations.all {
 
 dependencies {
     /// dependencies for app building
-    compile project(':touchImageView')
```
hannesa2: Btw, I was annoyed by no maintenance of this lib. But after going my own way and having a vital fork, Mike Ortiz made me to a maintainer and as you can see, there is now some life there https://github.com/MikeOrtiz/TouchImageView/graphs/contributors,
but anyway, if you are happy with `com.github.chrisbanes:PhotoView` it's perfect !
@michaelstingl @davigonz

davigonz: We decided to use PhotoView because, as you can see in the files changed tab, we have pretty less code.

hannesa2: It's a library, and in library it's one **one** relevant file.

Issue #2116, opened 2018-02-12
ownCloud notification always visible in Android 8


davigonz:
### Actual behaviour
- ownCloud app shows notification with message "ownCloud is using battery" all the time.

<img src="https://user-images.githubusercontent.com/15261363/36089114-81de5d8c-0fda-11e8-9685-2f8a83fa67ec.png" width="70%">

### Expected behaviour
- ownCloud app notification should be shown when it's needed and disappear after that.
 
### Steps to reproduce
1. Open the app
2. Press home button
3. Have a look at the Android top bar. 

Can this problem be reproduced with the official owncloud server? Yes
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Android version: 8.1.0

Device model: Nexus 6P

Stock or customized system: Stock

ownCloud app version: 2.6.0

ownCloud server version: it doesn't matter

PR #2125, opened 2018-02-26
Fix ownCloud notification always visible in Android 8

davigonz:
Solves #2116


jesmrec:
Fixed, notification disappears once the work in background finishes.

PR #2140, opened 2018-03-09
Use proper color for the icons in bottom sheets menu

davigonz:
According to the design of new bottom sheets menu, the icons to upload files or pictures from camera should be grey instead of black.

upload bottom sheets


davigonz:
@jesmrec just updating some icons and colors, no extra QA needed.

davigonz:
@theScrabi can you please review this code? It will be quick, just icons replacement.

Issue #2141, opened 2018-03-09
Upload pictures directly from devices camera, uploads the thumbnail of the image.


sd1998:
### Actual behaviour
When using the upload from camera feature the thumbnail of the picture is being uploaded.

### Expected behaviour
The actual picture file should be uploaded instead of the thumbnail.
 
### Steps to reproduce
1. Open the main FAB.
2. Select  Upload FAB
3. Select "Picture from camera" option in the Bottom Sheet

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes

### Environment data
Android version: 8.0.0

Device model:  Oneplus 3T

Stock or customized system:

ownCloud app version: 2.6.0

sd1998:
@davigonz while I working on this feature I did not notice this mistake. But I just saw it when I ws working on #2119. And I have fixed it will be sending a PR in some time.

davigonz:
Great @sd1998, thanks for reporting it, we will include it in the next 2.7 release

PR #2142, opened 2018-03-09
Fix to Issue #2141

sd1998:
Issue #2141 fixed. Now the file of the picture taken from the device's camera is uploaded. Instead of their thumbnails.

### DIFF
--- a/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -251,7 +251,7 @@ public class UploadFilesActivity extends FileActivity implements
         } else if(requestCode == REQUEST_IMAGE_CAPTURE && resultCode == RESULT_CANCELED){
             if(image != null){
                 image.delete();
-                Log_OC.d(TAG,getResources().getString(R.string.file_deleted));
+                Log_OC.d(TAG,"File deleted");
             }
             setResult(RESULT_CANCELED);
             finish();
--- a/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -2,7 +2,8 @@
  *   ownCloud Android client application
  *
  *   @author David A. Velasco
- *   Copyright (C) 2016 ownCloud GmbH.
+ *   @author Shashvat Kedia
+ *   Copyright (C) 2018 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License version 2,
--- a/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -89,6 +89,7 @@ public class UploadFilesActivity extends FileActivity implements
     private Account mAccountOnCreation;
     private DialogFragment mCurrentDialog;
     private String capturedPhotoPath;
+    private File image = null;
 
     private RadioButton mRadioBtnCopyFiles;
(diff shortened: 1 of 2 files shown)


sd1998:
@davigonz @jesmrec I have fixed it.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -194,11 +197,39 @@ public static void startUploadActivityForResult(Activity activity, Account accou
...
+            Uri photoUri = FileProvider.getUriForFile(getApplicationContext(),
+                    getResources().getString(R.string.file_provider_authority),photoFile);
+            pictureIntent.putExtra(MediaStore.EXTRA_OUTPUT,photoUri);
+        }
         if (pictureIntent.resolveActivity(getPackageManager()) != null) {
             startActivityForResult(pictureIntent, REQUEST_IMAGE_CAPTURE);
         }
     }
 
+    private File createImageFile(){
```
davigonz: I guess that with this method you are creating the image file before getting it from the camera, aren't you? Do you really need it or you can just create it once you have the picture from the camera?

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -34,6 +34,7 @@
 import android.preference.PreferenceManager;
```
davigonz: Please update the comments in the bottom of the activity by including your name as @author and updating the year to 2018.

davigonz:
@sd1998 I requested a minor change and included a question.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
sd1998: @davigonz That is needed so that we can set the address to which the picture will be saved once it has been taken.

sd1998:
@davigonz I have made the change.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
sd1998: I can another condition in the onActivityResult() to check if the resultCode was RESULT_CANCELED if it is true then I will delete the file that I have created.

davigonz: Yes, if you are keeping the recently created image when taking picture is cancelled, you should definitely delete the unneeded file.

sd1998: @davigonz I have done that change.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -248,8 +248,14 @@ private String getCapturedImageName(){
     protected void onActivityResult(int requestCode,int resultCode,Intent capturedData){
         if(requestCode == REQUEST_IMAGE_CAPTURE && resultCode == RESULT_OK){
             new CheckAvailableSpaceTask(getCapturedImageFile().getAbsolutePath()).execute();
-        }
-        else if(resultCode == RESULT_CANCELED){
+        } else if(requestCode == REQUEST_IMAGE_CAPTURE && resultCode == RESULT_CANCELED){
+            if(image != null){
+                image.delete();
+                Log_OC.d(TAG,getResources().getString(R.string.file_deleted));
```
davigonz: We do not need to create string resources (that will be automatically translated) to log something, you can include the string itself here.

sd1998:

davigonz:
> I have made the changes. 

Thanks @sd1998 , code approved, ready to pass QA @jesmrec

jesmrec:
Fixed!! thanks @sd1998

Issue #2119, opened 2018-02-19
As an user, I would like to take a picture of my handwritten notes or printed documents, scan them and upload it directly to my ownCloud


davigonz:
The current ownCloud Android app has different options to upload files to an ownCloud server, such as selecting them from the internal storage, selecting content from other apps or uploading photos directly from the camera.

It would be great to include a new option which enables the user to take a new picture with the camera and select the edges of handwritten notes or printed documents. After that, the image will be processed, adjusted to a better view and uploaded to a folder on the server.

sd1998:
@davigonz I would like to work on this.

davigonz:
Great @sd1998 , glad to hear you want to contribute with this. Just for you to know, there's a bounty for implementing this feature.

sd1998:
@davigonz Where can I get the information about the bounty for this feature?

michaelstingl:
@sd1998 https://www.bountysource.com/issues/55217024

sd1998:
@michaelstingl  Thanks.

sd1998:
@davigonz  For this feature should I use a 3rd party dependency or should I try to implement the feature from scratch using opencv etc.?

Unsigno:
Hi guys , when you talk about "the image will be processed", do you mean using OCR to get plain text or adjust the marked edges to get a rect and aligned document ?

davigonz:
> Hi guys , when you talk about "the image will be processed", do you mean using OCR to get plain text or adjust the marked edges to get a rect and aligned document ?

Actually Is not OCR itself, is just about taking a picture and adding some kind of filters so that the picture seems a scanned document.

Unsigno:
Oki , i am looking for a PR and have some questions.

1 - Could be added OpenCV as dependency to do this ?
2 - If i send a PR, it has to contain all the translations of the texts used ?

sd1998:
@Unsigno I am already working on this. Could you work on any other issue? Thanks.

Unsigno:
@sd1998  You could mark it on bountysource as started and give an aproximated time to prevent bounty hunters :p . give some news if you leave it please .

If some one can asnwer 2 i still curious .

davigonz:
> 1 - Could be added OpenCV as dependency to do this ?
> 2 - If i send a PR, it has to contain all the translations of the texts used ?

@Unsigno 

1. Yes, OpenCV could be used to detect the document edges for example.
[...]

Issue #1727, opened 2016-06-30
"Select all" files option in multiselection mode


jabarros:
In #1176 multi select feature was developed and to avoid wasting time selecting all files in a folder one by one, an option to select all the files in a folder is necessary. This will let us select all files in a unique click.

davivel:
An option in the overflow menu should be enough.

sd1998:
Is anyone working on this?

davigonz:
> Is anyone working on this?

@sd1998 No, contributions are welcome 😉

sd1998:
@davigonz I would like to work on this.

jesmrec:
@sd1998 for me it is ok, go on!! first of all, we should agree the way it is going to be displayed. I think in several options:

- Set a global checkbox in multiselection mode. By selecting it, all checkboxes will be selected 
- Set an option in tool bar or three-dot-menu to select all the stuff. 

Also, it would be cool to include an additional option to "deselect" all the stuff

If you are going to work on this, maybe is a good idea to create mock-ups before starting to be sure if all are on the same page, and we agree in the UI.

sd1998:
@jesmrec I will add the mockups here is some time. I think the second options will be better and I think that the option in three-dot-menu should only be available if the user has already selected an item(i.e is already in multiselection mode).

sd1998:
@jesmrec I have made the mockup. The mockup is not very good as I still learning to make these so Sorry for the inaccuracy to replicate the UI completely but the mockup still represents vaguely the option in tree-dot menu to select all the items and deselect all the items. Is this type of representation of the options ok?

![image](https://user-images.githubusercontent.com/22047421/41172486-a2b2282a-6b71-11e8-829c-afbca308b2a7.png)

sd1998:
@jesmrec Is [this](https://github.com/owncloud/android/issues/1727#issuecomment-395836030) alright? If it is then I will start working on it.

jesmrec:
looks good for me. Are he options "Select all" and "Deselect all" always in the three-dot-menu, and not in the toolbar (although there is room for this)? 

CC @davigonz @michaelstingl

davigonz:
Thanks @sd1998 for the mockup, I have a couple of questions:
1. Does the option **select all** appear when the user has already selected one or more files or when there's no one selected? In my opinion it could appear in both cases.
2. Does the option **deselect all** make sense when there's no file selected? Think it should appear just when the **select all** option is being used.

What do you think?

[...]

PR #2143, opened 2018-03-10
Documents can be scanned using the device's camera and then uploaded

sd1998:
Issue #2119 Fixed. Now a user can scan a document from the owncloud app and upload it.


sd1998:
@davigonz I have implemented the feature.

davigonz:
> I have implemented the feature.

👏 Thanks @sd1998 for implementing this, I will have a look at the code and ping you with the requested changes, if any.

sd1998:
@davigonz Have you started the code review?

davigonz:
> When will the code review start?

@sd1998 we are currently focused on releasing 2.7.0 beta, we will take care of this after that.

davigonz:
@sd1998 I've just started to have a look at your code, do we really need to compile the ocv project and include it as a separate module? Isn't possible to use just the classes or methods we need without importing all the ocv code?

sd1998:
@davigonz the standard way is to compile the ocv project. Not using the ocv project would require the user to download the opencv manager app to use this feature. Thus to avoid that I used the project.

sd1998:
@davigonz has the code review started?

theScrabi:
@sd1998 davidgonz is not available for the next two weeks, he will review it then.

jesmrec:
@sd1998 is the development finished?

sd1998:
@jesmrec Yes.

davigonz:
Resuming the work here, let's continue with the code review

davigonz:
@sd1998 could you please rebase with master so that I can review the code with the latest changes? Thanks

sd1998:
@davigonz I have done the rebase.

davigonz:
I have cloned your code @sd1998 to give it a try and I have noticed that the edges of a document are not being properly detected, that is the main purpose of opencv, isn't it?

| Current behaviour  | Expected behaviour |
| ------------- | ------------- |
| bad | good |

sd1998:
@davigonz When I tested the feature this was not the case. I think I will have to further optimize the feature I will work on this.

davigonz:
> When I tested the feature this was not the case. I think I will have to further optimize the feature I will work on this. 

Ok, thanks, I will continue with the code review meanwhile

sd1998:
@davigonz I have found another bug in the cancel button I will fix that too.

src/com/owncloud/android/ui/activity/CropActivity.java
```diff
@@ -0,0 +1,406 @@
...
+        List<PointF> points = findPoints();
+        if(points != null){
+            mSelectionImageView.setPoints(points);
+        } else{
+            Log_OC.d(TAG,getResources().getString(R.string.no_contour_found));
+        }
+        mCropButton.setOnClickListener(this);
+        mCancelButton.setOnClickListener(this);
+        mResultDialog = new MaterialDialog.Builder(this)
+                .title("Scan Result")
```
davigonz: Please use here string resources

src/com/owncloud/android/ui/activity/CropActivity.java
```diff
@@ -0,0 +1,406 @@
...
+                .positiveText("Save")
```
davigonz: Same

src/com/owncloud/android/ui/activity/CropActivity.java
```diff
@@ -0,0 +1,406 @@
...
+                .negativeText("Cancel")
```
davigonz: Same

src/com/owncloud/android/ui/activity/CropActivity.java
```diff
@@ -0,0 +1,406 @@
...
+                .customView(R.layout.dialog_document_scan_result, false)
+                .onPositive(new MaterialDialog.SingleButtonCallback() {
+                    @Override
+                    public void onClick(@NonNull MaterialDialog materialDialog, @NonNull DialogAction dialogAction) {
+                        // TODO Saving
```
davigonz: Delete this comment if is already implemented

settings.gradle
```diff
@@ -1,3 +1,5 @@
 include ':owncloud-android-library'
 include ':'
 include ':oc_jb_workaround'
+include ':touchImageView'
```
davigonz: Are you using this library? It was deleted when GIF support was implemented [here](https://github.com/owncloud/android/pull/2123/files#diff-c197962302397baf3a4cc36463dce5eaL43)

davigonz:
@sd1998 performed the first code review, I will continue after applying the requested changes.

theScrabi:
@sd1998 have you checked this library here?
https://github.com/Credntia/CVScanner

sd1998:
@theScrabi  No I did not I will check it now. Thanks.

sd1998:
@theScrabi The distinction between the edge of the paper and other edges is also being done using OpenCV. Currently, I am looking into CVScanner and I think using that would be a better option. I will try implementing the feature using that.

sd1998:
@davigonz I have made the changes. And I removed all the other architectures other than the ones mentioned in [Fix-2119-comment](https://github.com/owncloud/android/pull/2143#pullrequestreview-111198993) this has reduced the size of APK to 69.95 MB. To further decrease the APK size I can make different APK for different architectures (as current I have made a single APK which supports the 4 architectures mentioned in [Fix-2119-comment](https://github.com/owncloud/android/pull/2143#pullrequestreview-111198993)) this will further reduce the size of the APK. Should I do it?

sd1998:
@davigonz Ok I will work on that then.

sd1998:
@davigonz If I use  https://github.com/Credntia/CVScanner  then the feature will work only for devices with API level 19 or above, on the other hand, my implementation supports devices with API level 14 hence should I still use CVScanner or continue with my implementation?

davigonz:
> If I use https://github.com/Credntia/CVScanner then the feature will work only for devices with API level 19 or above, on the other hand, my implementation supports devices with API level 14 hence should I still use CVScanner or continue with my implementation?

@sd1998 API 19, a.k.a Android 4.4 or Kit Kat was released 5 years ago which is enough time in my opinion and according to the latest data collected by Google, Android 4.3 and above represent less than 5%  of Android devices. Therefore, I would continue with your implementation using CVScanner

screen shot 2018-05-17 at 08 43 38

Any other opinion? @jesmrec @theScrabi

jesmrec:
agreed. Set the minimum API as 19 for this feature is absolutely enough.

sd1998:
@davigonz I have implemented the feature using CVScanner even though using this has increased the accuracy of the feature but I still have to use OpenCV as a separate module as the gradle dependency used in the CVScanner project is not viable anymore. And so to decrease the size of the APK I can remove all the classes form the OpenCV module that are not being used. This along with making separate APK's for different architectures can reduce the size of the APK significantly.

sd1998:
@davigonz Please test the feature now (it has been re-implemented using CVScanner).

davigonz:
Hi @sd1998 , I am little worried about the new two modules included, `cvscanner` and `openCVLibrary340`, are you modifying any class of those modules or are you just using them as they are? If we don't need to modify those libraries, it would be great to include the dependencies to those libraries without adding submodules to the app.

To do so, we can use Jitpack and copy the dependency in our build.gradle, you have documentation about it [here](https://jitpack.io/docs/). I'm trying to do it following the next steps: 

1. Go to [https://jitpack.io/](https://jitpack.io/)
2. Write Credntia/CVScanner in `Git repo url` field and press look up. 
3. Several commits we can use appear but there's an error 

screen_shot_2018-05-21_at_10_42_47

[Here](https://jitpack.io/com/github/Credntia/CVScanner/9c1126bd2c/build.log) you have the link of the error. It seems that CVScanner doesn't build properly, maybe because of some changes performed there, I haven't managed to build it locally either, have you? 

We would need CVScanner itself building properly when getting it from the original repository, with no modifications so we can include it as a dependency and get rid of submodules. That would ease our development effort for sure  😀

sd1998:
@davigonz I am able to build the cvscanner library successfully (there is some fault in the file structure of the library on making some changes and defining a settings.gradle file I was able to successfully build the library) to add this as a grade dependency I am making a sperate repository and pushing the code for the library (the one which is building successfully) there and then will follow these [steps](https://github.com/owncloud/android/pull/2143#issuecomment-390593874) to add this as a gradle dependency. I think this would be a decent plan of action what do you think?

davigonz:
> I am making a separate repository and pushing the code for the library

@sd1998 I would have sent a PR to fix the problem to the original library but since it seems the owner is currently not too active (having a look at his issues), forking is not a bad idea.

> I have removed both the submodules and added the following as gradle dependency.

For me is ok but instead of using `-SNAPSHOT` I would use a specific commit or even a tag since `SNAPSHOT` will automatically build the latest commit on the master branch and that commit might not work for any reason and break the document scanner.

sd1998:
@davigonz Alright I will make the changes.

sd1998:
@davigonz I have changed the dependency to point to the last commit (the library is building successfully up to this commit) that I have done to the repository.

sd1998:
@davigonz Have you tested the feature?

davigonz:
> Have you tested the feature?

Ok @sd1998 , but can you first rebase with master? You might have to update some naming according to the latest version of Gradle we included.

sd1998:
@davigonz I have done the rebase.

sd1998:
@davigonz Has the code review started?

davigonz:
> Has the code review started?

@sd1998 I've just cloned your repository to have a look at the code but I get this error when trying to build the project:

screen shot 2018-05-28 at 10 04 15

Any idea? This error doesn't appear when building your master branch.

sd1998:
@davigonz I don't understand why this is happening I am having no issues in building the Fix-2119 branch. I will look into it further.

davigonz:
> I don't understand why this is happening I am having no issues in building the Fix-2119 branch. I will look into it further.

It seems there's some errors in jcenter, where we get some dependencies from so let's wait a bit...

sd1998:
@davigonz Did it start working?

davigonz:
> Did it start working?

Hi @sd1998 , @theScrabi is having a look at it, he will include here more information

jesmrec:
I've put both answers together, what do you think?

Hi @sd1998, thank you for your effort in this contribution, we have tested your most recent changes but the feature is not usable enough from an end user point of view since is able to detect papers in perfect lighting conditions but that’s not the most likely scenario. When the camera is a bit tilted, the scanner is not able to detect the papers properly, the same as with poor lighting conditions or shadows and we can’t release this feature without making it work in most cases.

I still think that this is a nice feature that should be included in our app but is not sort of trivial and it could take a long time in terms of reviewing or testing and we can not pay too much attention to it in this moment so we have to stop. Anyway, you can continue working on it and maybe you manage to have a good solution by directly contributing CVScanner library o researching about image processing as long as you are interested in it of course. Next year ownCloud will apply for GSoC again, and this will be a new choice to work in this feature as a formal project. Other incoming features (like https://github.com/owncloud/android/issues/1727) are good choices to contribute in ownCloud.

sd1998:
@jesmrec All right then I will be closing this PR.

hannesa2:
Thank you @sd1998 ! This cool pull request is very inspiring.

But on my side I only capture such images
image

image

Do you have any ideas what's wrong here ?

Btw, it would be very helpful, when you restore your branch. Thank you in advance !

jesmrec:
we'd like to have such feature, but the first try did not work as we expected... maybe the 2.0 will work better.

vibhuti019:
@hannesa2 can we implement something like the image to text translate so that the handwritten image is translated to word files immediately. I think that would be an awesome feature..

Issue #2066, opened 2017-11-27
Adapt "public link" permissions to oC server 10.0.4 UI


michaelstingl:
UI for "public link" permissions has been improved with oC server 10.0.4:
https://github.com/owncloud/core/pull/29376

![bildschirmfoto vom 2017-10-27 16-30-06](https://user-images.githubusercontent.com/12717530/32109172-3460e192-bb34-11e7-9f91-6f02710c95dc.png)

@davigonz What is the effort to bring this to the Android client?

davigonz:
@michaelstingl This is for public link, not for private links, isn't it? 

I really like the new distribution of sharing permissions in core, it is much clearer for the user.

> What is the effort to bring this to the Android client? 

Well, in terms of UI, we just have to replace the two switch buttons (Allow editing & Show file listing) with a new dropdown menu, or something like that, which includes the new three options. I will create a mockup.

In terms of logic, it shouldn't take too much because we just have to move the same functionality from two buttons to a list with three options. 

So, to sum up, I would say that is easy to implement, one or two days at most.

theScrabi:
This is how it will look like
![screenshot_1521037265](https://user-images.githubusercontent.com/1891273/37408233-ef01c11e-279b-11e8-927a-12b2e4ae9213.png)

jesmrec:
The links to files has also incuded the "Read Only" option:

![screen shot 2018-05-03 at 14 54 18](https://user-images.githubusercontent.com/14894746/39577565-fbc901b0-4ee1-11e8-9c4a-bd0c87a72da7.png)

will we include in the scope of this issue, or create a separate one? @theScrabi @davigonz

theScrabi:
Is the readonly a radio button here or a check box?

theScrabi:
I think this should be in scope of this. However I think this option is not working on server side because it looks like a radio button and it behaves like one, tho it should be a checkbox. This is why clicking on it has no action.

![gurken](https://user-images.githubusercontent.com/1891273/39624866-2d5cd4c2-4f9b-11e8-88c2-277c69f78497.gif)

However maybe it's not intended to make a single file share writable.

jesmrec:
let's check with core to take a decision.

pmaier1:
> However I think this option is not working on server side because it looks like a radio button and it behaves like one, tho it should be a checkbox

[...]

PR #2150, opened 2018-03-14
make permission selection for link sharing a radio group

theScrabi:
Implements https://github.com/owncloud/android/issues/2066

____

BUGS & IMPROVEMENTS

- [x] (1) Edit options show `Read Only`always https://github.com/owncloud/android/pull/2150#issuecomment-386292171 [FIXED]
- [x] (2) publicUpload permission https://github.com/owncloud/android/pull/2150#issuecomment-395373196
- [X] (3) Edition of public link broken by name with ' &'  https://github.com/owncloud/android/pull/2150#issuecomment-399404634 -> [WONT FIX HERE]

### DIFF
--- a/src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
+++ b/src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
@@ -49,7 +49,6 @@ import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.shares.OCShare;
 import com.owncloud.android.lib.resources.status.OCCapability;
 import com.owncloud.android.lib.resources.status.OwnCloudVersion;
-import com.owncloud.android.operations.UpdateSharePermissionsOperation;
 import com.owncloud.android.ui.activity.FileActivity;
 import com.owncloud.android.ui.dialog.ExpirationDatePickerDialogFragment;
 import com.owncloud.android.utils.DateUtils;
@@ -69,8 +68,6 @@ public class PublicShareDialogFragment extends DialogFragment {
     private static final String ARG_SHARE = "SHARE";
     private static final String ARG_ACCOUNT = "ACCOUNT";
     private static final String ARG_DEFAULT_LINK_NAME = "DEFAULT_LINK_NAME";
-    private static final int CREATE_PERMISSION = OCShare.CREATE_PERMISSION_FLAG;
-    private static final int UPDATE_PERMISSION = OCShare.UPDATE_PERMISSION_FLAG;
     private static final String KEY_EXPIRATION_DATE = "EXPIRATION_DATE";
 
     /**
@@ -246,7 +243,7 @@ public class PublicShareDialogFragment extends DialogFragment {
                         | OCShare.READ_PERMISSION_FLAG):
                     mReadWriteButton.setChecked(true);
                     break;
-                case CREATE_PERMISSION:
+                case OCShare.CREATE_PERMISSION_FLAG:
                     mUploadOnlyButton.setChecked(true);
                     break;
                 default:
@@ -279,10 +276,10 @@ public class PublicShareDialogFragment extends DialogFragment {
         initPasswordToggleListener();
 
         view.findViewById(R.id.saveButton)
-                .setOnClickListener(v -> dismiss());
+                .setOnClickListener(v -> onSaveShareSetting());
 
         view.findViewById(R.id.cancelButton)
-                .setOnClickListener(v -> onSaveShareSetting());
+                .setOnClickListener(v -> dismiss());
 
         return view;
     }
--- a/src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
+++ b/src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
@@ -36,7 +36,6 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.WindowManager;
 import android.view.inputmethod.InputMethodManager;
-import android.widget.Button;
 import android.widget.CompoundButton;
 import android.widget.EditText;
 import android.widget.LinearLayout;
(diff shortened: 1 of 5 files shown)


src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -281,24 +263,13 @@ public void onClick(View v) {
...
-
-                    // If edit permissions are checked, the server will set the read permission
-                    // but we need create permissions at this point
-                    publicLinkEditPermissions = false;
-
+                if(publicLinkUploadFilePermissionGranded(getView())) {
                     publicLinkPermissions = OCShare.CREATE_PERMISSION_FLAG;
                 }
 
+
```
davigonz: Too much space here

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -822,4 +733,14 @@ private TextView getExpirationDateExplanation(View view) {
     private TextView getErrorMessage() {
         return (TextView) getView().findViewById(R.id.public_link_error_message);
     }
+
+    private boolean publicLinkEditPermissionGranded(View view) {
+        return getPermissionsRadioGroup(view).getCheckedRadioButtonId()
+                == R.id.shareViaLinkEditPermissionReadAndWrite;
+    }
+
+    public boolean publicLinkUploadFilePermissionGranded(View view) {
```
davigonz: This method could be private

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -35,6 +35,8 @@
 import android.widget.CompoundButton;
```
davigonz: Year and author

davigonz:
@theScrabi some changes requested

davigonz:
@theScrabi thanks for applying the requested changes but there's some conflicts in PublicShareDialogFragment.

jesmrec:
### (1)

1. Create a Public Link with the option `Read & Write` (`Upload Only` also valid to reproduce)
2. Save it
3. Tap on the `>` to edit the link

Current: `Read Only` (by default)
Expected: The option selected in creation  (in `permissions` property from GET request)

Tested with Nexus 6P v7.1.2

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -239,7 +240,10 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
...
 
             switch (mPublicShare.getPermissions()) {
-                case (UPDATE_PERMISSION | CREATE_PERMISSION):
+                case (OCShare.CREATE_PERMISSION_FLAG
+                        | OCShare.DELETE_PERMISSION_FLAG
+                        | OCShare.UPDATE_PERMISSION_FLAG
+                        | OCShare.READ_PERMISSION_FLAG):
                     mReadWriteButton.setChecked(true);
                     break;
                 case CREATE_PERMISSION:
```
davigonz: Please use `OCShare.CREATE_PERMISSION_FLAG` as you did in the previous case. Besides, delete the constants `CREATE_PERMISSION` and `UPDATE_PERMISSION` defined at the top of the class since you are using directly `OCShare.whatever`

davigonz:
This is one of our priorities for the next 2.8 release, @theScrabi can you have a look at the requested changes and conflicting files? #2195 is already merged, so no blocker anymore.

theScrabi:
done

davigonz:
Great, ready to test then @jesmrec

jesmrec:
Core issue: https://github.com/owncloud/core/issues/31676

jesmrec:
### (2)

The core issue mentioned before was caused by the inclusion of the parameter `publicUpload` in the POST request performed in the client. Such parameter is not necessary, because the permissions are handled by the `permissions`

theScrabi:
should be fixed. @jesmrec pls take a look.

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -313,13 +314,23 @@ private void onSaveShareSetting() {
                 break;
         }
 
+        // since the public link permission foo got a bit despagetified in the server somewhere
+        // at 10.0.4 we don't need publicUploadPermission there anymore. By setting it to false
+        // it will not be send to the server.
```
davigonz: it will not be **sent** to the server

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
```diff
@@ -45,6 +45,7 @@
 
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.OCFile;
+import com.owncloud.android.lib.common.OwnCloudAccount;
```
davigonz: This dependency is not needed

src/com/owncloud/android/ui/fragment/PublicShareDialogFragment.java
theScrabi: fixed

theScrabi:
done

jesmrec:
### (3) 

Steps:

1. Create a public link which name contains &
2. Edit the public link (no necessary to change any field)

Current: & is taken as a parameter separator, so the request is partially broken
Expected: & character escaped.

jesmrec:
bug (3) also reproducible in 2.7.0, so it will be handled in a different issue

theScrabi:
> bug (3) also reproducible in 2.7.0, so it will be handled in a different issue

also refactoring in the library fixed that issues.

jesmrec:
All stuff passed, the remaining bug will be fixing within next version

PR #2154, opened 2018-03-19
fix exception during animation

theScrabi:
This will introduce a (ugly) fix for the animation but that triggers an exception if the sub folder contains more folders.

- [X] Browse to folders with subfolders [FIXED]
- [x] Browse back from folder which contains files and folders https://github.com/owncloud/android/pull/2154#issuecomment-374197004

### DIFF
--- a/res/anim/dir_fadein_back.xml
+++ b/res/anim/dir_fadein_back.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <scale
-        android:duration="200"
+        android:duration="@integer/folder_animation_duration"
         android:fromXScale="0.95"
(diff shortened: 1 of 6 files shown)


jesmrec:
Subfolders can be browsed, but check this one:

1. Create a folder structure like that one: A/B/C
2. Browse  to C folder, and create another folder D and upload a file. 
3. Browse back  to C

Current: app crashes
Expected: browsing back is correct.

Tested with Nexus 6P v7

theScrabi:
>Current: app crashes
Expected: browsing back is correct.

Thst's basically the same error just in reverse order.

Provided fix.

jesmrec:
Fixed.

davigonz:
@theScrabi please next time wait for my code review as well 😄

theScrabi:
Oh two times the same error :/ I will definitely take more care :)

Issue #2036, opened 2017-09-22
[OAuth2] Improve trust/security when login in embedded web view


michaelstingl:
As discussed at the [ownCloud Conference 2017](https://occon17.owncloud.org/), there some best practice recommendations to improve trust and security when user login in the embedded web view.

This is is an article from Carnegy Mellon CERT that describes the motivation:
https://insights.sei.cmu.edu/cert/2016/08/the-risks-of-google-sign-in-on-ios-devices.html

Another article describes possible solutions with a contribution from Google:
https://www.pingidentity.com/en/blog/2016/03/10/using_appauth_to_enable_your_apps_with_mobile_sso.html

There is also a video recording available from the Google Team:
https://youtu.be/DdQTXrk6YTk

You will find very detailed information in a new IETF draft from OAuth Working Group:
~~https://tools.ietf.org/html/draft-ietf-oauth-native-apps (June 9, 2017)~~
https://tools.ietf.org/html/rfc8252 (October 2017)

@davivel @davigonz Could you check how the ownCloud Android client could be improved following the linked recommendation?

Related: https://github.com/owncloud/ios/issues/942

00008274

davivel:
Reading...

davivel:
Is there anything else in the e-mail? I remember a mention to a Surfnet SDK in the live meeting. Or maybe it was a mention to AppAuth, the library in the second link?

davivel:
Summarizing, the main idea there is: "do not use webviews in your app to authenticate the user".

And the alternatives available are:

 - redirect the log-in to the user web-browser
 - redirect the log-in to an app/service/framework owned by the authentication provider

IMHO, the idea is not about improving security itself, but about granting to the user that the security is real. In the end, the OC app is not storing the credentials when SAML or OAuth are used; delegating the log-in to a web browser or an external app would not improve that, just would demonstrate to the user that there is no way we sniff them from the embedded web view.

michaelstingl:
Not exactly. AppAuth would be part of your app, but wouldn't look like DIY and it would have an option to leave your app and use your system browser instead for login.

michaelstingl:
For Android, this would mean, use [AppAuth-Android](https://github.com/openid/AppAuth-Android) instead of our selfmade web view.

codedsun:
I want to contribute! need help..

michaelstingl:
FYI: just updated the the link to a new version of the RFC

davigonz:
[...]

PR #2157, opened 2018-03-21
make oauth request using custom tabs

theScrabi:
implements https://github.com/owncloud/android/issues/2036

____

BUGS & IMPROVEMENTS

- [X] (1) Chrome disabled makes the app crash https://github.com/owncloud/android/pull/2157#issuecomment-383081026 [FIXED]
- [x] (2) Edit credentials does not ask for new credentials https://github.com/owncloud/android/pull/2157#issuecomment-384254486


src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -305,7 +333,56 @@ public void onClick(View v) {
...
+            CustomTabsClient.bindCustomTabsService(this, mCustomTabPackageName, mCustomTabServiceConnection);
+        }
+    }
+
+    @Override
+    public void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+        // check if its a oauth SALS callback intent
+        if(intent.getAction().equals(ACTION_VIEW) &&
+                intent.getData().getHost().equals("android.owncloud.com")) {
```
davigonz: Isn't there a different way to get "android.owncloud.com" instead of using a hardcoded string?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -305,7 +333,56 @@ public void onClick(View v) {
...
+
+        if(mCustomTabPackageName != null) {
+            CustomTabsClient.bindCustomTabsService(this, mCustomTabPackageName, mCustomTabServiceConnection);
+        }
+    }
+
+    @Override
+    public void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+        // check if its a oauth SALS callback intent
```
davigonz: What is the meaning of SALS here?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -305,7 +333,56 @@ public void onClick(View v) {
...
+    }
+
+    private String getCustomTabPackageName() {
+
+        PackageManager pm = getPackageManager();
+        // Get default VIEW intent handler.
+        Intent activityIntent = new Intent(ACTION_VIEW, Uri.parse("https://owncloud.org"));
+        ResolveInfo defaultViewHandlerInfo = pm.resolveActivity(activityIntent, 0);
+
+
```
davigonz: Too much space here

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -987,35 +1071,43 @@ private void startOauthorization() {
         OAuth2Provider oAuth2Provider = OAuth2ProvidersRegistry.getInstance().getProvider();
         oAuth2Provider.setAuthorizationServerUri(mServerInfo.mBaseUrl);
 
-        OAuth2RequestBuilder builder = oAuth2Provider.getOperationBuilder();
+        OAuth2RequestBuilder builder =
+                oAuth2Provider.getOperationBuilder();
```
davigonz: This fits properly in the previous line

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -987,35 +1071,43 @@ private void startOauthorization() {
...
+            openUrlInBrowser(builder.buildUri());
+        }
+    }
 
-        // Show OAuth web dialog
-        ArrayList<String> targetUrls = new ArrayList<>();
-        targetUrls.add(getString(R.string.oauth2_redirect_uri));    // target URI for successful authorization
-        targetUrls.add(mServerInfo.mBaseUrl + OwnCloudClient.FILES_WEB_PATH);
+    private void openUrlWithCustomTab(String url) {
+        TypedValue tvBackground = new TypedValue();
```
davigonz: Not sure if tvBackground is a descriptive or a good name for this variable, what do you mean? Isn't it anything related to 📺 , is it?

src/com/owncloud/android/ui/dialog/LoginWebViewDialog.java
```diff
@@ -122,23 +121,17 @@ public void onAttach(Activity activity) {
...
-                    mOAuthWebViewClientListener = (OAuthWebViewClientListener) activity;
-                    mWebViewClient = new OAuthWebViewClient(activity, handler, mOAuthWebViewClientListener);
-                    break;
-                case SAML_WEB_SSO:
-                    mSsoWebViewClientListener = (SsoWebViewClientListener) activity;
-                    mWebViewClient = new SAMLWebViewClient(activity, handler, mSsoWebViewClientListener);
-                    break;
-                default:
+
+            if(SAML_WEB_SSO == authenticationMethod) {
```
davigonz: I guess this part will be deleted when implementing Chrome custom tabs with SAML, have you found any blocker and decided not to take advantage of this PR and get rid of webview in SAML as well?

davigonz:
@theScrabi some requested changes

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: well its the redirect for oauth.
https://github.com/owncloud/android/blob/1c40bad0dd70a76046fabec3c4a9ae2e87477f9c/AndroidManifest.xml#L203

As long as i know it was hard coded before. We could put it into a constant if thats better.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: I added this because I thought SALS would work the same way. But now the are handled completely different, so we can remove it.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: yea probably better to change ... copypasta error

src/com/owncloud/android/ui/dialog/LoginWebViewDialog.java
theScrabi: SAML as it is implemented for ownCloud requires WebView. We therefore can't remove Webview.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1084,11 +1080,11 @@ private void startOauthorization() {
     }
 
     private void openUrlWithCustomTab(String url) {
-        TypedValue tvBackground = new TypedValue();
-        getTheme().resolveAttribute(R.attr.colorPrimary, tvBackground, true);
+        TypedValue backgroundCollor = new TypedValue();
```
davigonz: backgroundColor instead of backgroundCollor

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -343,9 +343,7 @@ public void onClick(View v) {
     @Override
     public void onNewIntent(Intent intent) {
         super.onNewIntent(intent);
-        // check if its a oauth SALS callback intent
-        if(intent.getAction().equals(ACTION_VIEW) &&
-                intent.getData().getHost().equals("android.owncloud.com")) {
+        if(intent.getAction().equals(ACTION_VIEW)) {
```
davigonz: Don't you need `intent.getData().getHost().equals("android.owncloud.com"))` finally?

davigonz:
A couple of things more @theScrabi

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: This statement should already filter the allowed statement:
https://github.com/owncloud/android/blob/96c141cc899f9c4e689a81618982be7284885567/AndroidManifest.xml#L203

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: done

davigonz:
Code approved, ready to pass QA @jesmrec

jesmrec:
### (1)

Steps:

1. Go to  settings of the device -> Apps -> Chrome
2. Disable the app
3. Open ownCloud app and type an OAuth2 URL
4. Connect

Current: App crashes
Expected: No crash

No Firefox installed in the device

Tested with Nexus 6P v 7.1.2

theScrabi:
The problem is that if you remove chrome on your device no browser is installed with which you can do the login.

So I fixed it by displaying an alert telling the user to install a browser.

jesmrec:
### (2)

Steps:

1. Create an OAuth2 session
2. Open "Manage accounts" 
3. Tap on the key to edit credentials

Current: Credentials are not asked again, only authorization
Expected: Credentials asked (`master` behaviour)

Tested with Nexus 6P v7.1.2 and Google Chrome

theScrabi:
@jesmrec thats expected behavior since the user is still logged in with his browser. However, adding an account on the same cloud could be troubling, since the user first had to log out in his browser.

We need to talk about that with @michaelstingl.

jesmrec:
Then, the option "Edit Credentials" makes no sense for OAuth2 accounts, so this only involves the creation of a new token, by tapping on `Authorize`. The user will only type again the credentials if the token is revoked and is redirected to login view.

michaelstingl:
> @jesmrec thats expected behavior since the user is still logged in with his browser. However, adding an account on the same cloud could be troubling, since the user first had to log out in his browser.

Needs better handling on the server side. @jesmrec @SamuAlfageme Could you link to the related issue?

jesmrec:
A fix (in core/OAuth2 side), could be detecting if other session is alive in browser. If it is, set a `Logout` option or similar. Something like that will fit needs of all clients.

jesmrec:
Log in using OAuth2 already performed through device browser, to improve user reliability in the authentication method. 

Approved

davigonz:
Merged, it will be included in 2.8.0

Issue #2135, opened 2018-03-08
[2.7.0] Release new version


davigonz:
### TASKS
- [x] [WIZARD] Create new steps with important features included in 2.7.0 #2136 
- [x] [COORDINATION] Create task in marketing board to publish a blog post, at least one week before releasing.
 - [X] [GIT] Create branch release_0.9.19 in owncloud/android-library from master  (almost feature-frozen) [DONE, not library changes for this version] 
 - [x] [GIT] Create branch release_2.7.0 in owncloud/android from master (almost feature-frozen) 
 - [x] [DEV] Update version number and name in build.gradle in android module
 - [x] [DEV] Update version number and name in build.gradle in oc_jb_workaround module 
 - [x] [DIS] Create changelog file (< 500 chars) and add to CHANGELOG.md in owncloud/android 
 - [x] [QA] Design Test plan @jesmrec
 - [X] [QA] Regression Test plan  @jesmrec 
 - [X] [GIT] Merge branch release_0.9.19 in owncloud/android-library, into master [DONE, not library changes for this version] 
 - [x] [GIT] Merge branch release_2.7.0 in owncloud/android, into master 
 - [x] [DIS] Generate test APKs files from branch master in owncloud/android and perform some basic operations.
 - [x] [GIT] Create and sign tag oc-android-2.7.0 in HEAD commit of master branch, in owncloud/android 
 - [X] [GIT] Create and sign tag oc-android-library-0.9.19 in HEAD commit of master branch, in  #owncloud/android-library [DONE, not library changes for this version] 
- [x] [DOC] Include new changelog on Releases section of github android repo by drafting the new release 2.7.0
 - [x] [DIS] Generate final APK files (including oc_jb_workaround) from signed commit in owncloud/android
 - [x] [DIS] Upload & publish release APK and changelog in Play Store
 - [x] [DIS] Upload APKs to repo
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [DIS] push OC JB workaround app to Play Store.
 - [x] [COORDINATION] Write in release-coordination chat.
 - [x] [COORDINATION] Write email to release-coordination mailing list.
 - [x] [SPREADTHEWORD] Write a topic in central about the release. https://central.owncloud.org/t/owncloud-android-app-2-7-0-released/
 - [x] [SPREADTHEWORD] Write a post in the blog
 - [x] [DOC] Update documentation with the new features. https://github.com/owncloud/documentation/issues/4035
[...]

PR #2162, opened 2018-03-26
fix no requthentification return form requesting other apps

theScrabi:
@jesmrec I've created a pull request for this bug because I wan't your opinion about this. Can it be seen as a whole punched into the idea of security fingerprint unlock tries to create?

Related to this https://github.com/owncloud/android/issues/2135#issuecomment-376111631 and this https://github.com/owncloud/android/issues/2135#issuecomment-374153626

### DIFF
--- a/src/com/owncloud/android/authentication/FingerprintManager.java
+++ b/src/com/owncloud/android/authentication/FingerprintManager.java
@@ -2,6 +2,7 @@
  *   ownCloud Android client application
  *
  *   @author David González Verdugo
+ *   @author Christian Schabesberger
  *   Copyright (C) 2018 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
@@ -155,5 +156,15 @@ public class FingerprintManager {
                     "Could find out if finger prints are enroded due to lack of android.permission.INTERACT_ACROSS_USERS");
             return false;
         }
+    }
+
+    /**
+     * This can be used for example for onAcitivtyResult, where you don't want to re authenticate
+     * again.
+     *
+     * USE WITH CARE
+     */
+    public void bayPassUnlockOnce() {
+        setUnlockTimestamp();
     }
 }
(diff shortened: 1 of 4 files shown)


jesmrec:
that's a good question, as i have thought in https://github.com/owncloud/android/issues/2135#issuecomment-376131011. The ideal fix should be to avoid the security method only in the case of "Upload from Camera", not to bother user if he/she wants to take several pics. Do you think it is suitable to develop this fix?

CC @michaelstingl

theScrabi:
In my opinion we should not restrict that to the camera only as seen here:
mar-26-2018 15-01-41

theScrabi:
@michaelstingl yesterday we said we should take a look at how dropbox is doing this. Well turns out dropbox/google drive/etc don't have such an authentication feature.

I've now added a commit that deactivates re authentication on "onActivityResult" for all authentication methods. @jesmrec @michaelstingl please tell me what you think about that.

jesmrec:
@theScrabi i will have a look from QA point of view. I mean, if it works as expected, works for me.

michaelstingl:
Sounds okay. Maybe have a look in next weeks demo with @cdamken and @pmaier1 …

jesmrec:
@theScrabi 

checks performed:

- [X] Upload from Camera
- [ ] Content for other apps -> passcode/pattern asked (device with no fingerprint support)
- [ ] Send -> passcode/pattern asked (device with no fingerprint support)

theScrabi:
>  Send -> passcode/pattern asked

as mentioned earlier, I would not fix that now, but rather create an additional issue for that. It's because this is not quite as trivial as the return of activities.

theScrabi:
Further work on this will require https://github.com/owncloud/android/pull/2168 to be merged otherwise I had to rewrite several parts.

jesmrec:
>as mentioned earlier, I would not fix that now, but rather create an additional issue for that. It's because this is not quite as trivial as the return of activities.

Could you create the issue?

theScrabi:
> Content for other apps -> passcode/pattern asked

I could not confirm that, however i noticed that you will have to enter passcode/pattern/finger print if you turn the device while you shoot a photo/select a file

PR #2172, opened 2018-04-03
Skip Welcome Wizard to let run the Espresso tests

jesmrec:
App has currently three modes:

- Debug
- Beta
- Release

currently, the welcome wizard is shown in all of them when the app is installed from scratch. To skip it, an automatic click on SKIP is performed in the first test, so only once per execution of the suite. After that, the check over the credential fields' visibility is moved to the first test as well.

The best approach would be to show the welcome wizard only in Release and Beta, and avoid it in Debug mode. Maybe this tweak can be faced in the next version.


theScrabi:
Even not beta, since beta got removed.

Issue #1890, opened 2017-01-23
Disable preview on tablet in wide-screen.


mbrinkmann:
### Actual behaviour
- ownCloud APP on tablet shows a preview-area if you hold the tablet vertically

### Expected behaviour
- I would like to have the possibility to disable this to see longer file-names.
 
### Steps to reproduce
1. Open ownCloud App
2. Browse any Folder
3. Turn the tablet

### Environment data
Android version:
Every tested (5.0 and upwards)

Device model: 
Acer Iconia A3-40

Stock or customized system:
Stock

ownCloud app version:
2.2.0

ownCloud server version:
9.1.2 (stable)

davivel:
>  if you hold the tablet vertically

I guess that's a typo and you meant 'horizontally' :)

I know what you mean. We need a better, more flexible layout for tablets.

mbrinkmann:
of course you are right, sorry ;p
Thanks for your understanding.

fokkejacobi:
+1
Also when using raster view to review your images, the preview panel is useless.

theScrabi:
> Also when using raster view to review your images, the preview panel is useless.

Absoloutly, and its also completely out of trend. No other files app does two panels anymore. So I remove it.

theScrabi:
@michaelstingl Your opinion pls:
![screenshot_2018-04-05-12-33-59](https://user-images.githubusercontent.com/1891273/38361717-a21fdd6a-38ce-11e8-9d6d-c899543b1003.png)

michaelstingl:
Sounds like a good approach. Go for it! 👍

davigonz:
It will be included in 2.8.0

PR #2173, opened 2018-04-05
make one pannel for tablets

theScrabi:
implements https://github.com/owncloud/android/issues/1890


davigonz:
@theScrabi the code looks good to me but there are several uses of dual panel in other classes such as https://github.com/owncloud/android/blob/b6d80ecf481d2fa6dff967e0b2a4cf15384291e0/src/com/owncloud/android/ui/preview/PreviewTextFragment.java#L342 or https://github.com/owncloud/android/blob/b6d80ecf481d2fa6dff967e0b2a4cf15384291e0/src/com/owncloud/android/ui/fragment/FileDetailFragment.java#L251, what about deleting them as well?

theScrabi:
> https://github.com/owncloud/android/pull/2173#issuecomment-380059255

fixed

davigonz:
Code approved, ready to pass QA

jesmrec:
Tested with Grid/List over tablet with Android 7, and check correct behaviour in phone as well.

Approved

Issue #2120, opened 2018-02-19
Floating buttons and bottom sheets reorder


davigonz:
We have recently included a new contribution to upload pictures directly from the camera and created a new bottom sheets menu in the app that appears when pressing the upload button and includes the new option and upload files one:

![bottom_sheet](https://user-images.githubusercontent.com/15261363/36371109-73dea4b2-1561-11e8-988d-0a2ddf4cb0bf.gif)

Following the new bottom sheet menu shown above and taking into account the features **New file** #2118 and **Scanned document** #2119, I would suggest the next UI changes:

1. Move **Content from other apps** option from floating action buttons to the new bottom sheet.
2. Create a new floating action button for **New text file** option, close to **New folder** option. 
3. Add the **Scanned document** option to the new bottom sheet.

| Floating buttons current design  | Floating buttons new design | 
| ------------- |:-------------:| 
| ![floating_buttons_old](https://user-images.githubusercontent.com/15261363/36375087-2878757c-156e-11e8-959a-43b914211e02.png) | ![floating_buttons_new](https://user-images.githubusercontent.com/15261363/36375093-2f97975c-156e-11e8-9ccf-fa4d82cb1078.png) |

| Bottom sheets current design  | Bottom sheets new design  | 
| ------------- |:-------------:| 
| ![screen shot 2018-02-19 at 12 17 06](https://user-images.githubusercontent.com/15261363/36375321-2394d310-156f-11e8-9b1a-e14371dacf2b.png) | ![screen shot 2018-02-19 at 12 16 27](https://user-images.githubusercontent.com/15261363/36375327-2c4a32d4-156f-11e8-94c7-f07eacae5313.png)

michaelstingl:
👍

theScrabi:
I researched a bit about the "create new file" features, and I found out that other apps basically create an empty file, and than call another app to handle that file.
Like create empty excel document, then call excel.
However for this to implement we'd have to put empty office files into our app assets.
Also if we want to support MSO we should also consider to offer support for odf files.

@michaelstingl @davigonz what are your opinions about this.

michaelstingl:
Empty *.docx files are probably not what was requested from them. As I understood (@SamuAlfageme can verify), they were thinking about the Cloud Storage Provider integration. (see https://github.com/owncloud/ios/issues/1038#issuecomment-374165166 , also why it is difficult) 

[...]

PR #2174, opened 2018-04-06
[WIP] put import button into bootom select thing

theScrabi:
implements https://github.com/owncloud/android/issues/2120


jesmrec:
already done!!

Issue #1825, opened 2016-10-18
Blocking when an account is deleted in server side


jesmrec:
### Steps
1. In app, login in an account
2. In server, delete the account
3. In app, select the account in drawer or in account manager
4. Redirection to login view

The account does not exist anymore, so credentials will never work (username is non-editable). If you tap on "Go Back" button, the files view or settings view (depending whether the app has more accounts or not) is shown for a little while, but you are redirected again to login view. It is a bit annoying to switch to another account or tap on "new account".

Expected behaviour? Switch to another account, or to login view if the app has no accounts. The question is if server side answers in the same way if the credentials have changed and if the account is removed to detect these two different cases.

jesmrec:
In the case of authentication (OAuth2 or whatever), if the user credentials are revoked, the side-effect is the same, making the app annoying to use.

theScrabi:
So the idea was to redirect the user to the account select window instead of the login view. This way the user is still able to select another account or delete one if no other is there.

michaelstingl:
Great teamwork. Go for it!

davigonz:
> agreed with this new behaviour? 

Yes but just an improvement I came up with about the next thought

> If the account was removed, the user is redirected to account select view in case that more accounts are attached. Then, if the user selects again the server-removed account, he should be redirected again to account select view in a loop, until he selects a different account or remove the account from the device.

@michaelstingl @jesmrec @theScrabi  What about deleting the account when the app detects that the server account was removed? I mean, instead of keep showing it in the account select view in a loop, just delete it from the list if is not longer available in the server. IMHO, keep a loop for this doesn't make sense.

jesmrec:
@theScrabi and me were discussing about that. Is there any way to detect a removed account in server-side?

theScrabi:
> @theScrabi and me were discussing about that. Is there any way to detect a removed account in server-side?

One method that works the same for basic auth and oauth

jesmrec:
The case of SAML behaves in same way. SAML session expires itself and the expected behaviour is asking credentials again, so it should be redirected to login view.

PR #2175, opened 2018-04-09
fix block for deleted basic/oauth accounts

theScrabi:
fixes https://github.com/owncloud/android/issues/1825

____

BUGS & IMPROVEMENTS

- [x] (1) List of accounts twice https://github.com/owncloud/android/pull/2175#issuecomment-385927218

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -71,6 +71,8 @@
     <string name="recommend_subject">"Try %1$s on your smartphone!"</string>
     <string name="recommend_text">"I want to invite you to use %1$s on your smartphone!\nDownload here: %2$s"</string>
 
+    <string name="auth_failure_snackbar">Authentication failed</string>
+    <string name="auth_failure_snackbar_action">Change account</string>
     <string name="auth_check_server">Check server</string>
     <string name="auth_host_url">Server address https://…</string>
     <string name="auth_username">Username</string>
--- a/src/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileActivity.java
@@ -343,18 +343,6 @@ public class FileActivity extends DrawerActivity
     }
 
     protected void requestAccountChange() {
-        requestAccountChange(null);
-    }
-
-    protected void requestAccountChange(Account account) {
-        /// request credentials to user
-/*        Intent updateAccountCredentials = new Intent(this, AuthenticatorActivity.class);
-        updateAccountCredentials.putExtra(AuthenticatorActivity.EXTRA_ACCOUNT, account);
-        updateAccountCredentials.putExtra(
-                AuthenticatorActivity.EXTRA_ACTION,
-                AuthenticatorActivity.ACTION_UPDATE_EXPIRED_TOKEN);
-        updateAccountCredentials.addFlags(Intent.FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS);
-        startActivityForResult(updateAccountCredentials, REQUEST_CODE__UPDATE_CREDENTIALS); */
         Intent selectAccount = new Intent(this, ManageAccountsActivity.class);
         startActivity(selectAccount);
     }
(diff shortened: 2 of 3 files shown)


theScrabi:
gurken

davigonz:
@theScrabi there are several comments in the code, is this ready to review?

jesmrec:
### (1)

Steps:

1. Attach a basic auth account and open file list
2. In web UI, remove the account
3. Refresh account in device -> redirected to accounts list
4. Select the removed account

Current: List of accounts again displayed. A second tap will redirect to the file list
Expected: List of files, and then again to account list.

tested with Samsung Galaxy S7

jesmrec:
i guess there is no way to distinguish this 401 scenario (removed account) to other 401 ones. So, the user is suddenly redirected to the accounts list and he/she does not know why (for example, if the server admin removes the account). Is it correct @theScrabi ?

jesmrec:
Removing account in web UI, getting 401 from webdav endpoint: 

```
<?xml version='1.0' encoding='utf-8'?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotAuthenticated</s:exception>
  <s:message>Username or password was incorrect, Username or password was incorrect</s:message>
</d:error>
```
and for ocs

```
{
    "message": "Current user is not logged in"
}
```

Same error is received if the user changes password in webUI.

In case of OAuth2 token removal: 

```
<?xml version='1.0' encoding='utf-8'?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotAuthenticated</s:exception>
  <s:message>No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured, No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured, Bearer token was
incorrect</s:message>
</d:error>
```


At this point we are in the middle of an UX crossroad.

- If we forward the user to the login view -> https://github.com/owncloud/android/issues/1825
- If we forward the user to the account list -> user is confused, so he/she does not know what happened (credentials changed, session expired, account deleted, any other kind of error...).

A solution "in the middle": show a toast/snackbar pointing to a credentials error. This would be the same solution as before, but giving the user directly the option to choose other account in his list o create a new one. This message is not very accurate, but it is better than show nothing. What do you think.

davigonz:
> A solution "in the middle": show a toast/snackbar pointing to a credentials error, or showing the message included in the body response.

If we don't have more information from the server, I would go for the first option, show a toast/snackbar pointing to a credentials error.

theScrabi:
I take car about it.

theScrabi:
I got another odd error. If you have an __basic auth__ an an __oauth__ account, and you remove the __oauth__ one from the web and switch to the __basic auth__ you will get this error:
```
<?xml version='1.0' encoding='utf-8'?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotAuthenticated</s:exception>
  <s:message>No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured, No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured, Bearer token was
incorrect</s:message>
</d:error>
```
gurken

@jesmrec I din't know if this is what you reported, but for me this looks like a different error.

theScrabi:
@jesmrec I took care about that issue by showing a snack bar instead of redirecting the user when a login failed. You can actually see this behavior in the video [here.](https://github.com/owncloud/android/pull/2175#issuecomment-386255092)

jesmrec:
From the point of view of basic auth and OAuth2 it works correctly. Needs to be checked with SAML in order to know what happens when the SAML session expires.

jesmrec:
Approved for basic and OAuth2, so it is ready to be used for used in Play Store

Issue #1888, opened 2017-01-21
instant upload,  camera directory not selectable in root


seisen:
### Actual behaviour
go to instant upload abd change camera directory to /, you are stuck in ut,  no subdirectory is shown 
 
### Steps to reproduce
1.  change camera directory to /
2. now,  no subdirectory is shown,  you are stuck in /

### Environment data
Android version: 7.0

Device model: Sony XZ

Stock or customized system: stock

ownCloud app version: 2.2.0

davivel:
Yeah, Android is making weird things with file system in Android 7... we'll have to give a look.

kokangit:
I experience that same problem on ownCloud Android-app version 2.4.0 on Sony Xperia Z5. Android version 7.1.1.
Is there a work around to this?

brianmichalk1:
I am also having the same problem.  Where is the configuration file with this path setting?  That should be a workaround.

theScrabi:
This issue does not ocure to me in the current version 2.6.0. 
@seisen do you experience this still? Maybe this can be closed.

theScrabi:
![gurken](https://user-images.githubusercontent.com/1891273/38493806-c203dcbc-3bf3-11e8-9b36-a9f7ed80f3ca.gif)

peted2:
Android version: 7.1.2

Device model: Motorola Moto G 2014

Stock or customized system: LOS 14.1

ownCloud app version: 2.7.0


I experience the same issue.
Pretty annoying since the Phones internal storage is limited and i cannot use the preset internal Camera folder.
I need to select external storage. This is impossible due to this bug.

peted2:
until the fix is implemented, would it be feasable to create a symlink in the root dir pointing to the camera folder? which permissions would have to be set to this symlink?

theScrabi:
@peted2 my fix will only bring you back to home after you accesd root. You should already be able to access your sdcard. What is in your `/storage` folder?

peted2:
when i access it with any file explorer app, i can see /storage/external-sd-card-number /storage/emulated and storage/self 

Unfortunately inside the app i am already stuck to /

@theScrabi is there any way to contact you more directly?

jesmrec:
This fix will be released in 2.8.0

PR #2176, opened 2018-04-09
add home button for selecting camera directory

theScrabi:
fixes https://github.com/owncloud/android/issues/1888


.gitignore
```diff
@@ -37,4 +37,7 @@ tests/proguard-project.txt
 .gradle
 .idea
 *.iml
-build
\ No newline at end of file
+build
+
+# android sdk captures folder
+captures
```
davigonz: Where is this directory located? I do not see it in the app directories.

res/layout/files_folder_picker.xml
```diff
@@ -1,58 +1,68 @@
 <?xml version="1.0" encoding="utf-8"?>
-    <!--
-  ownCloud Android client application
+<!--
+ownCloud Android client application
 
-  Copyright (C) 2016 ownCloud GmbH.
+Copyright (C) 2016 ownCloud GmbH.
```
davigonz: Please update this year

src/com/owncloud/android/ui/activity/LocalFolderPickerActivity.java
```diff
@@ -2,6 +2,7 @@
  *   ownCloud Android client application
  *
  *   @author David A. Velasco
+ *   @author Christian Schabesberger
  *   Copyright (C) 2016 ownCloud GmbH.
```
davigonz: Same here

src/com/owncloud/android/ui/activity/LocalFolderPickerActivity.java
```diff
@@ -121,6 +123,17 @@ public void onClick(View view) {
                 finish();
             }
         });
+        mHomeBtn = findViewById(R.id.folder_picker_btn_home);
+        mHomeBtn.setVisibility(View.VISIBLE);
```
davigonz: Do you really need to take care of this button visibility?  You are defining it like gone in the xml and making it visible again.

davigonz:
@theScrabi please have a look at the requested changes

.gitignore
theScrabi: Its not there, but if you create captures it will be auto generated. However in order to not evidently upload captures we need to place it here.

src/com/owncloud/android/ui/activity/LocalFolderPickerActivity.java
theScrabi: Yes but its necesry because the layout is used twice in the code. Once for moving files within the cloud and once to select the folder for the camera upload.

We don't want the button to show up when moving files.

davigonz:
Code approved, ready to pass QA @jesmrec

jesmrec:
Fixed. Camera folder in instant uploads can be changed after selecting "/", by using the new Home button

Issue #2121, opened 2018-02-19
FR: Adaptive icon


mueller-ma:
Api 26 introduced adaptive icons where background and foreground are two different images.
https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html

theScrabi:
This is also useful because on android 8+ the icon does not blend in well.
![screenshot_1523272529](https://user-images.githubusercontent.com/1891273/38495040-6f9f3d04-3bf8-11e8-8544-79277413af1e.png)

PR #2177, opened 2018-04-09
add adaptive icon

theScrabi:
implements https://github.com/owncloud/android/issues/2121
screenshot_1523275413 screenshot_1523275429

@jesmrec what should I do about the own brander thing here?


davigonz:
> what should I do about the own brander thing here?

I'm just seeing vector resources, don't know how to put this there. 😕

theScrabi:
the `logo_icon.xml` can be replaced by non vector graphics. However the thing about adaptive icons is that they are a little more complicated to produce. Ownbrander should be able to generate the `icon.xml`, because someone needs to be able to set his own back and foreground for the icon.

davigonz:
LGTM, moving to ready to test @jesmrec

jesmrec:
this is only applicable to API >= 26, ins't it? in case of API < 26, the icon will be the same as usual?

theScrabi:
> this is only applicable to API >= 26, isn't it? in case of API < 26, the icon will be the same as usual?

Yes. But everyone useing android 8+, and these people get quite more don't see a good icon design.

jesmrec:
Tested with all available icon shapes:

- [X] Default
- [X] Square
- [X] Rounded square
- [X] Squircle
- [X] Teardrop

Approved

davigonz:
Merged, it will be included in 2.8

Issue #2153, opened 2018-03-16
Possible Code refactorings


theScrabi:
I've worked on the android app code for a little while now, and found some code which could be refactored or even removed. Here is my list:

- [ ] [ChangeLogDialog](https://github.com/owncloud/android/blob/master/src/com/owncloud/android/ui/dialog/ChangelogDialog.java) is not in use anymore since we now have the tutorial screen. Thus we can DELETE it.

theScrabi:
Also we might want to consider to upgrade to Java 8 since that comes very in handy with lambda expressions, which is good because less ugly Java code -> better readable.

dutta14:
I can work on this if it has not been taken up yet.

theScrabi:
@dutta14 if you like go ahead :)

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #2178, opened 2018-04-10
remove unused classes

theScrabi:
remove classes listed in https://github.com/owncloud/android/issues/2153


src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1879,9 +1879,6 @@ private void showUntrustedCertDialog(RemoteOperationResult result) {
 
     }
 
-    /**
-     * Called from SslValidatorDialog when a new server certificate was correctly saved.
```
davigonz: Is this method being used now that SslValidatorDialog has been also deleted?

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1879,9 +1879,6 @@ private void showUntrustedCertDialog(RemoteOperationResult result) {
 
     }
```
davigonz: Update the year and include your name as @author at the top of the class

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1891,10 +1888,6 @@ public void onSavedCertificate() {
         }
     }
 
-    /**
-     * Called from SslValidatorDialog when a new server certificate could not be saved 
-     * when the user requested it.
-     */
     @Override
     public void onFailedSavingCertificate() {
```
davigonz: And this method?

src/com/owncloud/android/ui/RadioButtonPreference.java
```diff
@@ -1,39 +0,0 @@
-/**
```
davigonz: This class was included in methods to delete in #2153, I've just moved it to classes to delete and checked it as done.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: yes, this is also an override. Overides should not be removed even if they are not used.
I've updated the issue because of that.

src/com/owncloud/android/authentication/AuthenticatorActivity.java
theScrabi: done

src/com/owncloud/android/ui/RadioButtonPreference.java
theScrabi: fixed that

src/com/owncloud/android/authentication/AuthenticatorActivity.java
davigonz: Yes, you're right, I didn't see the Override label

davigonz:
Code approved, no QA needed in my opinion since we are deleting non-used classes, what do you think @jesmrec?

jesmrec:
Only assuring that the app is not losing the basic functionality, only smoke testing. Checks to perform:

- [X] App built and installed 
- [X] Login
- [X] Upload
- [X] Download

All OK. Approved

PR #2179, opened 2018-04-10
remove unused methodes

theScrabi:
Remove unused methodes listed here:
https://github.com/owncloud/android/issues/2153

### DIFF
--- a/src/com/owncloud/android/authentication/AccountAuthenticator.java
+++ b/src/com/owncloud/android/authentication/AccountAuthenticator.java
@@ -2,8 +2,9 @@
  *   ownCloud Android client application
  *
  *   @author David A. Velasco
+ *   @author Christian Schabesberger
  *   Copyright (C) 2012  Bartek Przybylski
- *   Copyright (C) 2016 ownCloud GmbH.
+ *   Copyright (C) 2018 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License version 2,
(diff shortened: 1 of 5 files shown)


src/com/owncloud/android/services/OperationsService.java
```diff
@@ -521,7 +521,6 @@ private void nextOperation() {
                     }
                 }
                 
-                //sendBroadcastOperationFinished(mLastTarget, mCurrentOperation, result);
```
davigonz: sendBroadcastOperationFinished method is not currently being used, I guess you decided to keep it since it could be useful in the future.

src/com/owncloud/android/broadcastreceivers/ConnectivityActionReceiver.java
```diff
@@ -126,6 +126,7 @@ public void onReceive(final Context context, Intent intent) {
         //  and reproduced in Nexus 5 with Android 6.
```
davigonz: Same here

src/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -319,25 +319,6 @@ public UnsupportedAuthTokenTypeException() {
         }
     }
```
davigonz: Update the year at the top of the class

src/com/owncloud/android/services/OperationsService.java
```diff
@@ -521,7 +521,6 @@ private void nextOperation() {
                     }
```
davigonz: Same here

src/com/owncloud/android/services/OperationsService.java
theScrabi: I kept serveral methode, because it was logic to have them. However if you say we don't need those as well lets remove them.

davigonz: Ok, I would keep this method since is logic to have it.

davigonz:
Code approved, @theScrabi please don't squash the commits when applying the requested changes 😃 Should we perform some basic tests @jesmrec?

jesmrec:
Checks to perform:

- [X] App built and installed 
- [X] Login
- [X] Upload
- [X] Download

All OK. Approved

PR #2115, opened 2018-02-10
add password visibility toggle feature in link share dialog

harshitbansal05:
This pr solves #2112. It adds the feature to toggle visibility of password in public link share dialog.

______

BUGS & IMPROVEMENTS

- [ ] (1) Not working in edit public link https://github.com/owncloud/android/pull/2115#issuecomment-372291266

Issue #2112, opened 2018-02-06
No password visibility toggle feature in Create link share dialog


harshitbansal05:
### Actual behaviour
There is no password toggle feature in **Create Link Share** dialog.
### Expected behaviour
A password toggle feature should be added to the  dialog.
 
### Steps to reproduce
1. Long tap on a file and click share icon.
2. Click on create a link option.

Can this problem be reproduced with the official owncloud server?
Yes

### Environment data
Android version: 5.1.1

Device model: Redmi Note 3

Stock or customized system: Stock

ownCloud app version: 2.5.1-beta.1

jesmrec:
do you mean  the password to protect the public link? i can not reproduce with version 2.6.0

![screen shot 2018-02-07 at 08 29 10](https://user-images.githubusercontent.com/14894746/35903814-04c918e8-0be1-11e8-94e1-58e01792bf0b.png)

harshitbansal05:
After we switch on the password feature, while entering the password, there is no option to toggle visibility of the password.

jesmrec:
do you mean to set the password as visible (not hidden behind asterisks), like the password in login view?

jesmrec:
yep, i meant that. The eye icon, for example?

harshitbansal05:
Yes, should it be implemented?

harshitbansal05:
@jesmrec i would start working on it...

PR #2186, opened 2018-04-13
Password toggle

davigonz:
This pr solves #2112 and merge the code included in #2115. It adds the feature to toggle visibility of password in public link share dialog.

______

BUGS & IMPROVEMENTS

- [x] (1) Not working in edit public link https://github.com/owncloud/android/pull/2115#issuecomment-372291266 [WONT FIX]


jesmrec:
The bug (1) reported will not be fixed, so the password is not stored in the device, and there is no way to know if in car be toggled or not. So, this is approved and ready to merge

davigonz:
@harshitbansal05 thanks for this contribution, it will be included in next version 2.8.0

PR #2199, opened 2018-04-19
code cleaning: refactor small bits

theScrabi:
- Remove HockActivity
- made SqareImageView and ActionEditText be based on AppCompat Views
- made PreviwVideoFramgent.onConfigurationChaged() class its superclass


davigonz:
> I also notice changes in ActionEditText . What is this used for?

It seems is not being used anywhere, shall we delete it @theScrabi?

davigonz:
Let's move this forward, think that the tests you performed are enough @jesmrec , can we approve it then?

PR #2213, opened 2018-05-04
change edit share icon

theScrabi:
The pan fits better to what happens when you click that option.

### Before --- After
before after


res/layout/share_public_link_item.xml
```diff
@@ -64,7 +64,7 @@
             android:layout_height="wrap_content"
             android:layout_gravity="center_vertical"
             android:padding="@dimen/standard_half_padding"
-            android:src="@drawable/ic_navigate_next" />
+            android:src="@drawable/ic_lead_pencil_grey" />
```
mueller-ma: @theScrabi This pencil looks like it comes from the pre-Material time in Android. You might want to use [this icon](https://material.io/tools/icons/?icon=edit&style=baseline) as svg, not png.

theScrabi: I thought it would look a little slim with the new Icon. but yes i can give it a shot.

davigonz:
Code approved, ready to pass QA @jesmrec

jesmrec:
i guess that the pencil icon fits the material design.

But, the core uses the gear icon for the same aim:

screen shot 2018-08-01 at 18 33 30

what do you think?

davigonz:
In my opinion, the pencil fits better to our app since is widely used in Android and is what appears in Material design icons when searching "Edit".

screen shot 2018-08-02 at 11 07 46

I think is better from the user point of view to adapt our icons to each platform, and not to try to use the same icons as the web. It might sound odd but Android users are used to material design icons and the web is not using that sort of icons. 

There's several examples about this, such as the icon to copy a public link which is different from the web UI as well, and about editing a public link, iOS doesn't use the gear but a information icon what is iOS specific.

jesmrec:
@michaelstingl @pmaier1 agreed with using the pencil icon instead of the web's gear?

Issue #2205, opened 2018-05-02
As a user, I would like to know the space remaining in my account


davigonz:
It would be great to have this feature included in our side menu for example, as suggested [here](https://github.com/owncloud/android/issues/2204).

There's a pretty old PR, https://github.com/owncloud/android/pull/737 that follows a different approach, creating a specific activity to show the quota. Although we can use the library PR maybe, https://github.com/owncloud/android-library/pull/102 and take advantage of `GetRemoteUserQuotaOperation`

___

**TASKS**

- [X] [LIBRARY] Get quota from the server by using GetRemoteUserQuotaOperation. **Already merged via https://github.com/owncloud/android-library/pull/102**
- [X] [APP] Get quota when getting user profile info and stores it in DB. 
- [X] [APP] Include quota in drawer menu, handling different cases with quotas using MB or GB
- [X] [APP] Handle pending, unknown and unlimited quota, switch to WebDAV endpoint. Related: https://github.com/owncloud/core/issues/31376
- [X] [APP] Handle quota 0

theScrabi:
Related to this: https://github.com/owncloud/android/issues/2204

We can put a mark in the sidebar, an optional show more detailed information if one tabs on that mark.

davigonz:
Here we go with some screenshots so you can see how this feature is gonna look like:


| Limited quota | Unlimited quota |
| ------------- | ------------- |
| ![limited](https://user-images.githubusercontent.com/15261363/39805633-b553bf78-5377-11e8-872e-718322349fc8.png) | ![unlimited](https://user-images.githubusercontent.com/15261363/39805641-baeb4fe6-5377-11e8-81ad-b7a28e60ceef.png) |

PR #2214, opened 2018-05-04
User quota

davigonz:
Implements #2205 
Needs https://github.com/owncloud/android-library/pull/189

BUGS & IMPROVEMENTS

- [X] (1) set 1024 MB as 1 GB https://github.com/owncloud/android/pull/2214#issuecomment-389202364 [FIXED]
- [X] (2) Floating comma expressions https://github.com/owncloud/android/pull/2214#issuecomment-389523917 [FIXED]
- [X] (3) Incorrect output with "Other" option https://github.com/owncloud/android/pull/2214#issuecomment-389542375 [FIXED]
- [X] (4) Progress bar https://github.com/owncloud/android/pull/2214#issuecomment-389767361 [FIXED]

### DIFF
--- a/src/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -48,6 +48,7 @@ import com.owncloud.android.datamodel.UserProfile;
 import com.owncloud.android.datamodel.UserProfilesRepository;
 import com.owncloud.android.lib.common.OwnCloudAccount;
 import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.lib.resources.users.GetRemoteUserQuotaOperation.Quota;
 import com.owncloud.android.utils.DisplayUtils;
 
 /**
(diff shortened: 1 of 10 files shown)


SamuAlfageme:
@davigonz could you guys post a couple of screenshots with relative (know upper limit) and absolute (unknown space left) quotas?

davigonz:
> could you guys post a couple of screenshots with relative (know upper limit) and absolute (unknown space left) quotas?

@SamuAlfageme [Here](https://github.com/owncloud/android/issues/2205#issuecomment-387671732) you have a couple of screenshots with limited and unlimited quotas. 

Regarding to the second case you mention (unknown available space, code -2) , the message will be something like: 62.3 MB of Unknown used.

SamuAlfageme:
> the message will be something like: 62.3 MB of Unknown used 

The wording "unknown" might _scare_ end users, I'd rather display just the space used:

in use

Thanks a bunch for the screencaps!

src/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -412,6 +420,78 @@ private void repopulateAccountList(Account[] accounts) {
         showMenu();
     }
 
+    /**
+     * Updates the quota in the drawer
+     */
+    private void updateQuota() {
+
+        Account account = AccountUtils.getCurrentOwnCloudAccount(this);
```
theScrabi: I think it's better if there is no empty line after every new line.

src/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -412,6 +420,78 @@ private void repopulateAccountList(Account[] accounts) {
...
+        }
+    }
+
+    /**
+     * Retrieve free quota message for specific cases
+     * @param freeQuotaCode code to represent the free quota in specific cases, more details in {@link Quota}
+     * @return message to show depending on the specific case of quota storage
+     */
+    private String getQuotaTotalMsg(int freeQuotaCode) {
+
```
theScrabi: Here as well.

michaelstingl:
Data in shared folders or external storage aren't part of users quota. Theoretically more space is can be in use, then quota. Desktop client has not the perfect solution for it, but the numbers are shown correct: https://github.com/owncloud/enterprise/issues/2140 (internal link)

davigonz:
@michaelstingl 

The first approach I followed was to use `{{server}}/ocs/v1.php/cloud/users/{{user}}` endpoint, used by our original `GetRemoteUserQuotaOperation` and getting this information from the server:

```
"quota": {
    "free": 56690487296,
    "used": 5812174,
    "total": 56696299470,
    "relative": 0.01
}
```

Everything was going fine till I tried to show info about unlimited quotas and figured out that wasn't possible with that endpoint since it retrieves similar information with unlimited, unknown and pending quotas and I needed a flag to distinguish them from limited quotas (+info in https://github.com/owncloud/core/issues/31376) 

So I decided to follow the same approach as desktop client and start using the PROPFIND that returns the next data when a quota is unlimited:

```
<d:quota-available-bytes>-3</d:quota-available-bytes>
<d:quota-used-bytes>5731916</d:quota-used-bytes>
```

Here I have the code -3, that helps me to show different messages in quota depending on the specific case.

Is this solution good enough for now? The **perfect solution** you mean would be fixing https://github.com/owncloud/core/issues/31376 and have more reliable information about the quota at `{{server}}/ocs/v1.php/cloud/users/{{user}}` endpoint.

michaelstingl:
Okay, I understand. Thanks!

jesmrec:
### (1) Improvement

1. Set user quota = 1 GB

Current: 50 MB of 1024.0 MB used

It would be better to set this quota value in GB, is such case 1 GB:

50 MB of 1 GB used

jesmrec:
@davigonz right now, the expression appears in floating comma:

1E+1 GB

jesmrec:
### (3)

Steps:

1. In web UI, set "Other" as quota, and input any amount, f. ex 390 MB
2. Install the app from scratch and log in with the user

Current: %1$s of %2$s used
Expected: correct output

jesmrec:
### (4)

Steps:

1. Set a user with 0 B quota -> appears the correct message and no progress bar (correct behaviour)
2. Switch the quota to other value (1GB f. ex.)

Current: Progress bar does not appear until the app is killed and reopened
Expected: Progress bar appears

jesmrec:
All stuff fixed.

Approved.

PR #2218, opened 2018-05-07
Remove F-Droid beta from README

lefherz:
the F-Droid Beta was discontinued.


davigonz:
Hi @lefherz, the beta version in F-Droid has not been discontinued and can be installed from https://f-droid.org/tr/packages/com.owncloud.android/index.html (see screenshot below). What we did was delete the separate app and include the beta in a different entry since F-Droid is able to show different versions in parallel.


screen_shot_2018-05-08_at_08_54_52

davigonz:
Thanks for opening this PR but the README file is properly updated, so let's close this...

PR #2220, opened 2018-05-10
Fix stuck logs view

### LINKED ISSUE #2129: ANR when opening logs
### Actual behaviour
-Tell us what happens

### Expected behaviour
-Tell us what should happen
 
### Steps to reproduce
1. 
2. 
3. 

### Environment data
Android version:

Device model: 

Stock or customized system:

ownCloud app version:

ownCloud server version:

#### Web server error log
```
Insert your webserver log here
```

```
Insert your ownCloud log here
```

davigonz:
Fixes #2129


src/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -173,76 +179,83 @@ private void sendMail() {
...
+                dismissLoadingDialog();
             }
         }
 
         /**
          * Read and show log file info
          */
-        private String readLogFile() {
-
+        private ArrayList<Log> readLogFile() {
```
theScrabi: Please chop this methode down in order to make it better readable.

Suggestion:
- `readLogFiles()` which reads all log files
- `readLogFile()` which only reads one file
- `readLineFromLog()` which only reads one line

src/com/owncloud/android/logs/Log.java
```diff
@@ -0,0 +1,51 @@
...
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package com.owncloud.android.logs;
```
theScrabi: Do you think it necessary to add a new package just for a single class? Couldn't we put it eg. into `utils`?

src/com/owncloud/android/ui/activity/LogHistoryActivity.java
davigonz: Done

src/com/owncloud/android/logs/Log.java
davigonz: I didn't find a proper package to put it. This new class is not a `utils` class, is kind of a model class but it would't fit in `datamodel` package either since that package contains things to save in our local database and is not the case.

theScrabi: Hm and just keeping it in the root folder then? Tho I know this is not our style.

davigonz: I would rather keep it in the logs package, in the root folder we don't have model classes.

davigonz:
Code approved, moving to ready to test

jesmrec:
Approved

Issue #2158, opened 2018-03-22
Feature Request Show the video being played even if the app has been closed.


sd1998:
### Actual behaviour
When the user plays a video and closes the app using the home button the video stops playing and the app closes.

### Expected behaviour
When the user presses the home button the video should continue playing on a small window that can be drawn on other apps.
 
### Steps to reproduce
1. Play a video that has been uploaded to ownCloud
2. Press the home button.

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
 Yes

### Environment data
Android version: 8.0.0

Device model: Oneplus 3T

ownCloud app version: 2.6.0

sd1998:
@davigonz I think this would be a nice feature. I would like to work on this.

davigonz:
Hi @sd1998 , thanks for this suggestion, this feature would be awesome!! Glad to hear you want to work on this, we will help you in whatever you need.

sd1998:
@davigonz will there be a bounty for this feature?

michaelstingl:
@sd1998 currently there is no bounty: https://www.bountysource.com/issues/56314384-feature-request-show-the-video-being-played-even-if-the-app-has-been-closed

Propose on https://central.owncloud.org/ to get some attention from supporters?

theScrabi:
Over at NewPipe we use a non Pip based overlay mode that gives us the ability to to play videos while we are in another app.

![screenshot_20180323-102747](https://user-images.githubusercontent.com/1891273/37821964-746cbf8a-2e85-11e8-9002-4123fcff6edc.png)

Our implementation has several advantages/disadvantages over Oreo pip:

#### Advantages
- Windows is resizable
- Compatible with older android devices

#### Disadvantages
- legecy code once oreo is rolled out on more than 90% of all devices (may take a while)
- UI bug when enabling fullscreen desktop

However if you where interested I could ask one of the NewPipe developers @mauriciocolli or @karyogamy if they where interested in making the NewPipe player a seperate github repo/gradle plugin, so the ownCloud app can take advantage of this as well.

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #2230, opened 2018-05-27
PIP mode

shashvat-kedia:
Now the video being viewed by the user will keep on playing (in a minimized view (PictureInPicture mode)) even if the user has exit the activity(using home-button of the device).

Related issue: https://github.com/owncloud/android/issues/2158

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -274,6 +274,7 @@
         <activity
             android:name=".ui.preview.PreviewVideoActivity"
             android:configChanges="keyboard|keyboardHidden|orientation|screenSize|screenLayout|smallestScreenSize|uiMode"
+            android:supportsPictureInPicture = "true"
             android:theme="@style/Theme.ownCloud.Fullscreen" />
         <activity android:name=".ui.activity.PatternLockActivity" />
         <activity android:name=".ui.activity.FingerprintActivity"/>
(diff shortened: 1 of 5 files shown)


sd1998:
@davigonz I have made a basic implementation of the feature. Should I also add buttons to control the player (play/pause etc.) in PIP mode?

theScrabi:
I am using android 8.1 but i was not able to activate pip.

sd1998:
@theScrabi Did you press the home button when the video was being displayed in the full screen (PreviewVideoActivity) and not the PreviewVideoFragment(As in this case the video is not truly playing on the full screen)? Because I have implemented the feature in such a way that only when the user is viewing the video in full screen and presses the home button PIP is activated.

theScrabi:
I see, that works 👍 
However if you have an emulated navigation bar that a bit inconvenient, since that bar disappears on full screen view. Maybe you should consider to enable pip also on the fragment view.

sd1998:
@theScrabi Alright I will work on that.

jesmrec:
which status do we have here?

sd1998:
@jesmrec I could . not understand what other changes were required. I have implemented the pip mode for video preview.

sd1998:
@jesmrec Yes but some changes were suggested by @theScrabi [#2230Comment](https://github.com/owncloud/android/pull/2230#issuecomment-396554317) which in which I could not understand what was exactly required.

farhazulMullick (approved):

abelgardep:
Some classes were migrated to kotlin and now this PR is outdated

PR #2232, opened 2018-05-30
refactor SynchronizeFolderOperation

theScrabi:
This will make the SynchronizeFolderOperation better readable, since so far you had to read between the lines in order to understand what it does.


abelgardep:
SynchronizeFolderOperation was refactored entirely in the new architecture: https://github.com/owncloud/android/blob/new_arch%2Fsynchronization/owncloudApp/src/main/java/com/owncloud/android/usecases/synchronization/SynchronizeFolderUseCase.kt

Feel free to check the new code and suggest new changes there 👍

PR #2234, opened 2018-06-04
add powersave mode to animations

theScrabi:
This will disable Animations when phone is in power save mode.


jesmrec:
Tested with Android 7 and 8. Approved

davigonz:
Merged, it will be included in 2.8 version

Issue #2219, opened 2018-05-09
New login view


davigonz:
Our login view is not consistent with the most recent ownCloud versions, so I suggest to update it with something like this.

[<img src="https://user-images.githubusercontent.com/15261363/39826067-ce0e1694-53b3-11e8-8b0b-93f355874c15.png" width="400px"/>](https://user-images.githubusercontent.com/15261363/39826067-ce0e1694-53b3-11e8-8b0b-93f355874c15.png)

In comparison with the previous design:

| Old login view design | New login view design |
| ------------- | ------------- |
| ![old_login_device](https://user-images.githubusercontent.com/15261363/39826089-e0c1c402-53b3-11e8-8bda-74ded0c2b8db.png) | ![login_device](https://user-images.githubusercontent.com/15261363/39826067-ce0e1694-53b3-11e8-8b0b-93f355874c15.png) | 

The **background image** should be customizable of course 

CC / @michaelstingl @jesmrec @ChrisEdS @theScrabi

priyanktyagi89:
@davigonz : Can I pick this task?

jesmrec:
@pmaier1 do you agree to use the mentioned look for the login view? i think is congruent with the current web UI in the product.

The point is, that this pic changes in different versions of the product, so it should be maintained. 

About branding, the pic is the same used in https://github.com/owncloud/android/issues/2204, so a combined option with the current one for branding the color of the drawer fits (maybe) better.

CC @theScrabi  @michaelstingl

michaelstingl:
@jesmrec sounds cool 👍

pmaier1:
I kind of dislike the headline "Server address https://...". Would rather like to have it as a placeholder in the input field. Apart from that it's only the background/colors that has changed, right?

pmaier1:
> @pmaier1 the issue about the android design is that the place holder turns into the headline once you tap into the input field.

Ah ok, I see, thanks :+1:

davigonz:
> Apart from that it's only the background/colors that has changed, right?

@pmaier1 Colors, background and a better handling of white spaces.

cdamken:
Question: Now is not necessary to type `http://` or `https://` in front of the URL, will be that autodetect?

theScrabi:
no its not automated.

theScrabi:
good point :D

davigonz:
> ;( then you have to still write it? in the screenshots at the top is now written.

[...]

PR #2235, opened 2018-06-06
add new login style

theScrabi:
This is the implementation of https://github.com/owncloud/android/issues/2219

_____

**BUGS & IMPROVEMENTS**

- [x] (1) Avoid background resizing when opening the keyboard [WON'T BE FIXED]
- [X] (2) Branding https://github.com/owncloud/android/pull/2235#issuecomment-409601640
- [x] (3) Arrow in URL https://github.com/owncloud/android/pull/2235#issuecomment-411478878 [DONE] @jesmrec  
- [x] (4) Logo is cut in credentials mode https://github.com/owncloud/android/pull/2235#issuecomment-414738777 [WON'T BE FIXED]
- [x] (5) Background color https://github.com/owncloud/android/pull/2235#issuecomment-414944304 [FIXED]


res/values/styles.xml
```diff
@@ -246,4 +246,27 @@
         <item name="android:textSize">16sp</item>
     </style>
 
+    <!-- Authenticator Activitu -->
```
jesmrec: Activity

res/values/styles.xml
```diff
@@ -246,4 +246,27 @@
         <item name="android:textSize">16sp</item>
     </style>
 
+    <!-- Authenticator Activity -->
+    <style name="Theme.ownCloud.Authenticator"
+        parent="Theme.ownCloud">
+
```
davigonz: Too much space here

res/values/styles.xml
```diff
@@ -246,4 +246,27 @@
...
+        <!-- properties for action mode -->
+        <item name="windowActionModeOverlay">true</item>
+        <item name="actionModeBackground">@color/action_mode_background</item>
+
```
davigonz: Same here

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1253,10 +1253,15 @@ private void onGetServerInfoFinish(RemoteOperationResult result) {
             mServerIsValid = false;
         }
 
+
+
+
         if (mServerIsValid) {
             updateAuthenticationPreFragmentVisibility();
         }
```
davigonz: Same here

src/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1253,10 +1253,15 @@ private void onGetServerInfoFinish(RemoteOperationResult result) {
             mServerIsValid = false;
         }
 
+
```
davigonz: Same here

res/layout-land/account_setup.xml
```diff
@@ -17,203 +17,214 @@
...
+            <ScrollView
+                android:id="@+id/scroll"
+                android:layout_width="0dp"
+                android:layout_height="match_parent"
+                android:layout_weight="1"
+                android:fillViewport="true"
+                android:orientation="vertical">
+
+                <LinearLayout
+                    android:id="@+id/LinearLayout1"
```
davigonz: Please use a more descriptive name for this linear layout

res/layout-land/account_setup.xml
```diff
@@ -17,203 +17,214 @@
...
+                            android:layout_height="wrap_content"
+                            android:layout_gravity="bottom"
+                            android:contentDescription="@string/auth_host_address"
+                            android:drawablePadding="5dp"
+                            android:hint="@string/auth_host_url"
+                            android:inputType="textUri"
+                            android:paddingRight="55dp"
+                            style="@style/EditText.Authenticator"
+                            android:theme="@style/EditText.Authenticator">
+
```
davigonz: Too much space here

davigonz:
Would it be possible to distribute better the white spaces as the original design? https://github.com/owncloud/android/issues/2219

[](https://user-images.githubusercontent.com/15261363/42682430-e0175b48-868a-11e8-9321-14519117fc88.png)

davigonz:
I like so much the new design but I miss a bit of intuitivity here and this problem comes from previous versions. 

I'm talking about the connect button, which appears as soon as the login view is opened and is useless till the username and password fields come out. 

Think we could use the same button but with different names:

- **Check server** to click on it after introducing the server address.
- **Log in** to click on it after introducing username and password.

jesmrec:
My choice:

- **Connect** in the first
- **Log in** or **Authorize** in the second (same wording used in OAuth2)

Take in account that the first step behaves a bit different depending on the authentication method. For example, in OAuth2, when you tap the button you are performing more than a "check". 

as reference: iOS-neo uses "Continue" to move forward to the following step (URL -> credentials -> log in)

In any case, we can face this UX issue in a separated discussion.

jesmrec:
### (2)

On the other side, the current feature seems to have branding concerns. At least, moving the changes in `setup.xml`.

davigonz:
About what was commented in https://github.com/owncloud/android/pull/2235#issuecomment-409293117 and https://github.com/owncloud/android/pull/2235#issuecomment-409599937, we could do something like the gif below:

[](https://user-images.githubusercontent.com/15261363/43789705-e32574a6-9a70-11e8-9fff-6a0d896a196d.gif)

Here we have hidden the `Connect button` till the server is properly checked.

What do you think? @jesmrec @michaelstingl @pmaier1

michaelstingl:
@davigonz What does trigger the server check? Every keystroke? What happens if I have `demo.owncloud.org` AND `demo.owncloud.org/testnextversion`? How can I connect to the 2nd one?

davigonz:
> What does trigger the server check? Every keystroke?

@michaelstingl No, the server check will start after pressing any of the parts highlighted in the screenshot below. The most used option would likely be the **Next** button in the keyboard.

[](https://user-images.githubusercontent.com/15261363/43832701-8eec4764-9b08-11e8-8dae-62f2c4826c84.png)

> What happens if I have demo.owncloud.org AND demo.owncloud.org/testnextversion? How can I connect to the 2nd one?

You can repeat the same steps to connect the 2nd one

[](https://user-images.githubusercontent.com/15261363/43833000-878af12c-9b09-11e8-9ffa-097789321d36.gif)

davigonz:
> On the other side, the current feature seems to have branding concerns. At least, moving the changes in setup.xml.

@jesmrec done

michaelstingl:
hm, okay. let's try this… (can we test with a public beta?)

@pmaier1 @voroyam @cdamken @patrickjahns @pako81 How does this button-less behaviour feel for you as everyday Android users?

voroyam:
From my pow:

There are no apps to my knowledge that behave the same - button-less. So it would be unexpected to look for buttons on the screen and then have to find it on the key board.

Also this is the login screen / connection screen. I have seen this once I configured my new phone and never again, so I don't think I can provide important feedback aside from what I already wrote.

jesmrec:
@davigonz does it affect to credential edition mode?

theScrabi:
Well there are apps similar to our one that show a little arrow on the end of the eddit field. This arroe behaves as the connect button.

davigonz:
> There are no apps to my knowledge that behave the same - button-less. So it would be unexpected to look for buttons on the screen and then have to find it on the key board.

Thank you @voroyam for your opinion, I agree with you but the previous login behaviour was even worse in my opinion, since there was a "Connect" button that could not be used to check the server and was visible, making the users press on it and see that nothing happens. 

> Well there are apps similar to our one that show a little arrow on the end of the eddit field. This arrow behaves as the connect button.

Let's try with something like this

[](https://user-images.githubusercontent.com/15261363/43852329-3c640dee-9b3d-11e8-9354-88de6cacb69d.gif)

jesmrec:
I think that the arrow does not add more value to the view, since the user can tap everywhere in the screen, apart of the key in the keyboard.  The approach here: https://github.com/owncloud/android/pull/2235#issuecomment-411121913 is the one that straightforward the user during the process in my opinion.

davigonz:
@jesmrec all the login icons are already prepared to be brandeable, so what is left is whether we use the arrow in the server address field or not.

jesmrec:
Regarding the use of the arrow to confirm the typed URL, more thoughts are welcome: @michaelstingl @sd1998 @pmaier1 ...

michaelstingl:
@jesmrec let's try this…

jesmrec:
@michaelstingl you mean with or without the arrow

michaelstingl:
Let's try the arrow…

jesmrec:
ok, we will include it also for the beta.

jesmrec:
### (4)

1. Enter correct URL
2. When credentials fields are displayed (soft keyboard), scroll the view

Current: ownCloud logo is cut
Expected: ownCloud logo is not cut:


aug-21-2018 18-29-37 1

jesmrec:
### (5)

Suggestion: with the changes in `setup.xml` i suggest to set:

`<color name="login_background_color">@color/owncloud_blue</color>`

instead of 

`<color name="login_background_color">@color/white</color>`

because in case of `false` in the `use_login_background_image` parameter, the login view is totally white.

theScrabi:
I think the alternative for [(4)](https://github.com/owncloud/android/pull/2235#issuecomment-414738777) would be to remove the could once the keyboard is shown, however that's rather weird.

I see it like this. If you get to the login screen for the first time, you see all including the cloud. But once you tab into a filed the keyboard shows up. here it's important to enter your credentials, not to have the cloud fully visible. By letting it "cut" you may also show the user that he can scroll the layout.

davigonz:
> If you get to the login screen for the first time you see all, including the cloud, but once you tab into a filed the keyboard shows up. here it's important to enter your credentials, not to have the cloud fully visible. By letting it "cut" you may also show the user that he can scroll the layout.

Agree with this, for me is not a bad behaviour.

jesmrec:
@theScrabi of course, it is not an important issue because in the moment the keyboard raises, the focus is on credentials and not in background colors and images. But take in mind that we should not include regression issues in the app, and all suggestions and behaviours are susceptible to be reported regardless they will or won't be fixed.

davigonz:
@jesmrec bugs and improvements done

jesmrec:
all stuff fixed. Approved.

davigonz:
To be included in 2.9 version

Issue #2238, opened 2018-06-12
[2.8.0] New release


jesmrec:
### TASKS
- [x] [WIZARD] Create new steps with important features included in 2.8.0 https://github.com/owncloud/android/issues/2244
- [x] [COORDINATION] Ping community guys to publish a blog post, at least one week before releasing.
 - [x] [GIT] Create branch release_0.9.20 in owncloud/android-library from master, if needed
 - [x] [GIT] Create branch release_2.8.0 in owncloud/android from master 
 - [x] [DEV] Update version number and name in build.gradle in android module
 - [x] [DEV] Update version number and name in build.gradle in oc_jb_workaround module 
 - [x] [DIS] Create changelog file (< 500 chars) and add to CHANGELOG.md in owncloud/android 
 - [x] [QA] Design Test plan @jesmrec
 - [x] [QA] Regression Test plan  @jesmrec 
 - [x] [GIT] Merge branch release_0.9.20 in owncloud/android-library, into master, if needed
 - [x] [GIT] Merge branch release_2.8.0 in owncloud/android, into master 
 - [x] [DIS] Generate test APKs files from branch master in owncloud/android and perform some basic operations.
 - [x] [GIT] Create and sign tag oc-android-2.8.0 in HEAD commit of master branch, in owncloud/android 
 - [x] [GIT] Create and sign tag oc-android-library-0.9.20 in HEAD commit of master branch, in  #owncloud/android-library 
- [x] [DOC] Include new changelog on Releases section of github android repo by drafting the new release 2.8.0 
 - [x] [DIS] Generate final APK files (including oc_jb_workaround) from signed commit in owncloud/android
 - [x] [DIS] Upload & publish release APK and changelog in Play Store
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [DIS] push OC JB workaround app to Play Store.
 - [x] [DIS] Upload APKs to repo
 - [x] [SPREADTHEWORD] Write a topic in central about the release. https://central.owncloud.org/t/owncloud-android-app-2-8-0-released/
 - [x] [COORDINATION] Write in corresponding channels.
 - [x] [COORDINATION] Write email to release-coordination mailing list.
 - [x] [SPREADTHEWORD] Write a post in the blog
 - [ ] [DOC] Update documentation with the new features. @settermjd 
 - [x] Publish app in F-Droid 
 - [x] [GIT] merge master branch into stable, in owncloud/android-library
[...]

PR #2239, opened 2018-06-13
direcly show login screen on saml expiration

theScrabi:
fixes https://github.com/owncloud/android/issues/2238#issuecomment-396660454

### DIFF
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -56,6 +56,8 @@ import android.view.View;
 import com.owncloud.android.AppRater;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.authentication.AccountAuthenticatorActivity;
+import com.owncloud.android.authentication.AuthenticatorActivity;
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -56,6 +56,8 @@
 import com.owncloud.android.AppRater;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.authentication.AccountAuthenticatorActivity;
```
davigonz: Please delete this dependency and the one below since you are not using them

davigonz:
@theScrabi please review the requested changes and rebase with master

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -56,6 +56,7 @@
 import com.owncloud.android.AppRater;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.authentication.AuthenticatorActivity;
```
davigonz: This is not being use either

jesmrec:
SAML expirations redirects right now to the login view instead of the list of servers. 

Approved

Issue #2247, opened 2018-06-22
Inappropriate Message when creating a folder with an existing folder name


sd1998:
### Actual behaviour
When trying to create a folder with the same name as one that has already been created. The snackbar message states "Method not allowed". 

### Expected behaviour
I think the message should be a bit more informative like "Folder with same name exists" of another forlder should be created with suffix "<folder name> + 1".

### Steps to reproduce
1. Create a folder named "xyz" by selecting option from the FAB in FileDisplayActivity.
2. Create another folder with the name "xyz" following the similar procedure.

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes.

![41779120-6f65b99a-764e-11e8-9eaf-2eaa7eb93e5c](https://user-images.githubusercontent.com/1891273/41834718-8217ef50-7855-11e8-891e-79c32fe2b65d.png)

sd1998:
@davigonz @jesmrec  I think this is an issue and if you agree with this then I would like to make the change its just a minor one.

jesmrec:
go ahead

sd1998:
@jesmrec Which one of the following should it be? The message should be a bit more informative like "Folder with same name exists"? or another folder should be created with suffix " + 1".?

michaelstingl:
What error does the server send in the response? Could you post it here?

jesmrec:
The response of the MKCOL request is a 405 error with:

`<s:message>The resource you tried to create already exists</s:message>`

PR #2248, opened 2018-06-26
Message fixed

sd1998:
Fix to Issue #2247

### DIFF
--- a/src/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/src/com/owncloud/android/operations/CreateFolderOperation.java
@@ -55,10 +55,6 @@ public class CreateFolderOperation extends SyncOperation {
         mCreateFullPath = createFullPath;
     }
 
-    public String getNewFolderName(){
-        return new OCFile(mRemotePath).getFileName();
-    }
-
     @Override
     protected RemoteOperationResult run(OwnCloudClient client) {
         CreateRemoteFolderOperation createRemoteFolderOperation = new CreateRemoteFolderOperation(
--- a/src/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/src/com/owncloud/android/operations/CreateFolderOperation.java
@@ -24,6 +24,7 @@ package com.owncloud.android.operations;
 
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.lib.common.OwnCloudClient;
+
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.files.CreateRemoteFolderOperation;
@@ -52,6 +53,10 @@ public class CreateFolderOperation extends SyncOperation {
     public CreateFolderOperation(String remotePath, boolean createFullPath) {
         mRemotePath = remotePath;
         mCreateFullPath = createFullPath;
+    }
+
+    public String getRemotePath(){
+        return this.mRemotePath;
     }
 
     @Override
--- a/src/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/src/com/owncloud/android/operations/CreateFolderOperation.java
@@ -3,10 +3,7 @@
  *
  *   @author David A. Velasco
  *   @author masensio
-<<<<<<< HEAD
  *   @author David González Verdugo
-=======
->>>>>>> Code review fixes
  *   Copyright (C) 2018 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
--- a/src/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/src/com/owncloud/android/operations/CreateFolderOperation.java
@@ -3,7 +3,10 @@
  *
  *   @author David A. Velasco
  *   @author masensio
+<<<<<<< HEAD
  *   @author David González Verdugo
+=======
+>>>>>>> Code review fixes
  *   Copyright (C) 2018 ownCloud GmbH.
  *
  *   This program is free software: you can redistribute it and/or modify
@@ -55,8 +58,8 @@ public class CreateFolderOperation extends SyncOperation {
         mCreateFullPath = createFullPath;
     }
 
-    public String getRemotePath(){
-        return this.mRemotePath;
+    public String getNewFolderName(){
+        return new OCFile(mRemotePath).getFileName();
     }
 
     @Override
--- a/src/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileActivity.java
@@ -53,6 +53,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode;
 import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.operations.CreateFolderOperation;
 import com.owncloud.android.operations.CreateShareWithShareeOperation;
 import com.owncloud.android.operations.GetSharesForFileOperation;
 import com.owncloud.android.operations.RemoveShareOperation;
@@ -339,6 +340,10 @@ public class FileActivity extends DrawerActivity
 
             result.getData();
 
+        } else if(operation instanceof CreateFolderOperation){
+            if(result.isException()){
+                showSnackMessage("Hey");
+            }
         }
     }
 
--- a/src/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileActivity.java
@@ -340,10 +340,6 @@ public class FileActivity extends DrawerActivity
 
             result.getData();
 
-        } else if(operation instanceof CreateFolderOperation){
-            if(result.isException()){
-                showSnackMessage("Hey");
-            }
         }
     }
 
--- a/src/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileActivity.java
@@ -53,7 +53,6 @@ import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode;
 import com.owncloud.android.lib.common.utils.Log_OC;
-import com.owncloud.android.operations.CreateFolderOperation;
 import com.owncloud.android.operations.CreateShareWithShareeOperation;
 import com.owncloud.android.operations.GetSharesForFileOperation;
 import com.owncloud.android.operations.RemoveShareOperation;
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1660,7 +1660,6 @@ public class FileDisplayActivity extends FileActivity
                 showSnackMessage(
                         ErrorMessageAdapter.getResultMessage(result, operation, getResources())
                 );
-
             } catch (NotFoundException e) {
                 Log_OC.e(TAG, "Error while trying to show fail message ", e);
             }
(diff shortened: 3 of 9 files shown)


src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -339,6 +340,10 @@ public void onRemoteOperationFinish(RemoteOperation operation, RemoteOperationRe
 
             result.getData();
 
+        } else if(operation instanceof CreateFolderOperation){
+            if(result.isException()){
+                showSnackMessage("Hey");
```
theScrabi: :D ... what is the purpose of this?

sd1998: Really Sorry forgot to fix this. Actually was trying to see where the snackbar message was generated.

sd1998: @theScrabi I have fixed it now.

src/com/owncloud/android/operations/CreateFolderOperation.java
```diff
@@ -55,9 +52,11 @@
     public CreateFolderOperation(String remotePath, boolean createFullPath) {
         mRemotePath = remotePath;
         mCreateFullPath = createFullPath;
-        
     }
 
+    public String getRemotePath(){
+        return this.mRemotePath;
```
davigonz: You don't need `this.mRemotePath`, `return mRemotePath` is enough.

src/com/owncloud/android/operations/CreateFolderOperation.java
```diff
@@ -21,12 +21,9 @@
```
davigonz: Please, update the copyright year at the top of the class.

src/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -53,6 +53,7 @@
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode;
 import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.operations.CreateFolderOperation;
```
davigonz: This import is not being used.

src/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -571,7 +570,7 @@ private void onCreateFolderOperationFinish(CreateFolderOperation operation,
         } else {
             try {
                 showSnackMessage(
-                    ErrorMessageAdapter.getResultMessage(result, operation, getResources())
+                        FileStorageUtils.getNewFolderName(((CreateFolderOperation) operation).getRemotePath()) + " " + getResources().getString(R.string.folder_already_exists)
```
davigonz: Same here

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1632,7 +1632,7 @@ private void onCreateFolderOperationFinish(CreateFolderOperation operation,
         } else {
             try {
                 showSnackMessage(
-                    ErrorMessageAdapter.getResultMessage(result, operation, getResources())
+                        FileStorageUtils.getNewFolderName(((CreateFolderOperation) operation).getRemotePath()) + " " + getResources().getString(R.string.folder_already_exists)
```
davigonz: Can't you get the folder name from https://github.com/sd1998/android/blob/master/src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java#L106 by calling `operation.getFileName()` and avoid creating a new method in FileStorageUtils?

src/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -396,7 +397,7 @@ private void onCreateFolderOperationFinish(
         } else {
             try {
                 showSnackMessage(
-                    ErrorMessageAdapter.getResultMessage(result, operation, getResources())
+                        FileStorageUtils.getNewFolderName(((CreateFolderOperation) operation).getRemotePath()) + " " + getResources().getString(R.string.folder_already_exists)
```
davigonz: Same here

davigonz:
Thanks @sd1998 for fixing this message, have a look at the requested changes

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
sd1998: @davigonz The getFileName() method is applicable to UploadFileOperation.class and not for the CreateFolderOperation.class thus cannot use it.

davigonz: I see...but since what you are doing is generating an error message, don't you think it would be better to add a new if case in https://github.com/sd1998/android/blob/master/src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java#L105  - `if (operation instanceof CreateFolderOperation)` and create a new method in `CreateFolderOperation` with something like:

```java
public String getFolderToCreateName() {
    return new OCFile(mRemotePath).getFileName();
}
```

davigonz:
@sd1998 any update on this? If you need any help, ping us again :blush:

sd1998:
@davigonz Sorry for the delay on this. Was busy with some work. Have started working on it now.

jesmrec:
Thanks @sd1998. Ping us when it is ready!

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
sd1998: @davigonz I just changed the getRemotePath() function in CreateFolderOperation.java to getNewFolderName() which would do exactly as you suggested in the getFolderToCreateName() function in previous [comment](https://github.com/owncloud/android/pull/2248#discussion_r202600158). Hence not requiring the getNewFolderName() in FileStorageUtils.java hence string manipulation.

sd1998:
@jesmrec I have made the code review changes.

jesmrec:
@davigonz, could you check the code? From QA point of view,  it is already fixed.

sd1998:
@davigonz Has the code review started?

davigonz:
I think this is not a proper solution since the error message "Folder already exists" appears whenever create folder operation fails, even if there's no folder with the same name. 

Have a look at this scenario:

1. Disable wifi connection.
2. Create folder with a non existing name.
3. Error is "Test already exists" when it should be something like "There's no connection"

[](https://user-images.githubusercontent.com/15261363/45144475-89e21d80-b1be-11e8-96b0-2b23cf586454.gif)

Take into account that errors are usually handled in [ErrorMessageAdapter](https://github.com/owncloud/android/blob/master/src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java) as I suggested in https://github.com/owncloud/android/pull/2248#discussion_r202600158. 

Therefore, you could use the response body from the `MKCOL` executed within CreateRemoteFolderOperation, which already includes the message "The resource you tried to create already exists" when the operation fails because that reason.

jesmrec:
@davigonz good catch

sd1998:
@davigonz I will fix this.

jesmrec:
Thanks, we will review the new code and pass QA with the changes.

src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
```diff
@@ -126,6 +127,12 @@ public static String getResultMessage(final RemoteOperationResult result,
             return result.getData().get(0).toString();
         }
 
+        if(operation instanceof CreateFolderOperation && result.getHttpPhrase() != null && result.getHttpPhrase().length() > 0){
```
davigonz: It's possible that other 405 errors are triggered when creating a folder but not necessarily be related to a duplicated folder. To take into account all the existing 405 errors retrieved by the server, we could read the specific message from the response body as I suggested in https://github.com/owncloud/android/pull/2248#issuecomment-419012250. 

[Here](https://github.com/owncloud/android-library/blob/8e357d78119921afdf8bb3466c168b6165f480a0/src/com/owncloud/android/lib/common/operations/RemoteOperationResult.java#L263) you have some specific codes that we created in similar scenarios and for this case we could check whether `mHttpCode` is `HttpStatus.SC_METHOD_NOT_ALLOWED` and if so, parse the error message and set a new code like SPECIFIC_METHOD_NOT_ALLOWED.

If everything went well, the error message "The resource you tried to create already exists" should appear in the snackbar.

davigonz:
Some changes requested here as well

src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
sd1998: @davigonz As you have suggested I could get the message from the CreateFolderOperation itself so that could be instead right? What I mean is that instead of checking `result.getHttpParse().length() > 0` I could instead check `result.getHttpParse().equals("The resource you tried to create already exists")` that would also work right?

davigonz: > What I mean is that instead of checking result.getHttpParse().length() > 0 I could instead check result.getHttpParse().equals("The resource you tried to create already exists") that would also work right?

You mean to set a more specific message than "The resource you tried to create already exists" with the name of the folder and so on. Think that using the message coming from the server is enough and we don't need to be aware about the changes in that server message, what could break the condition you said above.

jesmrec:
To include this PR in 2.9.0, we need to finish it within the following days. Otherwise, we will move to next version.

CC @sd1998 @davigonz

src/com/owncloud/android/ui/errorhandling/ErrorMessageAdapter.java
sd1998: @davigonz I don't think the condition above would break as the output the response being returned is "The resource you tried to create already exists" checking for this response along with checking if the operation type was CreateFolderOperation should work.

davigonz: If server decides to change that message to _The resource you are trying to create already exists_ for example, what you suggested in https://github.com/owncloud/android/pull/2248#discussion_r223694676 would stop working. What I mean is that to make this error handling easier, we could just use the message retrieved from the server, following the steps I defined in https://github.com/owncloud/android/pull/2248#discussion_r223656714. 

Have a look at [this PR](https://github.com/owncloud/android-library/pull/166), in which we are reading antivirus message from response body, it might be helpful

sd1998: @davigonz Yes I tried this but for this to work I had to add 
```
case HttpConstants.HTTP_SERVICE_UNAVAILABLE:
                parseErrorMessageAndSetCode(
                        httpMethod.getResponseBodyAsString(),
                        ResultCode.SPECIFIC_SERVICE_UNAVAILABLE
                );
                break;
```
to the RemoteOperationResult.java file in the owncloud-android-library so that I could get the error message from the server response which can they be displayed on the snackbar directly  but I am unable to commit these changes.

davigonz: I can help you with this, you would need to follow the next steps:
1. Go to owncloud-android-library: `cd owncloud-android-library` from ownCloud Android app code, I guess you will be in branch `sd1998:Fix-2247`.
2. Create branch in owncloud-android-library
3. Add + commit + push changes
4. Go back to ownCloud Android app code (`cd ..`)
5. Update the library reference: `git add owncloud-android-library`
6. Commit + push changes

sd1998: @davigonz Thanks for the help. I have made the [PR](https://github.com/owncloud/android-library/pull/207) to the owncloud-android-library repo.

sd1998:

davigonz:
Thanks for applying the changes, ready to QA @jesmrec

jesmrec:
@sd1998 there is an error in the reference to the android-library commit. The app is pointing to a non-existing commit 

`Unable to checkout '106b4fd9e55d208c2089bea89db81daf326fc69f' in submodule path 'owncloud-android-library'`

Could you have a look?

jesmrec:
Correct error messages when:

- Create a folder with an existing name
- Rename an item with an existing item
- Create a folder with no connection 

Approved on my side

PR #2261, opened 2018-07-18
make adaptive icon based on png

theScrabi:
This will make the adaptive Icon ownBrander ready.


davigonz:
One question more, wouldn't it be a better practice to place the app icons in **mipmap** folders instead of **drawable** ones? 

In adaptive icons documentation, they are using mipmap folders to place the foreground icon: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive

PR #2267, opened 2018-07-26
add fdroid aditional styles

theScrabi:
This will improve how the android app is precented in fdroid.

This is what it looks like atm:
screenshot_1532599628

This is what it will look like after:
fdroid_foo


davigonz:
Code approved, no QA needed, this will appear in F-Droid and should be merged within 2.9 release tasks, not before.

fastlane/metadata/android/en-US/full_description.txt
```diff
@@ -0,0 +1,6 @@
+ownCloud is a free software package you can install on a server to manage files,
```
mueller-ma: Why do you have all these newlines here? IMO it's better to let the end user program decide when to make a newline.

davigonz: This description is what will appear in the ownCloud tab in F-Droid, I don't get this comment, sorry

mueller-ma: This text contains a lot of line breaks. Wouldn't it be better to remove all line breaks and let the F-Droid client and the F-Droid Website decide where to insert line breaks based on the text size and screen size?

davigonz:
No QA needed , changes will be applied in F-Droid

fastlane/metadata/android/en-US/full_description.txt
mueller-ma: Screenshot_20190314-172249
@davigonz Please see my screenshot. With all the needlessly newlines it looks awful.

davigonz: Understood, thanks for the screenshot, more clear now, would you like to open a PR to fix it? Thanks

PR #2274, opened 2018-08-06
Implement local search

davigonz:


res/values/strings.xml
```diff
@@ -3,6 +3,7 @@
 
     <string name="about_android">%1$s Android app</string>
     <string name="about_version">version %1$s</string>
+    <string name="actionbar_search">Search</string>
```
jesmrec: i would use here "search in the current folder" to avoid misunderstandings

davigonz: Done

jesmrec:
### (1) 

Suggestion: the current hint message is `Search in the current folder`. This message is a bit long and it is truncated in devices with a not very wide screen. On the other hand, this feature performs a filter, so  the word `Searck`, maybe make users confused.

Suggestions:

- `Search this folder` (iOS option)
- `Filter this folder`
- `Filter file name`
- ... (other suggestions)

jesmrec:
### (2)

Steps:

1. Perform a search
2. Open drawer
3. Close drawer

Current: blank page even though the search bar was hidden
Expected: either results of the search displayed with search bar visible, or file list displayed with search bar hidden and lens icon visible.

Samsung galaxy s7 edge android v7

jesmrec:
### (3)

1. Perform a search which results contain one folder
2. Browse into the folder

Current: searching bar visible with the typed pattern, but the displayed list after browsing does not match the pattern
Expected: Closing the search bar when browsing. This is a local search (in the current filter), so only make sense to keep until the next browsing

Samsung galaxy s7 edge android v7

jesmrec:
### (4) 

Improvement:

If the typed pattern does not match any item in the folder, the message `Nothing in here. Upload something!` appears. Such message fits for the case of empty folder, in this case a different message would fit better like:

- `No matches` (iOS)
- `There is no results for this search` (iOS)
-  `No results found`
- ... (suggestions)

davigonz:
@jesmrec all the bugs are fixed

jesmrec:
Approved

Issue #1911, opened 2017-03-02
Include sorting method in file picker view when uploading


hutber:
### Actual behaviour
When in the `Upload` file view all files are sorted `Oldest First`. This means a very long time to scrolling to the bottom of the view.

### Expected behaviour
Have the option to sort how desired
or simply sort by newest first
 
### Steps to reproduce
1. Select a folder
2. Click the `+`
3. Have to scroll down

### Environment data
Android version:
All
Device model: 
N/a

davivel:
> When in the Upload file view all files are sorted Oldest First

Really? This could be a bug. 

Didn't we implement as 'Newest first', @jesmrec ?

jesmrec:
@davivel i think that @jamiehutber means the folder picker view.

The uploads view is sorted by "Newest first".

jamiehutber:
Ye, I was just going to say. You have updated the file view, but its when you are picking files as very often my instant upload fails the upload. I'd say around 40% of the time.

davivel:
Oh, I see. Seems an interesting proposal. 

> very often my instant upload fails the upload. I'd say around 40% of the time.

Sorry to read that. What is the problem, they fail in the upload or they are not even detected by the app?

What's your model device and Android version? There are some issues open about instant uploads in the repository, maybe one of them is of your interest, just to track improvements in the instant uploads.

jamiehutber:
Ye, its very `upsetting` :p when having to scroll for 40 seconds to get to the bottom and even then, I end up at the bottom of the videos when I want to manually upload the images that haven't sync'd properly.

ye :D I don't think those problems are in scope for this ticket. I will do some searching.

Just to say, they sometimes aren't detected, other times they fail mid way through upload I believe. All through wifi, `Nexus 5x` and latest version of android.

jamiehutber:
[...]

PR #2305, opened 2018-10-04
Sorting menu in file picker when uploading.

sd1998:
Sort menu added in UploadFilesActivity.
Fix to issue #1911.


sd1998:
@jesmrec I have implemented the feature.

sd1998:
@davigonz The sorting implementation exists for OCFile object as the sorting options are for files that have been uploaded. While implementing sorting for files that have to be selected from the device's internal memory I have to do deal with File object for which the previous implementations were not working due difference in the way both the objects are implemented. Hence I had to do the implementation from scratch. I could try implementing the sorting functionality using generics and could use "instanceof" for distinction between File object and OCFile object and based on that call appropriate methods this would I allow us to use single implementation with some modifications should I do this instead?

davigonz:
@sd1998 You will need to have several menus because the preferences might be different, a user can have his list of files sorted by name and his list of upload sorted by date but you could reuse the sorting logic for both menus.

sd1998:
@davigonz Alright I will make separate preferences for both the menus. And for reusing the logic so lets say that If the user has sorted the list of files order by name in that case to get the path of the file(OCFile object) we have to use the getRemotePath() function on the other hand for sorting the files in case of list of upload to get the file path(File object) we will need the getAbsolutePath() function thus requiring us to have different implementations for both the cases.

sd1998:
@davigonz I have added different preference for sorting during uploads thus allowing the user to sort the content uploaded and the content they want to upload in a different order and have also reused the previous implementation of sorting using the "instanceof" approach as I previously discussed [#2305Comment](https://github.com/owncloud/android/pull/2305#issuecomment-427336109).

davigonz:
Thanks @sd1998 , let's have a look

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -211,86 +217,163 @@ public static RemoteFile fillRemoteFile(OCFile ocFile){
...
-        case SORT_NAME:
-            files = FileStorageUtils.sortByName(files);
-            break;
-        case SORT_DATE:
-            files = FileStorageUtils.sortByDate(files);
-            break;
-        case SORT_SIZE:
-            files = FileStorageUtils.sortBySize(files);
-            break;
+    public static Vector<OCFile> sortFolderFileDisp(Vector<OCFile> files) {
```
davigonz: I think that instead of including methods to work with files, maybe is better to keep using OCFile everywhere and create OCFile from files by using https://github.com/dlech/owncloud-android/blob/master/src/com/owncloud/android/datamodel/OCFile.java#L69, what do you think?

davigonz:
@sd1998 I requested a few changes, most of them tiny

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -574,8 +576,8 @@ public boolean onOptionsItemSelected(MenuItem item) {
             case R.id.action_sort_descending: {
                 item.setChecked(!item.isChecked());
                 final boolean sortAscending = !item.isChecked();
-                PreferenceManager.setSortAscending(sortAscending, this);
-                switch (getSortOrder(this)) {
+                PreferenceManager.setSortAscending(sortAscending, this,FileStorageUtils.FILE_DISPLAY_SORT);
```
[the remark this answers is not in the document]
sd1998: Should it not be space before FileStorageUtils.FILE_DISPLAY_SORT instead of space after this

src/com/owncloud/android/utils/FileStorageUtils.java
sd1998: @davigonz I have already tried this and this already has a problem which is:-
When we convert File object to OCFile object the name of the file can be accessed with `getFileName()` method but the other attributes like last modified time and file size cannot be obtained using the corresponding methods `getModificationTimestamp()` and `getFileLength() `of OCFile object hence affecting the sortBySize and sortByDate functionality. Here is a small experiment:- 
This is the LocalFileListAdapter.java where I have used the OCFile constructor to create a OCFile object of the first File object present in the mFiles Array.
screenshot 2018-10-09 at 6 59 33 pm
Here are the logs.
screenshot 2018-10-09 at 6 59 40 pm
As you an see when being converted to OCFile the methods `getModificationTimestamp()` and `getFileLength()` giving output as 0 which is not the actual case. Hence the problem.

sd1998: @davigonz I have made rest of the changes other than this.

sd1998: @davigonz Yes this could work I will work on this implementation.

sd1998:

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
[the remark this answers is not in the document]
davigonz: Yes, that's what I meant

src/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.java
```diff
@@ -85,7 +85,7 @@ public PreviewImagePagerAdapter(FragmentManager fragmentManager, OCFile parentFo
         // TODO Enable when "On Device" is recovered ?
         mImageFiles = mStorageManager.getFolderImages(parentFolder/*, false*/);
 
-        mImageFiles = FileStorageUtils.sortFolderFileDisp(mImageFiles);
+        mImageFiles = FileStorageUtils.sortFolder(mImageFiles,FileStorageUtils.mSortOrderFileDisp,FileStorageUtils.mSortAscendingFileDisp);
```
davigonz: Some spaces between parameters here...

src/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -289,7 +291,8 @@ public void setSortOrder(Integer order, boolean isAscending) {
         FileStorageUtils.mSortOrderUpload = order;
         FileStorageUtils.mSortAscendingUpload = isAscending;
         if (mFiles != null && mFiles.length > 0) {
-            mFiles = FileStorageUtils.sortFolderUpload(mFiles);
+            mFiles = transformVecOCFile(FileStorageUtils.sortFolder(transformFileArray(mFiles),
```
davigonz: Same here. Besides, you could split the parameters into different lines to make it more readable:

```
mFiles = transformVecOCFile(
           FileStorageUtils.sortFolder(transformFileArray(mFiles),
           FileStorageUtils.mSortOrderUpload,
           FileStorageUtils.mSortAscendingUpload)
);
```

src/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -469,7 +469,7 @@ private void startSyncFolderOperation(OCFile folder) {
         FileStorageUtils.mSortOrderFileDisp = PreferenceManager.getSortOrder(this, FileStorageUtils.FILE_DISPLAY_SORT);
         FileStorageUtils.mSortAscendingFileDisp = PreferenceManager.getSortAscending(this, FileStorageUtils.FILE_DISPLAY_SORT);
 
-        files = FileStorageUtils.sortFolderFileDisp(files);
+        files = FileStorageUtils.sortFolder(files,FileStorageUtils.mSortOrderFileDisp,FileStorageUtils.mSortAscendingFileDisp);
```
davigonz: There's some pending spaces between parameters here.

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -206,12 +216,77 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 }
                 break;
             }
+            case R.id.action_sort_descending:
+                item.setChecked(!item.isChecked());
+                boolean isAscending = !item.isChecked();
+                com.owncloud.android.db.PreferenceManager.setSortAscending(isAscending,this,FileStorageUtils.UPLOAD_SORT);
```
davigonz: This is line exceeds 120 characters as well

davigonz:
Great job @sd1998 with the latest commits, we have now less and cleaner code. I've just requested a couple of changes more, once is ready, we can move forward to QA stage.

src/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -298,4 +301,25 @@ public void setSortOrder(Integer order, boolean isAscending) {
             notifyDataSetChanged();
         }
     }
+
+    private Vector<OCFile> transformFileArray(File[] files){
```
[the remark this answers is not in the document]
sd1998: @davigonz I have further changed the names to `transformFilesArrToVecOCFiles` and `transformVecOCFilesToFileArr` these provide much better clarity into what the function is doing.

sd1998:

davigonz:
> I have made the changes.

Thanks, approved and ready to pass QA @jesmrec

jesmrec:
The option to sort by size affects only files and not folders. Is this something fixable or depends on any system restriction? @sd1998 @davigonz

jesmrec:
To include this PR in 2.9.0, we need to finish it the following days. Otherwise, we will move to next version.

CC @sd1998 @davigonz

davigonz:
>  I just discovered a bug in my implementation and have fixed it.

Great @sd1998! Could you please describe the steps to reproduce so @jesmrec can test it?

jesmrec:
Not only reproduce it, also assuring that the fix did not break other stuff that worked before ;)

This is very close to be  approved and merged @sd1998

sd1998:
@jesmrec @davigonz The steps to reproduce the bug are:-

(i) Open the main FAB and select uploads
(ii) From the bottom sheet select files
(iii) Select the sort order from the options in sort menu
(iv) Once the files are sorted open any one of the non-empty directories 
(v) The files in the directory are not sorted in the order that was selected.

I have fixed this issue.

jesmrec:
Bug checked and correctly fixed. all sorting methods working properly. Approved from my side.

Thanks again @sd1998

davigonz:
Great, please @sd1998 rebase with master and we merge it

sd1998:
@davigonz Now the Travis CI build is passing previously the owncloud-android-library was not pointing to the correct commit hash. I have fixed it now.

davigonz:
Merged @sd1998 , awesome work 👏

PR #2306, opened 2018-10-04
Fix available offline in Android 8+ by using jobs

### LINKED ISSUE #2304: Available offline feature stops working after rebooting in Android 8 and above
### Actual behaviour
- Available offline files are not automatically uploaded after being edited.

### Expected behaviour
- Available offline files are automatically uploaded after editing them. 
 
### Steps to reproduce
1. Set a file, .txt for example, as available offline and wait till is properly synchronized. 
2. Reboot the device.
3. Edit the file and save it.

(url: https://demo.owncloud.org, user: test, password: test) Yes

### Environment data
Android version: 8.1.0

Device model: Nexus 5X

Stock or customized system: Stock

ownCloud app version: 2.8.0

ownCloud server version: it doesn't matter for this case

davigonz:
Fixes #2304 
Needs https://github.com/owncloud/android-library/pull/196
____
**BUGS & IMPROVEMENTS**

- [x] (1) Available offline job tries to synchronize folders themselves instead of the files inside. [FIXED] 
- [x] (2) https://github.com/owncloud/android/issues/2315 [FIXED]
- [X] (3) Sync of av. offline folders https://github.com/owncloud/android/pull/2306#issuecomment-431799933 
- [X] (4) Encoding of colon character in logs: https://github.com/owncloud/android/pull/2306#issuecomment-432283674
- [X] (5) [IMPROVEMENT] Notification instead of intrusive dialog to solve file conflicts. https://github.com/owncloud/android/pull/2306#issuecomment-432323908 [DONE] 
- [x] (6) Removing account with av. offline should stop the job https://github.com/owncloud/android/pull/2306#issuecomment-432333209 [FIXED] 
- [x] (7) [IMPROVEMENT] Show a notification for each file conflict [DONE]


davigonz:
This is ready to code review

jesmrec:
### (1)

1.Set  a folder as av. offline
2. Wait 15 minutes

Current: A notification about the folder failed to download
Expected: Folders are not downloaded, but its content. 

tested with huaweip20L v8

davigonz:
@jesmrec https://github.com/owncloud/android/pull/2306#issuecomment-428624116 ready to test

build.gradle
```diff
@@ -76,7 +76,11 @@ dependencies {
 
     //Zooming Android ImageView.
     implementation 'com.github.chrisbanes:PhotoView:2.0.0'
+
     implementation 'com.gitlab.ownclouders:dav4android:oc_support'
+
+    // Convert Java Objects into JSON and back
+    implementation 'com.google.code.gson:gson:2.8.5'
```
jcmontesmartos: I'd recommend you using Moshi instead

src/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -299,4 +299,4 @@ public void setSourcePath(String sourcePath) {
             mSourcePath = sourcePath;
         }
     }
-}
+}
```
jcmontesmartos: git is complaining because you have removed the latest empty blank line of the file ;-)

src/com/owncloud/android/files/services/AvailableOfflineSyncJobService.java
```diff
@@ -0,0 +1,189 @@
...
+                    execute(storageManager, mAvailableOfflineJobService);
+
+            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
+                // If the user is not running the app, this can be very intrusive so show a notification to solve
+                // the conflicts
+                Intent i = new Intent(mAvailableOfflineJobService, ConflictsResolveActivity.class);
+                i.setFlags(i.getFlags() | Intent.FLAG_ACTIVITY_NEW_TASK);
+                i.putExtra(ConflictsResolveActivity.EXTRA_FILE, availableOfflineFile);
+                i.putExtra(ConflictsResolveActivity.EXTRA_ACCOUNT, mAccount);
+                mAvailableOfflineJobService.startActivity(i);
```
jcmontesmartos: Launching an activity from job service....

This is not a good approach. You know Android restricts starting UI items from services. I'd move to another logic. For example, this job emits an event to start activity

src/com/owncloud/android/files/services/TransferRequester.java
```diff
@@ -126,7 +126,14 @@ private void uploadsUpdate(Context context, Account account, OCFile[] existingFi
         intent.putExtra(FileUploader.KEY_LOCAL_BEHAVIOUR, behaviour);
         intent.putExtra(FileUploader.KEY_FORCE_OVERWRITE, forceOverwrite);
 
-        context.startService(intent);
+        // Since in Android O and above the apps in background are not allowed to start background
+        // services and available offline feature may try to do it, this is the way to proceed
+        if (context instanceof AvailableOfflineSyncJobService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
jcmontesmartos: I'd use ContextCompat

    if (context instanceof AvailableOfflineSyncJobService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
        intent.putExtra(FileUploader.KEY_IS_AVAILABLE_OFFLINE_FILE, true);
    }
    ContextCompat.startForegroundService(context, intent)

jrodgut-sg: it's quite strange to change this logic based on the type f the context, especially when it was just used to start a service. Take into account that the Android context may be passed in very different ways and   Looks like an architecture smell. Probably overloading the uploadsUpdate method with a new flag and passing it around would help. Just to consider, not sure if feasible.

src/com/owncloud/android/operations/SynchronizeFileOperation.java
```diff
@@ -241,10 +243,18 @@ private void requestForUpload(OCFile file) {
...
-        i.putExtra(FileDownloader.EXTRA_FILE, file);
-        mContext.startService(i);
+        Intent intent = new Intent(mContext, FileDownloader.class);
+        intent.putExtra(FileDownloader.KEY_ACCOUNT, mAccount);
+        intent.putExtra(FileDownloader.KEY_FILE, file);
+
+        // Since in Android O and above the apps in background are not allowed to start background
+        // services and available offline feature may try to do it, this is the way to proceed
+        if (mContext instanceof AvailableOfflineSyncJobService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            intent.putExtra(FileDownloader.KEY_IS_AVAILABLE_OFFLINE_FILE, true);
```
jrodgut-sg: same as above

src/com/owncloud/android/operations/SynchronizeFileOperation.java
```diff
@@ -253,9 +263,7 @@ public boolean transferWasRequested() {
         return mTransferWasRequested;
     }
 
-
     public OCFile getLocalFile() {
         return mLocalFile;
     }
-
-}
+}
```
jrodgut-sg: same

src/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -697,4 +693,4 @@ private void shareLink(String link) {
         DialogFragment chooserDialog = ShareLinkToDialog.newInstance(intentToShareLink, packagesToExclude);
         chooserDialog.show(mFileActivity.getSupportFragmentManager(), FTAG_CHOOSER_DIALOG);
     }
-}
+}
```
jrodgut-sg: same

src/com/owncloud/android/operations/SynchronizeFileOperation.java
```diff
@@ -241,10 +243,18 @@ private void requestForUpload(OCFile file) {
...
-        i.putExtra(FileDownloader.EXTRA_ACCOUNT, mAccount);
-        i.putExtra(FileDownloader.EXTRA_FILE, file);
-        mContext.startService(i);
+        Intent intent = new Intent(mContext, FileDownloader.class);
+        intent.putExtra(FileDownloader.KEY_ACCOUNT, mAccount);
+        intent.putExtra(FileDownloader.KEY_FILE, file);
+
+        // Since in Android O and above the apps in background are not allowed to start background
+        // services and available offline feature may try to do it, this is the way to proceed
+        if (mContext instanceof AvailableOfflineSyncJobService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
jcmontesmartos: The same as before. Use ContextCompat


    if (context instanceof AvailableOfflineSyncJobService && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
        intent.putExtra(FileUploader.KEY_IS_AVAILABLE_OFFLINE_FILE, true);
    }
    ContextCompat.startForegroundService(context, intent)


BTW, Can this code be refactored so you only have a common section?

build.gradle
davigonz: This looked pretty easy to use, what are the advantages of Moshi?

build.gradle
```diff
@@ -82,8 +82,11 @@ dependencies {
         exclude group: "com.android.support"
     }
 
-    //Zooming Android ImageView.
+    // Zooming Android ImageView.
     implementation "com.github.chrisbanes:PhotoView:2.0.0"
+
+    // Convert Java Objects into JSON and back
+    implementation 'com.google.code.gson:gson:2.8.5'
```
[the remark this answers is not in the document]
davigonz: Done

src/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -2137,7 +2138,7 @@ private OCCapability createCapabilityInstance(Cursor c) {
                 do {
                     file = createFileInstance(cursorOnKeptInSync);
                     accountName = cursorOnKeptInSync.getString(
-                        cursorOnKeptInSync.getColumnIndex(ProviderTableMeta.FILE_ACCOUNT_OWNER)
+                            cursorOnKeptInSync.getColumnIndex(ProviderTableMeta.FILE_ACCOUNT_OWNER)
```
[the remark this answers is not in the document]
davigonz: Is automatically put after moving it from previous line by pressing Enter.

src/com/owncloud/android/files/services/AvailableOfflineSyncJobService.java
davigonz: > I'd move to another logic. For example, this job emits an event to start activity

Well, using a broadcast in this case would not be that trivial because I cannot register the receiver in `ConflictsResolveActivity` since `onCreate()` method is never called before the job detects conflicts, that activity is only started with the code above. 

Besides, I could register the receiver in other place, but how would I unregister it? In an activity `onDestroy()` method would not work because as soon as you kill the app, the broadcast from the job will not work.

> You know Android restricts starting UI items from services

Is there any official Android restrictions about this? I mean, I agree with you this is not the best approach, but think is not because of Android restrictions, the Android  framework allows you to start an Activity from a Service, but because we are interrupting what the user is currently doing, so the best approach would be to use a Notification instead and once the user decides to solve the conflict, open the activity.

src/com/owncloud/android/files/services/AvailableOfflineHandler.java
```diff
@@ -0,0 +1,87 @@
...
+        mContext = context;
+        mAccountName = accountName;
+        mJobScheduler = (JobScheduler) mContext.getSystemService(Context.JOB_SCHEDULER_SERVICE);
+    }
+
+    /**
+     * Schedule a periodic job to check whether recently updated available offline files need to be synchronized
+     */
+    public void scheduleAvailableOfflineJob() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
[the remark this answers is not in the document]
davigonz: Yes, Android Studio shows the next message `Call requires API level 21 (current min is 14)`

jesmrec:
### (3)

Only files are synced by the job. An av. offline folder is not checked

1. Set a file as. av. offline
2. Set a folder as av. offline
3. Wait (at least) 15 minutes

Current: file is checked to be synced. Folder isn't
Expected: Both are checked to be synced

Files individually are synced by the job, no matter where the location is (root or non-root folder)

Android 8.0 Nexus 5X

jesmrec:
### (3)

Only files are synced by the job. An av. offline folder is not checked

1. Set a file as. av. offline
2. Set a folder as av. offline
3. Wait (at least) 15 minutes

Current: file is checked to be synced. Folder isn't
Expected: Both are checked to be synced

Files individually are synced by the job, no matter where the location is (root or non-root folder)

jesmrec:
### (4)

When job is executed, in logs appear:

`File /storage/emulated/0/owncloud/user@10.40.40.199%3A10443/file.txt already synchronized, ignoring`

The colon character is not correctly encoded.

davigonz:
### (5) IMPROVEMENT

I've taken advantage of this issue to improve the behaviour when solving available offline conflicts in background synchronization. Instead of showing an intrusive dialog that could interrupt the user when using any other app, a notification will appear, showing the dialog after tapping it.

jesmrec:
### (6)

Steps:

1. Add two accounts
2. Set some items as av. offline in one of the accounts, The other one will not have any av. offline item
3. After 15 mins, account with av. offline items is correctly synced.
4. Remove the account with av. offline items from the app

Current: After 15 mins, job is executed and try to sync the av. offline items, even though they are already not in the app anymore

Expected: After removing the account, there is no more av. offline items in the app. Job should be stopped.

Tested with Android 8.1 Nexus 5x

jesmrec:
About (5):

Behaviour is correct, so the conflict is shown in a notification and not over the current view. In case there are more than one conflicts, only the last detected is shown in the notification. Some ideas about, not sure if they are feasible or not, or they are improvements to deal in the future:

- In case only one conflict is detected: current behaviour is perfect
- In case more than one conflict is detected: 

   - Show in the notification that many/multiple/several conflicts were detected (including or not the file names), and by clicking on the notification itself. user is lead to the app when any conflict happened.
   - Show a notification for each conflict. In case of a huge number, we can point the user to notification flooding. Maybe not the best idea.

davigonz:
> Show a notification for each conflict. In case of a huge number, we can point the user to notification flooding. Maybe not the best idea.

Let's for for this solution

jesmrec:
Behaviour in Android O is correct. Let's check now if older versions behave correctly as well.

davigonz:
> Therefore, users with Android 4.x can notice different or weird behaviours.

To be more specific, background synchronization - i.e. changes in files automatically detected and synchronized with the server when the app is not active - is not going to work in Android 4.x or lower.

PR #2320, opened 2018-10-19
Make login connection and credentials status texts brandable

davigonz:


jesmrec:
Approved.

Correct color-separation for connection and credential messages, as well as for login button background and primary background.

Issue #2322, opened 2018-10-22
App crashes creating private link


jesmrec:
### Steps to reproduce
1. Install the app
2. Create a private share

### Actual behaviour
App crashes. When app restarts, there is an authentication error and user needs to enter credentials again

### Expected behaviour
Private link created and life goes on

Reproducible in `master` branch (2.9.0-beta2) 
Not reproducible in 2.8.0

PVince81:
it seems the private link was created already, it just did not appear in the app.

jesmrec:
Fixed. Will be released in 2.9.0

PR #2323, opened 2018-10-22
Fix crash when creating public link

davigonz:
Implements #2322
Needs https://github.com/owncloud/android-library/pull/197


jesmrec:
Fixed

PR #2275, opened 2018-08-09
String improvements

mueller-ma:
* Change `ownCloud Android app` to `ownCloud app for Android`

> "Android", or anything confusingly similar to "Android", cannot be used in names of applications or accessory products, including phones, tablets, TVs, speakers, headphones, watches, and other devices. Instead, use "for Android".
    Incorrect: "Android MediaPlayer" or "Ndroid MediaPlayer"
    Correct: "MediaPlayer for Android"

from https://developer.android.com/distribute/marketing-tools/brand-guidelines

* Capitalize `Version 1.2.3`
* Remove commented strings. If you need removed lines of code, git is
your friend ;)
* Remove quotation marks as they aren't shown if unescaped
* Change … to it's unicode representation
* Remove whitespace between … and the previous words
* Remove ellipses on buttons

>  Omit ellipses from menu items or buttons that open a dialog or start a process.

from https://web.archive.org/web/20180103183733/https://material.io/guidelines/style/writing.html

* Use "ok" and "cancel" from the android sdk

PR #2324, opened 2018-10-24
Android Studio 3.2.1

hannesa2:
Needs https://github.com/owncloud/android-library/pull/198

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -23,8 +23,6 @@
     android:versionName="${versionName}">
 
     <uses-sdk
-        android:minSdkVersion="14"
-        android:targetSdkVersion="26"
         tools:overrideLibrary="android.support.customtabs"/>
 
     <!--
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.2'
+        classpath 'com.android.tools.build:gradle:3.2.1'
     }
 }
 
@@ -97,6 +97,9 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
+        minSdkVersion 14
+        targetSdkVersion 26
+
         manifestPlaceholders.versionCode = "20900300"
         manifestPlaceholders.versionName = "2.9.3"
 
--- a/oc_jb_workaround/build.gradle
+++ b/oc_jb_workaround/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.2'
+        classpath 'com.android.tools.build:gradle:3.2.1'
     }
 }
 
@@ -25,6 +25,9 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
+        minSdkVersion 14
+        targetSdkVersion 26
+
         versionCode = 100040
         versionName = "1.0.40"
     }
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,5 +3,5 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
-distributionSha256Sum=7a2c66d1a78f811d5f37d14630ad21cec5e77a2a4dc61e787e2257a6341016ce
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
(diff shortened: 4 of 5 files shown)


hannesa2:
maybe depends on https://github.com/owncloud/android-library/pull/198

AndroidManifest.xml
```diff
@@ -23,7 +23,6 @@
     android:versionName="${versionName}">
 
     <uses-sdk
-        android:minSdkVersion="14"
         android:targetSdkVersion="26"
```
davigonz: Please move `targetSdkVersion` to build.gradle as well, as we did in https://github.com/owncloud/android-library/pull/198/

build.gradle
```diff
@@ -11,7 +11,7 @@ buildscript {
         jcenter()
```
davigonz: Think travis is not working because `minSdkVersion` has not been moved here after being deleted from AndroidManifest.xml

oc_jb_workaround/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
```
davigonz: Same with oc_jb_workaround module

davigonz:
@hannesa2 please review the requested changes

oc_jb_workaround/build.gradle
hannesa2: done

davigonz:
Hi @hannesa2 , as you said in https://github.com/owncloud/android/pull/2324#issuecomment-432531956, this PR depends on owncloud/android-library#198 to properly work so we should point from here to https://github.com/owncloud/android-library/pull/198 last commit.

For these cases, library changes must be reflected in the app by an app commit that points to those library changes. 

To do that:
1. From `AndroidStudio-3.2.1` branch in the app, go to owncloud-android-library and checkout `AndroidStudio-3.2.1`.
2. Go back to app project and `git add owncloud-android-library`.
3. Commit, push...

hannesa2:
>  so we should point from here to owncloud/android-library#198 last commit.

There is no need doing so. It builds

davigonz:
> There is no need doing so. It builds

@hannesa2 building or not is not the point, if we develop something in the library we need to update the reference to the library in the app and it makes sense to use this PR for that purpose because here we are updating to Android Studio 3.2.1 the app module as well.

hannesa2:
@davigonz I do not agree, the lib can be more advanced than the app, that's the advantage of submodules.
Anyway I'm tired to rebase and rebase, and no merge is done, since weeks I wait for a merge https://github.com/owncloud/android-library/pull/198 and currently it seems to be blocked. 
So, I wait till a progress is done there.

Normally a PR needs weeks, when a complex issue with a lot of influence is in the room, but I don't see any of this preconditions.

Or with other words, please trigger me, when a merge can be done within the next hours, then I do further steps

davigonz:
>  I do not agree, the lib can be more advanced than the app, that's the advantage of submodules.

You are right although we are not used to that kind of scenarios. All the features or fixes we've been including in the library from time ago, we wanted/needed to have them in the ownCloud app. 

In this case, I clearly see the connection between the changes included in https://github.com/owncloud/android-library/pull/198 and this PR, so for me makes sense to merge first the library PR and after that, the app one pointing to the latest in the library. Be aware we will have to update that reference at some point anyway.

davigonz:
> Or with other words, please trigger me, when a merge can be done within the next hours, then I do further steps

No worries, I will ping you by then, likely at the end of this week, thanks for your patience

jesmrec:
> In this case, I clearly see the connection between the changes included in owncloud/android-library#198 and this PR, so for me makes sense to merge first the library PR and after that, the app one pointing to the latest in the library. Be aware we will have to update that reference at some point anyway.

👍 👍 👍  absolutely agreed. Latest commit in te app must take the latest commit in the library, so new branches will get the reference and the latest changes in the app

build.gradle
```diff
@@ -97,6 +97,7 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
+        minSdkVersion 14
         manifestPlaceholders.versionCode = "20900300"
```
davigonz: You have removed targetSdkVersion from AndroidManifest.xml but not included it in here.

res/values/strings.xml
```diff
@@ -123,6 +123,7 @@
     <string name="common_remove_upload">Remove upload</string>
     <string name="common_retry_upload">Retry upload</string>
     <string name="common_cancel_sync">Cancel sync</string>
+    <string name="common_cancel">Cancel</string>
```
davigonz: This is no longer needed, it was deleted in https://github.com/owncloud/android/pull/2275/files#diff-4691bafd7791e471b30c492e6f43d65bL130

davigonz: This is no longer needed, it was deleted in https://github.com/owncloud/android/pull/2275/files#diff-4691bafd7791e471b30c492e6f43d65bL130

davigonz:
@hannesa2 just merged https://github.com/owncloud/android-library/pull/198, please apply the pending requested changes , update the library reference to the latest in `master` of the library and I will merge this, thanks!

jesmrec:
App builds correctly with the tool update.

Approved

davigonz:
Merged, thanks for the contribution @hannesa2

PR #2343, opened 2018-11-08
Get rid of old ant tests in library

davigonz:
Needed to merge https://github.com/owncloud/android-library/pull/198

Needs https://github.com/owncloud/android-library/pull/211


hannesa2:
Yes it's an option to delete them, an other option is to ignore them and use it as base ground for converting them to Espresso tests later (when ?)
But I've no opinion about this

davigonz:
> Yes it's an option to delete them, an other option is to ignore them and use it as base ground for converting them to Espresso tests later (when ?)

We've been discussing that option as well but we'd rather delete them if are not being currently used and if we need to have a look at it, these old tests will be always in the git history.

Issue #2329, opened 2018-10-29
Re-enter password while creating share link


sd1998:
### Actual behaviour
The password option in the create share link options request the user to enter the password once.

### Expected behaviour
The user should be asked to re-enter the password while creating the share link which happens in case of setting a pattern lock or passcode lock on the app.
 
### Steps to reproduce
1. Select a file or a directory
2. Select the share option from the toolbar
3. Select the option to create a share link
4. Enable the password option in the dialog

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes.

### Environment data
Android version: 8.0

ownCloud app version: Version 2.9.0-beta.2

davigonz:
It won't be implemented, reasons in https://github.com/owncloud/android/pull/2352

PR #2352, opened 2018-11-13
Fix to Issue #2329

sd1998:
Re-enter password check added while creating public share links.

### DIFF
--- a/res/layout/share_public_dialog.xml
+++ b/res/layout/share_public_dialog.xml
@@ -115,6 +115,15 @@
                 android:inputType="textPassword"
                 android:visibility="gone" />
 
+            <EditText
+                android:id="@+id/share_via_link_re_password_value"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginLeft="@dimen/standard_half_margin"
+                android:layout_marginRight="@dimen/standard_half_margin"
+                android:inputType="textPassword"
+                android:visibility="gone"/>
+
             <RelativeLayout
                 android:id="@+id/shareViaLinkExpirationSection"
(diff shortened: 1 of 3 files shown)


davigonz:
Thanks for this contribution @sd1998 , I'd like to know what's the opinion of @jesmrec and @michaelstingl about this improvement since is not implemented in the web version of ownCloud, do you think we really need to re-enter the password when creating public links?

[](https://user-images.githubusercontent.com/15261363/48784123-52294300-ece2-11e8-9484-f7fc60a31aa0.png)

michaelstingl:
Hm, not sure. Add the "reveal" eye-icon to show the password before clicking the "Share" button instead? After creating the share, we can't show it of course… /cc @pmaier1

pmaier1:
I would also vote against creating an inconsistency with the web interface. Also these passwords are very easy to change so I can't really see the need to type it twice. A "reveal" icon does make sense, though.

jesmrec:
@sd1998 could yo add the "reveal" instead?

davigonz:
@michaelstingl @pmaier1 thanks a lot for your opinion , I agree with you, think that writing the password twice is not needed. 

The reveal icon, a.k.a eye icon was already implemented and included in 2.8 version, have a look at https://github.com/owncloud/android/issues/2112. 

@sd1998 maybe you have not seen it in the latest version of the app because there's a UI glitch we've just noticed, have a look at https://github.com/owncloud/android/issues/2374. The eye icon is also used in login view and since the color is white now, it's visible in the login view but not in the share dialog. Maybe you want to work on that

davigonz:
Anyway, I will close this since is not going to be implemented, thanks anyway @sd1998

Issue #2353, opened 2018-11-13
[2.9.2] New release


davigonz:
### TASKS
 - [x] [GIT] Create branch `release_2.9.2` in owncloud/android from master
 - [x] [GIT] Create branch `release_0.9.23` in owncloud/android-library from master 
 - [ ] [DEV] Update version number and name in build.gradle in android module
 - [ ] [DIS] Create changelog file (< 500 chars) and add to CHANGELOG.md in owncloud/android
 - [ ] [WIZARD] Show wizard when upgrading to 2.9 or 2.9.2
 - [ ] [QA] Design Test plan @jesmrec
 - [ ] [QA] Regression Test plan  @jesmrec 
 - [ ] [GIT] Merge branch `release_0.9.23` in owncloud/android-library, into master, if needed
 - [ ] [GIT] Merge branch `release_2.9.2` in owncloud/android, into master 
 - [ ] [DIS] Generate test APKs files from branch master in owncloud/android and perform some basic operations.
 - [ ] [GIT] Create and sign tag `oc-android-2.9.2` in HEAD commit of master branch, in owncloud/android 
 - [ ] [GIT] Create and sign tag `oc-android-library-0.9.23` in HEAD commit of master branch, in  #owncloud/android-library 
- [ ] [DOC] Include new changelog on Releases section of github android repo by drafting the new release 2.9.2
 - [ ] [DIS] Upload & publish release APK and changelog in Play Store
 - [ ] [SPREADTHEWORD] Write a topic in central about the release.
 - [ ] [COORDINATION] Write in corresponding channels.
 - [ ] [SPREADTHEWORD] Write a post in the blog @lefherz 
 - [ ] Publish app in F-Droid https://gitlab.com/fdroid/fdroiddata/merge_requests/3912
 - [ ] [GIT] merge master branch into stable, in owncloud/android-library
 - [ ] [GIT] merge master branch into stable, in owncloud/android

_____
### RELATED

https://github.com/owncloud/android/issues/2347

jesmrec:
Checks to do:

- [ ] https://github.com/owncloud/android/issues/2347 fixed
- [ ] Upgrade 2.8.0 -> 2.9.2 (Wizard)
- [ ] Upgrade 2.9.0 -> 2.9.2 (no Wizard)
- [ ] Upgrade 2.9.1 -> 2.9.2 (no Wizard)

PR #2354, opened 2018-11-13
Release 2.9.2

### LINKED ISSUE #2347: "Folder doesn't exist anymore" on OwnCloud account with a space
### Actual behaviour
- When I'm logged in to an OwnCloud account with a space in the username, I get the error in Dutch:
"Map /remote.php/dav/files/Folder name/Folder name/ bestaat niet meer". Translated: "Folder /remote.php/dav/files/Folder name/Folder name/ doesn't exist anymore".

We already tried to create a new account without a space, in that case it works. If we create another new account also with a space in it, it also doesn't work.

This doesn't work since 2 days so I expect the issue came with the 2.9.0 update.

It works just fine on the iOS app.

### Expected behaviour
- I should be able to browse files normally.
 
### Steps to reproduce
1. Create an account with a space in the username, for example "User name".
2. Log in to this account on the OwnCloud Android app.
3. It should now display the error at the bottom of the page.

- No, I can't create an extra user on this server.

### Environment data
Android version: 7.0

Device model: Samsung Galaxy J5

Stock or customized system: Stock

ownCloud app version: 2.9.1

ownCloud server version: 10.0.10

#### Web server error log
- No lines are appearing when the error occurs.

- No lines are appearing when the error occurs.

davigonz:
Fixes #2347 via #2353

Needs https://github.com/owncloud/android-library/pull/213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.9.2 (November 2018)
+- Bug fixes for users with username containing spaces
+
 ## 2.9.1 (November 2018)
 - Bug fixes for LDAP users using uid:
   + Fix login not working
--- a/build.gradle
+++ b/build.gradle
@@ -97,8 +97,8 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
-        manifestPlaceholders.versionCode = "20900100"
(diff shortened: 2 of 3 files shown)


jesmrec:
Checks passed. Approved.

Issue #2335, opened 2018-11-04
Batched deletions


sd1998:
### Actual behaviour
When selecting multiple files as deleting each file is deleted individually due to which the snackbar message is shown multiple times(it is shown after each file has been deleted).

### Expected behaviour
The files should be deleted in batches  or the message should be shown once after all the selected files have been deleted
 
### Steps to reproduce
1. Select multiple files
2. Select delete from the three-dot menu

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes

### Environment data
Android version:8.0

jesmrec:
Agreed. This is an interesting improvement.

davigonz:
This would be great @sd1998, thanks for the suggestion

PR #2360, opened 2018-11-18
Fix to issue #2335

sd1998:
The snackbar message is shown once after all the selected files have been deleted.


src/com/owncloud/android/operations/RemoveFileOperation.java
```diff
@@ -51,6 +51,11 @@ public RemoveFileOperation(String remotePath, boolean onlyLocalCopy) {
         mOnlyLocalCopy = onlyLocalCopy;
     }
 
+    public RemoveFileOperation(String remotePath,boolean onlyLocalCopy,boolean lastFile){
```
davigonz: Think this constructor is not being used

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1501,9 +1502,11 @@ public void onRemoteOperationFinish(RemoteOperation operation, RemoteOperationRe
     private void onRemoveFileOperationFinish(RemoveFileOperation operation,
                                              RemoteOperationResult result) {
 
-        showSnackMessage(
-                ErrorMessageAdapter.getResultMessage(result, operation, getResources())
-        );
+        if(getListOfFilesFragment().getNoOfItems() == 2 || result.isException()) {
```
davigonz: Just tried your code and after having a look at this part, the deletion message doesn't appear when deleting one file or more than 2 and it should appear.

sd1998:
@davigonz I have fixed the bug.

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -439,6 +440,10 @@ public void onFocusChange(View v, boolean hasFocus) {
         }
     }
 
+    public boolean isCheckedItemsZero(){
```
davigonz: The name for this method doesn't fit with what you are doing inside since you are verifying whether there's one file checked or not, is that correct?

sd1998: @davigonz I will make the function name change.

sd1998:

davigonz:
Thanks @sd1998 , code approved, ready to test @jesmrec

PR #2375, opened 2018-11-29
Get rid of unused translations

davigonz:


res/values-af-rZA/strings.xml
```diff
@@ -50,9 +50,7 @@
   <string name="action_share">Deel</string>
   <string name="common_yes">Ja</string>
   <string name="common_no">Nee</string>
-  <string name="common_ok">OK</string>
-  <string name="common_cancel">Kanseleer</string>
-  <string name="common_back">Terug</string>
+    <string name="common_back">Terug</string>
```
jesmrec: indentation is not correct in the added resources.

davigonz: That's the correct indentation - same as default strings.xml -  and has been set by Android Studio automatically but Transifex seems not to respect that. We can face this in a separate issue.

jesmrec: 👍

PR #2398, opened 2018-12-13
Fixed waiting for wifi after connection changing

### LINKED ISSUE #2316: Pause manual upload when wifi is off.
### Actual behaviour
-Manual upload does not pause when wifi is disconnected or turned off. There is no option to change this behaviour. This is different to auto-upload, which does have wifi/charging option. 

### Expected behaviour
-We need an option to pause upload when wifi is off.
 

### Environment data
Android version: 9

Device model: pixel 2 xl

Stock or customized system:stock

ownCloud app version2.9 beta 1

ownCloud server version:10.0.10.4

```

### PR DESCRIPTION
**Pull request by BigAdam2005**
Fix #2316

### CODE REVIEW

**Comment by BigAdam2005** [comment]
@jesmrec May you check this PR.

**Comment by davigonz** [comment]
> Let's have a look...

File: src/com/owncloud/android/operations/UploadFileOperation.java [hunk 1]
```diff
@@ -477,16 +477,16 @@ protected void moveTemporalOriginalFiles(File temporalFile, File originalFile, S
...
-        boolean delayCameraUploadsVideo = (
-                isCameraUploadsVideo() && PreferenceManager.cameraVideoUploadViaWiFiOnly(mContext)
-        );
-        return (
-                (delayCameraUploadsPicture || delayCameraUploadsVideo) &&
-                        !ConnectivityUtils.isAppConnectedViaWiFi(mContext)
-        );
+        switch(getCreatedBy()){
+            case CREATED_AS_CAMERA_UPLOAD_PICTURE:
+                return isCameraUploadsPicture() &&  PreferenceManager.cameraPictureUploadViaWiFiOnly(mContext);
```
davigonz: Think you do not need to call `isCameraUploadsPicture()` since inside that method we are checking `mCreatedBy == CREATED_AS_CAMERA_UPLOAD_PICTURE`, same as we do with `switch(getCreatedBy()) case CREATED_AS_CAMERA_UPLOAD_PICTURE`

BigAdam2005:
@davigonz fixed

davigonz:
> fixed

Thanks @BigAdam2005 , please solve the conflicts and we are ready to test @jesmrec

BigAdam2005:
There is a problem because i cannot find any conflicts.

davigonz:
> There is a problem because i cannot find any conflicts.

Git doesn't warn you about conflicts when performing the rebase?

BigAdam2005:
Everything fixed

BigAdam2005:
@davigonz is it ready?

davigonz:
> @davigonz is it ready?

Thanks @BigAdam2005, the conflicts are fixed, now @jesmrec can start with QA

jesmrec:
It is not clear for me what this PR fixes. I read https://github.com/owncloud/android/issues/2316, and i understood that the scenario before the fix is 

1. Select a file to upload
2. Wifi connection is lost during the upload
3. Upload does not stop, so mobile connection keeps it alive. 

We do not want to waste our data plan, so the upload must stop once wifi is lost. am i correct?

in any case, as it was commented if the issue, we should need an additional option as camera uploads to let the user to decide whether the uploads are allowed to run under mobile connection or not. 

Please, light my way, maybe i'm messing myself.

@BigAdam2005 @davigonz

owncloudApp/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -477,16 +477,16 @@ protected void moveTemporalOriginalFiles(File temporalFile, File originalFile, S
...
-        );
+        switch(getCreatedBy()){
+            case CREATED_AS_CAMERA_UPLOAD_PICTURE:
+                return PreferenceManager.cameraPictureUploadViaWiFiOnly(mContext);
+            case CREATED_AS_CAMERA_UPLOAD_VIDEO:
+                return PreferenceManager.cameraVideoUploadViaWiFiOnly(mContext);
+            case CREATED_BY_USER:
+                return !ConnectivityUtils.isAppConnectedViaWiFi(mContext);
+        }
+        //cannot happen
```
jesmrec: if it can not happen (the three options above are the only options) , it is not needed to add the `return false`.
i'd also add a `break` inside each `case` to avoid checking the other options, once one of the is matched. What do you think? CC @davigonz

davigonz: > if it can not happen (the three options above are the only options) , it is not needed to add the return false.

You are right, I missed that part of the code, I guess @BigAdam2005 put that return there because delayForWifi has to return a boolean even if any of the cases do not match the switch. To fix this, please add a `default` case at the end of the switch. @BigAdam2005 

`break` would not be needed since the method finish when returning in any of the cases.

BigAdam2005: It isn`t possible because i got compilation error
`error: missing return statement`

davigonz:
```
private boolean delayForWifi() {
        switch(getCreatedBy()){
            case CREATED_AS_CAMERA_UPLOAD_PICTURE:
                return PreferenceManager.cameraPictureUploadViaWiFiOnly(mContext);
            case CREATED_AS_CAMERA_UPLOAD_VIDEO:
                return PreferenceManager.cameraVideoUploadViaWiFiOnly(mContext);
            case CREATED_BY_USER:
                return !ConnectivityUtils.isAppConnectedViaWiFi(mContext);
            default:
                return false;
        }
    }
```

Have you tried to include the `default` case below the other conditions?

davigonz: You shouldn't have `error: missing return statement.` then.

jesmrec:
@BigAdam2005 thanks for checking @davigonz 's review. Could you have a look to my comment https://github.com/owncloud/android/pull/2398#issuecomment-454707763 and share your thoughts?

BigAdam2005:
@jesmrec I added preference but i don't know how to modify ui.

jesmrec:
I guess that this PR is not very scoped. We do not really know what is the target, the expected behaviour and how does it fixes it. I will remove from the current milestone till we have clearer idea about it. Of course, we are still open to discussion etc...

Issue #2394, opened 2018-12-08
auto upload of pictures is broken


hannesa2:
With the current version, it seems the auto sync of pictures is broken. On my device there was nothing synced since months

hannesa2:
I did a test on a real device from 15:05 till 18:55

### Problem A:
It seems `CameraUploadsSyncJobService` does not start

```
adb pull /storage/emulated/0/owncloud/log/currentLog.txt .
cat currentLog.txt |grep CameraUploadsSyncJobService

2018/12/08 15:03:20 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService
2018/12/08 15:03:38 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService
2018/12/08 15:03:45 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService
2018/12/08 15:03:50 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService
2018/12/08 15:05:40 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService

```
Looks like L53 is never reached
![image](https://user-images.githubusercontent.com/3314607/49689026-1c29f280-fb1b-11e8-85fb-c9eac1bb62e1.png)

### Problem B:
Last entry at 15:47:47. After this time app seems to stop doing anything in background. No Job started, nothing

### Problem C:
I see a lot of actions in background, but server directory `Kamera-Upload` has still only **one** file instead if > 800 files 
```
status:UPLOAD_FAILED and result:NETWORK_CONNECTION
Socket timeout exception Exception
UPLOAD_IN_PROGRESS and result:UNKNOWN
updateCameraUploadSync returns with: 1
Unrecovered transport exception Exception
```

Does someone understand what's going on. Here is the complete log
[currentLog.txt](https://github.com/owncloud/android/files/2659927/currentLog.txt)

jesmrec:
Thanks a lot for reporting. We have had some issues of people who automatic uploads do not work for them. I guess you selected the correct source folder. We will have a look.

davigonz:
Thanks for reporting this @hannesa2, in the `currentLog.txt` you attach I see several logs saying `Scheduling a CameraUploadsSyncJobService` in different hours but there's no 15 minutes wait.

`2018/12/08 15:03:20 CameraUploadsHandler : Scheduling a CameraUploadsSyncJobService
[...]

Issue #934, opened 2015-03-29
No verification of the provided SSL certificate


wforums:
### Expected behaviour
The SSL certificate should be checked against a stored one, to prevent MITM attacks that have a wildcard/global valid certificate. This is implemented by a lot of companies, including competitors like DropBox (https://blogs.dropbox.com/dropbox/2014/06/weve-got-your-back/).

### Actual behaviour
OwnCloud app runs as usual, not doing anything.

### Steps to reproduce
Prerequisites: 
- Charles Proxy
- Android which has the proxy set and Charles root certificate installed

1) Open up the OwnCloud Android client
2) Check in Charles Proxy for requests (over HTTPS) to the status.php, which contains the basic authorization

![owncloud](https://cloud.githubusercontent.com/assets/7365586/6886261/b4fe8f60-d636-11e4-8507-082fab3ac7e4.png)

### Environment data
Android version: global (4.2.2 in my case)
Device model: global (Acer V370)
Stock or customized system: Stock
ownCloud app version: 1.7.0
ownCloud server version: 8.0.2

#### Web server error log
N/A

#### ownCloud log (data/owncloud.log)
N/A

### Proposed solution

My proposal to fix this issue would be storing the certificate (or signatures of them at least) when you are configuring the app (first login), and subsequently check the provided certificate of the server before sending any important data. If the certificate does not match, warn the user or fail the connection.

wforums:
@davivel @masensio Any plans/comments on this? This is a quite serious issue in my opinion...

akalypse:
:+1: I would even recommend to prepackage the root SSL certificate in a KeyStore weithin the APP and use it for SSL certificate pinning. An example project can be found here: 
https://github.com/ikust/hello-pinnedcerts

rugk:
:+1: for [pinning certs](https://developer.android.com/training/articles/security-ssl.html#Pinning).

It would be nice to also be able to pin certificates of your own OwnCloud servers by putting the hash in a settings option. This way you do not have the proposed "Trust on first use" approach as suggested by @canihavesomecoffee.

The TOFU way has some problems, because web admins may switch the certificates.
Actually the TOFU should still be used - IMHO *in addition* to the first approach, but at least implement this one:
**Add HTTP Public Key Pinning support**

[...]

PR #2402, opened 2018-12-19
Android X (P)

hannesa2:
I was forced to use Android P as well, so it's the move to P as well
 
-----------

Edited by @davigonz

About **Android P**, I'm gathering some changes from [here](https://developer.android.com/about/versions/pie/android-9.0-changes-28) that might affect the ownCloud Android app when targeting Android P:

- **Network TLS enabled by default**: if your app targets Android 9 or higher, the isCleartextTrafficPermitted() method returns false by default. If your app needs to enable cleartext for specific domains, you must explicitly set cleartextTrafficPermitted to true for those domains in your app's Network Security Configuration. Reproducible and fix in https://github.com/owncloud/android/pull/2402#issuecomment-456881009

- **Foreground services:**  apps that target Android P or higher and use foreground services must request the FOREGROUND_SERVICE permission. E.g.: https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java#L306

CC @jesmrec @hannesa2 

___

BUGS & IMPROVEMENTS

- [x] (1) Camera uploads https://github.com/owncloud/android/pull/2402#issuecomment-460961991


hannesa2:
Ok, there are some `-alpha02` included, either 

- we wait till they are non alpha or
- we go with `-alpha02` (my company uses it in production without any issues , 5.000.000+ installations) or
- we uses older stable versions, not sure if it work for all libs

davigonz:
Thanks a lot for this contribution @hannesa2 , we'd rather not use alpha versions, would it be possible to keep using stable versions just in the alpha ones?

hannesa2:
I reverted to stable libs. 
But I'm not convinced, because it's just number cosmetic. After 2.0.0 a 2.0.1 will come too. On the other side, using old ones means to waiver some improvement in beta

owncloudApp/build.gradle
```diff
@@ -3,48 +3,43 @@ apply plugin: 'com.android.application'
...
-    implementation "com.android.support:support-v4:${supportLibraryVersion}"
-    implementation "com.android.support:design:${supportLibraryVersion}"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "androidx.legacy:legacy-support-v4:$androidX"
+    implementation 'com.google.android.material:material:1.0.0'
     implementation 'com.jakewharton:disklrucache:2.0.2'
     implementation 'com.google.android.exoplayer:exoplayer:r2.2.0'
     implementation 'com.andrognito.patternlockview:patternlockview:1.0.0'
-    implementation "com.android.support:appcompat-v7:${supportLibraryVersion}"
+    implementation 'androidx.appcompat:appcompat:1.0.0'
```
davigonz: Why don't you get the version from `$androidX` here?

davigonz:
Apart from the requested changes above, I have compiled the code, install the app and I get the next error when checking the server in login view, which appears as `Unknown error` in the app, any idea?

```
2019-01-23 17:53:27.454 25087-25119/com.owncloud.android.debug E/GetRemoteStatusOperation: Connection check at http://192.168.0.173:10000: Unrecovered transport exception
    java.net.UnknownServiceException: CLEARTEXT communication to 192.168.0.173 not permitted by network security policy
        at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:147)
        at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:257)
        at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
        at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
        at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
```

davigonz:
About Android P, I'm gathering here some changes that might affect the ownCloud Android app when targeting Android P:

- **Network TLS enabled by default**: if your app targets Android 9 or higher, the isCleartextTrafficPermitted() method returns false by default. If your app needs to enable cleartext for specific domains, you must explicitly set cleartextTrafficPermitted to true for those domains in your app's Network Security Configuration.

- In res/xml/network_security_config.xml:

```xml
<?xml version="1.0" encoding="utf-8"?>
<network-security-config>
    <base-config cleartextTrafficPermitted="true" />
</network-security-config>
```

- In AndroidManifest.xml

```xml
android:networkSecurityConfig="@xml/network_security_config"
```

- **Foreground services:**  Apps that target Android P or higher and use foreground services must request the FOREGROUND_SERVICE permission. E.g.: https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java#L306

michaelstingl:
> But a move in the other direction could be an introduce of (dynamic) cert pinning which prevents from a man-in-the-middle attacks -just as an idea-

👍 

@hannesa2 you mean like in https://github.com/owncloud/android/issues/934 ? Or better open another issue to discuss details?

hannesa2:
Exactly https://github.com/owncloud/android/issues/934 ! (but I it's almost 3 years just an idea)
I will not promise anything, but maybe I can deliver something, when I generalize similar somewhere else 

And one more additional step could be a 2 phase authentication, like Google does when a new IP/device tries to connect with a mail notification/confirmation. But with this, the backend needs some work. But to be honest, this needs too much live/spare time from my side.

When it comes to my focus I can open a ticket for sure, but currently my main interest is the non working (already closed) upload https://github.com/owncloud/android/issues/2394 
Btw you can open a ticket too, if you see a sense to have an additional ticket

owncloudApp/build.gradle
hannesa2: because 
`androidX` is `1.0.0` and 
`androidx.appcompat:appcompat` has currently `1.0.2`
But my fault I forgot to apply `1.0.2` here

done

hannesa2:
What should be done here ?
Allow 'http' and be more un-secure or follow the Goggle suggestion and only allow `https` and do nothing ?
@davigonz @jesmrec @michaelstingl

davigonz:
> What should be done here ?
Allow http and be more un-secure or follow the Goggle suggestion and only allow https and do nothing ?

I would not get rid of http IMHO, is insecure for sure but ownCloud is a self-hosted solution and therefore, each end user/company should be the responsible for the security of his ownCloud instance. 

Disallowing http will likely trigger complains since we are making decisions that do not belong to us.

> But a move in the other direction could be an introduce of (dynamic) cert pinning which prevents from a man-in-the-middle attacks -just as an idea-

About this, I think that we are already checking the SSL certificate against a stored one, as is suggested in #934. Have a look at https://github.com/owncloud/android-library/blob/940f0b9bb042e3c318e4cbeff71782e0d8733f93/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/HttpClient.java#L78.

davigonz:
@hannesa2 can you please apply the changes about [http](https://github.com/owncloud/android/pull/2402#issuecomment-456881009) and [foreground services ](https://github.com/owncloud/android/pull/2402#issue-239732799) so we can push this to be included in 2.10.0? Thanks

hannesa2:
[http](https://github.com/owncloud/android/pull/2402#issuecomment-456881009) :
I followed https://developer.android.com/training/articles/security-config and applied the default configuration for apps targeting Android 6.0 (API level 23) and lower

hannesa2:
Foreground Service:

I added the missing ones, but to be honest I did only code changes and didn't test it.

In following files, in my point of view, there are too much conditions to run into `startForeground()` but you already have it on master, so I didn't changed it.
This would be ok
`Build.VERSION.SDK_INT >= Build.VERSION_CODES.O`  but the other ones ?

FileDownloader.java
`if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && (isAvailableOfflineFile || retryDownload))`

FileUploader.java
`if ((createdBy == CREATED_AS_CAMERA_UPLOAD_PICTURE || createdBy == CREATED_AS_CAMERA_UPLOAD_VIDEO ||
                isAvailableOfflineFile || isRequestedFromWifiBackEvent) &&
                Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {`

davigonz:
> FileDownloader.java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && (isAvailableOfflineFile || retryDownload))

You have a comment just below that line that explains why do we need those checks. We create the service to check available offline files and retry downloads by using `startForegroundService` and we have to call `startForeground`  within five seconds only for those cases.

https://github.com/owncloud/android/blob/2b15b5bab69d83e22f7ccb70b6868eb49edd3341/owncloudApp/src/main/java/com/owncloud/android/files/services/FileDownloader.java#L193-L197

> if ((createdBy == CREATED_AS_CAMERA_UPLOAD_PICTURE || createdBy == CREATED_AS_CAMERA_UPLOAD_VIDEO || isAvailableOfflineFile || isRequestedFromWifiBackEvent) && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {

Have a look at the comment below: https://github.com/owncloud/android/blob/2b15b5bab69d83e22f7ccb70b6868eb49edd3341/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java#L300-L304

davigonz:
@hannesa2 regarding the foreground services, I think you should not need to touch any code, as you did in https://github.com/owncloud/android/pull/2402/commits/1d843409951e2b636b2f6c226d88d80ec1f4aebe. Including [FOREGROUND_SERVICE](https://developer.android.com/reference/android/Manifest.permission#FOREGROUND_SERVICE) permission in AndroidManifest would be enough, as the [Android P documentation](https://developer.android.com/about/versions/pie/android-9.0-changes-28) says.

hannesa2:
- I reverted 1d84340
- I added FOREGROUND_SERVICE 
- I removed one lint warning
image

davigonz:
@hannesa2 ok thanks

owncloudApp/build.gradle
```diff
@@ -60,19 +55,19 @@ tasks.withType(Test) {
 }
 
 android {
-    compileSdkVersion 26
+    compileSdkVersion 28
 
     defaultConfig {
-        minSdkVersion 14
-        targetSdkVersion 26
+        minSdkVersion 15
```
davigonz: Let's keep this in 14, I know is a pretty old version but we will upgrade it after discussing it in https://github.com/owncloud/android/issues/2431

davigonz:
@hannesa2 just a minor change and we can move it forward, could you make the Android X and P upgrade in the library as well? Thanks

owncloudApp/build.gradle
hannesa2: Google removed 14 from any statistic, because is as rare like a unicorn.
It's even not worth to discuss to revert to 14, you will loose 0,0001 % of users
It tried to revert
image

Let' go 15. Even 15 is like a unicorn

davigonz:
Ready to test @jesmrec , though the library doesn't have Android X(P) yet, I suggested @hannesa2 to do it

hannesa2:
I guess you mean this https://github.com/owncloud/android-library/pull/231.

davigonz:
> I guess you mean this [owncloud/android-library#231](https://github.com/owncloud/android-library/pull/231).

Yes, thanks

jesmrec:
Let's start a QA round

jesmrec:
### (1)

Camera uploads are not working properly. Check these steps out:

1. Go to `Settings` and enable camera uploads for videos and pic
2. Minimize the app, open the camera and take a pic and a video
3. Wait 15 minutes, the app is not in foreground

Current: Items are not uploaded. They are in the uploads view with `App terminated`. The app finally leads to ANR:

>2019-02-06 10:45:45.085 1174-1206/? E/ActivityManager: ANR in com.owncloud.android.debug
    PID: 20006
    Reason: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{d7a9b0e u0 com.owncloud.android.debug/com.owncloud.android.files.services.FileUploader}
    Load: 4.4 / 4.23 / 4.29`

Expected: Items correctly uploaded after max. 15 minutes.

Google Pixel 2 v9 API 28

hannesa2:
It seems https://github.com/owncloud/android/pull/2402#issuecomment-459262343 is not true.
I will reintroduce  1d84340 again, @jesmrec please have a look on it.

hannesa2:
Interesting ! Maybe this is the reason for https://github.com/owncloud/android/issues/2394 ?
But unfortunately I can't build now. It seams google maven has an issue now, because 3h before PR was able to build. I will try it tomorrow

jesmrec:
@hannesa2 with the latest commit the app crashes from scratch (Android P)

> Process: com.owncloud.android.debug, PID: 10795  android.app.RemoteServiceException: Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification: Notification(channel=null pri=0 contentView=null vibrate=null sound=null defaults=0x0 flags=0x40 color=0xff1d2d44 vis=PRIVATE)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1738)

hannesa2:
Do you have to complete stacktrace ?
I run my personal build https://github.com/hannesa2/owncloud-android/releases and don't see this (till now)

owncloudApp/build.gradle
```diff
@@ -3,48 +3,44 @@ apply plugin: 'com.android.application'
...
-}
-
 dependencies {
     /// dependencies for app building
     implementation project(':owncloud-android-library:owncloudComLibrary')
-    implementation "com.android.support:support-v4:${supportLibraryVersion}"
-    implementation "com.android.support:design:${supportLibraryVersion}"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "androidx.legacy:legacy-support-v4:$androidX"
+    implementation "androidx.annotation:annotation:1.0.1"
```
hannesa2: without this line, build "release" run into an issue. That's why build in "debug" was too less

hannesa2:
btw, I made a test picture and since 45 minutes nothing is transferred to my Owncloud.
You can close https://github.com/owncloud/android/issues/2394 but this doesn't help to fix it.

hannesa2:
Now it syncs new pictures !

davigonz:
> Camera uploads are not working properly. Check these steps out:
> 1. Go to `Settings` and enable camera uploads for videos and pic
> 2. Minimize the app, open the camera and take a pic and a video
> 3. Wait 15 minutes, the app is not in foreground

> Current: Items are not uploaded. They are in the uploads view with `App terminated`. The app finally leads to ANR:
> 2019-02-06 10:45:45.085 1174-1206/? E/ActivityManager: ANR in com.owncloud.android.debug
> PID: 20006
> Reason: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{d7a9b0e u0 com.owncloud.android.debug/com.owncloud.android.files.services.FileUploader}
> Load: 4.4 / 4.23 / 4.29`

I think this was not tested with the proper commit. It should've been tested with 406494134e35e13c2e1e6b80baf5e0842257a804 but I suspect it was tested with the commit before that (6ed4d309bd0c854780f4707ff46708e6beb423e4) cause I'm getting the error `Process: Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification:`. 

Can you check that @jesmrec ?

> It seems [#2402 (comment)](https://github.com/owncloud/android/pull/2402#issuecomment-459262343) is not true.
I will reintroduce [1d84340](https://github.com/owncloud/android/commit/1d843409951e2b636b2f6c226d88d80ec1f4aebe) again, @jesmrec please have a look on it.

@hannesa2 Foreground services always need a notification and it does not make any sense to start every service in foreground as you did in https://github.com/owncloud/android/pull/2402/commits/e07901b9b2dc0f3db641a2dbad98cb119eb6a2f8, they make sense when uploading, downloading... 

The main problem here is related to the notifications. In Android P and after using FOREGROUND_SERVICE permission it seems we can not build notifications without parameters as we did in https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java#L306 and that was causing the error `Process: Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification:`.

To sum up, we do not need https://github.com/owncloud/android/pull/2402/commits/e07901b9b2dc0f3db641a2dbad98cb119eb6a2f8 , using FOREGROUND_SERVICE in manifest and 72a4297c14c1d30780c4902a18045935bfd9e75e would be enough.

hannesa2:
> To sum up, we do not need [e07901b](https://github.com/owncloud/android/commit/e07901b9b2dc0f3db641a2dbad98cb119eb6a2f8) , using FOREGROUND_SERVICE in manifest and [72a4297](https://github.com/owncloud/android/commit/72a4297c14c1d30780c4902a18045935bfd9e75e) would be enough.

I agree, [72a4297](https://github.com/owncloud/android/commit/72a4297c14c1d30780c4902a18045935bfd9e75e) fixes now the [e07901b](https://github.com/owncloud/android/commit/e07901b9b2dc0f3db641a2dbad98cb119eb6a2f8).
I can squash  (together with introduce 6dd0a4cc and revert d84f0bf02) it on request to summarize necessary changes, but in review you see summarize too 

Btw, `master` was not working for me as well, now with this PR it's working

hannesa2:
> Can you check that @jesmrec ?

My opinion is, it's just necessary to test latest commit

davigonz:
> I agree, [72a4297](https://github.com/owncloud/android/commit/72a4297c14c1d30780c4902a18045935bfd9e75e) fixes now the [e07901b](https://github.com/owncloud/android/commit/e07901b9b2dc0f3db641a2dbad98cb119eb6a2f8).
I can squash (together with introduce [6dd0a4c](https://github.com/owncloud/android/commit/6dd0a4ccc16fc576da3c3a6b3baf1c1ef925d026) and revert [d84f0bf](https://github.com/owncloud/android/commit/d84f0bf024c664076652c7a09c557d1e53e1fa18)) it on request to summarize necessary changes, but in review you see summarize too

@hannesa2 the `startForegroundService` stuff is still in this PR, can you have a look at it?

hannesa2:
please test it first. Without it, we've seen a crash https://github.com/owncloud/android/pull/2402#issuecomment-461133730

davigonz:
> please test it first. Without it, we've seen a crash [#2402 (comment)](https://github.com/owncloud/android/pull/2402#issuecomment-461133730)

As I explained in https://github.com/owncloud/android/pull/2402#issuecomment-462302042, I tested without it and I'm not getting that crash but the notification error. 

@jesmrec can you please test af756ba283513ab4a3c9a94268ebde08b9039e34 and confirm there's no crash there? If so, we can remove the `startForegroundService` stuff

jesmrec:
With commit 406494134e35e13c2e1e6b80baf5e0842257a804:

 - Installation: OK
 - URL check and authentication:  OK
 - Camera uploads: Fails with https://github.com/owncloud/android/pull/2402#issuecomment-460961991

With commit 6ed4d30

 - Installation: OK
 - URL check and authentication:  OK
 - Camera uploads: Fails. In this case, pics are not either enqueued (in the previous one, they were). I stumbled upon this trace:

> 2019-02-12 11:01:29.270 22838-22838/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.owncloud.android.debug, PID: 22838
    java.lang.RuntimeException: Unable to start service com.owncloud.android.files.services.FileUploader@a120a3d with Intent { cmp=com.owncloud.android.debug/com.owncloud.android.files.services.FileUploader (has extras) }: java.lang.SecurityException: Permission Denial: startForeground from pid=22838, uid=10482 requires android.permission.FOREGROUND_SERVICE
        at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3687)


with commit 72a4297c14c1d30780c4902a18045935bfd9e75e  (the latest one)

 - Installation: OK
 - URL check and authentication:  OK
 - Camera uploads: Works fine!!!
 - Additional check: av, offline files: OK

so, good news. In any case, a deeper testing should be performed to assure all scenarios are OK, but this is out of the scope of the current one.

jesmrec:
@hannesa2 i have rechecked the whole feature by dropping the commit c2db38e849b9fbdf54fea4e57f055c46ddfab682, and it works properly. No crashes in authentication and uploads run correctly (camera uploads and manual uploads). 

Maybe we can get rid of such commit, so it does not introduce anything sensible for the described feature, and will make a cleaner history. 

Sorry for the noise.

hannesa2:
ok, I removed the 
* introduce
* revert
* re-intoduce

commit, and kept all except my fix. (To be honest, because it worked for me, I simply followed your suggestion and didn't tested it)
@jesmrec please take over

jesmrec:
After a new check:

- Services job-based (camera uploads and av. offline syncing) are working properly in Android 9
- Same checks works fine in Android 8 and 7. 

We can move forward, and new checks will be performed in the following release stage.

Approved

jesmrec:
Please, update the branch for the final merge

hannesa2:
Sounds good ! Rebase is done

PR #2391, opened 2018-12-08
Install release and debug version on same device

hannesa2:
It's useful when you have production and debug version on same device

PR #2408, opened 2018-12-28
Show commit SHA1 in preferences

hannesa2:
Let a click jump to Github commit

I split https://github.com/owncloud/android/pull/2391
Now you can see commit sha1 and a click opens the last changes

owncloudcommit 2

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,8 +14,8 @@ ext {
 
 allprojects {
     repositories {
-        jcenter()
         google()
+        jcenter()
         maven {
             url "https://jitpack.io"
         }
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -1,5 +1,8 @@
 apply plugin: 'com.android.application'
 
+def commitSHA1 = 'COMMIT_SHA1'
+def gitRemote = 'GIT_REMOTE'
+
 configurations.all {
     resolutionStrategy {
         force "com.android.support:support-annotations:${supportLibraryVersion}"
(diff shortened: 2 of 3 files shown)


michaelstingl:
Make "click jump to Github commit" a theming option, so it can be disabled in some cases?

hannesa2:
"theming option" I have only a clue what you mean.
But I would never disable/hide it, because it increase transparency what we install on our device

michaelstingl:
There are various parameters you can set in [`owncloudApp/src/main/res/values/setup.xml`](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/res/values/setup.xml). This way you can build your own flavor of the app with some features enabled or disabled. I can imagine use cases, where the link to a repo on github.com isn't helpful.

hannesa2:
I would like to increase transparency by automatic upload to GooglePlay Alpha on tagging with https://github.com/Triple-T/gradle-play-publisher and show a menu entry to point to Travis build like this PR to last commit. 

sample: eg https://github.com/hannesa2/BitcoinCash-Wallet/commit/b0aa1eeec7b4c409d0162d30155695b41bee36f3

hannesa2:
In my point of view it's never not-helpful.
Imagine, when you are forced to investigate on a long run behavior, eg https://github.com/owncloud/android/issues/2394 after days you forget which version I use, release or debug ? which commit ? With this PR you see immediate what's the last relevant commit, not only an anonymous SHA1

davigonz:

jesmrec:
> In my point of view it's never not-helpful.

it is not a matter of helpfulness. The app is not only distributed via GitHub or only oriented to community. There are customers who brand the app to show only their stuff. In such case, there is no place for a link to a code repository. The SHA commit will point you to the installed version in any case.

jesmrec:
Checks performed:

- [X] Debug version -> correct commit hash and build type
- [X] Release version -> correct commit hash and build type

will create a separate issue to handle the click option that forwards to the GH commit.

Approved

hannesa2:
Ok introduce a multiple-click-add-more-menu-items behavior like on all Android devices for developer menu.

davigonz:
@hannesa2 please rebase with master and we can merge

davigonz:
Thanks @hannesa2 for this contribution, pretty useful!

PR #2412, opened 2018-12-28
fix a possible crash

hannesa2:


owncloudApp/src/main/java/com/owncloud/android/files/services/CameraUploadsSyncJobService.java
```diff
@@ -182,6 +181,7 @@ private synchronized void handleFile(File localFile) {
 
             if (mOCCameraUploadSync == null) {
                 Log_OC.d(TAG, "There's no timestamp to compare with in database yet, not continue");
+                return;
```
hannesa2: without the return, it will crash on L188
`if (isImage && localFile.lastModified() <= mOCCameraUploadSync.getPicturesLastSync()`

btw, the db layer was as expected (to be diplomatic) a surprise

jesmrec:
in which case can `mOCCameraUploadSync` be `null` ?

owncloudApp/src/main/java/com/owncloud/android/files/services/CameraUploadsSyncJobService.java
davigonz: > btw, the db layer was as expected (to be diplomatic) a surprise

Please if you consider there's something to improve, put it directly, just complaining is not useful for anyone.

hannesa2:
When the database can't deliver it. Simply first time never added ?

davigonz:
> Checking the feature with the fix, it works for both pics and videos.
Approved

Great! Let's merge it then...

PR #2414, opened 2018-12-29
scroll in log to bottom by default

hannesa2:
(cherry picked from commit c86aee80c78ce657da66a3a42ab3208617a5880c)

On my real Device Pixel 2 (Android P) the delete log button has no effect. So my logfile is very huge, that's why I've to scroll down for minutes till I see most recent entries. 
This PR solves the force to scroll


davigonz:
@hannesa2 not sure if this is the best solution, maybe showing the most recent logs at the top of the view would be better?

hannesa2:
If I've done it with a reverse list, the next guy come with "what's happen with the list ? I miss my entries.... or similar"

davigonz:
Showing the list of logs and scrolling down to the bottom is not the best solution in my opinion, maybe a new option to order the logs by date would work better. Please open a new PR if you want to implement that.

PR #2416, opened 2018-12-29
autofill server url with https

hannesa2:
During test I've to login that often and always I've to specify `https://` 
That's annoying, so I auto fill it, if it's missing. 
In short: `https://` is no more necessary


davigonz:
@hannesa2 really? [demo.owncloud.org](https://demo.owncloud.org) works with `https://` and is automatically filled. 

jan-02-2019 12-42-14

hannesa2:
Indeed, I checked once in GUI and it was not working (maybe miss written url )...

Issue #2284, opened 2018-08-28
User/group share: Show <shareWithAdditionalInfo> in autocompletion of shares


michaelstingl:
oC users can have same displayname, only username or email is different:

![users_-_owncloud_enterprise_edition](https://user-images.githubusercontent.com/214010/44736144-bcb95100-aaee-11e8-83a4-35e4d9d97a6b.png)

This make sharing UI difficult, but oC server got a solution for this:

![settings_-_owncloud_enterprise_edition](https://user-images.githubusercontent.com/214010/44736170-d65a9880-aaee-11e8-8af8-3764900a3624.png)

![com_bjango_istatmenus_network_and_item-0_and_item-0_and_item-0_and_item-0_and_item-0_and_item-0_and_item-0_and_item-0](https://user-images.githubusercontent.com/214010/44736184-dd81a680-aaee-11e8-8303-205bdfb9f499.png)

### Expected behaviour
Users with same displayname should be easier to distinguish. oC server can add `User ID` or `Email address` in brackets after the displayname

### Actual behaviour
Users with same displayname can't be distinguished

### Server configuration
ownCloud version:
10.0.9

### API

Here you can find the API responses if this feature is enabled on the oC server:
https://github.com/owncloud/client/issues/6749

michaelstingl:
@davigonz @jesmrec Is this something easy for the 2.8.0, or better schedule for 2.8.1 or 2.9.0 ?

jesmrec:
i would add it to 2.10.0, so current milestone is pretty full.

davigonz:
@michaelstingl the web UI shows the additional info even after saving the share, not only when searching.

<img width="283" alt="screenshot 2019-01-22 at 19 01 29" src="https://user-images.githubusercontent.com/15261363/51555589-3bd16c00-1e78-11e9-9873-dc6940edc40d.png">

But I would say that the code included in https://github.com/owncloud/client/pull/6837/files only shows the additional info when searching, not in the list of shares after saving it.

Guess the Android app should follow the web UI approach

PR #2428, opened 2019-01-23
Show shareWithAdditionalInfo in shares autocompletion and list

davigonz:
Implements #2284 
Needs https://github.com/owncloud/android-library/pull/229


davigonz:
@sd1998 @hannesa2 since you are pretty active contributors and we are grateful for that, you may want to start reviewing some PRs also, think is a worthy task in which everyone can learn from each other, what do you think? Thanks 😉

sd1998:
@davigonz I think this is a great idea. I will be starting with reviewing this PR.

jesmrec:
@sd1998 when you finish your review, ping us to start the next step (QA). Thanks!

jesmrec:
Works a expected. It was only tested with regular servers, and will be pending to check against special servers (LDAP or Shibboleth).

Checking the code it should not be a problem, since the `shareWithAdditionalInfo` is a capability, transparent from the authentication flow.

Approved.

PR #2432, opened 2019-01-30
Avoid NPE in RenameFileOperation on IOException

sjvs:
If/when an IOException occurs during a `FileRenameOperation`, the `result` variable is guaranteed to be `null`. I don't think there's any more information to send to the logger when this happens.

This issue was flagged up by LGTM.com: https://lgtm.com/projects/g/owncloud/android/alerts/

There are a few other alerts for this repository (including a few more potential NPEs and resource leaks), but I'm not familiar enough with the code base to assess whether these issues should be addressed (and if so, how).

It's possible to have LGTM perform automated code review for every pull request; here's an example how Google's AMPHTML open source project use that to catch potential security vulnerabilities: https://github.com/ampproject/amphtml/pull/13060

There are some more alerts on other ownCloud projects that are worth a look. For example: a few [XSS vulnerabilities have been flagged up in the imagelib extension](https://lgtm.com/projects/g/owncloud/apps/alerts/) in `owncloud/apps`.

(full disclosure: I'm an active ownCloud user myself, an am also part of the team that built LGTM.com)

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
@@ -112,7 +112,7 @@ public class RenameFileOperation extends SyncOperation {
             
         } catch (IOException e) {
             Log_OC.e(TAG, "Rename " + mFile.getRemotePath() + " to " + ((mNewRemotePath==null) ?
-                    mNewName : mNewRemotePath) + ": " + result.getLogMessage(), e);
+                    mNewName : mNewRemotePath) + " failed", e);
         }
 
         return result;


owncloudApp/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
```diff
@@ -112,7 +112,7 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             
         } catch (IOException e) {
             Log_OC.e(TAG, "Rename " + mFile.getRemotePath() + " to " + ((mNewRemotePath==null) ?
-                    mNewName : mNewRemotePath) + ": " + result.getLogMessage(), e);
+                    mNewName : mNewRemotePath) + " failed", e);
```
hannesa2: It makes sense to prevent form crash. -absolute-
But to see the reason makes maybe sense ? What's about this ?

` mNewName : mNewRemotePath) + ": " + (result==null ? "failed" : result.getLogMessage()), e);`

sjvs: As far as I can tell, `result` will **always** be `null` on line 115. It's assigned `null` at the start of the the method, then only gets assigned a value on line 102. The exception will happen on or above line 102, I believe.

hannesa2: Indeed L77 `RemoteOperationResult result = null;` you are right.
And I like Kotlin, then you have to handle it
Thank you

davigonz:
@jesmrec is this still reproducible?

davigonz:

jesmrec:
Approved

davigonz:
@sjvs can you please rebase with master so we can merge? Thanks

davigonz:
Hi @svjs , could you please solve the conflicts? Otherwise, we will do it ourselves in a different PR and merge it.

sjvs:
> Hi @svjs , could you please solve the conflicts? Otherwise, we will do it ourselves in a different PR and merge it.

Hi @davigonz. I currently don't have the time to work on this, so please go ahead!

PR #2435, opened 2019-02-06
Update with new codestyle and Android studio 3.3 in library

davigonz:
This PR is just to use the recent changes in library: Android Studio 3.3 and codestyle.


davigonz:
> Just to mention it, this doesn't apply codestyle to app

Yes, is just to update the library reference in the app, can you approve the changes @hannesa2 so we can merge? Thanks

Issue #2359, opened 2018-11-16
Changes in our Git Branching flow


davigonz:
Considering the following scenario with our current Git branching flow:

![current_branching_flow](https://user-images.githubusercontent.com/15261363/48621255-3a308700-e9a3-11e8-8674-a734e4a11b46.png)

1. We create branches to work on 2.9.2 features that are eventually merged into `master` branch.
2. Once 2.9.2 development is finished, we create `release_2.9.2` branch. At this point,`master` branch is blocked and if I or other contributors want to keep working on features for 2.10 release and merge them is not possible because we have to wait for merging and tagging `release_2.9.2`(step 3).
3. Once `release_2.9.2` is approved and fully tested, me merged it into master, tag the commit with `oc-android-2.9.2` and make the release public via GitHub and PlayStore. 
4. Merge `master` into `stable`.

This branching flow comes from long time ago and to speed up and not block our development process, including contributions, and according to what [GitFlow](https://es.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) defines, I would suggest the next changes: 

![new_branching_flow](https://user-images.githubusercontent.com/15261363/48621258-3b61b400-e9a3-11e8-83f3-90d48583892d.png)

1. Same as before
2. Once 2.9.2 development is finished, we create `release_2.9.2` branch but `master` is not blocked anymore and we can keep merging features for coming releases because of step 3 changes. 
3. Once `release_2.9.2` is approved and fully tested, me merged it into `stable`, tag the commit `oc-android-2.9.2` and make the release public via GitHub and PlayStore. 
4. Merge `stable` into `master`.

Think this workflow is widely used and fits properly to our case, what are your thoughts about? @felixboehm @PVince81 @michaelstingl @jesmrec

michaelstingl:
@davigonz we use [OneFlow](https://www.endoflineblog.com/oneflow-a-git-branching-model-and-workflow) (simple version) in the [`ios-app`](https://github.com/owncloud/ios-app/). Could you also have a look?

hannesa2:
When I read "At this point,master branch is blocked ..." this is really bad !
Maybe this is the reason why PR needs that much time here.

It's your decision what to do, but you should definite avoid blocking

hannesa2:
I vote at least for a removement of the PR-must-based-on-recent-master rule and accept such a git flow
[...]

PR #2440, opened 2019-02-09
Search log

hannesa2:
Now we can search log. It's helpful to investigate what's happen with a specific file https://github.com/owncloud/android/issues/2394

It's based/depends on https://github.com/owncloud/android/pull/2402 . That's why it's just a preview and needs a merge of `AndroidX` first. 

searchlog


hannesa2:
With additional commit "_colorize log_" your see colorized

- errors
- warnings
- verbose
- debug and info are like before to avoid too much ugly colors

image

The color comes only when https://github.com/owncloud/android-library/pull/233 is merged, but it's not mandatory.

Btw, you see immediate some missusages of `Log_OC.e(foo)`

hannesa2:
Because app crashes and I don't see on my real device what's wrong, I added a option to look into `logcat` in release version too.
image

Now you can see crashes which are not catched by `Log_OC`

hannesa2:
Btw, I encapsulate the functionality to an own repo https://github.com/hannesa2/Logcat
But it's use https://github.com/JakeWharton/timber and log to file with https://github.com/hannesa2/Logcat/blob/master/LogcatLib/src/main/java/info/hannes/timber/FileLoggingTree.kt and makes 'Log_OC.java' obsolete

If you want to go the way to use this lib, please let me know, otherwise I leave PR like it is

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -64,6 +71,8 @@
 
     private RecyclerView mLogsRecycler;
     private LogListAdapter mLogListAdapter;
+    private SearchView searchView;
```
davigonz: Use mSearchView in here.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -106,16 +117,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        MenuInflater inflater = getMenuInflater();
+        inflater.inflate(R.menu.log_menu, menu);
+        MenuItem searchItem = menu.findItem(R.id.menu_search);
+        SearchManager searchManager = (SearchManager) getSystemService(Context.SEARCH_SERVICE);
+
+        if (searchItem != null) {
+            searchView = (SearchView) searchItem.getActionView();
+            //            searchView.setQueryHint(getString(R.string.text));
```
davigonz: Can we delete this?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -106,16 +117,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        }
+        if (searchView != null) {
+            searchView.setSearchableInfo(searchManager.getSearchableInfo(getComponentName()));
+        }
+
+        if (searchView == null) {
```
davigonz: What about using `else` along with the previous condition `if (searchView != null)` ?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -106,16 +117,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            return true;
+        }
+        SearchView.SearchAutoComplete searchAutoComplete = searchView.findViewById(R.id.search_src_text);
+
+        if (searchView != null && mCurrentFilter != null && !mCurrentFilter.equals("")) {
```
davigonz: Not sure if this check is needed, you are already checking whether `searchView is null` above and returning true if so.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -106,16 +117,99 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+                @Override
+                public boolean onMenuItemActionExpand(MenuItem item) {
+                    // Do something when expanded
+                    return true;  // Return true to expand action view
+                }
+            });
+        }
+
+        if (null != searchView) {
```
davigonz: Same here

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -166,7 +166,7 @@ public void onCreate(Bundle savedInstanceState) {
     @Override
     public void onAttach(Context context) {
         super.onAttach(context);
-        Log_OC.e(TAG, "onAttach");
+        Log_OC.v(TAG, "onAttach");
```
davigonz: Maybe you used this to test the different log modes? Can you keep it as it was before? Thanks

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/LogListAdapter.kt
```diff
@@ -0,0 +1,92 @@
+/*
+ * ownCloud Android client application
+ *
+ * @author David González Verdugo
+ * Copyright (C) 2018 ownCloud GmbH.
```
davigonz: Update the year to 2019

davigonz:
@hannesa2 I requested some changes, can you have a look?

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
hannesa2: This makes a lot of noise ! If you see a red logcat and nothing is wrong, then it's senseless

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
hannesa2: I converted it back from Kotlin

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/LogListAdapter.kt
hannesa2: This complete entry is wrong
This is the right information, I would delete it from complete repo

image

davigonz: I guess the year is not about knowing in which moment was modified the file (git history) but copyright requirement. @michaelstingl @hosy Should we keep the year along with ownCloud Copyright in top of all the files or we can remove it? That's there from long time ago. Thanks

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -184,10 +177,10 @@ public boolean onMenuItemActionExpand(MenuItem item) {
             });
         }
 
-        if (null != searchView) {
-            searchView.setSearchableInfo(searchManager.getSearchableInfo(getComponentName()));
-            searchView.setIconifiedByDefault(true);
-            searchView.setOnQueryTextListener(queryTextListener);
+        if (null != mSearchView) {
```
davigonz: And if you keep `if (mSearchView == null) { return true }` in https://github.com/owncloud/android/pull/2440/commits/d2520555466d395603d89e6ba0303acb9ffe2465#diff-4100ef3dfbd51a1d654425db433334c2L141, you would not need to check it here again

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -131,28 +131,21 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
         if (searchItem != null) {
-            searchView = (SearchView) searchItem.getActionView();
-            //            searchView.setQueryHint(getString(R.string.text));
+            mSearchView = (SearchView) searchItem.getActionView();
         }
-        if (searchView != null) {
-            searchView.setSearchableInfo(searchManager.getSearchableInfo(getComponentName()));
-        }
-
-        if (searchView == null) {
```
davigonz: If you remove this null check it could end up with null pointer exceptions when using `mSearchView.setSearchableInfo` or even `mSearchView.setIconified(false);` so I would keep it as it is and remove the coming null checks over mSearchView.

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/LogListAdapter.kt
davigonz: The main reason of keeping this is that the source code could easily be - someday - used outside of github and the actual structure ... in which case the information would be lost.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
hannesa2: The first way was fine too.
Done

jesmrec:
Log Search -> OK
Colorize -> OK
Logcat -> OK

before merging, please update the PR reference to the `owncloud-android-library` to `master` where https://github.com/owncloud/android-library/pull/233 is already merged. If not, after merging the current PR will have the library reference outdated.

hannesa2:
done

jesmrec:
Transifex pushed before... we need another try.

jesmrec:
#2442 outdated this one. New rebase and will be merged.

hannesa2:
and again

michaelstingl:
> and again

I understand, this is annoying. @jesmrec proposed to change how the Transifex pushes work. We'll check and improve here.

hannesa2:
@michaelstingl thank you to see the issue
An other way could be to change the git flow eg https://github.com/owncloud/android/issues/2359#issuecomment-458815868

hannesa2:
@michaelstingl 
Interesting, this rebase policy is not valid for everyone

image

This is not bad, it just just helps me, to see my position

PR #2443, opened 2019-02-10
remove useless log entries concerning wrong version

hannesa2:
It removes this useless log entries, it concern to beta-version. This entries were miss leading.
Now it writes out a **warning** only on wrong real versionCode

image

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java
@@ -1,4 +1,4 @@
-/**
+/*
  * ownCloud Android client application
  *
  * @author Bartosz Przybylski
@@ -87,8 +87,9 @@ public class FeatureList {
         return features.toArray(new FeatureItem[features.size()]);
     }
 
+    @SuppressWarnings("WeakerAccess")
     static public class FeatureItem implements Parcelable {
-        public static final int DO_NOT_SHOW = -1;
+        static final int DO_NOT_SHOW = -1;
         private int image;
         private int titleText;
         private int contentText;
@@ -96,13 +97,13 @@ public class FeatureList {
         private int betaVersion;
         private boolean showOnInitialRun;
 
-        public FeatureItem(int image, int titleText, int contentText, String version, String betaVersion,
-                           boolean showOnInitialRun) {
+        FeatureItem(int image, int titleText, int contentText, String version, String betaVersion,
+                    boolean showOnInitialRun) {
             this.image = image;
             this.titleText = titleText;
             this.contentText = contentText;
             this.versionNumber = versionCodeFromString(version);
-            this.betaVersion = versionCodeFromString(betaVersion);
+            this.betaVersion = versionBetaCodeFromString(betaVersion);
             this.showOnInitialRun = showOnInitialRun;
         }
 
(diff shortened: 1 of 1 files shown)


owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java
```diff
@@ -160,16 +161,24 @@ public Object createFromParcel(Parcel source) {
...
+        int versionCode = versionBetaCodeFromString(version);
+        if (versionCode == 0) {
+            Log_OC.w(TAG, "Version string is incorrect " + version);
+            return 0;
+        }
+
+        return versionCode;
+    }
+
+    private static int versionBetaCodeFromString(String version) {
```
davigonz: Why do you need this new method to remove the useless log entries? I would keep `versionCodeFromString` as it was before but adding just a new check `(versionCode == 0)` to log or not. Besides, you are not checking `versionCode == 0` for those direct calls to `versionBetaCodeFromString`.

hannesa2: nop, with your suggestion I would ignore versionNumber=0, what's not correct
Current solution is a one check-code, but only warning for non-beta

owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java
```diff
@@ -90,22 +89,23 @@
...
-        public FeatureItem(int image, int titleText, int contentText, String version, String betaVersion,
-                           boolean showOnInitialRun) {
+        FeatureItem(int image, int titleText, int contentText, String version, String betaVersion,
+                    boolean showOnInitialRun) {
             this.image = image;
             this.titleText = titleText;
             this.contentText = contentText;
             this.versionNumber = versionCodeFromString(version);
-            this.betaVersion = versionCodeFromString(betaVersion);
+            this.betaVersion = versionBetaCodeFromString(betaVersion);
```
davigonz: I think that with this call you are ignoring `versionCode == 0` check

owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java
davigonz: @hannesa2 see https://github.com/owncloud/android/pull/2443#discussion_r256507431

hannesa2:
This is my solution, which should help this project, and I want to prevent to apply review changes, I do no support, till my tested fix stop working and causes new issues.

So you have 4 options:

- add your solution
- make an own solution
- merge it, like it is
- close it unmerged

I'm done and I'm fine with any solution

davigonz:
Ok, let's give it a try, code approved, ready to test.

jesmrec:
Two issues to comment here:

1.- By installing the app from scratch, the mentioned log `Version string is incorrect` is there, caused by the welcome wizard.
2- If you delete the history in the log view, no more logs are printed anymore.

jesmrec:
sorry, behaviour 2.- also happens in `master`  branch. I will open a new issue.

hannesa2:
image

I pushed branch to ˚upstream` remote. So you can do what you want

I'm done here

jesmrec:
Ok, thanks a lot for your effort, but the PR did not pass the CR & QA needed for moving forward. Let's check another ones then.

Issue #2422, opened 2019-01-10
Download progress bar is still visible after successfully download


hosy:
### Actual behaviour
I click on a file, which is locally not available. Download view does appear and file will be downloaded.

### Expected behaviour
Progress bar should be hidden, after file was successfully downloaded (progress 100%)
 
### Steps to reproduce
1. Click on a file which is not downloaded on device
2. Download view is opened and shows download progress
3. After file was downloaded, progress bar is still visible

### Environment data
Android version:
9
Device model: 
Mi A2
Stock or customized system:
Stock
ownCloud app version:
2.9.3
ownCloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### ownCloud log (data/owncloud.log)
```
Insert your ownCloud log here
```

hosy:
![screenshot_20190110-115407](https://user-images.githubusercontent.com/736109/50970415-93b7bc80-14e1-11e9-9dcb-79f94f91c565.png)

vikene:
Hey @hosy and @jesmrec , Can I look into this issue ?

PR #2445, opened 2019-02-12
Update local file path after download finishes(Issue #2422)

svvashishtha:
After file download finishes, the progress bar stays visible because the file was not updated in the file detail ui. I've updated the file in detail ui. Issue [https://github.com/owncloud/android/issues/2422](2422)

### DIFF
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1245,9 +1245,11 @@ public class FileDisplayActivity extends FileActivity
                 if (linkedToRemotePath == null || isAscendant(linkedToRemotePath)) {
                     refreshListOfFilesFragment(true);
                 }
+                String localFilePath = intent.getStringExtra(Extras.EXTRA_FILE_PATH);
                 refreshSecondFragment(
                         intent.getAction(),
                         downloadedRemotePath,
+                        localFilePath,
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1245,9 +1245,11 @@ public void onReceive(Context context, Intent intent) {
                 if (linkedToRemotePath == null || isAscendant(linkedToRemotePath)) {
                     refreshListOfFilesFragment(true);
                 }
+                String localFilePath = intent.getStringExtra(Extras.EXTRA_FILE_PATH);
                 refreshSecondFragment(
                         intent.getAction(),
                         downloadedRemotePath,
+                        localFilePath,
```
sd1998: @svvashishtha I don't find the `localFilePath` variable to be used more then once in this function hence I suggest instead of creating a variable you could just do:-
`refreshSecondFragment(
                        intent.getAction(),
                        downloadedRemotePath,
                        intent.getStringExtra(Extras.EXTRA_FILE_PATH),
                        intent.getBooleanExtra(Extras.EXTRA_DOWNLOAD_RESULT, false)
                );`

davigonz:
Sorry guys but this was already fixed via https://github.com/owncloud/android/pull/2448, so I guess we can close it.

Issue #2446, opened 2019-02-13
[2.10.0 beta v1] Release new beta version


davigonz:
### TASKS:
 - [x] [COORDINATION] Get in touch with community/marketing team to publish a blog post, several days before releasing.
 - [x] [DEV] Create **release/2.10.0_beta_v1** branch from master and:  
         - Update version code and name in build.gradle.
         - Update changelog file (< 500 chars)
 - [x] [QA] Perform some tests @jesmrec
 - [x] [DEV] Merge **release/2.10.0_beta_v1** branch into master in owncloud/android 
 - [x] [DIS] Generate test APKs files from branch master in owncloud/android 
 - [x] [GIT] Create and sign tag `oc-android-2.10.0-beta.1` in HEAD commit of master branch, in owncloud/android 
- [x] [DOC] Include new changelog on Releases section of github android repo by drafting the new pre-release 2.10.0 beta
 - [x] [DIS] Generate final APK files from signed commit in owncloud/android and perform some basic operations
_____
### PLAY STORE:
 - [x] [DIS] Upload & publish beta release APK and changelog
_____
### F-DROID:
 - [x] [DEV] Create MR in F-Droid repo to include 2.10.0 beta build. https://gitlab.com/fdroid/fdroiddata/merge_requests/4483
-  [x] [DEV] Merge MR into F-Droid. 
_____
### FINAL TASKS:
 - [x] [SPREADTHEWORD] Write a topic in central about the beta release. 
https://central.owncloud.org/t/owncloud-2-10-beta-app-for-android/18523
 - [x] [SPREADTHEWORD] Share new beta on social media.
_____
_____
### BUGS & IMPROVEMENTS:

davigonz:
@jesmrec you can start with some tests

PR #2449, opened 2019-02-14
2.10.0 beta v1

davigonz:
Implements https://github.com/owncloud/android/issues/2446


jesmrec:
Some tests performed over the new features to include.

Main focus on services that run in background (using jobs) due to changes performed in #2402 . Focus in android 7, 8 and 9. 

Approved on my side. Whole regression test will be performed for the official 2.10.

Issue #2361, opened 2018-11-19
Username with blanks incomplete in drawer


jesmrec:
Steps

1. Create an account which username contain  blanks
2. Add the account and another one (no matter how the username is) to the app
3. Open the second account
4. Open the drawer and tap on the triangle

Current: account name created in 1. appears trimmed
Expected:  account name created in 1. appears correctly in format user@server

Nexus 5X v8

sd1998:
@jesmrec I would like to work on this.

svvashishtha:
I was unable to reproduce this issue. 
I already had an account with a username without spaces. Added another one with spaces. Removed the first one and added it again.
The username for the one with spaces appears as <firstname secondname@server_ip>

jesmrec:
@abelgardep is currently working on it. Could you reproduce?

abelgardep:
Yes, i was able to reproduce it.
I created an account with this username: 'a 1 b 2 c 3'  (without the quotes).
In drawer, this username appears as 'a 1 b 2 c' but it should appear as 'a 1 b 2 c 3@server'
I have not found a fix yet.

svvashishtha:
@jesmrec @abelgardep Able to reproduce now. Any username with more than one space will be trimmed.

davigonz:
@svvashishtha we are not able to see where the problem comes from, you might want to have a look at it as well, thanks!

svvashishtha:
@davigonz I was able to find the root of the problem. Textview clips the text set to it when text  is too large to fit in the given width. 
If there's only one space in the username, the username is clipped at "@" or "/" etc. But if there are more than 1 spaces in the username, the username is clipped at the position of last space. 
Example 
 "s vashisht@192.168.0.65/owncloud"  will be shown as "s vashisht@192.168.0.65"
but 
"s v ashisht@192.168.0.65/owncloud" will be shown as "s v".

To confirm this, you can replace spaces with any other character at this line ([https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java#L419](url)) and the username will be displayed normally. (except the space replaced by other character thingy)

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #2465, opened 2019-02-28
Username with blanks incomplete in drawer #2361

svvashishtha:
The username is too large to be shown completely in the navigation menu for any screen size. Space can be added after the trailing slash in the username.
This allows the username with multiple spaces to not be clipped at the position of last space.

### DIFF
--- a/src/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -400,11 +400,15 @@ public abstract class DrawerActivity extends ToolbarActivity {
         for (int i = 0; i < accounts.length; i++) {
             if (!getAccount().name.equals(accounts[i].name)) {
 
+                String accountName = accounts[i].name;
+                if (accountName.contains("/")) {//add an extra space at <username@serverip/>^<servername>
+                    accountName = accountName.replace("/", "/ ");
+                }
                 MenuItem accountMenuItem = mNavigationView.getMenu().add(
                         R.id.drawer_menu_accounts,
                         Menu.NONE,
                         MENU_ORDER_ACCOUNT,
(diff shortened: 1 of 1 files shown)


davigonz:
Thanks @svvashishtha for this contribution , we will have a look at it for 2.11 release

davigonz:
Code approved, ready to test

jesmrec:
I still reproduce the problem.

With the following user:

Screen Shot 2019-03-21 at 12 05 44

i see this:

Screen Shot 2019-03-21 at 12 04 24

so, the account name is still not completely displayed

jesmrec:
@svvashishtha could you take a look to the reported problem when you have a while? thanks!

svvashishtha:
> @svvashishtha could you take a look to the reported problem when you have a while? thanks!

Sure. Will do.

jesmrec:
thanks!! ping me when it is ready, please.

jesmrec:
@svvashishtha anything new?

svvashishtha:
@jesmrec I have tried a few other things but it appears that the account name will always be clipped if it's too long. In fact, it is being clipped in all devices and all account names.
The only solution I can think of is to introduce a space in front of "@<server_ip>". But even then the username would be clipped if it's too long.

jesmrec:
is it not posible to ellipsize?

svvashishtha:
To ellipsize, a custom layout file will have to be supplied with menu items. 
Would that be acceptable?

davigonz:
> To ellipsize, a custom layout file will have to be supplied with menu items.
Would that be acceptable?

Could you send us your code with that custom layout so we can see how it looks? Thanks

jesmrec:
@svvashishtha are you still interested? we'd like more about your suggestion.

PR #2488, opened 2019-03-15
Use Play Store description for F-Droid

mueller-ma:
Fixes https://github.com/owncloud/android/pull/2267#discussion_r265654755

Can someone please look up what short descriptions are set in the Play Store (English and German)?


davigonz:
The code looks good to me, I guess it does not need QA @jesmrec so a rebase with master would be enough if you agree

fastlane/metadata/android/de-DE/short_description.txt
```diff
@@ -0,0 +1 @@
+File synchronization client
```
mueller-ma: This is still in English. "Client zum synchronisieren von Daten"?

jesmrec: @mueller-ma you are the PR owner. Please, could you change it? After this, if you could rebase the branch against master, we will merge it.

davigonz:
Hi @mueller-ma , could you please apply the change requested by you on https://github.com/owncloud/android/pull/2488#pullrequestreview-234681510?

mueller-ma:
@davigonz Sorry, I totally forgot this PR. I applied the change.

jesmrec:
thanks@mueller-ma! if you can rebase the branch with master to update it, we will merge directly.

mueller-ma:
I don't get it why you can't just use the "Rebase and Merge" button. IMO manual rebasing makes only sense if there are merge conflicts.

davigonz:
> I don't get it why you can't just use the "Rebase and Merge" button. IMO manual rebasing makes only sense if there are merge conflicts.

Because that button uses fast-forward by default and does not include an additional merge we use for knowing which commits belong to a feature for example, it's like a separator when reading master branch history.

Screenshot_2019-05-21_at_13_12_50

Issue #2478, opened 2019-03-10
Canceling dialog in settings turns on setting


mueller-ma:
### Steps to reproduce
1. Open settings
2. Click on "Touches with other visible windows"
3. Cancel the dialog by clicking somewhere outside the dialog
4. Touches with other visible windows is now turned on

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
This is most likely a client-only issue.

### Environment data
Android version: 7.1.2

Device model: Oneplus X

Stock or customized system: Lineage 14

ownCloud app version: 2.1.0

ownCloud server version: 10

### PR
https://github.com/owncloud/android/pull/2492
https://github.com/owncloud/android/pull/2491

jesmrec:
Got it. Same effect in disabling camera uploads

sd1998:
@jesmrec I would like to work on this.

PR #2491, opened 2019-03-16
Fixed ConfirmationDialog behaviour on outside the dialog touch

sd1998:
Fix to issue #2478.


sd1998:
@jesmrec I have fixed #2478.

jesmrec:
we finally went for the fix in #2492 to fix the issue #2478. Anyway, thanks for you effort!

PR #2492, opened 2019-03-16
Fix and cleanup confirmation dialogs in settings

### LINKED ISSUE #2478: Canceling dialog in settings turns on setting

mueller-ma:
Fixes #2478
Closes #2491

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -331,32 +331,23 @@ public class Preferences extends PreferenceActivity {
 
         if (mPrefTouchesWithOtherVisibleWindows != null) {
             mPrefTouchesWithOtherVisibleWindows.setOnPreferenceChangeListener((preference, newValue) -> {
-                        SharedPreferences.Editor appPrefs = PreferenceManager.
-                                getDefaultSharedPreferences(getApplicationContext()).edit();
+                    SharedPreferences.Editor appPrefs = PreferenceManager.
+                            getDefaultSharedPreferences(getApplicationContext()).edit();
 
-                        if ((Boolean) newValue) {
-                            showConfirmationDialog(
-                                    getString(R.string.confirmation_touches_with_other_windows_title),
-                                    getString(R.string.confirmation_touches_with_other_windows_message),
-                                    (dialog, which) -> {
-                                        if (which == DialogInterface.BUTTON_POSITIVE) {
-                                            appPrefs.putBoolean(
-                                                    PREFERENCE_TOUCHES_WITH_OTHER_VISIBLE_WINDOWS,
-                                                    true
-                                            );
-                                        } else if (which == DialogInterface.BUTTON_NEGATIVE) {
-                                            mPrefTouchesWithOtherVisibleWindows.setChecked(false);
-                                        }
-                                        dialog.dismiss();
-                                    });
-                        } else {
-                            appPrefs.putBoolean(PREFERENCE_TOUCHES_WITH_OTHER_VISIBLE_WINDOWS, false);
-                        }
-
-                        appPrefs.apply();
-
-                        return true;
+                    if ((Boolean) newValue) {
+                        new AlertDialog.Builder(this)
+                                .setTitle(getString(R.string.confirmation_touches_with_other_windows_title))
+                                .setMessage(getString(R.string.confirmation_touches_with_other_windows_message))
+                                .setNegativeButton(getString(R.string.common_no), null)
+                                .setPositiveButton(getString(R.string.common_yes), (dialog, which) -> {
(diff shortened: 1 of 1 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -331,32 +331,23 @@ public void onCreate(Bundle savedInstanceState) {
...
-                        }
-
-                        appPrefs.apply();
-
-                        return true;
+                    SharedPreferences.Editor appPrefs = PreferenceManager.
+                            getDefaultSharedPreferences(getApplicationContext()).edit();
+
+                    if ((Boolean) newValue) {
+                        new AlertDialog.Builder(this)
```
davigonz: Why have you removed the showConfirmationDialog method and decided to create the dialogs directly several times?

davigonz:
Thanks for this contribution @mueller-ma , can you have a look at what I asked above?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
mueller-ma: First of all the current implementation of showConfirmationDialog() mis-uses the Builder class.
````
        AlertDialog alertDialog = new AlertDialog.Builder(this).create();
        alertDialog.setTitle(title);
        alertDialog.setMessage(message);
        alertDialog.setButton(DialogInterface.BUTTON_NEGATIVE, getString(R.string.common_no), listener);
        alertDialog.setButton(DialogInterface.BUTTON_POSITIVE, getString(R.string.common_yes), listener);
        alertDialog.show();
````
The intent use of builders is:
````
        AlertDialog alertDialog = new AlertDialog.Builder(this)
        .setTitle(title);
        .setMessage(message);
        .setButton(DialogInterface.BUTTON_NEGATIVE, getString(R.string.common_no), listener);
        .setButton(DialogInterface.BUTTON_POSITIVE, getString(R.string.common_yes), listener);
        .show();
````

After fixing this I realized that showConfirmationDialog() is now a simple wrapper for AlertDialog.Builder() which adds nearly no "static" information, like the strings for the buttons. Thus I inlined the function.
Also IMO it's nice to have two listener for both buttons instead of one listener that has a `if (which == DialogInterface.BUTTON_POSITIVE)...`.

davigonz: > First of all the current implementation of showConfirmationDialog() mis-uses the Builder class.

Yes, good catch

> showConfirmationDialog() is now a simple wrapper for AlertDialog.Builder() which adds nearly no "static" information, like the strings for the buttons. 

The aim of that is to save code and avoid using `setTitle, setMessage` and `setButton` multiple times. Besides, if we wanted to change the title of the buttons at some point, we would just need to do it in a single place.

About using two listeners, you could keep the `showConfirmationDialog` and pass two different listeners to it instead of just one.

mueller-ma: IMO calling the Builder is much easier to read. You instantly know what `setTitle()` does, but you'll have to look at showConfirmationDialog() implementation to know what the third parameter is.

davigonz: > You instantly know what setTitle() does

I think that if we followed this approach, we would never create methods to avoid duplicated code since using constructors or builders directly is always easier to read. 

About the third parameter, maybe creating a listener variable before calling `showConfirmationDialog()` let readers know better what the third parameter is.

IMO, avoiding duplicated code by calling methods which create/build objects is a good practice.

mueller-ma: I understand your arguments, but in this case the duplicate code is `android.R.string.ok` and `android.R.string.ok`.
If you really want to have a `showConfirmationDialog()` function I'll create it anyway.

davigonz: No worries, since you have included a new listener (setOnCancelListener) maybe is too much for `showConfirmationDialog` and your option fits better.

jesmrec:
Approved. Now the option is not toggled till the correct option is tapped. Thanks for the contribution!

jesmrec:
@mueller-ma could you rebase to update the branch?

mueller-ma:
PR rebased

jesmrec:
thanks!! merging...

Issue #601, opened 2014-08-27
[Enhancement] Remove files from phone after upload


Bugsbane:
I mainly use ownCloud as a place to store and share photos and videos. I like to use high quality compression settings, meaning my phone uses it's storage space quickly.

Given that I already use instant upload, I would be helpful if there was an option to remove files successfully uploaded (especially by instant upload) so that I could avoid constantly running out of space and having to manually delete files that have already been uploaded to ownCloud.

Bugsbane:
I'm not sure this is a duplicate.

As far as I can tell, #346 is about not duplicating uploaded files on the device so there aren't 2 copies on the local device. That is is definitely needed for my usecase, however this issue was about removing the *original* copy from the device after use.

Right now, uploaded files exist twice on the local device.
#346 would make sure they only existed once.
In this issue I want them to not exist at all (on the local device).

tobiasKaminsky:
You are correct.
Re-opening it.

tobiasKaminsky:
@Bugsbane with the current version you could just delete your instantupload folder from time to time.
Is this enough for you?
Else we could a third option to "original file will be..." called "deleted.
@jancborchardt @davivel

Bugsbane:
I see in the beta app, there's an option that says "Original file will be..  Moved to the app folder". If " the app folder " means the ownCloud instant upload folder and moved is being used in the normal meaning of the word (ie not duplicated), and if uploads can be relied on to upload every file requested eventually, ( #919? ) then as far as I can see, my use case should already be covered without needing to manually go in and delete anything.

Maybe I'm misunderstanding something though.

tobiasKaminsky:
This is also in current version.
And yeah, your assumptions are all correct.

Bugsbane:
Cool. Well, sounds like we can close this then, as it seems the functionality I was asking for has been implemented...

tobiasKaminsky:
Please try it first and confirm that it is working they way you want it :)

Bugsbane:
Will do. I'm actually having some non oc related issues with my server at the moment that make this difficult, but when they're resolved, I'll give it a shot.

PanicDriven:
[...]

PR #2509, opened 2019-04-02
[ft] add remove file on source when uploading items

mstojanovic:
see issue #601

upload from camera dir didnt work on my tests for copy and move so i havent implemented the remove part either

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java
@@ -153,6 +153,7 @@ public class FileUploader extends Service
     public static final int LOCAL_BEHAVIOUR_COPY = 0;
     public static final int LOCAL_BEHAVIOUR_MOVE = 1;
     public static final int LOCAL_BEHAVIOUR_FORGET = 2;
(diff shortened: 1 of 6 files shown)


jesmrec:
Thanks for your contribution. We will take a look and let you know our findings

jesmrec:
@mstojanovic we plan to include this during the following days. There are some conflicts with our `master`, if you can take a look it will be appreciated. Anything you need, please ping us. Thanks

mstojanovic:
this is obsolete due to the switch to a native picker on android devices

jesmrec:
> this is obsolete due to the switch to a native picker on android devices

Native android picker does not affect the camera uploads. That is only for manual uploads.

mstojanovic:
see first post, camera upload didnt work for me during testing on emulator or physical device. pull request can be discarted

jesmrec:
ok, thanks. Out of the box, we have also received other reports from users which camera uploads are not working either. It would be very helpful if you can send us logs (available in `Settings`), or try top reproduce any scenario that help us to detect where the problem is. Let us know if you are interested. Thanks in advance anyway!

jesmrec:
After moving to native picker, we can close this. Thanks anyway for contributing!!

Issue #2500, opened 2019-03-25
[Document Provider] Create new folder


abelgardep:
New improvement to Document Provider : Create new Folder -> [Doc](https://developer.android.com/reference/android/provider/DocumentsProvider.html#createDocument(java.lang.String,%20java.lang.String,%20java.lang.String)) API 19

| Actual behaviour | Expected behaviour | 
| ------------- |-------------| 
| ![ActualBehaviour](https://user-images.githubusercontent.com/47524927/55476817-db00a580-5617-11e9-8f54-bc423d36b8e6.png) | ![ExpectedBehaviour](https://user-images.githubusercontent.com/47524927/55476921-1ac78d00-5618-11e9-97f7-2be044938ecd.gif) |

### TASKS

- [x] Research (if needed)
- [x] Create branch document_provider/create_folder
- [x] Development tasks
  - [x] Add flag to FileCursor
  - [x] Implement createDocument in DocumentsProvider
  - [x] Implement createFolder when mimetype is Dir
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch document_provider/create_folder into master

### PR
#2510

PR #2510, opened 2019-04-03
[Document Provider] Create Folder

abelgardep:
Implements #2500 
___

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.11.0/2510-document_provider_create_folder.md
___

Bugs & improvements:

- [X] (1) Invalid characters https://github.com/owncloud/android/pull/2510#issuecomment-487899325 [FIXED]
- [X] (2) P2R to make the folder visible https://github.com/owncloud/android/pull/2510#issuecomment-487900644 -> https://github.com/owncloud/android/issues/2536

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
@@ -29,7 +29,6 @@ import android.content.res.AssetFileDescriptor
 import android.database.Cursor
 import android.graphics.Point
 import android.net.Uri
-import android.os.Build
 import android.os.CancellationSignal
 import android.os.ParcelFileDescriptor
 import android.provider.DocumentsContract
@@ -211,7 +210,7 @@ class DocumentsStorageProvider : DocumentsProvider() {
         Log_OC.d(TAG, "Trying to rename ${file.fileName} to $displayName")
 
         RenameFileOperation(file.remotePath, displayName).apply {
-            execute(currentStorageManager, context).also { checkOperationResult(it, file) }
+            execute(currentStorageManager, context).also { checkOperationResult(it, file.parentId.toString()) }
         }
 
         return null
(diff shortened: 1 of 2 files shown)


abelgardep:
@davigonz Ready to CR

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -209,20 +225,32 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
         }
     }
 
-    private fun checkOperationResult(result: RemoteOperationResult<Any>, file: OCFile) {
+    private fun checkOperationResult(result: RemoteOperationResult<Any>, folderToNotify: String) {
         if (!result.isSuccess) {
-            notifyChangeInFolder(file.parentId.toString())
+            if (result.code != RemoteOperationResult.ResultCode.WRONG_CONNECTION) notifyChangeInFolder(folderToNotify)
             throw FileNotFoundException("Remote Operation failed due to ${result.exception.message}")
         } else {
```
davigonz: You might not need the else condition since you are throwing an exception just before and the code below else will be only reachable when the result is successes.

davigonz:
@abelgardep just a tiny suggestion

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
abelgardep: Done

abelgardep:
Done

davigonz:

jesmrec:
Let's QA!

jesmrec:
### (1) (improvement)

Invalid characters:

In the app, if you create a folder with "/" or "\\" , it is forbidden

If you do the same in document provider app...

Current behaviour: folder is created without the forbidden character. 
Expected behaviour: maybe, not allowed (same behaviour as the app).

The current behaviour is not wrong, so it is avoiding bad experience in synchronization, but if it'd handle in the same way as the app, it would nice

jesmrec:
### (2)

When a folder is created in document provider, it appears in the app but it is vanished at the moment. A pull to refresh is needed to make it visible:

ezgif com-optimize (1)

could it be improved?

Nexus 6P v7

davigonz:
[(2) bug](https://github.com/owncloud/android/pull/2510#issuecomment-487900644) is reproducible in master so I would not address it here

[](https://user-images.githubusercontent.com/15261363/57215733-16630b00-6fee-11e9-8a98-6fa8099117ec.gif)

abelgardep:
(1) Ready to Test

jesmrec:
(1) fixed. 

(2) will be addresed in a separate issue because this is not a document provider problem: https://github.com/owncloud/android/issues/2536

approved

Issue #2499, opened 2019-03-25
[Document Provider] Delete Files or Folders


abelgardep:
New improvement to Document Provider : Delete Files or Folders -> [Doc](https://developer.android.com/reference/android/provider/DocumentsProvider.html#deleteDocument(java.lang.String)) API 19

Delete a file or folder including all child files.

Current status:
- Option disabled in OwnCloud Provider

### TASKS

- [x] Add 'Delete Document' functionality to Document Provider

### PR
#2511 

### Test plan: 
https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.11.0/2511-document_provider_delete.md

sd1998:
@abelgardep I would like to work on this.

abelgardep:
@sd1998 I'm migrating DocumentsProvider to kotlin. This way, new functionality related with DocumentsProvider will be developed in kotlin. When i finish the migration i can ping you if you want to work on this 👍

abelgardep:
Hi @sd1998 , i will work on this because i have been working on the Document Provider for the last two weeks. Thanks anyway

PR #2511, opened 2019-04-03
[Document Provider] Delete File/Folder

abelgardep:
Implements #2499 

____

Bugs & improvements

- [x] (1) With no connection, success message is shown https://github.com/owncloud/android/pull/2511#issuecomment-483673397

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
@@ -29,6 +29,7 @@ import android.content.res.AssetFileDescriptor
 import android.database.Cursor
 import android.graphics.Point
 import android.net.Uri
+import android.os.Build
 import android.os.CancellationSignal
 import android.os.ParcelFileDescriptor
 import android.provider.DocumentsContract
@@ -40,6 +41,7 @@ import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.files.services.FileDownloader
 import com.owncloud.android.lib.common.utils.Log_OC
 import com.owncloud.android.operations.RefreshFolderOperation
+import com.owncloud.android.operations.RemoveFileOperation
 import com.owncloud.android.operations.RenameFileOperation
 import com.owncloud.android.providers.cursors.FileCursor
 import com.owncloud.android.providers.cursors.RootCursor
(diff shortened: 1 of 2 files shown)


owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -190,18 +193,50 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
-            throw java.lang.UnsupportedOperationException("Rename failed")
+            throw FileNotFoundException("Remote Operation failed due to ${result.exception.message}")
+
         } else {
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                revokeDocumentPermission(file.fileId.toString())
+            }
             syncRequired = false
             context?.contentResolver?.notifyChange(toNotifyUri(toUri(file.parentId.toString())), null)
```
davigonz: This line appears several times in the class, you might create a new method called `notifyChangeInDocument(documentId)` or something like that

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -190,18 +193,50 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+
+    override fun deleteDocument(documentId: String) {
+        val docId = documentId.toLong()
+
+        updateCurrentStorageManagerIfNeeded(docId)
+
+        val file = currentStorageManager?.getFileById(docId) ?: throw FileNotFoundException("File $docId not found")
+        Log_OC.d(TAG, "Trying to delete ${file.fileName} with id ${file.fileId}")
+
+        RemoveFileOperation(file.remotePath, false).apply {
```
davigonz: Wow, taking advantage of Kotlin power, just a couple of lines to execute the operation and check the result, love it 👏

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -190,18 +193,50 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+
+        //Documents can only have one parent
+        deleteDocument(documentId)
+
+    }
+
+    private fun checkOperationResult(result: RemoteOperationResult<Any>, file: OCFile) {
 
         if (!result.isSuccess) {
+
```
davigonz: I would not include this empty line here nor after `else {` but up to you

davigonz:
@abelgardep a few changes requested

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -190,18 +193,50 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+        val file = currentStorageManager?.getFileById(docId) ?: throw FileNotFoundException("File $docId not found")
+        Log_OC.d(TAG, "Trying to delete ${file.fileName} with id ${file.fileId}")
+
+        RemoveFileOperation(file.remotePath, false).apply {
+            execute(currentStorageManager, context).also { checkOperationResult(it, file) }
+        }
+    }
+
+    @TargetApi(24)
```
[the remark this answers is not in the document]
abelgardep: Remove is available from API24 -> (https://developer.android.com/reference/android/provider/DocumentsContract.Document.html#FLAG_SUPPORTS_REMOVE)
Delete is available from API19 -> (https://developer.android.com/reference/android/provider/DocumentsContract.Document.html#FLAG_SUPPORTS_DELETE)

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -190,18 +193,50 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+    override fun removeDocument(documentId: String, parentDocumentId: String) {
```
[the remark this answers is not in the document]
abelgardep: It is not, is it? Files are hierarchical, so they only have one parent. I suppose this function is not needed at all.

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
abelgardep: Ok

abelgardep:
Changes applied.

jesmrec:
Let's QA

jesmrec:
### (1)

1. Remove the connection of the device, stop the server or set the maintenance mode
2. Try to delete content using document provider

Current: File is not deleted, but snackbar shows success message
Expect: Some kind of error, since the file was not deleted. Is this something feasible or is it handled directly by Android?

Android 8

abelgardep:
@jesmrec 

In those cases, in Android 7.1.1 appears a notification error: 'These files weren't deleted: [List of files]'

jesmrec:
The matter was in the notification. There, we can see correctly that the files were not deleted.

jesmrec:
Approved.

PR #2525, opened 2019-04-25
Update issue templates

davigonz:
I've just created new issue templates from the old ones and placed them under `.github/ISSUE_TEMPLATE` folder as [About issue and pull request templates](https://help.github.com/en/articles/about-issue-and-pull-request-templates) suggests. This way they'll be available for contributors/us to use when opening new issues in the repository, instead of using manual copy-paste, as we did with release template for example.

This is how the issues view in our repo should look like from now on

new-issue-page-with-multiple-templates

I've replaced the old templates with three new ones (Feature request, bug report and release issue), please have a look at them and suggest whatever you consider.

CC / @jesmrec @michaelstingl @abelgardep


jesmrec:
For me it is OK to have separate templated for each kind of issue. These templates are a good start point.

Issue #2498, opened 2019-03-25
[Document Provider] Edit files and save changes


abelgardep:
New improvement to Document Provider: Open File Write Mode (API 19) [Doc](https://developer.android.com/reference/android/provider/DocumentsProvider.html#openFile(android.net.Uri,%20java.lang.String,%20android.os.CancellationSignal))

Open a document, edit it, and then save changes.

Current status:
- Allow users to open files within provider
- Changes are not saved

Expected status:
- Allow users to open files within provider
- Save changes and update server with them
- Notification if there is a conflict with file in server

### TASKS

- [x] Research (if needed)
- [x] Create branch document_provider/open_document_write_mode
- [x] Development tasks
  - [x] Enable write mode on openDocument
  - [x] Sync file with server
  - [x] Notify when conflict
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch document_provider/open_document_write_mode into master

### PR
#2527

PR #2527, opened 2019-04-29
[Document Provider] Edit files and save changes

abelgardep:
Implements #2498 

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.11.0/2527-document_provider_edit_file.md

Bugs & improvements:

- [x] (1) Behaviour with no connection on non-downloaded files https://github.com/owncloud/android/pull/2527#issuecomment-489058679 [WONT FIX]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
@@ -30,6 +30,7 @@ import android.database.Cursor
 import android.graphics.Point
 import android.net.Uri
 import android.os.CancellationSignal
+import android.os.Handler
 import android.os.ParcelFileDescriptor
 import android.provider.DocumentsContract
 import android.provider.DocumentsProvider
@@ -45,10 +46,12 @@ import com.owncloud.android.operations.CreateFolderOperation
 import com.owncloud.android.operations.RefreshFolderOperation
 import com.owncloud.android.operations.RemoveFileOperation
 import com.owncloud.android.operations.RenameFileOperation
+import com.owncloud.android.operations.SynchronizeFileOperation
 import com.owncloud.android.providers.cursors.FileCursor
 import com.owncloud.android.providers.cursors.RootCursor
 import java.io.File
 import java.io.FileNotFoundException
+import java.io.IOException
 import java.util.HashMap
 import java.util.Vector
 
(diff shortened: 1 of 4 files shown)


owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -87,7 +92,40 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+        val accessMode: Int = ParcelFileDescriptor.parseMode(mode)
+
+        val isWrite: Boolean = mode.contains("w")
+        return if (isWrite) {
+            val handler = Handler(context?.mainLooper)
+            // Attach a close listener if the document is opened in write mode.
+            try {
+                ParcelFileDescriptor.open(File(file.storagePath), accessMode, handler) {
+                    // Update the file with the cloud server. The client is done writing.
+                    Log_OC.d(TAG, "A file with id $documentId has been closed! Time to update the server.")
```
davigonz: I would say "time to synchronize it with server"

owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -87,7 +92,40 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+                            }
+                        }
+                    }.apply { start() }
+                }
+            } catch (e: IOException) {
+                throw FileNotFoundException(
+                    "Failed to open document with id $documentId and mode $mode"
+                )
+            }
+        } else {
```
davigonz: You could avoid this `else` condition by doing something like 

```
if (!isWrite) return ParcelFileDescriptor.open(File(file.storagePath), accessMode)

// Attach a close listener if the document is opened in write mode.
try {
    val handler = Handler(context?.mainLooper)
    
    return ParcelFileDescriptor.open(File(file.storagePath), accessMode, handler) {
        // Update the file with the cloud server. The client is done writing.
        Log_OC.d(TAG, "A file with id $documentId has been closed! Time to update the server.")
        Thread {
            SynchronizeFileOperation(
                file,
                null,
                currentStorageManager?.account,
                false,
                context,
                false
            ).apply {
                val result = execute(currentStorageManager, context)
                if (result.code == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
                    NotificationUtils.notifyConflict(file, currentStorageManager?.account, context)
                }
            }
        }.apply { start() }
    }
} catch (e: IOException) {
    throw FileNotFoundException(
        "Failed to open document with id $documentId and mode $mode"
    )
}
```

davigonz:
@abelgardep just a couple of changes...

abelgardep:
@davigonz Changes applied

davigonz:

jesmrec:
Starting QA...

jesmrec:
### (1)

Check these steps:

1. Remove connection from device
2. From downloads/files app, try to open a non-downloaded file

Current behaviour: file is tried to be opened forever
Expected: i reproduced the same steps over `master` branch. There, when the file is retrieved for the 3rd party app, an error message is displayed

Notification is OK, showing the connection problem

With `master`: 

May-03-2019 12-46-39

With current branch:

May-03-2019 12-56-53

Nexus 6P v7

jesmrec:
(1) is finally reproducible in master, so it will be moved to a separate issue. 

apart of that, everything OK

Approved

Issue #2518, opened 2019-04-10
[Document Provider] Account management & Multiaccount


jesmrec:
This is the current behaviour after installing the app from scratch

1. Add one account to ownCloud  app -> account appears in Files (OK)
2. Add a second (or more) accounts to ownCloud app -> accounts do not appear in Files (NOK, should appear as many accounts as the ones attached to the app)
3. Delete from the app the account added in 1. -> the account is still available in Files (NOK, the account is removed)

As improvement, the account is displayed with the user name, but it would be nice to use the display name, if feasible. 

### TASKS

- [x] Research (if needed)
- [x] Create branch document_provider/account_management
- [x] Development tasks
  - [x] Notify Document Provider when user logs in
  - [x] Notify Document Provider when user logs out
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch document_provider/account_management into master

### PR

#2528

abelgardep:
Multiaccount is supported, but Document Provider is not listening to logins or logouts. I agree that we need to listen to those events for security, so i include this improvement in DocumentsProvider epic.

> As improvement, the account is displayed with the user name, but it would be nice to use the display name, if feasible.

We can display user name or display name or whatever, we can change it without problems.

PR #2528, opened 2019-04-30
[Document Provider] Account management

abelgardep:
Implements #2518 

____

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.11.0/2528-document_provider_account_manager.md

bugs & improvements: 

- [x] (1) https://github.com/owncloud/android/pull/2528#issuecomment-489969621 [FIXED]
- [x] (2) https://github.com/owncloud/android/pull/2528#issuecomment-492258721 [NOT REPRODUCIBLE]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/RemoveAccountDialogFragment.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/RemoveAccountDialogFragment.java
@@ -20,11 +20,11 @@
 package com.owncloud.android.ui.dialog;
 
 /**
- *  Dialog requiring confirmation before removing an OC Account.
- *
- *  Removes the account if the user confirms.
- *
- *  Container Activity needs to implement AccountManagerCallback<Boolean>.
+ * Dialog requiring confirmation before removing an OC Account.
+ * <p>
+ * Removes the account if the user confirms.
+ * <p>
+ * Container Activity needs to implement AccountManagerCallback<Boolean>.
  */
 
 import android.accounts.Account;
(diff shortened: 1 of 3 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/dialog/RemoveAccountDialogFragment.java
```diff
@@ -103,6 +107,11 @@ public void onConfirmation(String callerTag) {
         AccountManager am = AccountManager.get(parentActivity);
         AccountManagerCallback<Boolean> callback = (AccountManagerCallback<Boolean>) parentActivity;
         am.removeAccount(mTargetAccount, callback, new Handler());
+
+        // Notify removal to Document Provider
+        String authority = getResources().getString(R.string.document_provider_authority);
+        Uri rootsUri = DocumentsContract.buildRootsUri(authority);
+        Objects.requireNonNull(getContext()).getContentResolver().notifyChange(rootsUri, null);
```
davigonz: Just a question, why do you use `Objects` and not something like:

```
if (getContext() != null) {
    getContext().getContentResolver().notifyChange(rootsUri, null);
}
```

jesmrec:
### (1)

1. Add an account
2. In oC, open drawer -> manage accounts
3. Add account from account manager
4. Add correct credentials of a new account -> account added and account manager displayed with both accounts
5. Minimize oC app and open Downloads/Files app, and select the account added in step 3

Current: app crashes (maybe because file list was not fetched yet)
Expected: not crash. To display, here depends on what can be showed at such point.

Nexus 5X v8

davigonz:
@jesmrec https://github.com/owncloud/android/pull/2528#issuecomment-489969621 is ready to test

jesmrec:
### (2)

By repeating the steps in (1), the app does not crash anymore. But, even browsing in the app through the account, in "Downloads" is shown `Can't load content at the moment`

jesmrec:
The issue explained in (2) is not always reproducible by using the same steps. Let's give a try to the current implementation and in case of someone or somewhere can get more specific steps, we will open a new issue.

Approved.

PR #2532, opened 2019-05-02
Add TOC Table Of Contents

mmattel:
This PR adds a Table Of Contents to the documentation pages.

We did this in docs too, live example see: https://doc.owncloud.com/server/admin_manual/installation/system_requirements.html

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -1,6 +1,7 @@
 = Using the ownCloud Android App
 :toc: right
 :toclevels: 1
+:experimental:
 
 == Introduction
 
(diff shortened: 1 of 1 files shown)


michaelstingl:
Could you post a screenshot?

mmattel:
Here we go:
image

Please also take a look to life docs in server (example): https://doc.owncloud.com/server/admin_manual/installation/system_requirements.html

michaelstingl:
Better add this to the left navigation? (then it's always present while scrolling)

mmattel:
`:toc: right` is prepared for exactly doing that, but this is a matter of how the css rule is made. Currently there is none, so rendering is as you see it. At the moment where (in docs-ui) the css rule is made, the toc is automatically set where the rule defines it :smile:

michaelstingl:
> a toc should be present in all pages (where necessary) of the repros included and look the same.

@mmattel Sounds like we should make this part of the template, maybe with a parameter to switch it off. @LukasHirt is working on new templates. Let's see…

mmattel:
drone: Could not resolve host: github.com .... huuuuu

Issue #2539, opened 2019-05-08
[Document Provider] Create Files


abelgardep:
This feature allows client apps to create files within document provider

Official Docs about it: [Support Document Creation](https://developer.android.com/guide/topics/providers/create-document-provider.html#creating)

### TASKS

- [x] Research (if needed)
- [x] Create branch document_provider/create_files
- [x] Development tasks
  - [x] Enable Create Documents at Root Cursor
  - [x] Implements createFiles at Document Provider 
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch document_provider/create_files

### PR

- App #2541 
- Library (if needed)

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.13.0/2541-create_file_provider.md

- [X] (1) File is not uploaded after updating https://github.com/owncloud/android/issues/2539#issuecomment-523788486 [FIXED]

jesmrec:
@abelgardep when a file is created and saved in oC, the is the file removed from original path (moved)?

abelgardep:
> @abelgardep when a file is created and saved in oC, the is the file removed from original path (moved)?

I suppose it depends on how you create that file. If you create a file and u didn't save anywhere, if you save it in OwnCloud it should save only in oC. If you open a file that exists locally and save in oC it should be like a copy.

jesmrec:
thanks for the clarification!

jesmrec:
### (1)

Please check these steps out (regression):

1. Open a file in oC in an external app
2. Change the content of the file

Current: file in oC is not uploaded till you open again the app and open the updated file.
Expected (`master` behaviour): In the moment the changes are saved in the external app, the file is uploaded to oC

MIA2 Android 9

davigonz:
> Open a file in oC in an external app

@jesmrec could you please detail what is the external app you are using?

jesmrec:
@abelgardep i checked with :

- MSWord
- Text Editor (https://play.google.com/store/apps/details?id=com.byteexperts.texteditor&hl=en_US)

michaelstingl:
> * MSWord

Create a new file in MSWord and "Save as…" in ownCloud doesn't work for me yet (with @abelgardep 's `*.APK`)

jesmrec:
details are always welcome. I was able to create and save.

jesmrec:
[...]

PR #2541, opened 2019-05-08
[Document Provider] Create Files

abelgardep:
Implements #2539

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
@@ -39,6 +39,8 @@ import com.owncloud.android.authentication.AccountUtils
 import com.owncloud.android.datamodel.FileDataStorageManager
 import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.files.services.FileDownloader
+import com.owncloud.android.files.services.FileUploader
+import com.owncloud.android.files.services.TransferRequester
 import com.owncloud.android.lib.common.operations.RemoteOperationResult
 import com.owncloud.android.lib.common.utils.Log_OC
 import com.owncloud.android.lib.resources.files.FileUtils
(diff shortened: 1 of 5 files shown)


owncloudApp/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.kt
```diff
@@ -65,66 +66,84 @@ class DocumentsStorageProvider : DocumentsProvider() {
...
+                    if (!waitOrGetCancelled(signal)) {
+                        return null
+                    }
+                    ocFile = currentStorageManager?.getFileById(docId)
+                        ?: throw FileNotFoundException("Failed to open document with id $documentId and mode $mode")
+
+                } while (!ocFile.isDown)
+            }
+        } else {
+            ocFile = fileToUpload
```
davigonz: If we are able to know when the provider uses this method to upload or download a file, maybe we do not need to use SynchronizeFileOperation to perform a simple upload and a call to `uploadNewFile` of `TransferRequester.java` from here is enough. This way we reduce some logic in SynchronizeFileOperation by removing the `uploadOnly` parameter and do not overload it with extra responsibilities. What do you think?

You could create a transferRequest in this else condition

```
val transferRequester = TransferRequester()
transferRequester.uploadNewFile(context, currentStorageManager?.account, ..., ...)
```

abelgardep: Yes, that would be definitively better

davigonz: Great, ping me when is ready so I can continue with the review, thanks!

abelgardep: But we'll need to create that transfer requester later, as it is done at that commit 63e19bf97d6c11af1705597ecc7bc36b0e1a8226

davigonz:

jesmrec:
Approved

Issue #2451, opened 2019-02-14
Login fails with "§" in password


Citrodata:
### Actual behaviour
- We can login with a User

### Expected behaviour
-Login in with a User
 
### Steps to reproduce
1.  Set the Password of a User to QWERTZqwert123456!"§$%&
2. Login with App

### Error
Wrong User
We are using LDAP

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

No only if the Password is QWERTZqwert123456!"§$%&

### Environment data
Android version:
8
Device model: 
All
Stock or customized system:
Stock Version
ownCloud app version:
2.9.3
ownCloud server version:
10.1.0

### PR

- App https://github.com/owncloud/android/pull/2542
- Library (if needed) https://github.com/owncloud/android-library/pull/249

michaelstingl:
> Can this problem be reproduced with the official owncloud server?
> (url: https://demo.owncloud.org, user: test, password: test)

@Citrodata You can login with admin:admin, and create a new user.

I played with it a bit, and it seems the length isn't a problem. This Password worked too:
`QWERTZqwert123456!"$%&evenlonger`
^ I removed the "`§`" character

On the other hand, short passwords with "`§`" character don't work:
`123§`

ownCloud 2.9.3 on Pixel 3 XL running Android 9

jesmrec:
Regression? https://github.com/owncloud/ios/issues/273

jesmrec:
I have checked the request sent to the server, and the character is not sent in the base64 encoded header. We will have to check if the current password management supports (or not) unicode characters.

PR #2542, opened 2019-05-08
Fix "Login fails with "§" in password " 

davigonz:
Fixes owncloud/android#2451
Needs https://github.com/owncloud/android-library/pull/249


davigonz:

jesmrec:
Fixed!

Tested with different special characters in password, apart of § (¨, ^, `, @, #...)

Approved.

Issue #2517, opened 2019-04-09
[New architecture] Delete public shares


davigonz:
New use case, Delete public share for a file

_____
### TASKS

- [x] Include delete public share query in OCShareDao
- [x] Include delete public share operation in OCLocalDataSource
- [x] Include delete public share operation OCRemoteDataSource
- [x] Include delete public share operation in OCShareRepository
- [x] Include delete public share function in OCShareViewmodel.
- [x]  Connect ShareFileActivity with OCShareViewmodel to delete shares.
- [x] Error handling, adapt ErrorMessageAdapter
- [x] Clean up some dependencies:
  - [x] Get shares
  - [x] Create shares
  - [x] Edit shares
  - [x] Remove shares
- [x] Fix UI tests after cleaning up dependencies in code
- [x] Use room implementation in Content Provider to delete shares
- [x] Tests:
  - [x] OCLocalDataSource
  - [x] OCRemoteDataSource
  - [x] OCShareRepository
  - [x] OCShareViewModel
  - [x] OCShareDao
  - [x] UI tests (To be completed within #2531)

### TASKS
- App: https://github.com/owncloud/android/pull/2544
- Library, when needed: https://github.com/owncloud/android-library/pull/250

PR #2544, opened 2019-05-10
[New arch] Delete public share

davigonz:
Implements https://github.com/owncloud/android/issues/2517
Needs https://github.com/owncloud/android-library/pull/250


owncloudApp/src/androidTest/java/com/owncloud/android/shares/db/OCShareDaoTest.kt
```diff
@@ -167,183 +163,143 @@ class OCShareDaoTest {
...
-            listOf(
-                TestUtil.createPublicShare(
-                    path = "/Texts/text1.txt",
-                    isFolder = false,
-                    name = "Text 1 link",
-                    shareLink = "http://server:port/s/1"
-                )
-            )
-        )
+        ocShareDao.insert(listOf(newPublicShare()))
```
jcmontesmartos: Haven't you created a method for a just one item instead of a list?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/PublicShareEditionDialogFragmentTest.kt
```diff
@@ -32,33 +30,31 @@ import androidx.test.espresso.matcher.ViewMatchers.withText
...
     @Rule
     @JvmField
     val activityRule = ActivityTestRule(TestShareFileActivity::class.java, true, true)
 
-    private val account = mock(Account::class.java)
-    private val capabilitiesLiveData = MutableLiveData<Resource<OCCapability>>()
-    private val sharesLiveData = MutableLiveData<Resource<List<OCShare>>>()
     private val file = mock(OCFile::class.java)
     private val publicShare = mock(OCShare::class.java)
+    private val expirationDate = 1556575200000
```
jcmontesmartos: When is it?

owncloudApp/src/main/java/com/owncloud/android/shares/datasource/OCRemoteSharesDataSource.kt
```diff
@@ -75,4 +76,11 @@ class OCRemoteSharesDataSource(
         updateRemoteShareOperation.retrieveShareDetails = true
         return updateRemoteShareOperation.execute(client)
     }
+
+    override fun deleteShare(
+        remoteId: Long,
```
jcmontesmartos: Do you use remoteId anywhere?

jcmontesmartos: Ok, I've seen it below

owncloudApp/src/main/java/com/owncloud/android/shares/ui/ShareActivity.kt
```diff
@@ -144,6 +178,119 @@ class ShareActivity : FileActivity(), ShareFragmentListener {
         getSharesForFileAsyncTask = null
     }
 
+    /**
+     * Updates the view, reading data from [com.owncloud.android.shares.viewmodel.OCShareViewModel]
+     */
+    private fun refreshSharesFromStorageManager() {
+        val shareFileFragment = shareFileFragment
+        if (shareFileFragment?.isAdded == true) {   // only if added to the view hierarchy!!
+            shareFileFragment.refreshUsersOrGroupsListFromDB()
```
jcmontesmartos: it would be better to have a livedata listening data from db. You will save to do this approach ;-)

owncloudApp/src/main/java/com/owncloud/android/shares/ui/fragment/ShareFileFragment.kt
```diff
@@ -425,76 +383,40 @@ class ShareFileFragment : Fragment(), ShareUserListAdapter.ShareUserAdapterListe
...
+     * Show or hide button for adding a new public share depending on the capabilities and the server version
+     */
+    private fun updatePublicLinkButton() {
+        // Since capabilities and publicLinks are loaded asynchronously, let's check whether they both exist
+        if (capabilities == null || publicLinks == null) {
+            return
+        }
+
+        if (!enableMultiplePublicSharing()) {
+            if (publicLinks?.size == 0) {
```
jcmontesmartos: why don't you use isNullOrEmpty() ?

jcmontesmartos: and the else block wouldn't have condition

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.owncloud.android.shares.db.OCShare
+import com.owncloud.android.utils.PreferenceUtils
+import kotlinx.android.synthetic.main.share_public_link_item.view.*
+import java.util.ArrayList
+
+/**
+ * Adapter to show a list of public links
+ */
+class SharePublicLinkListAdapter(
+    private val mContext: Context, resource: Int, private val mPublicLinks: ArrayList<OCShare>?,
```
jcmontesmartos: Why don't you put each param in a new line? Something like this:

```
class SharePublicLinkListAdapter(
    private val mContext: Context, 
    resource: Int, 
    private val mPublicLinks: ArrayList<OCShare>?,
    private val mListener: SharePublicLinkAdapterListener
)
```

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val mListener: SharePublicLinkAdapterListener
+) : ArrayAdapter<OCShare>(mContext, resource) {
+
+    override fun getCount(): Int {
+        return mPublicLinks!!.size
```
jcmontesmartos: it can be null!!!

jcmontesmartos: override fun getCount(): Int = mPublicLinks?.size ?: 0

better?

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+        val view = inflator.inflate(R.layout.share_public_link_item, parent, false)
+
+        // Allow or disallow touches with other visible windows
+        view.filterTouchesWhenObscured = PreferenceUtils.shouldDisallowTouchesWithOtherVisibleWindows(mContext)
+
+        if (mPublicLinks != null && mPublicLinks.size > position) {
+            val share = mPublicLinks[position]
+
+            // If there's no name, set the token as name
+            view.publicLinkName.text = if (share.name?.isEmpty()!!) share.token else share.name
```
jcmontesmartos: if name can be null, you can't set isEmpty as not null. That doesn't make any sense.

I would use isNullOrEmpty function instead

owncloudApp/src/androidTest/java/com/owncloud/android/shares/db/OCShareDaoTest.kt
davigonz: Done

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/PublicShareEditionDialogFragmentTest.kt
davigonz: I put a comment with the date, but do you think is really important?

owncloudApp/src/main/java/com/owncloud/android/shares/ui/ShareActivity.kt
davigonz: This is for private shares and has not been touched yet, it will be our next challenge in the architecture.

owncloudApp/src/main/java/com/owncloud/android/shares/ui/fragment/ShareFileFragment.kt
davigonz: Done

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
davigonz: Done

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
davigonz: Yes, done

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
davigonz: You are right, it does not make any sense and Android Studio keeps doing it when converting java files to kotlin 😄 , I will use `if (share.name?.isEmpty() == true)` instead

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/PublicShareEditionDialogFragmentTest.kt
jcmontesmartos: I don't know if it will some years later or some years ago.... I don't know... 
The problem is if the date is a coming date, it will fail tests just because we are in that date or it's past... 
I would do it that date dynamically in order to avoid that issue. For example:
`System. currentTimeMillis() + 2000000`

owncloudApp/src/main/java/com/owncloud/android/shares/ui/ShareActivity.kt
jcmontesmartos: ok

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/SharePublicLinkListAdapter.kt
jcmontesmartos: and why don't you use `if (share.name.isNullOrEmpty())` better?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/PublicShareEditionDialogFragmentTest.kt
davigonz: In this test we are setting a date in a view and checking if it appears, so I think is not affected by a coming or past date

jcmontesmartos: ok

davigonz:

jesmrec:
Approved

PR #2474, opened 2019-03-08
Batched snackbar updates added

sd1998:
Fix to issue #2471.

PR #2564, opened 2019-06-03
Fix batched permission errors

### LINKED ISSUE #2471: Batched permission errors when deleting if you are not granted
This is something like: https://github.com/owncloud/android/issues/2335

Here the steps are:

1. Share a folder with another user. Do not grant to delete
2. Sharee opens the folder  in the account and try to remove several files by multiselecting

Current: an snackbar appears from each file that is not allowed to be removed
Expected: only one snackbar 

### PRs
App: https://github.com/owncloud/android/pull/2564

davigonz:
Fix #2471 , code from https://github.com/owncloud/android/pull/2474


davigonz:
@jesmrec ready to code review

Issue #2565, opened 2019-06-03
[New architecture] Improve viewmodel behaviour


davigonz:
There's some scenarios in which we are sharing the livedata between different operations (getShares, createShares, removeShares()) , triggering some sort of duplicated calls when there's a successful operation for example. The app works fine so far but I think this approach may be improved.

The basic implementation of a viewmodel is based on a livedata that is explicitly used to get the data, not to edit or even delete it, since for those cases the changes should be automatically propagated to that livedata. 

Why are we using livedata for more cases apart from getting shares? Because we need to propagate network errors to UI for instance and that's not a basic case.

It would be great to work on splitting up the class responsible for warning the active observers into different ones so that creation and edition/deletion/other operations can be easily distinguishable and do not interfere between them.

- [x] Duplicate NetworkBoundResources: separate getShares and other operations
- [x] Check tests
- [ ] QA will be done in https://github.com/owncloud/android/pull/2558

### PR
App: https://github.com/owncloud/android/pull/2569

PR #2558, opened 2019-05-28
[New architecture] Whole flow tests for sharing

davigonz:
Implements https://github.com/owncloud/android/issues/2531

PR #2569, opened 2019-06-10
[New architecture] Improve viewmodel behaviour

davigonz:
Implements #2565


owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/usecases/CreatePublicShareTest.kt
```diff
@@ -194,32 +196,56 @@ class CreatePublicShareTest {
...
+            Resource.success(
+                publicShares
+            )
+        )
 
         // Check whether the dialog to create the public share has been properly closed
         onView(withText(R.string.share_via_link_create_title)).check(doesNotExist())
         onView(withText(newPublicShare.name)).check(matches(isDisplayed()))
+
+        unregisterCreateShareFragmentAsIdlingResource(createShareFragmentIdlingResource)
```
jcmontesmartos: why don't you include this statement in a "@After" method??

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/usecases/CreatePublicShareTest.kt
```diff
@@ -295,19 +357,29 @@ class CreatePublicShareTest {
...
-        sharesLiveData.postValue(
-            Resource.success(
-                sharesAfterCreation
-            )
+    private fun registerCreateShareFragmentAsIdlingResource(): FragmentVisibilityIdlingResource {
+        val idlingResource = FragmentVisibilityIdlingResource(
+            activityRule.activity.supportFragmentManager.findFragmentByTag(ShareActivity.TAG_PUBLIC_SHARE_DIALOG_FRAGMENT),
+            View.VISIBLE
         )
+        IdlingPolicies.setIdlingResourceTimeout(1, TimeUnit.HOURS);
```
jcmontesmartos: 1 hour???? really???

owncloudApp/src/main/java/com/owncloud/android/shares/viewmodel/OCShareViewModel.kt
```diff
@@ -74,11 +78,11 @@ class OCShareViewModel(
         expirationDateInMillis: Long,
         permissions: Int,
         publicUpload: Boolean
-    ): LiveData<Resource<List<OCShare>>> = shareRepository.updatePublicShareForFile(
+    ): LiveData<Resource<Void>> = shareRepository.updatePublicShareForFile(
```
jcmontesmartos: Why do you use Void? Shouldn't it be better use Unit?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/usecases/CreatePublicShareTest.kt
davigonz: Ups, I forgot to change this

owncloudApp/src/androidTest/java/com/owncloud/android/shares/ui/usecases/CreatePublicShareTest.kt
davigonz: Because the idlingresource to unregister is created for each test and not shared between different ones.

jcmontesmartos: "@before" and "@after" are functions are called before and after running one single test.

My idea would be to have your tests as following:

```
private lateinit var idlingResource: <Your type>

@Before
fun setUp() {
    idlingResource = // call constructor
    //register ...
}

@After  
fun tearDown() {
    // unregister
}
```

davigonz: Yes, I know what you mean, I tried that way but the tests were not working fine. Let me recheck it...

davigonz: > @Before
fun setUp() {
    idlingResource = // call constructor
    //register ...
}

I'm not using this because I register the idling resources in the middle of the tests, after executing other lines of code.

If you see the [Expresso idling resources docs](https://developer.android.com/training/testing/espresso/idling-resource#create-own) , there's two different ways to implement it (see below), I decided to use the second one and that's why I do not use `@After` method either.

Screenshot 2019-06-11 at 14 27 11

owncloudApp/src/main/java/com/owncloud/android/shares/viewmodel/OCShareViewModel.kt
davigonz: Done

davigonz:
Since we are merging this branch into `new_arch/flow_sharing_tests`, QA will be directly passed within https://github.com/owncloud/android/pull/2558

Issue #2515, opened 2019-04-09
[Release] 2.11


jesmrec:
### TASKS

 - [x] [COORDINATION] Ping community guys to publish a blog post, at least one week before releasing.
 - [x] [GIT] Create branch release_2.11.0 in owncloud/android from master
 - [x] [GIT] Create branch release_1.0.1 in owncloud/android-library from master 
 - [x] [DEV] Update versionCode and versionName in build.gradle in android module
 - [x] [DEV] Update versionCode and versionName in build.gradle in android library module
 - [x] [DEV] Disable feature to open github from the compilation info (clicks will be used to display logs section)
 - [x] [DIS] Create changelog file (< 500 chars) and add to CHANGELOG.md in owncloud/android 
 - [x] [QA] Design Test plan @jesmrec  https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.11.0/Regression.md
 - [x] [QA] Regression Test plan  @jesmrec 
 - [x] [DIS] Generate test APKs files from release branch in owncloud/android and perform some basic operations.
 - [x] [GIT] Create and sign tag `oc-android-2.11.0` in HEAD commit of release branch, in owncloud/android 
 - [x] [GIT] Create and sign tag `1.0.1` in HEAD commit of release branch, in  #owncloud/android-library 
 - [x] [DIS] Generate final APK files from signed commit in owncloud/android and perform some basic operations
- [x] [DOC] Include new changelog on Releases section of github android repo by drafting the new release 2.11.0
 - [x] [GIT] Merge branch release_1.0.1 in owncloud/android-library, into master
 - [x] [GIT] Merge branch release_2.11.0 in owncloud/android, into master 
 - [x] [DIS] Upload & publish release APK and changelog in Play Store
 - [x] [DIS] Upload release APK internal storage
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Storel 
 - [x] [SPREADTHEWORD] Write a topic in central about the release => https://central.owncloud.org/t/owncloud-2-11-for-android-released/
 - [x] [COORDINATION] Write in corresponding channels
 - [x] [COORDINATION] Write email to release-coordination mailing list.
 - [ ] [SPREADTHEWORD] Write a post in the blog @lefherz 
 - [X] [DOC] Create an issue in docs repository with the new features to be included in the user manual: https://github.com/owncloud/docs/issues/1323 
 - [x] Publish app in F-Droid [AUTOMATIC]
[...]

PR #2571, opened 2019-06-11
[Release] 2.11

davigonz:
Implements #2515 
It needs https://github.com/owncloud/android-library/pull/252


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -597,5 +597,9 @@
     <string name="welcome_feature_7_text">Select all files at once</string>
     <string name="welcome_feature_8_title">Available offline</string>
     <string name="welcome_feature_8_text">Your files automatically synchronized</string>
+    <string name="welcome_feature_9_title">Document provider</string>
+    <string name="welcome_feature_9_text">Access and edit your ownCloud files from other apps</string>
```
jesmrec: should be the `ownCloud` word brandable?

davigonz: Since the wizard is not brandable yet, I would not address this change here.

PR #2593, opened 2019-06-26
[Espresso] Tests over security section on "Settings"

jesmrec:
UI tests to cover the section "Security" of "Settings"

PR also includes a small improvement in "Camera Uploads" section, where the preferences are cleared after every test. In this way, we get rid of artificial methods to reset the status of the view after every execution.


jesmrec:
messed up after new_arch/shares rebase

PR #2601, opened 2019-07-05
Replace the Documentation Makefile with a Shell Script

settermjd:
This is a port of the cli script from the docs repository, streamlined for a single-component Antora setup. It also supports exporting the manual to docx format as well.

The manual generation can be tested with the following commands:

```console
cd docs
bin/cli -m && bin/cli -d
```


settermjd:
The request has come through a few times and it made sense to me to add extra functionality here to avoid having to remember it for next time.

michaelstingl:
@settermjd what are the recommended steps to test-drive this? `yarn antora` and `yarn serve` aren't enough?

tboerger:
@pmaier1 please clarify why you need this. Changes to a .docx document will never be put back into the repository. Why do we need that? I don't think that we want to maintain another container which includes the required tools just for fun or if there is no need for it.

settermjd:
@michaelstingl, I've updated the PR description with the relevant information. @tboerger, the docx generation command would only be used locally.

settermjd:
@pmaier1, I need your feedback to @tboerger's feedback.

pmaier1:
> I don't get why we want to generate a .docx format at all.

We needed the export once currently for some people to develop their own documentation based on ours. I don't know about further requirements or similar. At this point I don't find it necessary to automate this. I would rather want to check with those people whether they can also use Antora to create their own/branded docs.

knugz38277:
> cd docs
> bin/cli -m && bin/cli -d

settermjd:
The functionality for generating docx versions of the manual have been removed.

settermjd:
@pmaier1 / @tboerger, can this now be merged?

docs/books/ownCloud_Android_App_Manual.adoc
```diff
@@ -8,5 +8,4 @@ The ownCloud Team <docs@owncloud.com>
 :toclevels: 1
 :icons: font
 :icon-set: octicon
-
-include::modules/ROOT/pages/index.adoc[leveloffset=+1]
+:module_base_path: modules/ROOT/pages/
```
tboerger: That should not be required?

settermjd: Is that a question or a statement?

tboerger: The modules path is defined on the cli, so this should be dropped.

settermjd: The cli module path definition is for Antora. This line is used by asciidoctor-pdf. However, it can be removed and set, instead, with an attribute in bin/cli. I'll make that change in the next commit.

tboerger: https://github.com/owncloud/android/pull/2601/files#diff-d687c1392f67d7207b33141ff4979ef5R150-R153 these paths ARE applied to asciidoctor-pdf. There is nothing about Antora at all.

settermjd:
@tboerger, can I get another review, pls?

Issue #2556, opened 2019-05-28
[New arch] Edit and delete private shares


davigonz:
Delete and edit private shares

- [x] Show permissions and edit
- [x] Delete private share
- [x] Unit/UI Tests https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.12.0/new_architecture_edit_shares.md

### PRs
- App: https://github.com/owncloud/android/pull/2603

davigonz:
I've taken advantage of the new implementation to include the error when editing private share permissions in the dialog itself, before was being shown in a snackbar contained in the background activity and it didn't look that good. Besides, now can be included in a UI test.

| Error with previous implementation | Error with new arch |
| ------------- | ------------- |
| ![device-2019-07-09-183420](https://user-images.githubusercontent.com/15261363/60906793-4fcb2780-a278-11e9-8890-c01d72931398.png) |  ![device-2019-07-09-182402](https://user-images.githubusercontent.com/15261363/60906819-5eb1da00-a278-11e9-8b70-90650e75ceee.png) | 

CC / @jesmrec @michaelstingl

Issue #2523, opened 2019-04-24
[New arch] App modularization and decoupling


davigonz:
The status of the app and its code is a recurrent topic that's been discussed several times. A couple of months ago we started to work on a [new architecture](https://github.com/owncloud/android/issues/2351) that is transforming the app to be more testable, robust and future proof. 

Thanks to this new architecture, we should manage to have an app with much less dependencies but sometimes it's not easy to know and control, even less in an app as huge as ownCloud for Android.

So let's go through several solutions to modularize and decouple the app

### **Modularization with layers approach**

- Diagram
![2 3  oC app modules diagram_resized](https://user-images.githubusercontent.com/15261363/61300732-d33dc900-a7e2-11e9-8ecc-b956219d59ad.png)

- Advantages:
  - Scalability: several developers working independently in different modules.
  - Maintainability: gradle can build modules faster since the classes and resources are separated in different modules. CI builds will be accelerated too.
  - Less tests to run when there's a new change: we need to run the tests in the module affected by that change and anything else that depends on it. We do not need to run all the tests if other modules are not impacted by the changes.
  - Reduce dependencies => less coupled code.
  - SWITCH MODULES: if in the future we want an app with a different UI and a different way to handle the information to show, we would just need to replace the :ownCloudApp module. And the same for the rest of modules, if one day we want to handle data differently to the current implementation, replacing the :ownCloudData module should be enough.

- Video: https://youtu.be/PZBg5DIzNww

-----------------------------------------------

### OTHER TOOLS

I've found an open source tool called [Apk dependency graph](https://github.com/alexzaitsev/apk-dependency-graph) that could help us a lot on this matter.

This is how the **dependency graph for the ownCloud Android app** looks like, based on [New arch - capabilities branch](https://github.com/owncloud/android/pull/2508) (I put it live in https://davigonz.github.io/oCAndroidDependencyGraph/ so everyone can play with it) :

![Screenshot 2019-04-24 at 18 31 14](https://user-images.githubusercontent.com/15261363/56676690-29590f80-66bf-11e9-9a45-32b18ce15a3a.png)

Above we can notice some parts of new architecture (OCShareRepository, OCShareViewModel, OCCapabilityRepository, OCCapabilityViewModel...) along with old code.

And below there's an example of a **good architecture with low class coupling**:

[...]

PR #2603, opened 2019-07-08
[New arch] Private shares: edition and deletion

davigonz:
Implements #2556 

## Bugs and improvements

- [x] (1) Loading dialog stuck when trying to create an already existing private share [FIXED]


owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCLocalShareDataSourceTest.kt
```diff
@@ -236,8 +261,12 @@ class OCLocalDataSourceTest {
         assertEquals(8, updatedShareId)
     }
 
+    /**************************************************************************************************************
+     *************************************************** COMMON ***************************************************
```
davivel: Common to what?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCShareDaoTest.kt
```diff
@@ -284,15 +284,51 @@ class OCShareDaoTest {
...
+    fun updatePrivateShare() {
+        ocShareDao.insert(createDefaultPrivateShare())
+
+        ocShareDao.update(
+            createDefaultPrivateShare(permissions = 17)
+        )
+
+        val textShares = getValue(
+            ocShareDao.getSharesAsLiveData(
+                "/Texts/text1.txt", "admin@server", listOf(ShareType.USER.value)
```
davivel: This is a bit difficult to read. How should I know that "/Texts/text1.txt", "admin@server", etc... correspond to the fields of the share returned by 'createDefaultPrivasteShare()' ?

I would save the value of createDefaultPrivateShare() in a variable at the beginning, then use the fields of that variable here to pass file, user, etc... That way it's clear that we are getting the same share that we inserted first.

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCShareDaoTest.kt
```diff
@@ -284,15 +284,51 @@ class OCShareDaoTest {
...
+        assertThat(textShares, notNullValue())
+        assertEquals(1, textShares.size)
+        assertEquals(17, textShares[0].permissions)
+    }
+
+    @Test
+    fun deletePrivateShare() {
+        ocShareDao.insert(createDefaultPrivateShare())
+
+        ocShareDao.deleteShare(1)
```
davivel: Similar here, how do we know that the Share created by 'createDefaultPrivarteShare' is the one with id '1'?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/domain/privateShares/DeletePrivateShareTest.kt
```diff
@@ -0,0 +1,292 @@
...
+
+        `when`(intent.getParcelableExtra(FileActivity.EXTRA_FILE) as? Parcelable).thenReturn(file)
+        intent.putExtra(FileActivity.EXTRA_FILE, file)
+
+        `when`(ocCapabilityViewModel.getCapabilityForAccount(false)).thenReturn(capabilitiesLiveData)
+        `when`(ocCapabilityViewModel.getCapabilityForAccount(true)).thenReturn(capabilitiesLiveData)
+        `when`(ocShareViewModel.getPrivateShares(file.remotePath)).thenReturn(privateSharesLiveData)
+        `when`(ocShareViewModel.getPublicShares(file.remotePath)).thenReturn(MutableLiveData())
+
+        stopKoin()
```
davivel: I would create a method with annotation @After to put 'stopKoin()' there instead.

owncloudApp/src/androidTest/java/com/owncloud/android/shares/domain/privateShares/DeletePrivateShareTest.kt
```diff
@@ -0,0 +1,292 @@
...
+        )
+
+        onView(
+            allOf(
+                withId(R.id.unshareButton),
+                hasSibling(withText(privateShareList[0].sharedWithDisplayName))
+            )
+        ).perform(click())
+
+        privateSharesLiveData.postValue(
```
davivel: Sorry, I don't understand what is this for.

owncloudApp/src/main/java/com/owncloud/android/shares/domain/OCShareRepository.kt
```diff
@@ -108,16 +100,92 @@ class OCShareRepository(
...
+        shareTypes: List<ShareType>
+    ): MutableLiveData<Resource<List<OCShare>>> {
+        return object : NetworkBoundResource<List<OCShare>, ShareParserResult>(appExecutors) {
+            override fun saveCallResult(item: ShareParserResult) {
+                val sharesFromServer = item.shares.map { remoteShare ->
+                    OCShare.fromRemoteShare(remoteShare).also { it.accountOwner = accountName }
+                }
+
+                if (sharesFromServer.isEmpty()) {
+                    localShareDataSource.deleteSharesForFile(filePath, accountName)
```
davivel: Is this needed? Shouldn't be done inside 'replaceShares(...)' right below?

owncloudApp/src/main/java/com/owncloud/android/shares/presentation/ShareActivity.kt
```diff
@@ -114,7 +112,7 @@ class ShareActivity : FileActivity(), ShareFragmentListener {
         setContentView(R.layout.share_activity)
 
         // Set back button
-        supportActionBar!!.setDisplayHomeAsUpEnabled(true)
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
```
davivel: What happens if supportActionBar is null?

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCLocalShareDataSourceTest.kt
davigonz: Common to private and public shares

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCShareDaoTest.kt
davigonz: Done

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCShareDaoTest.kt
davigonz: Done

owncloudApp/src/androidTest/java/com/owncloud/android/shares/domain/privateShares/DeletePrivateShareTest.kt
davigonz: I can create it to stop Koin for the last executed test but this line would be still needed because since instrumented tests are launched on a device, the android application class is instantiated and it causes Koin initialization as a side effect Therefore, we need to stop it here before starting koin again, otherwise I get the exception below:


`org.koin.core.error.KoinAppAlreadyStartedException: A Koin Application has already been started`

owncloudApp/src/androidTest/java/com/owncloud/android/shares/domain/privateShares/DeletePrivateShareTest.kt
davigonz: We are mocking the viewmodel so the operation to delete the share from localShareDataSource, which takes place in the repository, is never executed. Therefore, privateSharesLiveData is not updated after the deletion so we have to do it here manually.

I know this makes these tests to be a little "fake", so I guess we would need to use mocks in layers below, i.e. datasources, to have an almost completed flow, I will face this in https://github.com/owncloud/android/issues/2523 to not change the tests twice

owncloudApp/src/main/java/com/owncloud/android/shares/domain/OCShareRepository.kt
davigonz: The `replaceShares` method in `OCShareDao.kt` includes the code below and its aim is replacing the still existing shares, not deleting the shares that no longer exist in the server.

``` kotlin
@Transaction
open fun replaceShares(ocShares: List<OCShare>): List<Long> {
    for (ocShare in ocShares) {
        deleteSharesForFile(ocShare.path, ocShare.accountOwner)
    }
    return insert(ocShares)
}
```

owncloudApp/src/main/java/com/owncloud/android/shares/presentation/ShareActivity.kt
davigonz: The arrow to go back would not appear in the action bar but you could still use the back button in the Android navigation bar.

owncloudApp/src/androidTest/java/com/owncloud/android/shares/data/OCShareDaoTest.kt
davigonz: Done

owncloudApp/src/androidTest/java/com/owncloud/android/shares/domain/privateShares/DeletePrivateShareTest.kt
davivel: You can create a full Koin tree to be used only in tests and load it in the @Begin methods with ` StandAloneContext.loadKoinModules` so that if overrides the full app module(s), then keep the @Begin/@After symmetry.

PR #2613, opened 2019-07-31
Fix trigger media scan

### LINKED ISSUE #2612: [BUG] Images are not detected in Android 9 gallery after being downloaded
### Actual behaviour
Picture does not appear in Gallery app. After rebooting the device, the picture appear.

### Expected behaviour
Picture appears in Gallery app
 
### Steps to reproduce
1. Download an image from the oC Android app.
2. Open Gallery app.

Android 9
Samsung S8/S8+/S9/S9+/S10/S10+

Version 2.11.1
Server 10.2.1

https://github.com/owncloud/enterprise/issues/3449

### PRs

https://github.com/owncloud/android/pull/2613

davigonz:
Fixes #2612 

QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.12.0/2613%20-%20trigger%20media%20scan.md

Bugs & improvements:

- [x] (1) mov files [NOT FIXEABLE]
- [x] (2) wrapped app images preview https://github.com/owncloud/android/pull/2613#issuecomment-517245571 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -1713,18 +1713,23 @@ public class FileDataStorageManager {
     public void triggerMediaScan(String path) {
         if (path != null) {
             Intent intent = new Intent(Intent.ACTION_MEDIA_SCANNER_SCAN_FILE);
-            try {
-                intent.setData(
-                        FileProvider.getUriForFile(
-                                mContext.getApplicationContext(),
-                                mContext.getResources().getString(R.string.file_provider_authority),
-                                new File(path)
-                        )
-                );
-            } catch (IllegalArgumentException illegalArgumentException) {
-                intent.setData(Uri.fromFile(new File(path)));
-            }
+            intent.setData(Uri.fromFile(new File(path)));
             MainApp.getAppContext().sendBroadcast(intent);
+
+            // For making uploads and downloads work in wrapped apps
+            Intent wrappedAppintent = new Intent(Intent.ACTION_MEDIA_SCANNER_SCAN_FILE);
+            try {
+                Uri uri = FileProvider.getUriForFile(
+                        mContext.getApplicationContext(),
+                        mContext.getResources().getString(R.string.file_provider_authority),
+                        new File(path)
+                );
+                wrappedAppintent.setData(uri);
+            } catch (IllegalArgumentException illegalArgumentException) {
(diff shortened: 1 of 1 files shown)


davigonz:
@jesmrec media scan triggering stopped working due to some changes we performed to make uploads and downloads with wrapped apps properly worked so could you test them to see that are still working?

davigonz:

jesmrec:
### (1)

I think this is not a problem in the oC app, but just in case: .mov files are not displayed in the Gallery (after checking different formats). Am i wrong or can we do something? @davigonz

jesmrec:
### (2)

> @jesmrec media scan triggering stopped working due to some changes we performed to make uploads and downloads with wrapped apps properly worked so could you test them to see that are still working?

about this. I compared the behaviour in version 2.11.0 and the current fix. With the fix, the pics are not displayed after being downloaded. Unfortunately, the behaviour is not the same (due to the fix?)

MI wrapper 8.5 Gen1

With Android 2.11.0, pics downloaded and displayed correctly:



With the fix, download error

2 11 1-fail

Tested with Android 5X v8 and Samsung S9 v9

I can provide as many wrapped apks as needed.

davigonz:
> I think this is not a problem in the oC app, but just in case: .mov files are not displayed in the Gallery (after checking different formats). Am i wrong or can we do something? @davigonz

.mov is a format created by Apple so I guess Android does not support it by default. 

Some users reported the same problem on S9 devices => https://www.reddit.com/r/GalaxyS9/comments/9t9dif/mov_files_not_showing_in_my_gallery/

jesmrec:
(2) is fixed. But here we have gone to the scratch point. Now, pics are correctly downloaded and previewed in both wrapped and non-wrapped apps. But the Samsung gallery only shows the pics when the app is non-wrapped. That means, the initial bug persists if the app is MI-wrapped. Any other workaround is posible here @davigonz?

CC @michaelstingl

davigonz:
> But here we have gone to the scratch point. Now, pics are correctly downloaded and previewed in both wrapped and non-wrapped apps. But the Samsung gallery only shows the pics when the app is non-wrapped. That means, the initial bug persists if the app is MI-wrapped. Any other workaround is posible here @davigonz?

This is how the wrapped apps work, files contained in the wrapped app container should not be accessible from other apps (including the gallery), but only from other wrapped apps. Otherwise, the wrapped apps would not be secure since they would be exposing their data to any app.

jesmrec:
Ok, then, everything fixed. Approved on my side.

PR #2649, opened 2019-08-29
Update Antora/Documentation dependencies to the latest versions

settermjd:
There are a number of vulnerabilities with the current versions of some of the dependencies. This change helps mitigate these problems. Thanks for reporting this in the documentation channel, @davigonz.


davigonz:
@settermjd code approved

Issue #2663, opened 2019-09-24
Changes in authentication flow


jesmrec:
Currently, when authentication is revoked (token removal etc..), app shows an error and give the user the choice to move to list of accounts

AC:

- Error message: more accurate message. 
- Redirection to login view instead of accounts view
- Shortcut to accounts view in login view. Check if it is feasible (optional) 

Related: https://github.com/owncloud/enterprise/issues/3451

### PRs: 
- App: https://github.com/owncloud/android/pull/2670

PR #2670, opened 2019-09-30
Improve oAuth flow UX and wording when token expires or becomes invalid

davigonz:
Implements #2663


davigonz:
@abelgardep ready to code review

jesmrec:
Checks performed:

- [X] Revoke token -> Correct snackbar and action
- [X] Session re-authorization
- [X] Token renewal (regression)

Approved

Issue #2453, opened 2019-02-18
[Release] 2.10.0


jesmrec:
### TASKS
 - [x] [COORDINATION] Ping community guys to publish a blog post, at least one week before releasing.
 - [x] [GIT] Create branch release_2.10.0 in owncloud/android from master
 - [x] [GIT] Create branch release_1.0.0 in owncloud/android-library from master 
 - [x] [DEV] Update min version to 19 + remove Jelly Bean workaround
 - [x] [DEV] Update version number and name in build.gradle in android module
 - [x] [DEV] Update version number and name in build.gradle in oc_jb_workaround module 
 - [x] [DIS] Create changelog file (< 500 chars) and add to CHANGELOG.md in owncloud/android 
 - [x] [QA] Design Test plan @jesmrec
 - [X] [QA] Regression Test plan  @jesmrec
- [x] [DOC] Include new changelog on Releases section of github android repo by drafting the new release 2.10.0 
 - [x] [GIT] Create and sign tag oc-android-2.10.0 in HEAD commit of release branch, in owncloud/android 
 - [x] [GIT] Create and sign tag oc-android-library-1.0.0 in HEAD commit of release branch, in  #owncloud/android-library 
 - [x] [GIT] Merge branch release_1.0.0 in owncloud/android-library, into master
 - [x] [GIT] Merge branch release_2.10.0 in owncloud/android, into master 
 - [x] [DIS] Generate test APKs files from branch master in owncloud/android and perform some basic operations.
 - [x] [DIS] Generate final APK files from signed commit in owncloud/android
 - [x] [DIS] Upload & publish release APK and changelog in Play Store
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [DIS] Remove OC JB workaround app from Play Store.
 - [X] [FEEDB] Creation of new specific mail addresses for feedback in  otrs mail @michaelstingl 
 - [x] [SPREADTHEWORD] Write a topic in central about the release. https://central.owncloud.org/t/owncloud-2-10-for-android-released/18900
 - [x] [COORDINATION] Write in corresponding channels.
 - [x] [COORDINATION] Write email to release-coordination mailing list.
 - [x] [SPREADTHEWORD] Write a post in the blog @lefherz 
 - [x] [DOC] Update documentation with the new features. @settermjd 
 - [x] Publish app in F-Droid [AUTOMATIC]
 - [x] [GIT] merge master branch into stable, in owncloud/android-library
[...]

PR #2679, opened 2019-10-04
[Camera Uploads] Sync/Upload for older file, which waits for WiFi too

hannesa2:
During sync, only newer files are handled, but files which were not uploaded because of missing and required WiFi are postponed currently **for ever**.

My mobile becomes wet and I missed a lot of pictures from the last year, because of this bug. 👎  
I hope this bugfix is working like expected, please give it a try. At least my tests were positive

### BUGS & IMPROVEMENTS:

- [x] (1) Double notification when uploading a file: https://github.com/owncloud/android/pull/2679#issuecomment-550198312 [FIXED WITH LATEST CHANGES]
- [x] (2) Does not retry if wifi is recovered with the app closed https://github.com/owncloud/android/pull/2679#issuecomment-554320203


hannesa2:
Btw, I've seen sometime a lot of 

> W/BroadcastQueue: Background execution not allowed: receiving Intent ... ConnectivityActionReceiver

In my point of view this complete class can be removed because with Android O implicit Intents are no more working. In my tests I never run into this `onReceive()` and on real device I see no entry in log

davigonz:
Hi @hannesa2 , thanks for this contribution, we will have a look at it.

> with Android O implicit Intents are no more working

Are you sure? [Implicit intents](https://developer.android.com/guide/components/intents-filters#ExampleSend) are widely used to invoke other apps on the device able to perform actions when our app is not able to perform them. Anyway, we will recheck what ConnectivityActionReceiver is currently doing.

hannesa2:
> Anyway, we will recheck what `ConnectivityActionReceiver` is currently doing.

Switch off/on Wifi and look for `ConnectivityActionReceiver` in log
I at least got no log entries with this name on my device and Emulator

pequesan:
Sorry. Is there any Play Store beta program I can join to test this fix?

hannesa2:
I'm not sure if owncloud has something like this. At least I see no branch/tag upload mechanism doing it. That's why I made my own "pre-deployment" with a deterministic build. Owncloud has no such deterministic build. 

If you are interested here it is https://github.com/hannesa2/owncloud-android/releases

pequesan:
Thanks for pointing me to the apk. I've downloaded and installed it (I had to uninstall play store apk first), but I can't pass login page... After introducing my owncloud url and credentials, app crashes and any further try to open it crashes directly.

Any idea?

hannesa2:
@pequesan https://github.com/hannesa2/owncloud-android/issues/3

jesmrec:
> Sorry. Is there any Play Store beta program I can join to test this fix?

ownCloud Android app is suscribed to the beta program of App Store. Right now, we do not have an active beta (latest release is an official one). If you are interested in beta testing, please suscribe the beta program of Google Play, and you will be notified in he moment we release a new one.

jesmrec:
I started to handle this, and some questions about the problem come to my head:

- You reproduce the problem with camera uploads, am i wrong? since "only wifi" are available only for camera uploads 
- You lost pics that were taken under mobile connection, and were not uploaded even when you recovered the wifi. Is this correct? that means, the pics are queued in `Uploads` but not retried. 

Does your fix make the pending stuff to be retried? @hannesa2

hannesa2:
@jesmrec  all right !
I fixed a follow up crash with missing forground permission too (Android O)

pequesan:
I can confirm I'm using @hannesa2 release with no further issues. All my files are being uploaded, with only wifi checked.

Well, I must say something weird happens when owncloud is uploading a file and you lose wifi... but I assume it should be treated as a separate issue.

jesmrec:
ok, we will move it forward in terms of code review & QA (CC @abelgardep @davigonz)

> Well, I must say something weird happens when owncloud is uploading a file and you lose wifi... but I assume it should be treated as a separate issue.

yes, please. Create a new issue including as many details as posible. 

Thanks a lot!!

hannesa2:
@pequesan you should see the reason here https://github.com/owncloud/android/pull/2440#issuecomment-462150721

pequesan:
Sorry to use this issue to ask this but... Where should I add a feature request? I'm not familiar with Github, sorry.

hannesa2:
@pequesan What's about to simply create an issue ?

But to be realistic, who should jump immediate into this issue ? 

Assuming a potential developer does it, you are fine ! 
Assuming nobody does it, you have to develop it by your own. Here I can write about a very long time period till it will be merged.

owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -290,8 +289,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
                 KEY_REQUESTED_FROM_WIFI_BACK_EVENT, false
         );
 
-        if ((isCameraUploadFile || isAvailableOfflineFile || isRequestedFromWifiBackEvent) &&
-                Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
abelgardep: This check was included to fix a problem with notifications here: https://github.com/owncloud/android/issues/2453#issuecomment-466322335 . But if everything is working, we can move this forward @jesmrec

jesmrec: If we'd remove that line, the problem you mention will arise again, causing a regression. We must avoid this. CC @hannesa2

hannesa2: @abelgardep , @jesmrec This if block is just to

 `startForeground(141, mNotificationBuilder.build());`

nothing else. And because of this if condition it was crashing. This `startForeground()` needs nothing else 

`if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {`

davigonz: @hannesa2 

We were checking `isCameraUploadFile || isAvailableOfflineFile || isRequestedFromWifiBackEvent` conditions to execute `startForeground(141, mNotificationBuilder.build())` because it needs to be called within five seconds after [calling `startForegroundService` from TransferRequester](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/TransferRequester.java#L95) for those cases (cameraUploads, available offline and requestdFromWifiBackEvent), otherwise, the app will crash. 

But you're also right, we should call `startForeground()` after [calling startService() from TransferRequester](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/TransferRequester.java#L98) to make FileUploader service run in foreground. Otherwise, the app could fail if the system decides to kill FileUploader (it does not usually happen). Could you tell me in which scenario is the previous code crashing for you?

What really concerns me most is the duplicated notification that appears when trying your solution, and we had to fix in https://github.com/owncloud/android/issues/2453#issuecomment-466322335 . Besides, we are also able to show a notification even when calling `startService() ` in a normal upload. It seems that many changes have happened here in the Android code itself.

Could you have a look at the [problem with duplicated notifications](https://github.com/owncloud/android/issues/2453#issuecomment-466322335) that seems to happen after applying your changes? Thanks

pequesan:
Well, FYI I'm running @hannesa2 apk with this fix, and I would say that I don't have double notifications. But It's true that from Android 9, notification system has changed a lot. I'm using Android 10 in a Pixel 4.

hannesa2:
Like @pequesan already wrote, I never observed a double notification too.
What's about @jesmrec ?

@davigonz 
To much if-conditions for `startForeground()` is simply wrong. I run into **crashes** when I do it !
If you ever observed too much notifications (I didn't) you have to rework the notification-channel. This is out of this PR

Summarize:
You have to choose between a crash-free and working synchronization with Wifi-on/off  (this is a **primary** function) or not-oberserved double notification prevention.
And if there are double notification, the place to solve are notification-channels !

@davigonz 
https://github.com/owncloud/android/issues/2453#issuecomment-466322335 where is the fix ?  Please point me to the commit, not to the release notes.

davigonz:
> Too much if-conditions for startForeground() is simply wrong. I run into crashes when I do it !

Ok, you talked about those crashes before and I asked you the steps to reproduce, could you give me those steps? Thanks

>  not-oberserved double notification

I've just tried this PR code and reproduced it in a Google Pixel 2 with Android 10

Nov-06-2019 08-48-52

> If you ever observed too much notifications (I didn't) you have to rework the notification-channel.

The problem with notifications is something that appears with your solution (this PR), I tried it in master and there's no additional notification. As we usually do with the rest of contributions and even with our own PRs, if something is broken, it should be fixed by the developer who included the new code. Otherwise, other developers would have to finish the code developed by other people.

Anyway, I can help you if there's any blocker.

davigonz:
**[BUG]  Double notification when uploading a file**
1. Upload file
2. Swipe down the top of the screen to see the notifications bar, two notifications appear:
- Notification with upload progress
- Empty notification

Current: two notifications for uploads
Expected: show only one notification

Pixel 2 Android 10 
Server 10.0.10

pequesan:
Wow. I've just tested to upload a file with @hanessa2 apk, and I must admit there's those two notifications noted by @davigonz. As I'm using owncloud app mainly for camera uploads, I never take a look at notifications behaviour in detail. So I can confirm that bug.

jesmrec:
Since the double notification and the lost uploads are different features, we should keep the checks

` if ((isCameraUploadFile || isAvailableOfflineFile || isRequestedFromWifiBackEvent) &&...`

in order to avoid regressions, and testing separately both things: this PR for the camera upload stuff, and the other thing, wherever you want (new issue or PR). 

Agreed? @abelgardep @davigonz @hannesa2

hannesa2:
> Since the double notification and the lost uploads are different features, we should keep the checks

@jesmrec Sorry, it will crash, when we keep this if conditions

jesmrec:
@hannesa2 steps to reproduce the crash?

davigonz:
I've been diving in the uploads and camera uploads code and it seems there was a case missing in [FileUploader.java](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/FileUploader.java#L891) for scheduling automatic retries for uploads that should wait for wifi to be uploaded when the user selects Only wifi. We were only scheduling automatic retries for uploads that were already launched and failed because lack of wifi connection, which is not the same. 

I've just pushed a commit with that missing case so we don't need to retry all the failed uploads from `CameraUploadsSyncJobService.java` as @hannesa2 was doing before. Now is more specific, we just schedule the retry for the camera upload that was waiting for wifi. I also removed the `if` that was triggering the double notification.

The automatic retries of failed uploads are handled by [RetryUploadJobService.java](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/files/services/RetryUploadJobService.java) but only work for Lollipop devices or above, the same task is handled by `ConnectivityActionReceiver.java` for devices below Lollipop. That's why you do not see logs there @hannesa2 , your device will be >= Lollipop for sure. It would be great to reopen the discussion about what to do with < Lollipop support, since 4.4 was released in 2013.

@jesmrec can you please recheck this and see if the crashes that @hannesa2 says appear? I'm not able to reproduce any crash with the latest changes I pushed

hannesa2:
I see, I've to make a _complete_ report for each detected crash.

Reintroduce ` if ((isCameraUploadFile || isAvailableOfflineFile || isRequestedFromWifiBackEvent)`
make a picture without Wifi, after 10 minutes switch on Wifi. After additional 5 minutes it tries to sync. 

--> You will see probable a crash because of missing foreground-something

jesmrec:
thanks @hannesa2. This is the same when you go to the doctor... telling him "i feel bad" is not enough. He needs to know where does it hurt ;)

jesmrec:
QA.

Catching events of wifi connection to upload from the camera:

- [ ] Roaming from mobile connection
- [ ] Switching from no connection (app foreground)
- [ ] Switching from no connection (app closed)
- [ ] Upload with no roaming

jesmrec:
Look at these steps:

1. Mobile data plan (no wifi) connection (disable the wifi option in the device)
2. Camera uploads with "only wifi" enabled
3. Take some pictures -> Pictures queued in Uploads view, waiting for wifi
4. Kill the app
5. Recover wifi (enable it on device settings)
6. Open the app, now under wifi connection

Current: pending uploads are not retried till you open the uploads view and retry them manually.
Expected: retried since wifi is recovered

Huawei P20Lite
Android 9

jesmrec:
I also notice ANR events when camera uploads take some time (not when only one upload is done). This is also reproducible in 2.13.1, so i will open a new issue with this.

jesmrec:
About report (2): i noticed different behaviours in different devices, so it needs more info and research. Since this problem is not a new one (no regression), i will open in a new issue as soon as i gather more info about it.  

In the same path, i will open an issue with https://github.com/owncloud/android/pull/2679#issuecomment-553374019 if i reproduce it. 

But this PR fixes at least the retry problem under "only wifi". So, we can move forward.

pequesan:
Does this "merge" means that next official release will address this delayed wifi upload issue?

davigonz:
> Does this "merge" means that next official release will address this delayed wifi upload issue? Did you manage to fix it while not reintroducing the double notification glitch?

Yes to both questions

pequesan:
That's great! Thx

PR #2689, opened 2019-10-30
remove pointless comments

hannesa2:
especially where I'm the last changed user and git blame points to me.

It's just noise

### DIFF
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -34,17 +34,11 @@ dependencies {
     androidTestImplementation 'androidx.test.ext:junit:1.1.1'
     // Android JUnit Runner
     androidTestImplementation 'androidx.test:runner:1.2.0'
-    // Espresso core
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.2.0'
-    // Espresso contrib
     androidTestImplementation 'androidx.test.espresso:espresso-contrib:3.2.0'
-    // Espresso intents
     androidTestImplementation 'androidx.test.espresso:espresso-intents:3.2.0'
-    // Espresso web
     androidTestImplementation 'androidx.test.espresso:espresso-web:3.2.0'
-    // UIAutomator - for cross-app UI tests, and to grant screen is turned on in Espresso tests
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
-    // fix conflict in dependencies; see http://g.co/androidstudio/app-test-app-conflict for details
     androidTestImplementation 'androidx.annotation:annotation:1.1.0'
     androidTestImplementation "androidx.room:room-testing:$roomVersion"
     androidTestImplementation "androidx.arch.core:core-testing:$archLifecycleVersion"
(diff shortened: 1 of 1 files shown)


hannesa2:
@michaelstingl It looks like your tests are broken, even removing pointless comments fails.
How can this happen ?

jesmrec:
@hannesa2 sometimes, the BitRise fails due to some timeouts or performance problems in its internal emulators. We are looking for the best solutions.

abelgardep:
From my point of view it is important to have gradle files organized. We are working on modularization, so we will have a build gradle file for each module. It is important to have them localized and commented in some way. What do you think @davigonz ?

hannesa2:
> From my point of view it is important to have gradle files organized. 

You can organise them like you want, but
image
but this comment (and the others) has no purpose, except noise. And worst, `git blame` shows me, I'm the commiter

owncloudApp/build.gradle
```diff
@@ -22,62 +22,44 @@ dependencies {
...
-    // Architecture lifecycle components
     implementation "androidx.lifecycle:lifecycle-extensions:$archLifecycleVersion"
     implementation "androidx.lifecycle:lifecycle-viewmodel:$archLifecycleVersion"
     kapt "androidx.lifecycle:lifecycle-compiler:$archLifecycleVersion"
 
-    // Room
     implementation "androidx.room:room-runtime:$roomVersion"
     kapt "androidx.room:room-compiler:$roomVersion"
 
-    // Koin dependency injector
```
davigonz: I wouldn't remove it, koin is not enough to explain that is related to dependency injection

owncloudApp/build.gradle
```diff
@@ -22,62 +22,44 @@ dependencies {
...
     implementation "org.koin:koin-androidx-viewmodel:$koinVersion"
 
-    // Kotlin extensions
```
davigonz: What is ktx about if we remove it?

davigonz:
@hannesa2 I'm in favor of not writing comments everywhere but providing that the code is self explanatory enough. As I wrote in the code review, some of the comments in the build.gradle can be removed for sure but there are others that would be very useful for someone new in the project who does not necessarily need to know what every dependency is for. 

Imagine that someone wants to work on the image loading and goes to the build.gradle file to research about the library being used. It would be impossible to find it without proper comments.

hannesa2:
Now I keep a lot of pointless comments. -sad-

It's always a question for whom you write comments. For pros or beginners. There are some who always scream for more documentation, but this ones you will never satisfy, because they have other issues

Documentation is like a newspaper: When it's printed, it's outdated. After some time documentation is just garbage

davigonz:
> It's always a question for whom you write comments. For pros or beginners.

I would say that comments should not explain everything from scratch, i.e. people who don't know what is an AndroidManifest.xml, it does not make any sense to explain with comments what is the AndroidManifest used for and as you say, that people will have many other issues that can be solved by reading/practicing more Android development.

In our case, this is an open source project and it's very common to have beginners developers. In deed, I have had to explain many things from scratch to some people and is not a problem for me. 

About this PR, I think that some comments that you wanted to remove from `build.gradle` are not only for beginners, someone with much experience might not know that glide library is used for image loading for example.

> Documentation is like a newspaper: When it's printed, it's outdated. After some time documentation is just garbage

I do not fully agreed with you, it's true that the documentation ends up being outdated

hannesa2:
Let's stop this senseless discussion. That means you keep my name on this pointless comment during git blame

davigonz:
I don't understand the reason for closing this PR now, when you have already applied the changes I requested yesterday.

The code is approved, @jesmrec no QA needed

PR #2703, opened 2019-11-06
Android Studio 3.5.2

hannesa2:


hannesa2:
Either Bitrise or s test is garbage

jesmrec:
> Either Bitrise or s test is garbage

Tests are not garbage. Bitrise either... usually, when you don't like something it is better to suggest an alternative to improve it. Being constructive is one of the best attitudes for a community product.

On the other hand, we work to improve the CI system.

Ah.. i forgot it. This is approved.

PR #2712, opened 2019-11-12
Point correct library commit

jesmrec:
Point to the latest commit in the library


davigonz:
Library reference is already updated, closing it

PR #2720, opened 2019-11-14
[FIX] Fix change device language leads to new image folder name

davigonz:
When changing the device language in Android app the whole configuration changes, too. The problem here is for example that a new directory for image uploads will create.
This means that all images will be uploaded again.

Fix for https://github.com/owncloud/enterprise/issues/3523

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -63,6 +63,8 @@ import androidx.annotation.LayoutRes;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatDelegate;
 
+import static com.owncloud.android.db.PreferenceManager.PREF__CAMERA_UPLOADS_DEFAULT_PATH;
+
 /**
  * An Activity that allows the user to change the application's settings.
  * <p>
@@ -873,8 +875,7 @@ public class Preferences extends PreferenceActivity {
      * Load picture upload path set on preferences
      */
     private void loadCameraUploadsPicturePath() {
-        mUploadPath = mAppPrefs.getString(PREFERENCE_CAMERA_PICTURE_UPLOADS_PATH,
-                getString(R.string.camera_upload_path));
+        mUploadPath = mAppPrefs.getString(PREFERENCE_CAMERA_PICTURE_UPLOADS_PATH, PREF__CAMERA_UPLOADS_DEFAULT_PATH);
         mPrefCameraPictureUploadsPath.setSummary(
                 DisplayUtils.getPathWithoutLastSlash(mUploadPath)
         );
@@ -891,8 +892,7 @@ public class Preferences extends PreferenceActivity {
      * Load video upload path set on preferences
      */
     private void loadCameraUploadsVideoPath() {
-        mUploadVideoPath = mAppPrefs.getString(PREFERENCE_CAMERA_VIDEO_UPLOADS_PATH,
-                getString(R.string.camera_upload_path));
+        mUploadVideoPath = mAppPrefs.getString(PREFERENCE_CAMERA_VIDEO_UPLOADS_PATH, PREF__CAMERA_UPLOADS_DEFAULT_PATH);
         mPrefCameraVideoUploadsPath.setSummary(DisplayUtils.getPathWithoutLastSlash(mUploadVideoPath));
     }
 
--- a/owncloudApp/src/main/res/values-ar/strings.xml
+++ b/owncloudApp/src/main/res/values-ar/strings.xml
@@ -339,8 +339,7 @@
   <string name="placeholder_media_time">12:23:45</string>
   <string name="camera_picture_upload_on_wifi">تحميل الصور عبر الواي فاي فقط</string>
   <string name="camera_video_upload_on_wifi">تحميل الفيديوهات عبر الواي فاي فقط</string>
-  <string name="camera_upload_path">/تحميل الكاميرا</string>
-  <string name="confirmation_disable_camera_uploads_title">تأكيد</string>
+    <string name="confirmation_disable_camera_uploads_title">تأكيد</string>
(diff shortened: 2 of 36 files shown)


owncloudApp/src/main/res/values-ar/strings.xml
```diff
@@ -339,8 +339,7 @@
   <string name="placeholder_media_time">12:23:45</string>
   <string name="camera_picture_upload_on_wifi">تحميل الصور عبر الواي فاي فقط</string>
   <string name="camera_video_upload_on_wifi">تحميل الفيديوهات عبر الواي فاي فقط</string>
-  <string name="camera_upload_path">/تحميل الكاميرا</string>
-  <string name="confirmation_disable_camera_uploads_title">تأكيد</string>
+    <string name="confirmation_disable_camera_uploads_title">تأكيد</string>
```
abelgardep: Indentation problem when deletion of above string in each strings file

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -873,8 +875,7 @@ private AppCompatDelegate getDelegate() {
      * Load picture upload path set on preferences
      */
     private void loadCameraUploadsPicturePath() {
-        mUploadPath = mAppPrefs.getString(PREFERENCE_CAMERA_PICTURE_UPLOADS_PATH,
-                getString(R.string.camera_upload_path));
+        mUploadPath = mAppPrefs.getString(PREFERENCE_CAMERA_PICTURE_UPLOADS_PATH, PREF__CAMERA_UPLOADS_DEFAULT_PATH);
```
abelgardep: This path is not the same as before, we are adding a path separator that before we didn't have. It is intended?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -891,8 +892,7 @@ private void saveCameraUploadsPicturePathOnPreferences() {
      * Load video upload path set on preferences
      */
     private void loadCameraUploadsVideoPath() {
-        mUploadVideoPath = mAppPrefs.getString(PREFERENCE_CAMERA_VIDEO_UPLOADS_PATH,
-                getString(R.string.camera_upload_path));
+        mUploadVideoPath = mAppPrefs.getString(PREFERENCE_CAMERA_VIDEO_UPLOADS_PATH, PREF__CAMERA_UPLOADS_DEFAULT_PATH);
```
abelgardep: Same here

owncloudApp/src/main/res/values-ar/strings.xml
davigonz: I removed it using the string editor, so I would say it's been properly indented (4 spaces) and the indentation problem (2 spaces) was already present in all the translations (maybe because of transifex?). We could indent all the strings.xml files but it's not this PR scope

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
davigonz: I defined `PREF__CAMERA_UPLOADS_DEFAULT_PATH` as a constant in `PreferenceManager` and all the constants in that file already had that path separator you mean, so I followed the same format.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
davigonz: Same as https://github.com/owncloud/android/pull/2720#discussion_r349570074

davigonz:
> A couple of comments here @davigonz

@abelgardep I answered them

owncloudApp/src/main/res/values-ar/strings.xml
abelgardep: I meant to keep same format in every string as it was before.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
abelgardep: I know, but previously camera uploads default path was '/CameraUpload' and now it is '/CameraUpload/'. And there were different uses, some with last file separator and some without. Now, we are using always with last path separator, So I meant that if that change was on purpose or not

davigonz: I see what you mean and just pushed a new commit with some changes, can you recheck it? Thanks

abelgardep:
Ready to QA @jesmrec

jesmrec:
### QA

There are two tests to do:

- Fresh install + Language switch: Installing the app from scratch, enable Camera Uploads, take some pictures, switch to other language, and take more pictures: Everything is in the same folder. This works correctly, even when the camera upload folder switches.

- Upgrade from older version + Language switch, as following of https://github.com/owncloud/enterprise/issues/3523#issuecomment-534469678 and below messages:

Check these steps:

1. Install 2.13.1 in your device, that is set in Spanish language
2. Enable camera uploads and let the default camera folder (/SubidasCamara)
3. Take some pictures to upload there
4. Upgrade to the current branch
5. Switch the device language to English
6. Take some pics with the camera

Current: Uploaded to /CameraUpload. At this moment you have /SubidasCamera and /CameraUploads. From now on, everything will be uploaded to /CameraUploads, regardless the language you set, but the older pics stays in /SubidasCamera till are manually moved. 

That only happens if the selected folder before upgrading is the default one. If you choose another folder in your folder structure, selection stays correctly after upgrading.

In order to fix such collateral effect, it would be needed some kind of migration (is it worthy???)

CC @ChrisEdS

michaelstingl:
> In order to fix such collateral effect, it would be needed some kind of migration (is it worthy???)

No hard requirement. Keep the solution simple please.

davigonz:
> In order to fix such collateral effect, it would be needed some kind of migration (is it worthy???)

@jesmrec I'm able to reproduce the steps you mentioned but as @michaelstingl says, I would not spend more time on this since is not a very common use case and the user could just move the images to the new camera uploads folder

jesmrec:
Ok, the approach is OK for me, so we are getting rid of the main issue. 

Approved.

davigonz:
Merged, to be included in 2.14

PR #2694, opened 2019-11-01
Better logging GUI

hannesa2:
It's about to change logging from self-made to [Timber](https://github.com/JakeWharton/timber).
This PR is the GUI part from https://github.com/owncloud/android-library/pull/275 and it relies on it.

It removes the logic to display log and simply uses a Fragment from logging-lib.

It keeps all previous functions like 
* searching

image

and adds 
* a tab for ownCloud Filelogging view
* a tab for Logcat view 
    * to see what happens on other processes too
    * to log what ownCloud forgot to log
    * within `release` build type !
* share log
* adds filter for `Verbose`, `Debug`, `Info`, `Warning`, `Error` 

Eg, when you select `Info` you will see only `Info`, `Warning`, `Error` entries and filters out `Verbose`, `Debug`

image

So it helps to investigate, what's wrong with ownCloud.

Btw, I recommend to add Firebase Crashlytics in ownCloud. If you want, I can help you with this task

PR #2680, opened 2019-10-04
Kotlinize

hannesa2:
During https://github.com/owncloud/android/pull/2679 I kotlinize some files and fixed some lint warnings. 
To keep https://github.com/owncloud/android/pull/2679 I separated my work into two PR.

PR #2725, opened 2019-11-15
Better logging gui

hannesa2:
It's the https://github.com/owncloud/android/pull/2694 but on ownCloud repo, instead of my fork

Implements #2730 
Needs https://github.com/owncloud/android-library/pull/283

QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.14.0/2725%20-%20Timber.md

- [x] (1) Searching is case sensitive https://github.com/owncloud/android/pull/2725#issuecomment-555977091 [NOT REPRODUCIBLE]
- [x] (2) Sharing logs with ownCloud https://github.com/owncloud/android/pull/2725#issuecomment-555982094 [FIXED]
- [x] (3) name of the log file should be branded [FIXED]
- [x] (4) Log rotation improvements https://github.com/owncloud/android/pull/2725#issuecomment-555998731 [FIXED]
- [x] (5) Share icon is used as clear log option when the clear log option has enough room in top bar. https://github.com/owncloud/android/pull/2725#issuecomment-558542870 [FIXED]
- [X] (6) [improvement] Add timestamp to log files name https://github.com/owncloud/android/pull/2725#issuecomment-558988872 [WONT DO HERE]


abelgardep:
I have found some problems @hannesa2 .

1- I got an infinity loop clearing logfile
infinity-loop

2- Some of those problems are still reproducible (It would be interesting to solve them)
https://github.com/owncloud/android/pull/2694#issuecomment-552332912

hannesa2:
1. I got an infinity loop clearing logfile -> fixed
2. stopLogging is not used anywhere -> I kept it, because when someone has the idea to re-intoduce stopLogging. If I immediate delete it, I expect the call for it
3. I write there

abelgardep:
Ready to QA @jesmrec

jesmrec:
@hannesa2 it is very appreciated. Also, it'd be appreciated if the library PR: https://github.com/owncloud/android-library/pull/275 is moved to the oC repo as well.

jesmrec:
### (1)

The log searching is case-sensitive. This is a bit tricky, since the main target is getting matches, regardless if capitals are there or not. Is it feasible to set the searching as non case sensitive or is it a Timber feature?

jesmrec:
### (2)

1. Open the Logs view 
2. Share the logs with ownCloud

Current: nothing happens
Expected: folder picker? 

This took my attention, since other apps in the same device displays the folder picker to share with ownCloud, but the Log app. But, not sure it this is a matter of the library itself. Sharing with other available providers works properly.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.kt
```diff
@@ -0,0 +1,84 @@
...
+                supportFragmentManager.beginTransaction()
+                    .add(R.id.fragment_container, it).commit()
+            }
+        }
+    }
+
+    companion object {
+        private const val logCatTargetFileName = "logfile.log"
+        private const val logCatSearchHint = "search logcat"
+        private const val bothLogsTargetFileName = "Owncloud.log"
```
jesmrec: Please, don't use `Owncloud` as hardcoded string inside the code, due to branding reasons. You can use the app_name or other parameter.

jesmrec:
### (4) 

About log rotation, every day is created a new file. Long term will flood the app folder with logs. Any way/chance to clean logs (automatically or manually)?

Logs are stored in the root folder of the app in the internal storage. Shouldn't be stored in a specific folder called `Logs`? (feasible?)

additional ideas welcome

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.kt
hannesa2: ok, let's call it log.log (or something other less-meaningful) before it was
image
even _current_ and _older_ has that much meaningful name
image

`app_name` needs context , so `companion object` is not the best place anymore then

But the branch in in your repo, you can do what you want

hannesa2:
> ### (2)
> 1. Open the Logs view
> 2. Share the logs with ownCloud
> 
> Current: nothing happens
> Expected: folder picker?
> 
> This took my attention, since other apps in the same device displays the folder picker to share with ownCloud, but the Log app. But, not sure it this is a matter of the library itself. Sharing with other available providers works properly.

No need for a file picker anymore. It has just one recent file

hannesa2:
> ### (4)
> About log rotation, every day is created a new file. Long term will flood the app folder with logs. Any way/chance to clean logs (automatically or manually)?
> 
> Logs are stored in the root folder of the app in the internal storage. Shouldn't be stored in a specific folder called `Logs`? (feasible?)
> 
> additional ideas welcome

This file logging is pointless since you see in `release` log. And additional you can share logcat with  others. 
Why is a file needed ?

But about flooting: Maybe we should delete all *.log files in that directory during `clear log` ?

hannesa2:
> ### (1)
> The log searching is case-sensitive. This is a bit tricky, since the main target is getting matches, regardless if capitals are there or not. Is it feasible to set the searching as non case sensitive or is it a Timber feature?

It's the behaviour like before. Make an issue and I apply it later. 
Please don't expand requirements, otherwise it will be an endless story

jesmrec:
(1) -> in the previous log implementation, searching was not case sensitive. For example:

device-2019-11-20-154618

(2)

> No need for a file picker anymore. It has just one recent file

if you open the Log view and you want to sync it in your oC server by sharing, you can't (in the former implementation you could). The option is there and does nothing. IMHO is confusing if you select `ownCloud` in the picker and nothing happens. Every action should have a reaction. It is not a matter of watching the file, but storing the file.

(3) -> will check the new names. pending.

(4)

> Maybe we should delete all *.log files in that directory during clear log ?

that works for me. What do you think @abelgardep @davigonz

jesmrec:
@hannesa2 in the report (1) you are right, ignore my answer, since the word `chunk` appears twice in the line, one in uppercase and other in lowercase. former filter worked in case sensitive as well.

So, the behaviour is the same and we are not losing functionality. We can go on with it.

hannesa2:
> (2)

A: during investigation, I solved **a lot** of lint issues, but my experience with https://github.com/owncloud/android/pull/2680 told me not to push this, because it causes too much afford. You prefer to go with a bad current state, instead of an improvement which only fits 95%

B: file chooser+
By the way this use case ! Ok, it should work.
I guess @davigonz knows, what the sense of this prevention it. (It blocks share from oC to oC)
I removed it, and it works. But please change this by your own. I see no sense in it
image

hannesa2:
(4)

>>    Maybe we should delete all *.log files in that directory during clear log ?
>
> that works for me. What do you think @abelgardep @davigonz

When you want to go this way, I'm ready prepared

abelgardep:
> (4)
> 
> > > Maybe we should delete all *.log files in that directory during clear log ?
> > 
> > 
> > that works for me. What do you think @abelgardep @davigonz
> 
> When you want to go this way, I'm ready prepared

Go ahead 👍

davigonz:
> A: during investigation, I solved a lot of lint issues, but my experience with #2680 told me not to push this, because it causes too much afford. You prefer to go with a bad current state, instead of an improvement which not fit 100% concerning automatic code format

About https://github.com/owncloud/android/pull/2680 , as @abelgardep pointed in https://github.com/owncloud/android/pull/2680#issuecomment-550986306 , there's some automatic transformations from Java to Kotlin that are a bit dangerous, mostly related to nullability.

> causes too much afford

Do you mean effort? Code reviews cause effort, it's something usual in a development process, you might probably know that. We do not merge code as it is, if some changes are needed, they should be applied. Otherwise, we will pile up not 100% proper code and it will end up in something not that good.

> I guess @davigonz knows, what the sense of this prevention is. (It blocks share from oC to oC)
I removed it, and it works. But please change this by your own. I see no sense in it

Yes, I can not give the details about that change but it was needed. I will have a look at it anyway.

By the way, I've noticed that there's a duplicated share icon when I install it in a Nexus 6P, any idea?

[](https://user-images.githubusercontent.com/15261363/69428681-1d572e00-0d32-11ea-9646-ebc0652c47d5.png)

hannesa2:
(4)

> Go ahead 👍

@abelgardep done

hannesa2:
> By the way, I've noticed that there's a duplicated share icon when I install it in a Nexus 6P, any idea?

Sorry, I've no such device.

I add this _share_ for each fragment separately with

    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {

        inflater?.inflate(R.menu.menu_log, menu)

My expectation is that it's only added for recent fragment

hannesa2:
With an Emulator Nexus 6P
image

I can't reproduce it
image

davigonz:
@hannesa2 I get this error when updating the submodule library from the last commit of this PR

```
cerror: Server does not allow request for unadvertised object 10a7bdbc4d06f0ff4f6ac477095eb8515f561293
Fetched in submodule path 'owncloud-android-library', but it did not contain 10a7bdbc4d06f0ff4f6ac477095eb8515f561293. Direct fetching of that commit failed.
```

It seems you did not push your [last commit from your library repo branch](https://github.com/hannesa2/android-library/commit/10a7bdbc4d06f0ff4f6ac477095eb8515f561293) to our library one.

I tried to do it myself but since you also have a branch in your library called `betterlogging`, git gets confused with `Betterlogging` branch and it mixes the branches. I advise that you use different names, not the same one with lower or upper cases. 😉 

Screenshot 2019-11-25 at 14 04 41

Screenshot 2019-11-25 at 14 08 58

hannesa2:
Before:
image

`git push upstream`

**Now:**
image

This has nothing to do with upper/lowercase branch names. You can call it, what you want, simply `sha1` rules
image

davigonz:
Yesterday I performed the next steps in your library repo:
1. `git checkout BetterLogging`
2. `git fetch`
3. `git rebase`

And the result was what you see in the image below

Screenshot 2019-11-25 at 14 04 41

The remote tracking reference is `origin/BetterLogging` and it was containing the commits of your remote `betterLogging` branch (`increase cache size`, `split log into debug and release`) instead of the commits of your remote `BetterLogging` branch. No idea about was going on under the hood, but is the first time I see something like this. Maybe I'm missing something.

davigonz:
(4) Share icon appears as the delete option when the delete option has enough room to appear in the top bar.

Nov-26-2019 10-28-17

davigonz:
@jesmrec all the bugs should already be fixed, could you check them?

jesmrec:
### (6) (improvement)

After fixing (2) and (3), the name of the logs file to share is always

<app name>.log

would it be posible to add the timestamp (same as the log stored in device). So, if some files are stored, can be distinguised. Something like

<app name>_20191127_101145.log

what do you think?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.kt
```diff
@@ -51,7 +50,7 @@ class LogHistoryActivity : ToolbarActivity() {
 
             fileLoggingTree()?.let {
                 logFragment = BothLogsFragment.newInstance(
-                    bothLogsTargetFileName,
+                    "${getString(R.string.app_name)}.log",
```
hannesa2: I didn't debugged it, but is it that smart to use `R.string.app_name` ? What's about when the app has a name like "my Cloud" with space ?
Leave it like it was and all is fine

hannesa2:
> ### (6) (improvement)
> would it be posible to add the timestamp (same as the log stored in device). So, if some files are stored, can be distinguised. Something like
> 
> _20191127_101145.log
> 
> what do you think?

https://github.com/hannesa2/Logcat/blob/5cba849239a0d71ef21f69d91c986c875c16b580/LogcatLib/src/main/java/info/hannes/timber/FileLoggingTree.kt#L23-L28 
I don't recommend it, then the name will be `_20191127_101145.2019-11-27.log` ...please not

jesmrec:
sorry, GH text processor played me a trick.

What i meant, is using the name of the app as name of the file (what i called there as "\<nameApp\>.log" and GH ellipsized):

ownCloud.log
MyApp.log
MyCompany.log

depending on the parameter in `setup.xml` for the appName.

so, the improvement will change those names for 

ownCloud_20191127_101145.log
MyApp_20191127_101145.log
MyCompany_20191127_101145.log

sorry for not reviewing the published comment

owncloudApp/src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.kt
davigonz: I tried it and no problem at all with spaces in the log to share

[](https://user-images.githubusercontent.com/15261363/43832186-df6aa322-9b06-11e8-99ce-f2d3d56a1542.png)

davigonz:
> ownCloud_20191127_101145.log
> MyApp_20191127_101145.log
> MyCompany_20191127_101145.log

This is not that trivial so I would keep it as it is, just `appname.log`

jesmrec:
Ok, let's keep it ftm... awaiting for more feedback.

From my side this is approved.

Thanks for the contribution @hannesa2!

PR #2731, opened 2019-11-18
update communication process

michaelstingl:


davigonz:
Thanks for this change @michaelstingl 

@jesmrec No QA needed so merging it

Issue #2732, opened 2019-11-20
[TESTING] Setting tests flakiness 


davigonz:
I was working on #2728 and at some point, I executed all the tests to check that are working as expected and noticed that some of the tests related to settings seem to be a bit flaky. I switched to master branch and try it there again but the result is pretty similar, could you check them @jesmrec ? 

- 1st execution, 5 tests failing:

![Screenshot 2019-11-20 at 13 02 03](https://user-images.githubusercontent.com/15261363/69237495-1eebef00-0b96-11ea-94a6-8b2662d68229.png)

- 2nd execution, 3 tests failing

![Screenshot 2019-11-20 at 13 03 51](https://user-images.githubusercontent.com/15261363/69237552-388d3680-0b96-11ea-8003-1a5188df13e5.png)

- 3rd execution, 4 tests failing

![Screenshot 2019-11-20 at 13 05 48](https://user-images.githubusercontent.com/15261363/69237685-80ac5900-0b96-11ea-90cd-8968368604ce.png)

- 4th execution, 5 tests failing

![Screenshot 2019-11-20 at 13 07 58](https://user-images.githubusercontent.com/15261363/69237820-cec15c80-0b96-11ea-9085-d613af482b9b.png)

Device: Samsung Galaxy S9
Android Version: Android 9

PR: https://github.com/owncloud/android/pull/2737

davigonz:
@jesmrec the setting tests are not being executed on Bitrise for now, in order to move forward pending PRs without relaunching the tests multiple times. Please ping me as soon as they work fine on your local (emulator or Samsung Galaxy S9) and I will put them back

jesmrec:
> @jesmrec the setting tests are not being executed on Bitrise for now, in order to move forward pending PRs without relaunching the tests multiple times. Please ping me as soon as they work fine on your local (emulator or Samsung Galaxy S9) and I will put them back

@davigonz with the PR https://github.com/owncloud/android/pull/2737, tests are not failing anymore in the real device. But, before merging, we can give a try in the PR branch to know what happens under CI.

Issue #2709, opened 2019-11-09
Espresso test failing


hannesa2:
> sometimes, the BitRise fails due to some timeouts or performance problems in its internal emulators. We are looking for the best solutions.

_Originally posted by @jesmrec in https://github.com/owncloud/android/pull/2689#issuecomment-549246165_

I took a look into Espresso test healthiness but the result is sobering.
I run locally `./gradlew :owncloudApp:connectedDebugAndroidTest` on master c2d0e860dabd0 and see this
![image](https://user-images.githubusercontent.com/3314607/68529535-0232fa00-0300-11ea-8939-a25841d7f3af.png)
![image](https://user-images.githubusercontent.com/3314607/68529545-1119ac80-0300-11ea-800f-9c3489c091c5.png)
![image](https://user-images.githubusercontent.com/3314607/68529558-2e4e7b00-0300-11ea-99f6-e6101e336a46.png)

Does is only run on specific Emulators ? I use 
![image](https://user-images.githubusercontent.com/3314607/68540117-6c45b080-038d-11ea-832b-05133ab2308b.png)

Or is there some magic to to on my machine, to have a constant 100% success result ?

---------------------------

Edited by @davigonz 

After writing down the tests which fail with more frequency, we can determine that `secondTryIncorrect` is the test that fail at most times. @jesmrec I think you wrote this test, is there anything that can be improved there?

jesmrec:
no, the magic is not needed. Let me tell you what the problem is. 

Before the current re-architecting of the app, there was not posible to develop test in every layer. So, the Espresso tests cover the whole stack (including the server). So, these tests were removed from CI, to be executed only locally. We should move then (not remove), in order to have a cleaner execution.

hannesa2:
Nice, when you take care about it. 👍 
Btw, I expected a constant failure rate, but the percentage is changing, that means there are flaky tests too

davigonz:
Hi @hannesa2 , thanks for these reports. UI tests in Android, as you probably already know, are highly dependent on the device in which are executed: different screens resolution, size, performance, animations. It seems that the Emulator you chose is a default one with everything enabled I guess, including the animations, so your results are the what I would expect. 

[...]

PR #2728, opened 2019-11-18
[New arch] [Modularization] Tests and cleanup

davigonz:
Implements https://github.com/owncloud/android/issues/2701
Needs https://github.com/owncloud/android-library/pull/282

PR #2737, opened 2019-11-27
Adjustments in Settings UI tests to avoid flakyness

jesmrec:
Issues: #2732 #2709

Some actions performed:

- Camera Uploads section:

For these tests, the whole preferences activity was loaded, taking advantage that the section is on the top (visible at a glance). Some tests were flaky because the activity takes a bit longer in loading, before the action over the camera uploads checkbox is done (this is something not desirable that the testing framework should take care). Therefore, the dialog warning users to set up correctly the camera uploads is not displayed, failing the matcher over the "OK" button.

As fix, the activity is only loaded with the camera uploads section. 

- LocalFolderPicker activity

Here, the fix is a bit hacky. An implicit wait till the UI componentes are visible to launch the tests. Although there is a loop, it will not be infinite since the `activityRule` always is created.

- Passcode

The test for cancelling the activity was removed. It was being flaky due to `isFinishing` assertion. Such test is not a critical one (too far from this.. it only checks that a cancelled view is dismissed). So, better to get rid of.

These fixes were tested with:

- Samsung Galaxy S9, Android 9
- Nexus 6P , Android 7
- Huawei P20 Lite, Android 9

### DIFF
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/camerauploads/OCSettingsCameraUploadsTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/camerauploads/OCSettingsCameraUploadsTest.kt
@@ -21,7 +21,9 @@ package com.owncloud.android.settings.camerauploads
 
 import android.os.Environment
 import android.preference.CheckBoxPreference
+import android.preference.PreferenceCategory
 import android.preference.PreferenceManager
+import android.preference.PreferenceScreen
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions.click
 import androidx.test.espresso.assertion.ViewAssertions.doesNotExist
@@ -36,14 +38,15 @@ import androidx.test.rule.ActivityTestRule
 import com.owncloud.android.R
 import com.owncloud.android.ui.activity.LocalFolderPickerActivity
 import com.owncloud.android.ui.activity.Preferences
-import org.junit.Assert
-import org.junit.Assert.assertTrue
-import org.junit.Assert.assertFalse
-import org.junit.Before
 import org.junit.After
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+
+
 
 @RunWith(AndroidJUnit4::class)
 class OCSettingsCameraUploadsTest {
@@ -65,6 +68,15 @@ class OCSettingsCameraUploadsTest {
 
         mPrefCameraPictureUploads = activityRule.activity.findPreference(CAMERA_PICTURE_UPLOADS) as CheckBoxPreference
         mPrefCameraVideoUploads = activityRule.activity.findPreference(CAMERA_VIDEO_UPLOADS) as CheckBoxPreference
+
+        //Only interested in "Camera Uploads" section, so we can get rid of the other categories.
+        val preferenceScreen = activityRule.activity.getPreferenceScreen() as PreferenceScreen
+        val securityCategory =
+            activityRule.activity.findPreference("security_category") as PreferenceCategory
+        val moreCategory =
+            activityRule.activity.findPreference("more") as PreferenceCategory
+        preferenceScreen.removePreference(securityCategory)
+        preferenceScreen.removePreference(moreCategory)
     }
 
     @After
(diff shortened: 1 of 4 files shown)


owncloudApp/src/androidTest/java/com/owncloud/android/settings/camerauploads/OCSettingsLocalFolderPickerTest.kt
```diff
@@ -72,4 +80,19 @@ class OCSettingsLocalFolderPickerTest {
...
+    fun waitForLocalPickerActivity(){
+        var textVisible = false
+        while (textVisible == false) {
+            try {
+                onView(withContentDescription(R.string.abc_action_bar_up_description)).check(matches(isDisplayed()))
+                onView(withId(R.id.folder_picker_btn_home)).check(matches(isDisplayed()))
+                onView(withId(R.id.folder_picker_btn_choose)).check(matches(isDisplayed()))
+                onView(withId(R.id.folder_picker_btn_cancel)).check(matches(isDisplayed()))
+                textVisible = true
+            } catch (e: NoMatchingViewException) {
```
hannesa2: That's nasty. There is no exit strategy to prevent an endless loop

hannesa2: This is nasty, I hope I find time to show you a proposal how you can solve this specific task. (Not all tests !)

jesmrec:
After merging https://github.com/owncloud/android/pull/2725, the Log menu UI has changed, so the tests were failing. Latest commit adapt it, by simple checks on the new options.

owncloudApp/src/androidTest/java/com/owncloud/android/settings/camerauploads/OCSettingsLocalFolderPickerTest.kt
jesmrec: I'm checking also better solutions. thanks.

jesmrec:
@hannesa2 i guess you trigger the tests in a local emulator. By executing in a real device, and also BitRise are working. Taking a look as well

davigonz:
I've been executing the tests in a Nexus 6P emulator and there's some intent errors triggered during `OCSettingsMoreTest` execution, something like `E/TestRunner: androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: Wanted to match 1 intents. Actually matched 0 intents.` and `java.lang.IllegalStateException: #init was called twice in a row`

I've performed some changes there and it increased a bit the success rate.

Screenshot 2019-11-28 at 10 59 26

Could you check them @jesmrec ? It would be great to have 100% even with emulator.

Anyway, this issue is about the Settings tests, not all of instrumented ones so first we want the Settings tests to be fixed and the other related to shares and capabilities are being fixed via https://github.com/owncloud/android/pull/2728

So as soon as the Setting tests are not flacky anymore (no more OCShareWhatever tests failing when executing `./gradlew :owncloudApp:connectedDebugAndroidTest`), we will merge them, close this issue and follow up
https://github.com/owncloud/android/issues/2709  for the whole instrumented test suite of presentation layer (owncloudApp module)

jesmrec:
I have get rid of the ugly implicit wait in the `LocalFolderPicker` suite, since it does not fix the problem and it is not desirable to have. 

I have updated and performed more testing, with the following results:

- Real devices: i have tested with different devices from Android 7 to Android 10, and all the Settings tests pass correctly. 

- CI: BitRise is passing correctly the tests over their emulators

- Local emulators: With Pixel emulators, tests are running correctly as well. But, other emulators like Nexus are running into issues. After researching a bit about it, i've realised that most of them are caused by some kind of weird race condition in the moment of loading the `ActivityRule`. Activity is loaded, but the first action (in most cases, a click) is not performed but there is no `NuSuchElementException` or similar either.  

Which emulators are you using @davigonz @hannesa2? do you trigger directly with `./gradlew test` or use additional options?

davigonz:
`./gradlew :owncloudApp:connectedDebugAndroidTest` with a Nexus6P API 28

davigonz:
@jesmrec thanks for this detailed report, the work done here is enough imho. We can not guarantee the tests passing in all emulators for now, Bitrise was the blocker here and the tests are passing there now and also in at least, a local emulator. It seems to be a problem with Nexus 6P emulators.

davigonz:
Tests are properly passing on Bitrise

hannesa2:
image

For the record: I can't confirm to work properly on an Emulator

image

jesmrec:
which emulator are you using? Pixel? Nexus? as i have written above, in different emulators i got different results. Also, it'd be helpful to know which tests are failing, since there are other PRs in which tests are being fixed.

hannesa2:
Ohh, I my idea was a independency from hardware profile, **just a dependency from system-image**. That's why I only wrote in previous mesage just system info, here is the complete info.
But maybe I'm wrong

image

jesmrec:
Ohh, my idea was the same! but... when things do not work, you have to know where the problem is (in order to fix it... ). Thanks for the additional info., will take a look.

PR #2741, opened 2019-11-28
Proposal of automatic versionname, just with taging it

hannesa2:
(cherry picked from commit 159b88260a80b4c35af3342cf6c7323e0749c2b5)

To not differ too much from my fork, I make this proposal, to get rid of manual version change in code.
Simply tag it, and the last used tag will be used as `versionName`

image

### DIFF
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -93,8 +93,8 @@ android {
         minSdkVersion 19
         targetSdkVersion 28
 
-        versionCode = 21300100
-        versionName = "2.13.1"
+        versionCode = getGitCommitCount()
+        versionName = getTag()
 
         buildConfigField "String", gitRemote, "\"" + getGitOriginRemote() + "\""
         buildConfigField "String", commitSHA1, "\"" + getLatestGitHash() + "\""
@@ -195,13 +195,8 @@ def setOutputFileName(variant, appName, callerProject) {
     }
 
     def versionName = "$variant.mergedFlavor.versionName"
-    if (variant.mergedFlavor.manifestPlaceholders.versionName != null) {
-        versionName = "$variant.mergedFlavor.manifestPlaceholders.versionName"
-    }
-    if (variant.buildType.manifestPlaceholders.versionName != null) {
-        versionName = "$variant.buildType.manifestPlaceholders.versionName"
-    }
-    newName += "_$versionName"
+    def versionCode = "$variant.mergedFlavor.versionCode".toInteger()
+    newName += "_$versionName-$versionCode"
 
     def buildNumber = System.env.OC_BUILD_NUMBER
     if (buildNumber) {
@@ -215,6 +210,16 @@ def setOutputFileName(variant, appName, callerProject) {
     variant.outputs.all {
         outputFileName = new File(".", newName)
     }
+}
+
+static def getGitCommitCount() {
+    def process = "git rev-list HEAD --count".execute()
+    return process.text.toInteger() + 21300100
+}
+
+static def getTag() {
+    def process = "git describe --abbrev=0".execute()
+    return process.text.toString().trim()
 }
 
 static def getLatestGitHash() {


davigonz:
I like the `getGitCommitCount()` approach but it would break our [wizard implementation](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java#L181), in which we convert the `versionName`, e.g. `2.13.1` to `versionCode`, e.g. `21300100` to know the feature slides the app must show when updating or installing a specific version.

About the `getTag()` method, it changes the `versionName` fom `2.13.1` to something like `oc-android-2.13.1_oem` and since that name is ["used as the version number shown to users"](https://developer.android.com/studio/publish/versioning) and I would say it's also what appears in Google Play, `oc-android-2.13.1_oem` is not a good name to be shown to our users, it works fine as an internal tag but not as a versionName for the app.

hannesa2:
>  in which we convert the versionName, e.g. 2.13.1, to versionCode, e.g. 21300100

Is this ready necessary ? Only as an idea, I use this changelog info automatically with https://github.com/hannesa2/ChangeLog 
It autogenerate changelog on each tag/CI-build, with collection all tags (with filter option, eg 'release' to filter only to customers)

> oc-android-2.13.1_oem

I don't know anything about your politics, but I would remove complete this `oc-android-....._oem`
Btw, it's easy to add such pre/suffix again.

As bigger idea, because customers don't know anything about this magic numbers `2`.`13`.`1`, you could to switch too an easy to understand logic: 

`2019-12-16`

michaelstingl:
> As bigger idea, because customers don't know anything about this magic numbers `2`.`13`.`1`, you could to switch too an easy to understand logic:

We'll try to better follow [Semantic Versioning](https://semver.org/) …

jesmrec:
> oc-android-2.13.1_oem is not a good name to be shown to our users, it works fine as an internal tag but not as a versionName for the app.

totally agreed. `oc-android-2.13.1_oem` is only an internal way to tag a commit, that we followed as standard. Of course, we can re-think for other options like the suggested here, but personally i don't like the idea of settings such string as version identifier for the final users.

To move forward this PR, we'd need to adjust our tagging/versioning policy. I agree with @davigonz that something like `oc-android-2.13.1_oem` is not suitable to set as version in the markets either. We can unify tags and versions as @hannesa2 suggested, but this is not straightforward. Not only in the current code, so other internal builders (CI) run scripts based in such numbers, so removing or changing them then can carry other problems/adjustments to do 

So, @michaelstingl @hosy, it is important o know what does oC expect from versioning:

- What to show in Settings view
- How to tag the release commits

without forgetting that customers can set their own numbering, that must fit with the way to calculate for public users.

hannesa2:
Btw, I'm fine when you still stay and close this pull request.
This means you keep magic numbers = Semantic Versioning and change version by a new commit here https://github.com/owncloud/android/blob/5c144d3b7f89e1a4a092d41c013334996a24a289/owncloudApp/build.gradle#L107-L108 every time

davigonz:
> Is this ready necessary ? Only as an idea, I use this changelog info automatically with https://github.com/hannesa2/ChangeLog

Not sure how that library could help us to change the wizard implementation. If you think you have a better solution to show/hide the wizard depending on the version (without [this method](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/features/FeatureList.java#L181)) please send us a PR with your solution and we will review it.

> This means you keep magic numbers = Semantic Versioning

Well, we have not re-invented the wheel, it's what most of the Android apps show in Google Play, even with more extensions than major, minor and patch.

[](https://user-images.githubusercontent.com/15261363/70980698-01864400-20b4-11ea-92c9-edd08a592837.png)

> change version by a new commit here every time

Do not worry, it's not that much, usually changing two lines once every two months.

Thanks for the contribution @hannesa2 , think we will keep what we have for now

PR #2746, opened 2019-12-01
remove twice log

hannesa2:
No need fom `Timber.plant(DebugTree(Log_OC::class.java))`
because
`Timber.plant(FileLoggingTree(...))`
is enough


abelgardep:
Thanks @hannesa2, moving this forward.

jesmrec:
Logs don't repeated anymore. Thanks for the fix @hannesa2

Issue #2738, opened 2019-11-27
Get rid of Android 4.4 support


davigonz:
Android 4.4 was released long time ago (31st October 2013) and we are still maintaining some parts of the code to support it. Taking into account the oC users using this version, not that many, we could upgrade the minimum version to Android 5.0 (Lollipop)

### PR
- App #2748 
- Library https://github.com/owncloud/android-library/pull/285

PR #2748, opened 2019-12-03
Get rid of Android 4.4 support

abelgardep:
Implements #2738 
Library https://github.com/owncloud/android-library/pull/285


jesmrec:
switch base branch to `master`?

abelgardep:
> switch base branch to `master`?

Yes, but needs modularization to be merged first. 👍

jesmrec:
Smoke needed.

owncloudApp/src/main/java/com/owncloud/android/utils/UriUtils.java
```diff
@@ -119,12 +119,9 @@ public static boolean isContentDocument(Uri uri) {
      * @param uri       The URI to resolve
      * @return The path in the file system to the content or null if it could not be found (not a file)
      */
-    @TargetApi(Build.VERSION_CODES.KITKAT)
     public static String getLocalPath(Uri uri, Context context) {
-        final boolean isKitKatOrLater = Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT;
-
         // DocumentProvider
-        if (isKitKatOrLater && DocumentsContract.isDocumentUri(context, uri)) {
+        if ( DocumentsContract.isDocumentUri(context, uri)) {
```
davigonz: Delete additional white space before `DocumentsContract.`

davigonz:
@abelgardep just a tiny change

abelgardep:
> @abelgardep just a tiny change

Applied

davigonz:

jesmrec:
just to be sure: users with a 4.4 android version...

- ...will not be able to install the app from scratch, because it will not appear neither in Google Play nor in FDroid. If they try to build the app, thay can not install because the target is older than the `sdkMinVersion ` set in the `build.gradle`
- ...will not be able to upgrade to 2.15 in case they had already installed the app. They will keep the 2.14 forever.

is this correct? do i miss any other scenario?

davigonz:
> will not be able to install the app from scratch, because it will not appear in Google Play

Correct

> will not be able to install the app from scratch, because it will not appear in FDroid

I think so, although I'm not sure how F-droid handles `minSdkVersion`. But I would expect the same behaviour as in Google Play.

> If they try to build the app, they can not install because the version is older than the sdkMinVersion set in the build.gradle

That's it, I've just tried it and ...

Screenshot 2019-12-17 at 10 22 06

> ...will not be able to upgrade to 2.15 in case they had already installed the app. They will keep the 2.14 forever.

Yes, I get the same error as in the previous image when updating the app.

jesmrec:
Checks performed:

- [X] Install the current branch in Android 4.4 -> not posible
- [X] Upgrade to the current branch from latest stable in Android 4.4 -> not posible
- [X] Smoke test over Android > 4.4

Everything OK.

PR #2772, opened 2019-12-22
Ability to change sync start date to the past

hannesa2:
As follow up from https://github.com/owncloud/android/pull/2679 I miss tons of pictures on my wife's phone in our ownCloud.
To be able to sync to missing pictures/videos I've to change change sync start date, the only way was to do it in database as a pro-feature
This pull request has hard coded strings, but for me it works. I want to share it with you, but it's up to you how you proceed with: close unmerged, improve or merge

image

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -23,6 +23,7 @@
 package com.owncloud.android.ui.activity;
 
 import android.app.AlertDialog;
+import android.app.DatePickerDialog;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.pm.PackageInfo;
@@ -43,11 +44,16 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Toast;
 
+import androidx.annotation.LayoutRes;
+import androidx.appcompat.app.ActionBar;
+import androidx.appcompat.app.AppCompatDelegate;
 import com.google.android.material.snackbar.Snackbar;
 import com.owncloud.android.BuildConfig;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.authentication.BiometricManager;
+import com.owncloud.android.datamodel.CameraUploadsSyncStorageManager;
+import com.owncloud.android.datamodel.OCCameraUploadSync;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.db.PreferenceManager.CameraUploadsConfiguration;
 import com.owncloud.android.files.services.CameraUploadsHandler;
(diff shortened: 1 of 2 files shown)


michaelstingl:
@hannesa2 How does this work? Is this a one-time action that adds older images to the upload queue?

hannesa2:
Exactly.
Imagine, my wife got the new version with https://github.com/owncloud/android/pull/2679 and this works for all pictures from now on. But what's about the past pictures ? Even a reinstall doesn't help.
I haven't this problem, because my phone was broken before https://github.com/owncloud/android/pull/2679 was solved

That's why this single event action tells the synchronization from which date to start.

I know, it's a super special pro feature and nothing for the mainstream, and when you like it, it should be only visible in developer mode.

When I write this, I recognize during update to 2.14 a question to the user, which has Wifi sync on, would be useful.

"_We fixed WiFi synchronization .... do you miss pictures....there you can set a new start date .._"

hannesa2:
Btw, before I tried to send/share missed images to ownCloud, but this wasn't possible.
Only chuck size of 200 are allowed by Android, then app was crashing .... so I start to solve it by my own

jesmrec:
Some comments before including this in the roadmap.

- Why did you place the option in the bottom of the `Settings` instead on the `Camera uploads` section?
- I installed the branch to play a bit with it (not official QA), and i was not able to set a date till both picture and video are enabled. Shouldn't be two different settings? 

Here we have also a product component. Not only Android, also the iOS app implements the "Camera uploads" feature as "enable, and start to upload at this moment". So, will check with PM is this is desired in that way. 

What do you think?

michaelstingl:
> also the iOS app implements the "Camera uploads" feature as "enable, and start to upload at this moment". So, will check with PM is this is desired in that way.
> 
> What do you think?

iOS app could get such a one-time action as well. No blocker.

hannesa2:
> * Why did you place the option in the bottom of the `Settings` instead on the `Camera uploads` section?

I don't spent any second about this question. And yes, maybe it makes sense to put it to an other place. But for me it was not important, my motivation was to make it work only

> * I installed the branch to play a bit with it (not official QA), and i was not able to set a date till both picture and video are enabled. Shouldn't be two different settings?

Of course there is always place for improvements. I put the branch to upstream repo.

hannesa2:
Keep in mind, this very special bug-fix feature of your non-working-sync between 2018-2019 was done only for me. 
The longer you wait, the more pointless it is for others.

Should I close it, or is someone interested ?

jesmrec:
It is OK to keep the PR, every fix minds. But we try to have the product as much correct in all its layers, including the presentation. From my pov, a very good fix or feature must be presented in the correct way. We have many users (and customers)... and giving him an option regarding the camera uploads, that is placed on the bottom of the Settings is not a good option, just for giving you an example (recheck [this](https://github.com/owncloud/android/pull/2772#issuecomment-572030793)) . I know that the feature/algorithm is where the magic happens, but from a quality pov is not enough. Are you open to make suggested improvements?

If anyone in the team has other better ideas, they will be welcome as well

Also, for me it is not clear if we are interested in a feature to upload older pictures than the ones after enabling the `Camera Uploads` feature. @michaelstingl your comment [here](https://github.com/owncloud/android/pull/2772#issuecomment-572032415) means that it is OK?

michaelstingl:
> your comment [here](https://github.com/owncloud/android/pull/2772#issuecomment-572032415) means that it is OK?

Yeah, I could imagine such a feature for both Android and iOS app.

hannesa2:
@jesmrec 
Thank you about the buzz word quality. 

I fixed the broken 

- sync on Android P
- non-working WiFi-only sync
- and this PR heals the resulting missing images upload

I placed it at the very most less in-accesable place I found, because it's a use-only-once option, and should only used when you have problems.
 
But when you have doubt about quality, and think it should be placed on a more prominent place, then I guess this PR has not a glorious future

hannesa2:
I solved only conflicts with a rebase to master

PR #2776, opened 2020-01-02
remove a crash

hannesa2:
This is an other crash fix
image


jesmrec:
thanks for the fix. any clue about if it is reproducible and how?

hannesa2:
It happened during mark a folder "offline available"

abelgardep:
Will be merged via #2796 

Thanks @hannesa2

PR #2779, opened 2020-01-03
update Room schema 28

hannesa2:
You changed with 348dcc0d35176a5956bfc201c83f03059ffd5eb4 for [2.14.1](https://github.com/owncloud/android/releases/tag/oc-android-2.14.1) the database structure by removing from 
`@ColumnInfo(name = CAPABILITIES_SHARING_PUBLIC_EXPIRE_DATE_DAYS, defaultValue = "0")` the `defaultValue`.
As follow up, room changed the database structure and **this changed file was ignored.** -sad-
Because ownCloud use `exportSchema = true` room is now no more able to open the database -sad-

The tests were fine, because I guess, you do only `InMemory` database tests. But I didn't checked this.
 
I don't use the GooglePlaystore and I don't know if you already rolled out this version. If you did it, I would not surprised about a disaster. But I don't know, so please double check this.

This pull request is to demonstrate the changes, and for me it works (because I'm able to make my fixes). But maybe you have to increase database version, make probably a migration fix too and ignore this pull request. But this is out of this pull request scope

PR #2780, opened 2020-01-03
no backup to have the ability to fix a damaged database

hannesa2:
Otherwise even after a deinstallation of the app, a damaged database is still on the filesystem and a reinstall doesn't help.

But maybe you fix it with a new database version, then this pull request is obsolete


hannesa2:
I guess, with https://github.com/owncloud/android/pull/2779 or a similar solution it's no more necessary.

PR #2782, opened 2020-01-05
Android Studio 3.5.3

hannesa2:
(cherry picked from commit e6d9b72a2b4840c3b549c62649e2a2b5dcdc16d6)

### DIFF
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -22,7 +22,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$appCompat"
     implementation 'com.getbase:floatingactionbutton:1.10.1'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
-    implementation "androidx.browser:browser:$androidX"
+    implementation "androidx.browser:browser:1.2.0"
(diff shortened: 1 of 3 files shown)


owncloudApp/build.gradle
```diff
@@ -22,7 +22,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$appCompat"
     implementation 'com.getbase:floatingactionbutton:1.10.1'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
-    implementation "androidx.browser:browser:$androidX"
+    implementation "androidx.browser:browser:1.2.0"
```
abelgardep: I think this is not related with Android 3.5.3, is it? Next time instead of replacing with a value, create a new variable in the build gradle as it is kotlinVersion.

hannesa2: It's not related.
A variable makes only sense, if you use it at least twice. Here it isn't.
Using everywhere a variable (without the need) only removes the ability to update with 
intention action

image

jesmrec:
Approved

davigonz:
Merged, thanks @hannesa2

PR #2783, opened 2020-01-05
Update Copyright year to 2020 (app)

hannesa2:
Btw, I see this inconsequent maintained @author still as redundant and pointless, because `git` has **much better** tools to see who did something

### DIFF
--- a/THIRD_PARTY.txt
+++ b/THIRD_PARTY.txt
@@ -1,7 +1,7 @@
 ###################################################################
   ownCloud Android client
 
-  Copyright (C) 2019 ownCloud GmbH.
+  Copyright (C) 2020 ownCloud GmbH.
   Copyright (C) 2012  Bartek Przybylski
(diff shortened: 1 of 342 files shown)


hannesa2:
What's the point to have such lines ?
`Copyright (C) 2011  Bartek Przybylski`

jesmrec:
> Btw, I see this inconsequent maintained @author still as redundant and pointless, because git has much better tools to see who did something

not everyone that get the code can manage with `git` (probably the most, for sure). 

About the year, we are wondering if the year that appears in every file should be the year in which it was modified the last or the current year. This issue can lead to some legal/copyright concerns.

@DeepDiver1975 do you know how should this be handled? are useful/needed the lines that @hannesa2 pointed out in the comment above?

davigonz:
@jesmrec code approved, no QA needed here

PR #2770, opened 2019-12-20
Replace Log_OC with timber

abelgardep:
Implements #2757 
Needs https://github.com/owncloud/android-library/pull/293

PR #2792, opened 2020-01-16
Use timber instead of printstacktrace

abelgardep:
Continuation of #2770 suggested in https://github.com/owncloud/android-library/pull/293#issuecomment-574991808
Needs https://github.com/owncloud/android-library/pull/297

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
@@ -97,8 +97,7 @@ public class AccountAuthenticator extends AbstractAccountAuthenticator {
             try {
                 validateAccountType(accountType);
             } catch (AuthenticatorException e) {
-                Timber.e(e, "Failed to validate account type " + accountType + ": " + e.getMessage());
-                e.printStackTrace();
+                Timber.e(e, "Failed to validate account type %s", accountType);
                 return e.getFailureBundle();
             }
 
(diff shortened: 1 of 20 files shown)


owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -97,8 +97,7 @@ public Bundle addAccount(AccountAuthenticatorResponse response,
             try {
                 validateAccountType(accountType);
             } catch (AuthenticatorException e) {
-                Timber.e(e, "Failed to validate account type " + accountType + ": " + e.getMessage());
-                e.printStackTrace();
+                Timber.e(e, "Failed to validate account type %s", accountType);
```
davigonz: We are duplicating code here, please create a method with:

```
try {
    validateAccountType(accountType);
} catch (AuthenticatorException e) {
    Timber.e(e, "Failed to validate account type %s", accountType);
    return e.getFailureBundle();
}
```

abelgardep: I think that it is out of this pr scope which is to stop using printStacktrace.

BTW, we are going to refactor this class in the short term future with the new arch. So we can face this kind of problems there. Can't we?

davigonz: I try to use the boy scout rule (leave the code better than you found it) if the changes are tiny. I thought this was a good example but it looks good to me.

davigonz:

jesmrec:
No much to test here, beyond the basic checks. Timber is making the magic.

Approved.

hannesa2:
@davigonz 
Only for the records:
You could get rid of such pointless commits (which are in the history just noise) by using `--fixup`

Instead of 
`git commit -m "Rebase with master"`

perform a 
```
git commit --fixup 94ed4f6edd3
git rebase -i master --autosquash
```

hannesa2:
Sorry, I mean this
image

davigonz:
Thanks for the tip @hannesa2 , I'll give it a try in coming rebases with master

PR #2814, opened 2020-02-16
Gradle verify

hannesa2:


hannesa2:
What prevents someone to push the approve button here ?
And as second step to push the merge button ?

Ok this is not important, it **just** security. It destroys nothing. 
But no action since 20 days is an indicator.

jesmrec:
Nothing to check here. When CI passes, we can merge.

Issue #2799, opened 2020-01-27
[BUG] Crash trying to move or copy a file into another one or selecting the folder for camera uploads


abelgardep:
### Actual behaviour
- Crash trying to move or copy a file into another one. It also occurs when selecting the folder for camera uploads feature.

### Expected behaviour
- Not crash
 
### Steps to reproduce
1. Select a file
2. Select move/copy
3. Try to move/copy to a file which is not a folder

![ezgif com-resize](https://user-images.githubusercontent.com/47524927/73178112-e249f000-4110-11ea-99a9-81897c158246.gif)

Can this problem be reproduced with the official owncloud server? Yes
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Device model: Pixel 3XL v10

#### ownCloud log (data/owncloud.log)
2020-01-27 14:16:45.923 26557-26557/com.owncloud.android.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.owncloud.android.debug, PID: 26557
    java.lang.ClassCastException: com.owncloud.android.ui.activity.FolderPickerActivity cannot be cast to com.owncloud.android.ui.activity.FileDisplayActivity
        at com.owncloud.android.ui.fragment.OCFileListFragment.onItemClick(OCFileListFragment.java:770)
        at android.widget.AdapterView.performItemClick(AdapterView.java:330)
        at android.widget.AbsListView.performItemClick(AbsListView.java:1190)
        at android.widget.AbsListView$PerformClick.run(AbsListView.java:3198)
        at android.widget.AbsListView.onTouchUp(AbsListView.java:4128)
        at android.widget.AbsListView.onTouchEvent(AbsListView.java:3886)
        at android.view.View.dispatchTouchEvent(View.java:13953)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3054)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2741)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3060)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2755)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3060)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2755)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3060)
[...]

PR #2820, opened 2020-02-19
Fix crash when moving/copying files to a folder and select a file

davigonz:
Fixes https://github.com/owncloud/android/issues/2799


jesmrec:
Tested with `Move` and `Copy` operations.

Correctly fixed

PR #2816, opened 2020-02-16
Github CI with action

hannesa2:
I know, currently there are no Espresso tests, but it's simple and fast.

PR #2832, opened 2020-03-05
Update libs

hannesa2:
It's only to be up to date and raise problems as soon an possible. 
Waiting to long makes things not easier.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,10 +7,10 @@ buildscript {
 
         // Android jetpack
         archLifecycleVersion = "2.1.0"
-        roomVersion = "2.2.2"
+        roomVersion = "2.2.4"
         appCompat = "1.1.0"
-        sqliteVersion = "2.0.1"
-        lifecycleLiveData = "2.0.0"
+        sqliteVersion = "2.1.0"
+        lifecycleLiveData = "2.2.0"
         biometricVersion="1.0.1"
 
         // Kotlin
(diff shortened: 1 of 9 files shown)


hannesa2:
Bitrise ! What should I say ?
image

image

And to be honest, I'm not interested in having an additional Bitrise account.
I made you a better solution https://github.com/owncloud/android/pull/2816 but this Espresso test are flake and not trust able like they are in current state

hannesa2:
And when I do local a `./gradlew :owncloudApp:lint` on master I run into 
image

This shows me, that you probably not perform a lint check in Bitrise. But I can't verify 👎 
The solution is #2816 and to be honest iOS and Android don't need always the same tools. You only collect the disadvantages from both systems. You could throw this paradigm to the waste basket @michaelstingl

abelgardep:
Hi @hannesa2, first of all, thanks for your contributions.

We perform a lint check in Bitrise, that's why this pr failed.
I need to research a little bit about GitHub Actions to check if it is viable. At the moment, bitrise is working for us, so we won't change if github actions don't.

jesmrec:
I have performed a smoke test over the branch and worked fine. Any additional suggestion for testing the stuff here @abelgardep ?

abelgardep:
Not at all. Just updating libraries and fixing lint errors.

jesmrec:
Ok, then it is approved, Please fix conflicts and move forward

abelgardep:
Merged, thanks @hannesa2 👍

PR #2833, opened 2020-03-06
Fix lifecycle owner

hannesa2:


davigonz:
Hi @hannesa2 , thanks for this contribution. A short description about what you are doing here would be appreciated.

Anyway, I've read about viewLifecycleOwner and maybe we do not need to use [Event](https://github.com/owncloud/android/blob/master/owncloudDomain/src/main/java/com/owncloud/android/domain/utils/Event.kt) anymore to deal with double observation issues when rotating the screen for instance, what do you think?

hannesa2:
> A short description about what you are doing here would be appreciated.

In code you see, you should use the `LifecycleOwner`
image

The other stuff are lint errors which raise/find on Android Studio 3.6.1
I cherry-picked them from my fork, where I'm already working with Android Studio 3.6.1

Concerning [Event](https://github.com/owncloud/android/blob/master/owncloudDomain/src/main/java/com/owncloud/android/domain/utils/Event.kt) give me some time, to understand what are you doing here

abelgardep:
Could you split this PR? Im not sure about lifecycle at the moment since it involves fragments and dialog fragments, it may need some research. But it would be great to merge those lint warnings soon. Thanks

hannesa2:
I removed the livecycle owner, and resulting is a simple lint fix
A: the title of branch is wrong now
B: This conflict is not right. I rebased on recent master. (I observed this sometimes in the project, I'm working for with an own hosted github)

So I close this and open a new PR with this simple lint fix 


image

Issue #2787, opened 2020-01-10
Problems found while migrating CERNBox to 2.13.1


diocas:
Hi,

We're working on the release of 2.13.1 to our CERN community and there were a couple of issues that we found.
Please keep in mind that we have our own implementation of the backend (REVA), which differs from the OC10 server, so some of these issues are not bugs _per se_ but what we see as limitations.
I can open some of them in separate issues if you prefer.

## 1. App crashes looking for parent
If I have a folder, I would expect dav to return `httpd/unix-directory` as `d:getcontenttype`. This is actually something that is present in the new REVA/OCIS ([propfind.go#L230](https://github.com/cs3org/reva/blob/e56f28d27e3d0aaedc3d193e4415c0248d0c38b3/internal/http/services/owncloud/ocdav/propfind.go#L230)).
But, if I return this, the app crashes [here](https://github.com/owncloud/android/blob/88089179b0f10b491174524850bac5e4874a6a2d/owncloudApp/src/main/java/com/owncloud/android/datamodel/OCFile.java#L517) with `parentPath` null.

## 2. App fails to load shares
Due to our storage infrastructure, where files might live in different storage instances and fileIDs are not shared between them,  `item_source` and `file_source` have the format `InstanceName:FileID`, for example `home:123456`. This works for both the web and sync clients.
But on the mobile app, when getting the shares of a file (`GET ocs/v2.php/apps/files_sharing/api/v1/shares`), the app gives an error and displays nothing. This happens because it tries to convert file/item source into `Long` ([ShareXMLParser.kt#L246](https://github.com/owncloud/android-library/blob/9a806469c5a0289494fbf8dae3491d632ae8c53e/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt#L246) and [ShareXMLParser.kt#L263](https://github.com/owncloud/android-library/blob/9a806469c5a0289494fbf8dae3491d632ae8c53e/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/ShareXMLParser.kt#L263))

## 3. App uses the old webdav endpoint
When we released 2.13 into our beta channel, our testers (including me) were updated from the version 2.6.0. Everything seemed to be working fine, until someone tried to install the app in a fresh device (and when I removed my account and added it again).
[...]

PR #2837, opened 2020-03-10
Parse file and item source as string

abelgardep:
Implements (Point 2) https://github.com/owncloud/android/issues/2787
Needs: https://github.com/owncloud/android-library/pull/312

### DIFF
--- a/owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
+++ b/owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
@@ -19,6 +19,7 @@
  */
 package com.owncloud.android.data.roommigrations
 
+import androidx.room.migration.Migration
 import androidx.room.testing.MigrationTestHelper
 import androidx.sqlite.db.SupportSQLiteDatabase
 import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
@@ -38,11 +39,28 @@ open class MigrationTest {
     fun getCount(db: SupportSQLiteDatabase, tableName: String): Long =
         db.compileStatement("SELECT COUNT(*) FROM `$tableName`").simpleQueryForLong()
 
+    fun performMigrationTest(
+        previousVersion: Int,
+        currentVersion: Int,
+        insertData: (SupportSQLiteDatabase) -> Unit,
+        recoverPreviousData: Boolean = true,
+        validateMigration: (SupportSQLiteDatabase) -> Unit,
+        listOfMigrations: Array<Migration>
+    ) {
+        helper.createDatabase(TEST_DB_NAME, previousVersion).run {
+            if (recoverPreviousData) insertData(this)
+        }
+
+        helper.runMigrationsAndValidate(
+            TEST_DB_NAME, currentVersion, true, *listOfMigrations
+        ).also { validateMigration(it) }
+    }
+
     companion object {
         const val TEST_DB_NAME = "migration-test"
 
         const val DB_VERSION_27 = 27
-        const val DB_VERSION_29 = 29
         const val DB_VERSION_28 = 28
+        const val DB_VERSION_29 = 29
     }
 }
--- a/owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
+++ b/owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
@@ -1,0 +1,48 @@
+/**
+ *   ownCloud Android client application
+ *
+ *   @author Abel García de Prada
+ *   Copyright (C) 2020 ownCloud GmbH.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License version 2,
+ *   as published by the Free Software Foundation.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        FrameworkSQLiteOpenHelperFactory()
+    )
+
+    fun getCount(db: SupportSQLiteDatabase, tableName: String): Long =
+        db.compileStatement("SELECT COUNT(*) FROM `$tableName`").simpleQueryForLong()
+
+    companion object {
+        const val TEST_DB_NAME = "migration-test"
+
+        const val DB_VERSION_27 = 27
```
davigonz: Why `DB_VERSION_29` is before `DB_VERSION_28`

owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
```diff
@@ -0,0 +1,48 @@
...
+package com.owncloud.android.data.roommigrations
+
+import androidx.room.testing.MigrationTestHelper
+import androidx.sqlite.db.SupportSQLiteDatabase
+import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
+import androidx.test.platform.app.InstrumentationRegistry
+import com.owncloud.android.data.OwncloudDatabase
+import org.junit.Rule
+
+open class MigrationTest {
```
davigonz: Love this parent class for migration tests ❤️

owncloudData/src/main/java/com/owncloud/android/data/sharing/shares/db/OCShareEntity.kt
```diff
@@ -115,5 +113,26 @@ data class OCShareEntity(
                 values.getAsString(ProviderTableMeta.OCSHARES_URL)
             )
         }
+
+        fun toContentValues(entity: OCShareEntity): ContentValues =
```
davigonz: Why do you create this function inside companion object, as a static method? This function is meant to transform an instance of this class so passing it as a parameter would not be needed, you could just create the function as an instance method and call it using `ocShareEntity.toContentValues()`

owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationToDB29.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        helper.runMigrationsAndValidate(
+            TEST_DB_NAME, DB_VERSION_29, true, *OwncloudDatabase.ALL_MIGRATIONS
+        ).also { validateMigrationTo29(it) }
+    }
+
+    @Test
+    fun migrationFrom28To29_containsCorrectData() {
+        helper.createDatabase(TEST_DB_NAME, DB_VERSION_28).run {
```
davigonz: This test and the previous one are almost the same, what about creating a function with the similar code?

owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationToDB28.kt
```diff
@@ -37,159 +33,56 @@ import com.owncloud.android.data.ProviderMeta.ProviderTableMeta.CAPABILITIES_VER
...
-
-        helper.closeWhenFinished(database)
-        return database
     }
 
-    private fun validateMigrationTo28() {
-        val dbCapability =
-            getMigratedRoomDatabase().capabilityDao().getCapabilitiesForAccount(OC_CAPABILITY.accountName!!)
-        with(dbCapability) {
-            assertEquals(OC_CAPABILITY.accountName, accountName)
```
davigonz: All these checks are no longer needed?

owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationTest.kt
abelgardep: My bad, i'll order them

owncloudData/src/androidTest/java/com/owncloud/android/data/roommigrations/MigrationToDB28.kt
abelgardep: I had some problems using Dao in migration tests, so i kept it simple.

If migration went well, the database's number of rows should be the same as it was before the migration.

owncloudData/src/main/java/com/owncloud/android/data/sharing/shares/db/OCShareEntity.kt
abelgardep: You mean creating an extension `OCShareEntity.toContentValues()` yes, it could be a good alternative

davigonz: It's not about extensions, for example in Kotlin would you convert an Int to String like below?

```
val number : Int = 2

Int.toString(number)
```

No, right? You would use `number.toString()` and this is because `toString()` is an instance method of `Int` not a class one.

jesmrec:
In this PR we do not have a new feature implementation, and the regular behaviour must keep. So, we will go for sanity checks:

- [ ] Creation of private shares
- [ ] Creation of public shares
- [ ] Edition of private shares
- [ ] Edition of public shares
- [ ] Deletion of private shares
- [ ] Deletion of public shares
- [ ] Fetching private shares
- [ ] Fetching public shares
- [ ] Upgrade from latest version

jesmrec:
From my side it is approved.

PR #2844, opened 2020-04-07
fix code style

hannesa2:
Sorry, I accidentally added an old markup plugin setting.
This should be removed on a more recent version of the plugin


jesmrec:
Nothing to do from my side. Go ahead @abelgardep

hannesa2:
Without it, every developer has a pointless git change locally.

What prevents a merge of it ?

Issue #2633, opened 2019-08-20
[New arch] OpenID connect implementation


davigonz:
### TASKS

- [x] Check [AppAuth](https://github.com/openid/AppAuth-Android)
- [x] Create `open_id_support` branch 
- [x] Development tasks
  - [x] Use `openid` as scope value
  - [x] Use service discovery or manual endpoints depending on a setup parameter
  - [x] Polish and clean the code
  - [x] Multiaccount
  - [x] Check flow (Oauth2 view or not)
  - [x] Refresh token
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge `open_id_support` branch into master

### Environments

- +Info in https://github.com/owncloud/openidconnect#openid-connect-for-owncloud
- Easily fire up an owncloud server with an open id connect provider https://github.com/owncloud/openidconnect/tree/master/tests/docker

### PRs

- App: https://github.com/owncloud/android/pull/2854
- Library (if needed): https://github.com/owncloud/android-library/pull/315
- iOS solution: https://github.com/owncloud/ios-app/pull/401

michaelstingl:
I'd propose to perform the following checks:

1) Login to https://ocis-latest.owncloud.com/ (`einstein:relativity`)
2) Survive token refresh after 10 Minutes (600s)

PR #2854, opened 2020-04-17
Open Id Connect along with new arch in login

davigonz:
Implements owncloud/android#2633
Needs https://github.com/owncloud/android-library/pull/315

----------------

### QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.15/2854-oidc.md

### BUGS AND IMPROVEMENTS
- [x] Fix 401 when connecting to oidc server 2.
- [ ] (1) Status message including OAuth2 https://github.com/owncloud/android/pull/2854#issuecomment-619942565


owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthUtils.kt
```diff
@@ -0,0 +1,89 @@
...
+    companion object {
+        fun buildOIDCAuthorizationServiceConfig(
+            context: Context,
+            serverBaseUrl: String,
+            onGetAuthorizationServiceConfiguration: RetrieveConfigurationCallback
+        ) {
+            Timber.d("OIDC, getting the auth and token endpoints from the discovery document (well-known)")
+
+            val urlPathAfterProtocolIndex = serverBaseUrl.indexOf(
+                PATH_SEPARATOR, serverBaseUrl.indexOf(PATH_SEPARATOR) + 2
```
abelgardep, line 42: Why +2 ?

owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -316,30 +318,40 @@ private void refreshToken(
...
+            Timber.d("No authorization configuration found, falling back to hardcoded oauth2 endpoints");
+            // The code below is for users (already logged in) updating the app from previous versions, which do not have
+            // an authState that is configured when doing a fresh log in
+            String baseUrl = accountManager.getUserData(
+                    account,
+                    AccountUtils.Constants.KEY_OC_BASE_URL
+            );
+            authorizationServiceConfiguration = new AuthorizationServiceConfiguration(
+                    Uri.parse(baseUrl + "/" + mContext.getString(R.string.oauth2_url_endpoint_auth)), // auth endpoint
+                    Uri.parse(baseUrl + "/" + mContext.getString(R.string.oauth2_url_endpoint_access)) // token endpoint
```
abelgardep, line 335: In OAuthUtils you use PATH_SEPARATOR, why don't you use it here?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -347,26 +367,79 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
         server_status_text.run {
             setCompoundDrawablesWithIntrinsicBounds(R.drawable.progress_small, 0, 0, 0)
             text = resources.getString(R.string.oauth_login_connection)
         }
 
-        mAuthService = OAuthUtils.createAuthorizationService(this)
+        val retrieveConfigurationCallback =
+            RetrieveConfigurationCallback { serviceConfiguration, exception ->
+                if (exception != null) {
+                    Timber.e(exception, "OIDC failed, try with normal OAuth");
```
abelgardep, line 382: Could you remove that `;` ? This is kotlin 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -347,26 +367,79 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+    /**
+     * OAuth step 1: Get authorization code
+     * If OIDC is not available, falling back to normal OAuth
+     */
+    private fun startNormalOauthorization() {
+        val retrieveConfigurationCallback =
+            RetrieveConfigurationCallback { serviceConfiguration, exception ->
+                if (exception != null) {
+                    Timber.e(exception, "Normal OAuth failed");
+                    updateOAuthStatusIconAndText(exception);
```
abelgardep, line 407: Same here, could you remove every `;` from this file?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -56,6 +56,8 @@ import com.owncloud.android.extensions.parseError
 import com.owncloud.android.extensions.showErrorInToast
 import com.owncloud.android.lib.common.accounts.AccountTypeUtils
 import com.owncloud.android.lib.common.accounts.AccountUtils
+import com.owncloud.android.lib.common.accounts.AccountUtils.Constants.OAUTH2_OIDC_SCOPE
+import com.owncloud.android.lib.common.authentication.oauth.OAuthConnectionBuilder
```
abelgardep, line 60: The idea of new arch is to remove dependencies between app and library module. Why are you adding constants in the library that are only used at app module?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -347,26 +367,79 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+            serverBaseUrl,
+            retrieveConfigurationCallback
+        )
+    }
+
+    /**
+     * OAuth step 1: Get authorization code
+     * If OIDC is not available, falling back to normal OAuth
+     */
+    private fun startNormalOauthorization() {
```
abelgardep, line 402: This method is pretty similar to `startOIDCOauthorization` can't they merge into one?

BTW, i cannot see here a `server_status_text` update, user can't see what's happening, can he?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -347,26 +367,79 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
     /**
      * OAuth step 1: Get authorization code
+     * Firstly, try the OAuth authorization with Open Id Connect, checking whether there's an available .well-known url
+     * to use or not
      */
-    private fun startOAuthorization() {
+    private fun startOIDCOauthorization() {
         server_status_text.run {
             setCompoundDrawablesWithIntrinsicBounds(R.drawable.progress_small, 0, 0, 0)
             text = resources.getString(R.string.oauth_login_connection)
```
abelgardep, line 376: I think we should have different messages for oauth and oidc, what do you think?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -403,11 +476,11 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
         TokenResponseCallback { tokenResponse: TokenResponse?, authorizationException: AuthorizationException? ->
             if (tokenResponse?.accessToken != null && tokenResponse.refreshToken != null) {
                 authenticationViewModel.loginOAuth(
-                    tokenResponse.additionalParameters[AccountUtils.Constants.KEY_USER_ID]!!,
+                    tokenResponse.additionalParameters[AccountUtils.Constants.KEY_USER_ID]?.let { it } ?: "",
```
abelgardep, line 479: You can remove that `let` and use simply `?:`, can't you?

owncloudDomain/src/main/java/com/owncloud/android/domain/authentication/usecases/LoginOAuthAsyncUseCase.kt
```diff
@@ -29,7 +29,6 @@ class LoginOAuthAsyncUseCase(
 
     override fun run(params: Params): String {
         require(params.serverInfo != null) { "Invalid server info" }
-        require(params.username.isNotEmpty()) { "Invalid username" }
```
abelgardep, line 32: Just curiosity, why username is not required?

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthUtils.kt
davigonz: You have a comment just below that with the explanation

 // OIDC Service Discovery Location is placed in urls like https://whatever and not https://whatever/others, so remove those subpaths

+2 refers to the first slash after `https://` , i.e. https://whatever**/**others

owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
davigonz: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: tenor

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: travolta-despistado-300x242

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthUtils.kt
```diff
@@ -0,0 +1,89 @@
...
+        fun buildOAuthorizationServiceConfig(
+            context: Context,
+            onGetAuthorizationServiceConfiguration: RetrieveConfigurationCallback,
+            baseUrl: String = ""
+        ) {
+            Timber.d("Trying normal OAuth instead")
+
+            val authorizationServiceConfiguration = AuthorizationServiceConfiguration(
+                Uri.parse(
+                    "$baseUrl/${context.getString(R.string.oauth2_url_endpoint_auth)}" // auth endpoint
```
abelgardep, line 75: BTW, if you change [this](https://github.com/owncloud/android/pull/2854/files#r412772570) you could change it here and below too. `"$baseUrl$PATH_SEPARATOR${context.getString(R.string.oauth2_url_endpoint_auth)}"`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: That's is the ideal scenario but since the AppAuth library is tied to the presentation layer since it takes care of all the OAuth process, including the UI stuff... I need these dependencies here...

You might have thought of moving those dependencies (constants and methods) to the presentation layer but I'd rather include stuff like http urls, ssl, trustmanager as least as possible in the presentation layer

abelgardep: I understand it, but i think that as app auth is an exception, its constants and methods should be included in app project. That way, we will avoid creating a constant in a submodule that will be only used in app project.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: > This method is pretty similar to startOIDCOauthorization

Not really, take a deeper look, the lines that are exactly the same, are just these three:

oidcSupported = true
performGetAuthorizationCodeRequest(serviceConfiguration)
authorizationServiceConfiguration = serviceConfiguration

And I've just realized `oidcSupported = true` should not be included in the normal OAuth case, copying-pasting issue

Other checks or call methods seem the same but are different either on the params used or the call itself.

> BTW, i cannot see here a server_status_text update, user can't see what's happening, can he?

We always try the oidc connection before calling this method, so the status update takes place in the startOIDCOauthorization method.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: I thought about and I kept it simple, OIDC is OAuth as well so there's nothing weird on using connecting to OAuth server, as well as we do not know if the server we are connecting to is OIDC or not at this point, we just have a bearer authentication header and it could be normal OAuth or OIDC till moving a bit forward, when we say **oidcSupported = true**

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: Sure 🤦

owncloudDomain/src/main/java/com/owncloud/android/domain/authentication/usecases/LoginOAuthAsyncUseCase.kt
davigonz: Because of the line of your previous comment 😄 

`tokenResponse.additionalParameters[AccountUtils.Constants.KEY_USER_ID]?.let { it } ?: "",`

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthUtils.kt
davigonz: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: I could do it with OAuthConnectionBuilder but not with `OAUTH2_OIDC_SCOPE` since `KEY_OAUTH2_SCOPE` is used in data layer.

owncloudDomain/src/main/java/com/owncloud/android/domain/authentication/usecases/LoginOAuthAsyncUseCase.kt
abelgardep: Yes, i mean why is not required now and it was required previously

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
abelgardep: A question related with oidcSupported, how is it handled with configuration changes?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: By the way, we would need to recheck more dependencies like the ones used in the code below

```
userAccount?.let {
    AccountUtils.getUsernameForAccount(it)?.let { username ->
        account_username.setText(username)
    }
}
```

abelgardep: If oauth is not handled by library, i would move oauth stuff to data layer. What do you think?

davigonz: Yep, let's move the constants I need to use in this PR but we still have other ones related to OAuth in AccountUtils, that are being used during synchronization, we will check it soon.

owncloudDomain/src/main/java/com/owncloud/android/domain/authentication/usecases/LoginOAuthAsyncUseCase.kt
davigonz: Because we always had a username before... it's something that appeared with oidc servers

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
davigonz: Well, oidcSupported is checked just after setting its value to true, as soon as the user press the log in button, it's a matter of milliseconds I would say, if you are able to tap the button and rotate the device in that time... I wasn't , maybe @jesmrec could manage to break it that way, let's see in QA stage

jesmrec:
### (1)

1. Type the OIDC url
2. URL is validated, then click on "Log In"

Current: `Connecting to OAuth2 server`.
Expected: Server is not OAuth2, so the message should be different. Specifying the OIDC server would be the best solution, if not do-able, we can skip the word `OAuth2` and let `Connectng to the server`. (more ideas welcome).

Android 10

jesmrec:
Some interesting points about the testing:

1) For the moment, there is no way to expire tokens manually and no way to change expiration time (no option in `konnectd` for it). So all the renewals are automatic after 10 minutes, default value. It was suggested to oCIS as improvement (hopefully it will be added to the OCIS cli)

2) Log out from idP is not posible. 

3) OIDC login process is based on service discovery (all sensible information is fetched from server's `well-know/openid-configuration` file). So we are assuming that the information fetched from the server is correct (correct endpoints, supported encryption algs etc...). I did not spend time on checking fields there. In case of OCIS, some of them are setup-able via environment variables, but will need more research and maybe is not the scope here. Let me know if i'm wrong.

4) Tests were mainly done over oCIS instances (public instance and local as well), in which front we have OIDC. OC10 with OIDC was tested with the known internal instance, since there were problems to create a local one. Such problems were reported in the compose playground repository some weeks ago.

5) Checked https://github.com/owncloud/enterprise/issues/3864 as well. Reported a comment about the Android behaviour: https://github.com/owncloud/enterprise/issues/3864#issuecomment-621065730 

If you are OK at this point, we can move forward. If not, please share your thoughts. 

@michaelstingl @hosy @abelgardep @davigonz

michaelstingl:
`(1)` is not a showstopper. OAuth 2.0 wording is okay for now.

jesmrec:
Then we can move it forward. Work will be iterative here. Let's wait for more feedback. 

Approved.

Issue #2857, opened 2020-04-20
[BUG] Crash when changing orientation in some operations


jesmrec:
### Steps to reproduce
1.  Select an item in the file list
2. Select copy or move operation
3. When the folder picker is displayed, change orientation

### Actual behaviour
App crashes

### Expected behaviour
App does not crash

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Android version: 10

Device model: Pixel 2

Stock or customized system: Stock

ownCloud app version: 2.14.2

ownCloud server version: 10.4

davigonz:
@jesmrec ready to test

PR #2859, opened 2020-04-21
Fix npe when rotating folder picker activity

abelgardep:
Implements #2857

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
@@ -144,7 +144,8 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     }
 
     private void createFragments() {
-        OCFileListFragment listOfFiles =  OCFileListFragment.newInstance(true, FileListOption.ALL_FILES, false, true, false);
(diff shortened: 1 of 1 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -162,8 +161,10 @@ private void setBackgroundText() {
             if (!mSyncInProgress) {
                 // In case folder list is empty
                 message = R.string.file_list_empty_moving;
-                listFragment.getProgressBar().setVisibility(View.GONE);
-                listFragment.getShadowView().setVisibility(View.VISIBLE);
+                if (listFragment.getProgressBar() != null && listFragment.getShadowView() != null) {
```
davigonz, line 164: Is it possible that `listFragment.getProgressBar()` is not null while `listFragment.getShadowView()` is null and we are not properly setting the progressBar visibility in that case?

abelgardep: I would say that it is not posible at all. Both are included in the same layout file (`top_progress_bar`) and are initialized at the same time.

jesmrec:
Crash is not reproducible anymore. Fixed.

Issue #2716, opened 2019-11-13
[New arch] Use Moshi to parse capabilities and get rid of RemoteCapability in library?


davigonz:
[`GetRemoteCapabilitiesOperation.kt`](https://github.com/owncloud/android-library/blob/master/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/GetRemoteCapabilitiesOperation.kt) takes care of parsing the capabilities retrieved from the server in JSON format and transform them into RemoteCapability object.

Similarly to what we detailed in https://github.com/owncloud/android/issues/2704, we could use Moshi to make the parsing of capabilities.

#### IN ADDITION

Since RemoteCapability object and OCCapability (domain model)  objects are going to represent basically the same (when [modularization](https://github.com/owncloud/android/issues/2607) is finished), we could get rid of `RemoteCapability` and parse the JSON directly to `OCCapability` from the data layer.

Parsing JSON and transforming it to domain models in data layer is the most common scenario when connecting with external APIs but ownCloud for Android is a special case, we have an intermediate layer between the oC server and our app, which is the ownCloud Android Library. This library is meant to make the life easier to those interested in communicate an Android app with ownCloud servers. 

If we remove `RemoteCapability` from the ownCloud Android Library, the apps using it will receive just JSON and would need to parse it, so maybe is not the best idea if the aim of the library is "integrating your Android application with ownCloud seamlessly." 🤔 

Tasks:

- [ ] Move capabilities parsing from library to data layer, that will parse the JSON and transform it into OCCapability object. This OCCapability object is the domain model that will be retrieved from the data layer.
- [ ] Get rid of RemoteCapability class in library since is basically the same as OCCapability

### PR
- App: #2860 
- Library: https://github.com/owncloud/android-library/pull/298

abelgardep:
### Capabilities
It would be good to know which capabilities are mandatory and are always sent to the client, and which are optional.

### Unused capabilities
There are two capabilities that are not being used anywhere. Remove them?
filesSharingPublicSendMail
filesSharingUserSendMail

### RemoteCapability
It is a simple facade of OCCapability inside Library. From my point of view, it can be removed. This way, the operation would return a CapabilityResponse which would be mapped to model(OCCapability) at OCRemoteCapabilitiesDataSource

davigonz:
@jesmrec ready to test

abelgardep:
> ### Unused capabilities
> 
[...]

PR #2860, opened 2020-04-22
Use moshi to parse capabilities (Library Update)

abelgardep:
Implement #2716 
Needs https://github.com/owncloud/android-library/pull/298

### DIFF
--- a/owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_28.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_28.kt
@@ -1,3 +1,21 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Abel García de Prada
+ * Copyright (C) 2020 ownCloud GmbH.
+ *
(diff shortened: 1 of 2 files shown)


jesmrec:
Checking a set of capabilities to find out regressions:

- [X] Allow apps to use the Share API
- [X] Allow users to share via link
- [X] Allow public uploads
- [X] Enforce password protection for read-only links
- [X] Enforce password protection for read + write links
- [X] Enforce password protection for upload-only (File Drop) links
- [X] Set default expiration date
- [X] Expire after X days
- [X] Enforce as maximum expiration date
- [X] Allow resharing
- [X] Allow sharing with groups
- [X] Restrict users to only share with users in their groups
- [X] Restrict users to only share with groups they are member of
- [X] Exclude groups from sharing
- [X] Extra field to display in autocomplete results
- [X] Federated
- [X] Check version

Capabilities not implemented yet (to open new issues):

- Default permissions (set of permissions by default in private shares, now is always 31 instead of the capability value)
- Permission "read/write/delete" in public shares for folders

From my side, we can move forward.

Issue #2827, opened 2020-02-27
[New arch] Increase test coverage


abelgardep:
We are working on the new architecture, and meanwhile we are creating tests. This way, it will be easier to debug and find potential bugs before merging a PR, or releasing a new version.

Android Studio let us know how many test coverage we have. But only for unit tests. We can check this coverage performing next action:

![how_to](https://user-images.githubusercontent.com/47524927/75443103-03e10600-5961-11ea-9590-c12112745ce2.png)

Once all tests finish, a new tab appears and shows the test coverage in that module.
For example, this would be the test coverage for domain module at the moment. 

### Domain module
This module can be tested completely using unit tests.
Coverage Result: 74% lines and 40% classes
We can improve this coverage a lot testing those exceptions. 

![domain](https://user-images.githubusercontent.com/47524927/75443097-017eac00-5961-11ea-9385-d2554bd6d9c4.png)

### Mobile module
Since this module is where old architecture resides, and most of the tests here are instrumented, the test coverage is almost 0%.

**But** if we move to the package where we have started working on new architecture, we can see that those new viewModels have a good test coverage.

![mobile](https://user-images.githubusercontent.com/47524927/75443101-02afd900-5961-11ea-993f-04f7da234b80.png)

### Data module
This module is a little bit tricky for test coverage. Since here we have some tests that are instrumented.
Result: 14% lines and 30% classes coverage

![data](https://user-images.githubusercontent.com/47524927/75443106-04799c80-5961-11ea-8643-df7281ba0041.png)

**But** we can move to shares package for example.
There, we can see that excluding db package where instrumented tests are, unit tests coverage is decent.

![data_shares](https://user-images.githubusercontent.com/47524927/75443109-05123300-5961-11ea-8926-74fb4ba0b4b3.png)

### To sum up
We have a decent unit test coverage for the new architecture modules but we can easily detect where we can increase it. The goal would be to have around 100% test coverage with unit tests.

PR #2868, opened 2020-04-29
Add some pending tests and general updates

abelgardep:
Add some pending tests related with #2827

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/sharing/fragments/ShareFileFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/sharing/fragments/ShareFileFragment.kt
@@ -91,7 +91,7 @@ class ShareFileFragment : Fragment(), ShareUserListAdapter.ShareUserAdapterListe
     /**
      * List of private shares bound to the file
      */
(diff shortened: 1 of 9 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/sharing/fragments/ShareFileFragment.kt
```diff
@@ -91,7 +91,7 @@ class ShareFileFragment : Fragment(), ShareUserListAdapter.ShareUserAdapterListe
     /**
      * List of private shares bound to the file
      */
-    private var privateShares: List<OCShare>? = null
+    private var privateShares: List<OCShare> = listOf()
```
davigonz, line 94: 🚀

davigonz:
Approved, please rebase before starting with QA

jesmrec:
as CI is passing with all the tests included here, not sure if there are something else to test. I'd say no, but...

abelgardep:
@jesmrec I added some pending tests, and updated some potential nullability problems related with shares. That would be the only thing to test here.

jesmrec:
will take the final testing in regression stage.

Issue #2851, opened 2020-04-15
[New Arch] Get Avatar


abelgardep:
Related with #2849, there are some missing operations in new architecture.
First is Get Quota Operation(#2850), next is this one(Get Avatar).

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/avatar
- [x] Development tasks
  - [x] Create GetAvatarForAccountUseCase
  - [x] Add getAvatarForAccount functions to UserRepository, UserDatasources and UserService
  - [x] Migrate RemoteOperation to kotlin
  - [x] Perform cleanup
  - [x] Call GetAvatarForAccountUseCase instead of GetRemoteUserAvatarOperation
  - [x] Remove GetUserProfileOperation if possible. Call usecases instead. (Intermediate step: Created a SyncProfileOperation)
  - [x] Add unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch new_arch/avatar into master

### PR

- App https://github.com/owncloud/android/pull/2893
- Library (if needed) https://github.com/owncloud/android-library/pull/324

abelgardep:
We are fetching Avatar each time we perform a refresh root folder operation.

Depending on user's avatar, it would be more or less data transferred. Let's set 29kb of image for example. That's like 29kb each time we refresh root folder.

How important is to keep avatar updated on each refresh?

jesmrec:
Not much. For me, it'd be enough to check every time you enter the account.

abelgardep:
BTW on each root folder refresh, oC performs next operations:
Is it needed a full profile sync on every root folder refresh? It take time and data

### Get User info(GetRemoteUserInfoOperation)
- Endpoint: /ocs/v2.php/cloud/user?format=json
- Size: ± 150B
### Get Avatar (GetRemoteUserAvatarOperation)
- Endpoint: /index.php/avatar/demo/448
- Size: Depending on avatar ± 20kb
### Get Quota (GetRemoteUserQuotaOperation)
- Endpoint: /remote.php/dav/files/demo/ (Same as refresh folder but with exclusive properties)
  - "quota-available-bytes"
  - "quota-used-bytes"
- Size: ±350B

### Refresh folder(ReadRemoteFolderOperation)
- Endpoint: /remote.php/dav/files/demo/ (Fetch all properties)
- Size: depending on files. Let's set 6Kb

### Fetch shares (GetRemoteSharesForFileOperation)
[...]

Issue #2850, opened 2020-04-15
[New Arch] Get Quota


abelgardep:
Related with #2849, there are some missing operations in new architecture.
First of them is Get Quota Operation.

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/get_quota
- [x] Development tasks
  - [x] Create GetQuotaForAccountUseCase
  - [x] Add getQuotaForAccount functions to UserRepository, UserDatasources and UserService
  - [x] Migrate RemoteOperation to kotlin
  - [x] Call GetQuotaForAccountUseCase instead of GetRemoteUserQuotaOperation
  - [x] Add unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch new_arch/get_quota into master

### PR

- App #2880
- Library (if needed) https://github.com/owncloud/android-library/pull/320

PR #2880, opened 2020-05-12
[New Arch] Quota

abelgardep:
Retrieve user quota using new architecture. It still requires a cleanup that will be done within #2851
Implements #2850 
Needs https://github.com/owncloud/android-library/pull/320

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RepositoryModule.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RepositoryModule.kt
@@ -21,31 +21,26 @@ package com.owncloud.android.dependecyinjection
 
 import com.owncloud.android.data.authentication.repository.OCAuthenticationRepository
 import com.owncloud.android.data.capabilities.repository.OCCapabilityRepository
-import com.owncloud.android.data.server.repository.OCServerInfoRepository
 import com.owncloud.android.data.files.repository.OCFileRepository
+import com.owncloud.android.data.server.repository.OCServerInfoRepository
 import com.owncloud.android.data.sharing.sharees.repository.OCShareeRepository
 import com.owncloud.android.data.sharing.shares.repository.OCShareRepository
 import com.owncloud.android.data.user.repository.OCUserRepository
 import com.owncloud.android.domain.authentication.AuthenticationRepository
 import com.owncloud.android.domain.capabilities.CapabilityRepository
-import com.owncloud.android.domain.server.ServerInfoRepository
 import com.owncloud.android.domain.files.FileRepository
+import com.owncloud.android.domain.server.ServerInfoRepository
 import com.owncloud.android.domain.sharing.sharees.ShareeRepository
 import com.owncloud.android.domain.sharing.shares.ShareRepository
 import com.owncloud.android.domain.user.UserRepository
 import org.koin.dsl.module
 
 val repositoryModule = module {
-    factory<CapabilityRepository> {
-        OCCapabilityRepository(
-            get(),
-            get()
-        )
-    }
-    factory<ShareRepository> { OCShareRepository(get(), get()) }
-    factory<ShareeRepository> { OCShareeRepository(get()) }
-    factory<UserRepository> { OCUserRepository(get()) }
+    factory<AuthenticationRepository> { OCAuthenticationRepository(get(), get()) }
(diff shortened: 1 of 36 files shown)


owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RepositoryModule.kt
```diff
@@ -21,31 +21,26 @@ package com.owncloud.android.dependecyinjection
...
-    factory<CapabilityRepository> {
-        OCCapabilityRepository(
-            get(),
-            get()
-        )
-    }
-    factory<ShareRepository> { OCShareRepository(get(), get()) }
-    factory<ShareeRepository> { OCShareeRepository(get()) }
-    factory<UserRepository> { OCUserRepository(get()) }
+    factory<AuthenticationRepository> { OCAuthenticationRepository(get(), get()) }
```
davigonz, line 39: You are following alphabetical order, right? If so, have a look at https://github.com/owncloud/android/pull/2880/files#diff-fe82a8d4a657ef2a30b6a669c7d30015R45 and the line below that one

owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_31.kt
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+package com.owncloud.android.data.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val MIGRATION_30_31 = object : Migration(30, 31) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        //Nothing to migrate at the moment
```
davigonz, line 27: You could use // TODO and see it from Android studio bottom menu, to the right of `Debug` option

owncloudData/src/main/java/com/owncloud/android/data/user/datasources/implementation/OCRemoteUserDataSource.kt
```diff
@@ -36,4 +39,9 @@ class OCRemoteUserDataSource(
         }.let { return remoteUserInfoMapper.toModel(it)!! }
     }
 
+    override fun getUserQuota(): UserQuota {
```
davigonz, line 42: You could use expression body `= whatever` instead of ` { return whatever }` and the same for the `getUserInfo` method

owncloudData/src/main/java/com/owncloud/android/data/user/repository/OCUserRepository.kt
```diff
@@ -19,12 +19,16 @@
...
 import com.owncloud.android.domain.user.UserRepository
 import com.owncloud.android.domain.user.model.UserInfo
+import com.owncloud.android.domain.user.model.UserQuota
 
 class OCUserRepository(
+    private val localUserDataSource: LocalUserDataSource,
     private val remoteUserDataSource: RemoteUserDataSource
 ) : UserRepository {
     override fun getUserInfo(): UserInfo = remoteUserDataSource.getUserInfo()
+    override fun getUserQuota(accountName: String): UserQuota = remoteUserDataSource.getUserQuota().also { localUserDataSource.saveQuotaForAccount(accountName, it) }
```
davigonz, line 33: This line is too long, could you add a line break?

owncloudDomain/src/main/java/com/owncloud/android/domain/user/model/UserQuota.kt
```diff
@@ -0,0 +1,37 @@
...
+) {
+    @VisibleForTesting
+    fun isLimited() = available > 0
+
+    fun getRelative() = if (isLimited()) {
+        val relativeQuota = (used * 100).toDouble() / getTotal()
+        (relativeQuota * 100).roundToLong() / 100.0
+    } else 0.0
+
+    fun getTotal() = if (isLimited()) available + used else 0
```
davigonz, line 36: I would have used braces here but up to you 😋  , too much text together `available + used else`...

owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_31.kt
abelgardep: It's not a TODO since there is nothing to migrate yet.

PR #2907, opened 2020-06-24
Use Android separator instead of homemade one

hannesa2:
There is no need of an homemade `PATH_SEPARATOR = "/"` 
when there is in Android a `File.separator`

Library: https://github.com/owncloud/android-library/pull/329


abelgardep:
Hi @hannesa2 , thanks for the contribution. We already use File.separator sometimes, so unify them would be great. However, i don't like so much to import a 2k lines file just to access a character. But for the moment it's fine.

hannesa2:
>  import a 2k lines file just to access a character.

Where is the problem ?

jesmrec:
Nothing to check here from my side. Please move forward.

PR #2909, opened 2020-06-26
Simplify version name display

hannesa2:
It's not a big deal, it's just to simplify version name display to a single line. Additional it adds `versionCode` as well

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -25,8 +25,6 @@ package com.owncloud.android.ui.activity;
 import android.app.AlertDialog;
 import android.content.Intent;
 import android.content.SharedPreferences;
-import android.content.pm.PackageInfo;
-import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.res.Configuration;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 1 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -154,16 +152,8 @@ public void onCreate(Bundle savedInstanceState) {
...
-        try {
-            PackageInfo pkg = getPackageManager().getPackageInfo(getPackageName(), 0);
-            temp = pkg.versionName;
-        } catch (NameNotFoundException e) {
-            temp = "";
-            Timber.e(e, "Error while showing about dialog");
-        }
-        final String appVersion = temp + " " + BuildConfig.BUILD_TYPE + " " + BuildConfig.COMMIT_SHA1;
+        final String appVersion =
+                BuildConfig.VERSION_NAME + " " + BuildConfig.BUILD_TYPE + " " + BuildConfig.COMMIT_SHA1 + " " + BuildConfig.VERSION_CODE;
```
abelgardep, line 156: VERSION_CODE would not be needed since we know it with VERSION_NAME. Just keeping the previous behaviour would be enough for the moment.

hannesa2: I see it as an necessary info, but I changed it

jesmrec:
Approved

PR #2912, opened 2020-06-29
[Fix] Refresh credentials

### LINKED ISSUE #2906: Permanent Authentification Error
### Actual behaviour
I want do synchronize my Android tablet to my owncloud installation. The data is already on the server and contains several 1000 files and is about 20GB of size. It is mainly a download-only setup (just keep Android downloading new files without changing them back to the server).

In previous versions of the owncloud Android client I received several authentification failures during synchronization. It was tedious but eventually I managed downloading almost all relevant data.
Now with the latest update the sync-process does not work anymore: authentification failure is a permanent error.
I can re-enter the password in the account settings dialog which it accepts before saying again that authentification fails.

Concluding, owncloud client is not working at all for me.
Note that the password is correct as login in through browser works from same device.
Also, the android app works for the same server on another device and another user name (same app version).

This might be related to this issue I had a while ago: https://github.com/owncloud/client/issues/6843
Server support for HTTP/2 is unfortunately still not available. But I have this issue only on one client...

### Expected behaviour
Accept password and synchronize without crashing or authentification errors.
 
### Steps to reproduce
1. Open owncloud app
2. Error appears 
3. (Re-enter password in change account dialog)
4. (Error appears)

No, not reproducible

### Environment data
Android version: 9

Device model: Samsung Galaxy Tab S5e

Stock or customized system: stock version

ownCloud app version: 2.15 (less severe with older versions)

ownCloud server version: 10.4.1.3

#### Web server error log
N/A

[ownCloud.log](https://github.com/owncloud/android/files/4829146/ownCloud.log)
Edit: Removed server, username

abelgardep:
Fixes #2906 
Needs https://github.com/owncloud/android-library/pull/331

### DIFF
--- a/owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCLocalAuthenticationDataSource.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCLocalAuthenticationDataSource.kt
@@ -44,6 +44,7 @@ import com.owncloud.android.lib.common.accounts.AccountUtils.Constants.KEY_OC_VE
 import com.owncloud.android.lib.common.accounts.AccountUtils.Constants.KEY_SUPPORTS_OAUTH2
 import com.owncloud.android.lib.common.accounts.AccountUtils.Constants.OAUTH_SUPPORTED_TRUE
 import com.owncloud.android.lib.common.authentication.OwnCloudBasicCredentials
+import com.owncloud.android.lib.common.authentication.OwnCloudBearerCredentials
 import timber.log.Timber
 import java.util.Locale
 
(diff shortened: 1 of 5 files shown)


owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCLocalAuthenticationDataSource.kt
```diff
@@ -65,8 +67,14 @@ class OCLocalAuthenticationDataSource(
             userName = userName,
             password = password,
             updateAccountWithUsername = updateAccountWithUsername
-        ).also {
-            updateUserAndServerInfo(it, serverInfo, userInfo)
+        ).also { account ->
+            updateAccountWithUsername?.let {
+                accountManager.setPassword(account, password)
+                SingleSessionManager.getDefaultSingleton().refreshCredentialsForAccount(
```
davigonz, line 73: Just curious, why don't you need this for OAuth?

abelgardep: Applied for OAuth too 👍

jesmrec:
In this case, the following specific scenarios are tested:

- [X] Change password (key button) / basic auth and create new session after removing all active sessions
- [X] Change password (key button) / OAuth and create new token
- [X] Revoke token / OAuth and create new token
- [X] Multiaccount

any other suggestion @abelgardep

abelgardep:
> any other suggestion @abelgardep

No, should be enough with those checks.

Issue #2915, opened 2020-07-03
[BUG] add ownCloud url to mobile -> version is lower than 10


cs35-owncloud:
Hello,

since the new android version (2.15), you cannot add the ownCloud URL to your app if the server has the parameter **'version.hide' => true,** in its config.php ...

All our servers are up to date:

```
# occ status
  - installed: true
  - first_install_version: 10.2.1.4
  - version: 10.4.1.3
  - versionstring: 10.4.1
  - edition: Community
```
Once I set the parameter to **'version.hide' => false,** it works again, I can add the URL of the ownCloud in android.

We tried on iOS or with the Desktop Client, none of them has a problem with hiding the version.hide ... why is it different ?

Could you fix this and revert to the previous behavior ? We shouldn't have to leak information for oC to work on android.
Thanks.

jesmrec:
If the parameter is true (version hidden), we must let the request go:

I checked which version `version.hide` exists from:

oC 9.1.8 (latests 9.x version) => `version.hide` not supported, string always visible.
oC 10.0.0 => `version.hide` supported

So, in case the version is hidden, we can assume it is an oC>=10, so requests will be allowed to go. This is the corner case to support.

Should we include in 2.15.1 @michaelstingl ?

abelgardep:
Thanks for the report. It will be included in 2.15.1

jesmrec:
Fixed. Ready to go in 2.15.1.

cs35-owncloud:
Hello,
thanks for your work :)

PR #2917, opened 2020-07-07
Support for version hide config

abelgardep:
Implements: https://github.com/owncloud/android/issues/2915
Needs: https://github.com/owncloud/android-library/pull/334


jesmrec:
Checks to do:

- [x] <= 9.1.8 & version.hide = true (not supported) => version string visible, version not supported
- [x] <= 9.1.8 & version.hide = false (not supported) => version string visible, version not supported
- [x] <= 9.1.8 & version.hide is not present in `config.php` => version string visible, version not supported

- [X] >= 10.0.0 & version.hide = true => version string visible, version supported. Log in is allowed
- [x] >= 10.0.0 & version.hide = false => version string not visible, version supported.  Log in is allowed
- [x] >= 10.0.0 & version.hide is not present in `config.php` => version string visible, version supported.  Log in is allowed

This is correct on my side

Issue #2873, opened 2020-04-30
[New arch] Fetch list of files of an specific path


jesmrec:
Synchronization epic: https://github.com/owncloud/android/issues/2818

- Once a folder is browsed, list of files is fetched. Important data:

  - File name
  - Extension (to set a thumbnail)
  - Size 
  - Date
  - Permissions?
  - Private/Universal link (if included in the response)
  - ...

### Improvement
- Remove individual Get Quota operation and use the quota retrieved in root propfind instead

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/fetch_list_files
- [x] Development tasks
  - [x] Create new domain model OCFile
  - [x] Add refresh files usecase
  - [x] Migrate RemoteOperation to kotlin
  - [x] Add function to FilesRepository
  - [x] Implement FilesRemoteDatasource
  - [x] Implement FilesLocalDatasource
  - [x] Create OCFileEntity and Dao
  - [x] Add tests on domain module
  - [x] Add tests on data module [Not completed because it requires changes]
  - [x] Handle basic conflicts
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [x] Merge branch new_arch/fetch_list_files into master

### PR

- App: https://github.com/owncloud/android/pull/2918
- Library (if needed): https://github.com/owncloud/android-library/pull/335

michaelstingl:
> * Extension (to set a thumbnail)

New iOS app just asks for thumbnail for EVERY file type… (plenty of 404)

This might help in the future: https://github.com/owncloud/core/issues/31267

I'd avoid hardcoded list of file types that support preview generation.

jesmrec:
We will not forget it... right now, only asking for picture thumbnails.

PR #2918, opened 2020-07-08
[New arch] Fetch list of files

abelgardep:
Implements https://github.com/owncloud/android/issues/2873
Needs https://github.com/owncloud/android-library/pull/335


owncloudData/src/main/java/com/owncloud/android/data/ProviderMeta.java
```diff
@@ -31,14 +31,15 @@
...
-    public static final int DB_VERSION = 31;
+    public static final int DB_VERSION = 32;
 
     private ProviderMeta() {
     }
 
     static public class ProviderTableMeta implements BaseColumns {
         public static final String OCSHARES_TABLE_NAME = "ocshares";
         public static final String CAPABILITIES_TABLE_NAME = "capabilities";
+        public static final String OCFILES_TABLE_NAME = "files";
```
davigonz, line 42: These constants names/values are a bit messy, if files table doesn't have `oc` before, you could use `FILES_TABLE_NAME`... It would be great to have everything similar `ocshares`, `occapabilities` and `ocfiles` or without the oc all of them, but I guess that will need a migration

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCFile.kt
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.testutil
+
+import com.owncloud.android.domain.files.model.OCFile
+
+val OC_FILE = OCFile(
+        id = 122,
```
davigonz, line 24: Not properly indented

owncloudData/src/main/java/com/owncloud/android/data/ProviderMeta.java
abelgardep: You guess right, i can rename it to FILES_TABLE_NAME and keep it the same format as capabilities

Issue #2902, opened 2020-06-12
Replace custom share dialog with Android Sharesheet


abelgardep:
Currently, there is a custom dialog to share files or text with other apps. We could replace it with [Android Sharesheet](https://developer.android.com/training/sharing/send)

Sharing text
| Custom | Android | 
| ------------- | ------------- |
| ![custom-text](https://user-images.githubusercontent.com/47524927/84505115-75995180-acbd-11ea-8e51-433f0d763ab6.jpg) | ![android-text](https://user-images.githubusercontent.com/47524927/84505129-7cc05f80-acbd-11ea-8355-2f4c63d2abcf.jpg)|

Sharing files
| Custom | Android | 
| ------------- | ------------- |
| ![custom-pdf](https://user-images.githubusercontent.com/47524927/84505221-a7121d00-acbd-11ea-9996-28437c7f2430.jpg) | ![android-pdf](https://user-images.githubusercontent.com/47524927/84505235-ae392b00-acbd-11ea-8148-be3fc3494200.jpg)|

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

ddopik:
@abelgardep 
can i work on this

abelgardep:
@ddopik Go ahead 👍

ddopik:
> @abelgardep
> can i work on this

i know it's a naive question but i have just started contributing to onCloud and i want to assure from you do  are we have to download and intilize a local server in order to run the app
or there is a development Link

michaelstingl:
@ddopik If you have Docker installed, you can run a local server very easy:
https://github.com/owncloud/ios-app/blob/master/SETUP.md#local-owncloud-server-for-testing

Otherwise you can connect the Android app to https://demo.owncloud.org/ with credentials `test:test` or `admin:admin`

ddopik:
> @ddopik If you have Docker installed, you can run a local server very easy:
> https://github.com/owncloud/ios-app/blob/master/SETUP.md#local-owncloud-server-for-testing
> 
> Otherwise you can connect the Android app to https://demo.owncloud.org/ with credentials `test:test` or `admin:admin`

i am **In** thank you for your fast reply

PR #2922, opened 2020-07-14
Replace custom share dialog with Android Sharesheet

ddopik:
Custom dialog has been replaced with android shareSheet 

after some research i have found this task could be implemented in two diffrent methods

First          : create A custom **BottomSheet** with ActivityIntent Adapter and follow the same implementation of the old 
                   DialogFragment 

Secound : make use of android **Sharesheet**
    **so**  my choice setteled with **ShareSheet** as it provide more consistency for users across apps
      **Please note** that android Sharesheet documentaion state that we should not customize our shareSheet to display our own 
      list of share targets , but it is considered but with alittle twist when we exluded **OwnCloud** package from the intent List 
      and **added** "CopyToClipBoardIntent" , but this twist won't effect the shareSheet standards as we **did not exclude** any of 
     Acquired **intents** of the **sharedIntent**

this approch has been tested in diffrent simulators with diffrent SDK vertion  and it's result has been compared with 
the old dialogFragment 

this approched has been refactored and applied  only for  the case illustrated by @abelgardep   #2902 
as the old DialgoFragment  **aka** "ShareLinkToDialog" is still getting used by another action **sendDownloadedFile(OCFile file)** 
but it is still applicable to adopt our soultion  if required

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToDialog.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToDialog.java
@@ -85,12 +85,13 @@ public class ShareLinkToDialog extends DialogFragment {
         PackageManager pm = getActivity().getPackageManager();
         List<ResolveInfo> activities = pm.queryIntentActivities(mIntent,
                 PackageManager.MATCH_DEFAULT_ONLY);
+
         Iterator<ResolveInfo> it = activities.iterator();
         ResolveInfo resolveInfo;
         while (it.hasNext()) {
             resolveInfo = it.next();
             if (packagesToExcludeList.contains(resolveInfo.activityInfo.packageName.toLowerCase())) {
-                it.remove();
+                 it.remove();
             }
         }
 
@@ -105,8 +106,13 @@ public class ShareLinkToDialog extends DialogFragment {
             }
         }
 
+
+
         Collections.sort(activities, new ResolveInfo.DisplayNameComparator(pm));
         mAdapter = new ActivityAdapter(getActivity(), pm, activities);
+
+
+
 
         return createSelector(sendAction);
 
@@ -136,6 +142,8 @@ public class ShareLinkToDialog extends DialogFragment {
 
                         // Send the file
                         getActivity().startActivity(mIntent);
+
+
                     }
                 })
                 .create();
@@ -171,5 +179,4 @@ public class ShareLinkToDialog extends DialogFragment {
             icon.setImageDrawable(getItem(position).loadIcon(mPackageManager));
         }
     }
-
 }
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToShareSheet.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToShareSheet.kt
@@ -1,0 +1,145 @@
+package com.owncloud.android.ui.dialog
+
+import android.app.Activity
+import android.content.ComponentName
+import android.content.Intent
+import android.content.pm.LabeledIntent
+import android.content.pm.PackageManager
+import android.content.pm.ResolveInfo
+import android.os.Build
+import android.os.Parcelable
+import com.owncloud.android.R
+import com.owncloud.android.ui.activity.CopyToClipboardActivity
+import java.util.ArrayList
+import java.util.Collections
+import java.util.HashSet
+
+class ShareLinkToShareSheet() {
+
+    private var intentToShareLink: Intent? = null
+
+    private var mFileActivity: Activity? = null
+    private var mComponentNameFilter: ComponentNameFilter? = null
+
+    companion object {
+        fun getInstance(
+            intentToShareLink: Intent,
+            mFileActivity: Activity,
+            mComponentNameFilter: ComponentNameFilter
+        ): ShareLinkToShareSheet {
+
+            val shareLinkToShareSheet = ShareLinkToShareSheet()
+            shareLinkToShareSheet.mFileActivity = mFileActivity
+            shareLinkToShareSheet.mComponentNameFilter = mComponentNameFilter
+            shareLinkToShareSheet.intentToShareLink = intentToShareLink
+
+            return shareLinkToShareSheet
+        }
+    }
+
+    private fun getIntentChooser(): Intent {
+
+        val resolveInfoList =
+            mFileActivity?.baseContext?.packageManager?.queryIntentActivities(intentToShareLink, PackageManager.MATCH_DEFAULT_ONLY)
+
+        val chooserIntent: Intent
+        val targetIntents: MutableList<Intent>
+
+        val titleId: Int
+        val sendAction: Boolean = intentToShareLink?.getBooleanExtra(Intent.ACTION_SEND, false)!!
+        titleId = if (sendAction) {
+            R.string.activity_chooser_send_file_title
+        } else {
+            R.string.activity_chooser_title
+        }
+        val chooserTitle = mFileActivity?.resources?.getString(titleId)
+
+
+        resolveInfoList?.let {
+            val excludedComponentNames = getExcludedComponentNames(resolveInfoList, mComponentNameFilter!!)
+
+            // add activity for copy to clipboard
+            if (!sendAction && !getCopyToClipboardResolveInfoList().isNullOrEmpty()) {
+                resolveInfoList.add(getCopyToClipboardResolveInfoList()?.get(0))
+            }
+
+            Collections.sort(resolveInfoList, ResolveInfo.DisplayNameComparator(mFileActivity?.baseContext?.packageManager))
+            targetIntents = getTargetIntents(intentToShareLink!!, resolveInfoList, excludedComponentNames)
+
+            // deal with M list separate problem
+            chooserIntent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                // create chooser with empty intent in M could fix the empty cells problem
+                Intent.createChooser(Intent(), chooserTitle)
+            } else {
+                // create chooser with one target intent below M
+                Intent.createChooser(targetIntents.removeAt(0), chooserTitle)
+            }
+
+            // add initial intents
+            chooserIntent.putExtra(Intent.EXTRA_INITIAL_INTENTS, targetIntents.toTypedArray<Parcelable>())
+            return chooserIntent
+        }
+
+        return Intent.createChooser(Intent(), chooserTitle)
+
+    }
+
+    private fun getExcludedComponentNames(
+        resolveInfoList: MutableList<ResolveInfo>,
+        componentNameFilter: ComponentNameFilter
+    ): HashSet<ComponentName> {
+        val excludedComponentNames: HashSet<ComponentName> = HashSet<ComponentName>()
+        for (i in resolveInfoList.indices) {
+            val activityInfo = resolveInfoList[i].activityInfo
+            val componentName = ComponentName(activityInfo.packageName, activityInfo.name)
+            if (componentNameFilter.shouldBeFilteredOut(componentName)) {
+                excludedComponentNames.add(componentName)
+            }
+        }
+        return excludedComponentNames
+    }
+
+    /**
+     * @param intent to assosiate target with
+     * @param resolveInfo list of All Target Intents
+     * @param excludedComponentNames HasSet of components to exclude
+     * return an List<intent> for each intent we explicitly  associate it  with desired App target
+     * */
+    private fun getTargetIntents(
+        intent: Intent,
+        resolveInfo: List<ResolveInfo>,
+        excludedComponentNames: HashSet<ComponentName>
+    ): MutableList<Intent> {
+        val targetIntents: MutableList<Intent> = ArrayList()
+        for (i in resolveInfo.indices) {
+            val activityInfo = resolveInfo[i].activityInfo
+            if (excludedComponentNames.contains(ComponentName(activityInfo.packageName, activityInfo.name))) {
+                continue
+            }
+            //---> adds every intent not excluded
+            val targetIntent = Intent(intent)
+            targetIntent.setPackage(activityInfo.packageName)
+            targetIntent.component = ComponentName(activityInfo.packageName, activityInfo.name)
+            // wrap with LabeledIntent to show correct name and icon
+            val labeledIntent = LabeledIntent(targetIntent, activityInfo.packageName, resolveInfo[i].labelRes, resolveInfo[i].icon)
+            // add filtered intent to a list
+            targetIntents.add(labeledIntent)
+        }
+        return targetIntents
+    }
+
+    fun show() {
+        mFileActivity?.startActivity(getIntentChooser())
+    }
+    private fun getCopyToClipboardResolveInfoList(): MutableList<ResolveInfo>? {
+        val copyToClipboardIntent = Intent(mFileActivity, CopyToClipboardActivity::class.java)
+        val copyToClipboard: MutableList<ResolveInfo>? =
+            mFileActivity?.baseContext?.packageManager?.queryIntentActivities(copyToClipboardIntent, 0)
+        return copyToClipboard
+    }
+
+    interface ComponentNameFilter {
+        fun shouldBeFilteredOut(componentName: ComponentName?): Boolean
+    }
+}
+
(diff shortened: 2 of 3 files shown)


abelgardep:
Thanks @ddopik for the PR !! We will take a look after 2.15.1 release.

> this approched has been refactored and applied only for the case illustrated by @abelgardep #2902
as the old DialgoFragment aka "ShareLinkToDialog" is still getting used by another action sendDownloadedFile(OCFile file)
but it is still applicable to adopt our soultion if required

Yes, the idea is to replace completely our custom implementation "ShareLinkToDialog" with the native one and remove ShareLinkToDialog and every related file. It would be great if you could do it within this PR, thanks!

jesmrec:
Thanks for your contribution!! i encourage you to have a look to the [CONTRIBUTING file](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md)

In that file, you'll notice that we use to rebasing instead of merging to update the branches, so we have a clear and clean history. If you have any question about this, please let the team know and we will help you.

As @abelgardep already commented, we will include in the following milestone. We will review your code, post some comments (if needed), and pass a QA stage to assure the most quality as posible.

ddopik:
> Thanks @ddopik for the PR !! We will take a look after 2.15.1 release.
> 
> > this approched has been refactored and applied only for the case illustrated by @abelgardep #2902
> > as the old DialgoFragment aka "ShareLinkToDialog" is still getting used by another action sendDownloadedFile(OCFile file)
> > but it is still applicable to adopt our soultion if required
> 
> Yes, the idea is to replace completely our custom implementation "ShareLinkToDialog" with the native one and remove ShareLinkToDialog and every related file. It would be great if you could do it within this PR, thanks!

thank you i will work on it right away

ddopik:
> Thanks for your contribution!! i encourage you to have a look to the [CONTRIBUTING file](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md)
> 
> In that file, you'll notice that we use to rebasing instead of merging to update the branches, so we have a clear and clean history. If you have any question about this, please let the team know and we will help you.
> 
> As @abelgardep already commented, we will include in the following milestone. We will review your code, post some comments (if needed), and pass a QA stage to assure the most quality as posible.

thanks i have followed your steps as mentioned i hope that i make it in right this time

jesmrec:
you did it perfect @ddopik !

ddopik:
> you did it perfect @ddopik !

really appreciate your respond

ddopik:
Hii @jesmrec 
i have made  Pull request 5 days Ago 
and i still have a messaeg of "Merging is blocked"  is my pull request has any issues ? or it's just  under review
thank you.....

hannesa2:
> Hii @jesmrec
> i have made Pull request 5 days Ago
> and i still have a messaeg of "Merging is blocked" is my pull request has any issues ? or it's just under review
> thank you.....

You need patient. 
It needs sometime long time. eg this one https://github.com/owncloud/android/pull/2076 it needed one year

owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToDialog.java
```diff
@@ -105,9 +106,14 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
             }
         }
 
+
+
```
hannesa2, line 110: What's the point to blow up the file ?

owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToDialog.java
```diff
@@ -136,6 +142,8 @@ public void onClick(DialogInterface dialog, int which) {
 
                         // Send the file
                         getActivity().startActivity(mIntent);
+
+
```
hannesa2, line 146: same here

owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToShareSheet.kt
```diff
@@ -0,0 +1,145 @@
...
+import com.owncloud.android.ui.activity.CopyToClipboardActivity
+import java.util.ArrayList
+import java.util.Collections
+import java.util.HashSet
+
+class ShareLinkToShareSheet() {
+
+    private var intentToShareLink: Intent? = null
+
+    private var mFileActivity: Activity? = null
```
hannesa2, line 21: `mXyz` is java style, Kotlin is with out `m`

hannesa2: every where in this Kotlin class !

owncloudApp/src/main/java/com/owncloud/android/ui/dialog/ShareLinkToShareSheet.kt
```diff
@@ -0,0 +1,145 @@
...
+        mFileActivity?.startActivity(getIntentChooser())
+    }
+    private fun getCopyToClipboardResolveInfoList(): MutableList<ResolveInfo>? {
+        val copyToClipboardIntent = Intent(mFileActivity, CopyToClipboardActivity::class.java)
+        val copyToClipboard: MutableList<ResolveInfo>? =
+            mFileActivity?.baseContext?.packageManager?.queryIntentActivities(copyToClipboardIntent, 0)
+        return copyToClipboard
+    }
+
+    interface ComponentNameFilter {
```
hannesa2, line 141: I vote for extract this interface to an own file

owncloudApp/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -442,9 +459,20 @@ private void shareLink(String link) {
                     )
             );
         }
+        //////////
+//        String[] packagesToExclude = new String[]{mFileActivity.getPackageName()};
+//        DialogFragment chooserDialog = ShareLinkToDialog.newInstance(intentToShareLink, packagesToExclude);
+//        chooserDialog.show(mFileActivity.getSupportFragmentManager(), FTAG_CHOOSER_DIALOG);
+        //////////
```
hannesa2, line 466: What's the point to keep this noise ?

abelgardep:
Hi @ddopik, as i told you some days ago, we will take a look after 2.15.1 release. That's our priority.

By the way, i suggested you to perform a cleanup of ShareLinkToDialog and related stuff if you were able to replace it completely.

jesmrec:
> Hii @jesmrec
i have made Pull request 5 days Ago
and i still have a messaeg of "Merging is blocked" is my pull request has any issues ? or it's just under review
thank you.....

The process is the following: 

we work in a sprint-based methodology. That means, several issues are included in every sprint depending on our needs and our availability. During the sprint, our focus is only the stuff in the sprint. 

Once the issue/PR is included in the sprint, it must pass a code review and then, a QA process before being merged. 

Your PR was not included in the current sprint, but, we'd like to include in the next one. 

What do you think about @abelgardep's suggestion? 

> i suggested you to perform a cleanup of ShareLinkToDialog and related stuff if you were able to replace it completely.

Let us know every question you have, please!

ddopik:
> > Hii @jesmrec
> > i have made Pull request 5 days Ago
> > and i still have a messaeg of "Merging is blocked" is my pull request has any issues ? or it's just under review
> > thank you.....
> 
> The process is the following:
> 
> we work in a sprint-based methodology. That means, several issues are included in every sprint depending on our needs and our availability. During the sprint, our focus is only the stuff in the sprint.
> 
> Once the issue/PR is included in the sprint, it must pass a code review and then, a QA process before being merged.
> 
> Your PR was not included in the current sprint, but, we'd like to include in the next one.
> 
> What do you think about @abelgardep's suggestion?
> 
> > i suggested you to perform a cleanup of ShareLinkToDialog and related stuff if you were able to replace it completely.
> 
> Let us know every question you have, please!
 
i think all comments @hannesa2 , @abelgardep  has mentioned are very convincing and considerable 
and i will update my code right away 
but regardless of the changes on **ShareLinkToDialog** should i delete this  file as it has no referance to any dependency , 
or i should delete it ?

abelgardep:
> i think all comments @hannesa2 , @abelgardep has mentioned are very convincing and considerable
> and i will update my code right away
> but regardless of the changes on **ShareLinkToDialog** should i delete this file as it has no referance to any dependency ,
> or i should delete it ?

If you replace the current ShareLinkToDialog with the native one completely, you should remove the ShareLinkToDialog file and related stuff like adapters or any file that it won't be used anymore.

Thanks in advance!

owncloudApp/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -442,9 +440,9 @@ private void shareLink(String link) {
...
             );
         }
-
-        String[] packagesToExclude = new String[]{mFileActivity.getPackageName()};
-        DialogFragment chooserDialog = ShareLinkToDialog.newInstance(intentToShareLink, packagesToExclude);
-        chooserDialog.show(mFileActivity.getSupportFragmentManager(), FTAG_CHOOSER_DIALOG);
+         ShareLinkToShareSheet shareLinkToShareSheet = ShareLinkToShareSheet.Companion.getInstance(intentToShareLink, mFileActivity, componentName -> {
+                    return componentName.getPackageName().equals(mFileActivity.getPackageName());
+                });
+        shareLinkToShareSheet.show();
```
abelgardep, line 446: I think you overcomplicated this. If you check this [documentation](https://developer.android.com/training/sharing/send#send-text-content), you will see that you could simply call

`        mFileActivity.startActivity(Intent.createChooser(intentToShareLink, null));
`

And it will open the sharesheet as expected. In newer versions you will see a copy to clipboard if you are sharing text etc

There are two things to take care when we are sharing text:

1- Do we need to show ownCloud as target? I don't think so
2- Do we need to add some extra text to the share link? For example: "demo @owncloud.com shared this link(http:server/link) with you" I think it would be enough to copy the link.

Opinions? @michaelstingl @jesmrec

owncloudApp/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -193,9 +190,10 @@ public void sendDownloadedFile(OCFile file) {
             sendIntent.putExtra(Intent.ACTION_SEND, true);      // Send Action
 
             // Show dialog, without the own app
-            String[] packagesToExclude = new String[]{mFileActivity.getPackageName()};
-            DialogFragment chooserDialog = ShareLinkToDialog.newInstance(sendIntent, packagesToExclude);
-            chooserDialog.show(mFileActivity.getSupportFragmentManager(), FTAG_CHOOSER_DIALOG);
+            ShareLinkToShareSheet shareLinkToShareSheet = ShareLinkToShareSheet.Companion.getInstance(sendIntent, mFileActivity, componentName -> {
+                return componentName.getPackageName().equals(mFileActivity.getPackageName());
+            });
+            shareLinkToShareSheet.show();
```
abelgardep, line 196: Similar to sharing text, you can simply use
`            mFileActivity.startActivity(Intent.createChooser(sendIntent, null));
`
And the Sharesheet will open without problems.

Taking into account that currently we can not share multiple files, you don't need to take care of [that](https://developer.android.com/training/sharing/send#share-multiple-content).

Adding [rich content to text previews](https://developer.android.com/training/sharing/send#adding-rich-content-previews) would be interesting, but not needed for now.

And similar to text, we need to take care of:
- Do we need to add ownCloud as target? I don't think so, but maybe i'm missing a special usecase..

If you want to exclude specific targets, you can use [EXTRA_EXCLUDE_COMPONENTS](https://developer.android.com/training/sharing/send#excluding-specific-targets-by-component)

ddopik: Hello @abelgardep  
i have tried to apply this approch at my first try but i had to twist it abit for the following reasons :
 1-- Intent.EXTRA_EXCLUDE_COMPONENTS not  support android below < 24
       so we had to filter Target intents and add the desired list ، that is why we have to **Explicitly** add CopyToClipBoard 
      intent 
2--honstly i wasn't fully aware of sharing  (single/multiple) files Logic as mentioned 👀   ,  as we filtering target intent 
      ,i was just following the same logic the old dialog was build on

appreciate your review , and waiting your feedback.....

abelgardep: 1- It would be great to use EXTRA_EXCLUDE_COMPONENTS for api>=24 and another behaviour for api<24. That way if we want to upgrade sdk in the future, we could simply remove old code instead of replace it again.

2- Yes, don't worry about that, it is enough to keep the previous functionality

Yesterday i tried your solution and i think that there were not as many apps as there were before. With the code i posted in my previous comment, i was able to share with some apps that i wasn't able with this solution.

Taking into account that it may add complexity to this task, i think it would be enough to use sharesheet with api>=24 using EXTRA_EXCLUDE_COMPONENTS and keep using old ShareLinkToDialog when api <24.

What do you think? Thanks for your interest 🥇

ddopik: i think your soultion is producing more target Intent cause we are not specifing the **MiMe** type 
according to the **old ShareDialog** , it was required to query intents with flag  MATCH_DEFAULT_ONLY to get most supported TargetIntetn 
` val resolveInfoList =
            fileActivity?.baseContext?.packageManager?.queryIntentActivities(intentToShareLink, PackageManager.MATCH_DEFAULT_ONLY)`

--- and having both shareDialog and shareSheet would't add complexty more than code duplacity 
      and one of the main purpose this ticket opened for was to remove dialog completly even from the code base and 
       only use shareSheet.

abelgardep: I know that one of the main purposes of this ticket was to remove ShareLinkToDialog. But taking into account that we cannot keep the same behaviour for <24 devices, i would not remove it yet. I think that the best solution for the moment is to check SDK and use old implementation for <24 and sharesheet for >=24. Something like:
```
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
        useShareSheet()
    } else {
        useShareLinkToDialog()
    }
```

jesmrec:
I did a first preliminary QA check, and it seems to be in a good shape. Once the code review is done, i will resume.

owncloudApp/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
abelgardep: Hi @ddopik in Android 7.0, there are no big differences between this two implementations.  But there are big differences in Android 10, where i can only share with 2 apps with this implementation which is not the expected behavior.

Using Android 7.0: 
| This solution | Using the one i suggested | 
| ------------- | ------------- |
current_7 | android-text|

ddopik: i have investigated in your issue when you said that you have only 2 app to share to when runnning on android(10) Q , and you are totaly right , i spent some time depugging and testign code again but meanwhile i have found this issue openedin [stackOverFlow](https://stackoverflow.com/questions/59786777/why-intent-extra-initial-intents-not-working-in-android-10) and also in [IssueTracker](https://issuetracker.google.com/issues/134367295). 

i think we shoud implement our own bottomSheet to match the desired behaviour and gurantee the compatibility 
what do you think @abelgardep @jesmrec @michaelstingl

abelgardep: I would go with a solution like the one i suggested you some days ago: https://github.com/owncloud/android/pull/2922#discussion_r463425724 

```   
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
    useShareSheet()
} else {
    useShareLinkToDialog()
}
```

Using native sharesheet with EXTRA_EXCLUDE_COMPONENTS for ownCloud when api>24 and the current implementation when api<24. Don't you think?

jesmrec: Moving stuff to native is something where we are moving to (SAF is the best example). If we can use the native implementation, i think it'd be better.

jesmrec: @ddopik what do you think about @abelgardep suggestion? we can take advantage of the native implementation from android N on, and let the old one for users with older devices. Once we upgrade our minimum API to 24(somewhen in the future), we can get rid of the old implementation and keep the feature as native. We can help you, whatever you need.

abelgardep:
Merged via #2968

Issue #2872, opened 2020-04-30
[New Arch] Download a single file


jesmrec:
Synchronization epic: https://github.com/owncloud/android/issues/2818

### Postponed for the moment. To do in upcoming PRs.
- Group notifications with progress. We need to decide how to show them when several downloads are performed parallelly. This will have the same behavior for uploads. To be done together. 
- No information will be shown in the adapters due to performance issues. (Icon downloading in the file list)
- Tests will be included in upcoming PR because several things will change till the final version.

AC:

- By tapping a file in the list, a copy of the remote file is downloaded and stored in the device. 
- If the file is currently downloaded, check before if  new version exists in the server.
- Notifications: many notifications can happen at the time if many files are downloaded. They can be grouped from 3 different (pending to check)
- Progress bar to be updated during the action with the %. Also visible in the notification. 
- Once the file finishes to download, notification is gone.
- Errors in download shown in notification
- File is previewed after being downloaded

Errors: 

- No connectivity
- Server unreachable
- Error servers (file was remotely deleted, moved, maintenance mode)

### TASKS

- [ ] Research which is the best way to do it
- [x] Create branch new_sync/download_file
- [ ] Development tasks
  - [x] Migrate remote operation to kotlin
  - [x] Create a file provider in data module to handle file storage [Not needed for downloads]
  - [x] Integrate workmanager to schedule downloads
  - [x] Create a one time worker to download a file, cancelable, and listenable
  - [x] Integrate this task with the new architecture
  - [x] Show notifications with download progress
  - [x] Use new worker to download and send files
  - [x] Use new worker to download and show files in documents provider
  - [x] Observe worker in the viewpager adapter
  - [x] Use new worker to download and show images
  - [x] Use new worker to download and open with
  - [ ] Use new worker and show something downloading in adapters
  - [x] Show a custom message in notification
  - [x] Create a transfer provider to handle transfers. Similar to TransferRequester but with new Arch
  - [x] Clean old downloader stuff
  - [ ] Error handling
  - [ ] Unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [x] Merge branch new_sync/download_file into new_arch/sync

### PR

abelgardep:
We will stop this issue for the moment. It is needed to complete #2955 first.

abelgardep:
I have some questions about how downloads work now and how they should work after this synchronization refactor.

### Let's start showing the current behavior.

[...]

Issue #2869, opened 2020-04-29
[New arch] OCFileListFragment isolation


abelgardep:
Currently, there are several fragments to list files, it would be great to unify them and get rid of activities dependencies.

(Good to have) Adapter refactors, use a recycler view to show the list of files, use diff util to increase performance, use already created FilesViewModel, get files from DB as Flow and Pagination for better performance.

### Tasks 
- [X] Create branch new_arch/file_list_isolation
- [ ] Development tasks
  - [X] New fragment MainFileListFragment #3451 
  - [X] New ViewModel (?) #3451 
  - [X] Show a simple list of files (GetFolderContentUseCase it will be updated to LiveData-Flow) #3451 #3461 
  - [x] Footer with number of files and folders https://github.com/owncloud/android/pull/3457
  - [x] Empty files view (Could be modified -> https://github.com/owncloud/android/issues/3026) https://github.com/owncloud/android/pull/3460
  - [x] Swipe to refresh (call RefreshFolderFromServerAsyncUseCase) #3469 
  - [x] Different options: All files, Shared by link and Av Offline (FileListOption -> Check if we can improve it) ( GetFilesSharedByLinkUseCase) (AvOffline use case does not exist yet) -> https://github.com/owncloud/android/pull/3471
  - [x] Bottom navigation bar -> https://github.com/owncloud/android/pull/3476
  - [x] Grid view mode #3493 
  - [x] Floating action button handling (Only shown in All files option) -> https://github.com/owncloud/android/pull/3492
  - [x] Search files within the list (Check if a new use case is needed for this. It is searched locally and only for this folder) -> https://github.com/owncloud/android/pull/3494
  - [x] Sort by name, size, and date (Check if a new use case is needed for this) #3478
  - [x] Consider adding diff util to improve performance
  - [x] Display images correctly in the list #3495 
  - [ ] Multiselection
    - [x] Change toolbar
      - [x] Control items selected #3508 
      - [x] Control checkbox -> https://github.com/owncloud/android/pull/3545
      - [x] Control Drawer #3511 
      - [x] Refresh option -> https://github.com/owncloud/android/pull/3545
      - [x] Share option -> https://github.com/owncloud/android/pull/3545
    - [ ] Show Menu Dialog (Keep in mind that we may change the UI -> https://github.com/owncloud/android/issues/2885)
      - [x] Select all -> https://github.com/owncloud/android/pull/3545
      - [x] Select inverse -> https://github.com/owncloud/android/pull/3545
[...]

Issue #2866, opened 2020-04-24
[New Arch] Copy file/folder operation


abelgardep:
Synchronization epic: https://github.com/owncloud/android/issues/2818

AC:

- Copy remote in server.
- Local: check if file is downloaded to move to new location. Check if new version is available in case of file already downloaded.
- In case of copy folder: check if items inside are downloaded, if they have newer versions, update.
- Check if the copied file is downloaded, in case the source one is downloaded.
### QA Checks
https://github.com/owncloud/QA/blob/master/Mobile/Android/Sync/Copy.md

### TASKS

- [x] Research (if needed)
- [x] Create branch new_sync/copy
- [ ] Development tasks
  - [x] Create new usecase -> CopyFileUseCase
  - [x] Add copyFile to repository
  - [x] Add copyFile to RemoteDataSource (Call CheckRemotePath before CopyRemoteFileOperation to check if folder already exists?)
  - [x] Migrate CopyRemoteFileOperation to kotlin
  - [x] Update database
  - [ ] Copy local files/folders
  - [x] Update DocumentsProvider
  - [x] Error handling
  - [x] Unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [x] Merge branch new_sync/copy into new_arch/sync

### PR

- App https://github.com/owncloud/android/pull/3253
- Library (if needed) https://github.com/owncloud/android-library/pull/406

abelgardep:
Related to AC:
> Copy remote in server.
> Local: check if file is downloaded to move to new location. Check if new version is available in case of file already downloaded.
In case of copy folder: check if items inside are downloaded, if they have newer versions, update.
Check if the copied file is downloaded, in case the source one is downloaded.

I think that it could be enough to copy files remotely and to update the database accordingly. We could avoid some problems that way instead of copying every downloaded file.

abelgardep:
Will be tested within synchronization epic

Issue #2865, opened 2020-04-24
[New Arch] Move file/folder Operation


abelgardep:
Synchronization epic: https://github.com/owncloud/android/issues/2818

### TASKS

- [x] Research (if needed)
- [x] Create branch new_sync/move
- [x] Development tasks
  - [x] Create new usecase -> MoveFileUseCase
  - [x] Add moveFile to repository
  - [x] Add moveFile to RemoteDataSource (Call CheckRemotePath before MoveRemoteFileOperation to check if folder already exists?)
  - [x] Migrate MoveRemoteFileOperation to kotlin
  - [x] Migrate MoveRemoteChunkFileOperation to kotlin
  - [x] Add move to FileService
  - [x] Add move to ChunkService
  - [x] Update database
  - [x] Move local files/folders
  - [x] Update DocumentsProvider
  - [x] Error handling
  - [x] Unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [x] Merge branch new_sync/move into new_arch/sync

### PR

- App https://github.com/owncloud/android/pull/3232
- Library (if needed) https://github.com/owncloud/android-library/pull/400

abelgardep:
Pending:
- Check behaviour when moving Av. Offline or moving to a folder marked as Av. Offline
- Check performance when moving A LOT OF files

abelgardep:
Will be tested within the synchronization epic.

PR #2934, opened 2020-08-12
[New arch] Synchronization

abelgardep:
Implements synchronization epic: https://github.com/owncloud/android/issues/2818
Library: https://github.com/owncloud/android-library/pull/339

### Already merged:
- Fetch list of files: https://github.com/owncloud/android/issues/2873
- Create folder: https://github.com/owncloud/android/issues/2861
- Get File/Folder by path/id : https://github.com/owncloud/android/issues/2957
- Get Folder content: https://github.com/owncloud/android/issues/2956
- Save file/folder: https://github.com/owncloud/android/issues/2958
- Migrate old database to room: #2955
- Download a single file: #2872 
- Remove file/folder #2864
- Move File/Folder #2865
- Rename file/folder #2863
- Copy File/Folder #2866
- File list isolation #2869


abelgardep:
Rebased with 2.15.3

abelgardep:
Rebased with master: 88e8289ef324df0a523d2b382203b7893f9aacb4

owncloudData/src/main/java/com/owncloud/android/data/files/db/OCFileEntity.kt
```diff
@@ -0,0 +1,126 @@
...
+    val length: Long,
+    val creationTimestamp: Long?,
+    val modificationTimestamp: Long,
+    val mimeType: String,
+    val etag: String?,
+    val permissions: String?,
+    val privateLink: String? = null,
+    val storagePath: String? = null,
+    var name: String? = null,
+    val treeEtag: String? = null,
```
hannesa2, line 70: I want to anyhow indicate duplicate files. For this I need a `hash` field. 
I know that having this field on server makes most sense, but I know nothing about the server code, so this small solution, to have in only in Android client, satisfy my needs too.

It would make my fork much easier, because I don't need to handle db changes.
Any chance to add this `val hash: String? = null` field here ?

abelgardep: AFAIK duplications are detected when synchronizing a file or folder in the client. 

We already have the etag and the remote id.

Why would you need an additional field to indicate duplicated files?

michaelstingl: Also checksums are available ownCloud server, and already used by desktop and iOS apps. But their use case is more data integrity, not so much duplicate detection.

hannesa2: @abelgardep No duplicates are found when they are in different directories.
Next to this, a rename could be also a source of duplicates.
@michaelstingl 
You could use it for duplicates detection as well

abelgardep:
Rebased with 2.20 29ebcb0f9cfc338801198836944cc3be83390799

abelgardep:
Rebased with 2.21 f02860c212f4745b7ab5bf4a010ede1fca02b2c0

abelgardep:
Rebased with master cfa018cf45f29f349e4b53534f8c747f59e7de33

JuancaG05 (approved):
LET'S GOOOOOOO 🚀

PR #2947, opened 2020-09-02
Fix npe when logging camera upload request

abelgardep:
According to google play console, account could be null at that point and it causes a null pointer exception when we log it.


davigonz (approved):

jesmrec:
Just new logs here. No QA attention needed

Issue #2956, opened 2020-09-10
[New arch] Get folder content


abelgardep:
Related with #2955, we need to get the folder content from the local database in order to show the file list, so this should be one of the first issues to take care of after the migration.

For the moment, it would be enough to retrieve a list of files and keep the previous behaviour.

BTW, we can include GetImagesFromFolder use case since it is simply an aditional filter in the database.

In the future, we could use Flow or LiveData to improve the performance and error handling.

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/get_folder_content
- [x] Development tasks
  - [x] Implement usecase for get folder content
  - [x] Implement usecase for get folder images
  - [x] Implement repository
  - [x] Implement localdatasource
  - [x] Implement dao
  - [x] Add tests for domain layer
  - [x] Add tests for data layer 
- [x] Code review and apply changes requested
- [x] Design test plan [no needed - after migration]
- [x] QA [no needed - after migration]
- [x] Merge branch new_arch/get_folder_content into new_arch/synchronization

### PR

- [app](https://github.com/owncloud/android/pull/2965)
- Library (if needed)

abelgardep:
Merged into new_arch/synchronization. It will be tested there.

PR #2965, opened 2020-09-14
[New arch] Get folder content

abelgardep:
Implements #2956


theScrabi (approved):

owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/UseCaseModule.kt
```diff
@@ -27,8 +27,12 @@ import com.owncloud.android.domain.capabilities.usecases.GetCapabilitiesAsLiveDa
 import com.owncloud.android.domain.capabilities.usecases.GetStoredCapabilitiesUseCase
 import com.owncloud.android.domain.capabilities.usecases.RefreshCapabilitiesFromServerAsyncUseCase
 import com.owncloud.android.domain.files.usecases.CreateFolderAsyncUseCase
+<<<<<<< HEAD
```
theScrabi: Make sure to remove these before a commit.
Please use 
`git diff HEAD` before committing.

abelgardep: I know, rebase problems 😢  It was fixed in next commit

owncloudData/src/test/java/com/owncloud/android/data/file/datasources/OCLocalFileDataSourceTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    @Test(expected = Exception::class)
+    fun `get folder content - ko`() {
+        every { dao.getFolderContent(any()) } throws Exception()
+
+        localDataSource.getFolderContent(dummy.id!!)
+    }
+
+    @Test
+    fun `get folder images - ok`() {
```
theScrabi, line 126: Why do you sometimes write `- ok` and sometimes `- ko`
Just a question of interest: what is the scheme for writing the description/names of the tests?

owncloudApp/src/main/java/com/owncloud/android/utils/UriUtils.java
```diff
@@ -226,13 +230,13 @@ private static String getDisplayNameFromContentResolver(Uri uri, Context context
         String mimeType = context.getContentResolver().getType(uri);
```
theScrabi, line 230: This method could be broken down into smaller ones.

owncloudApp/src/main/java/com/owncloud/android/utils/UriUtils.java
```diff
@@ -226,13 +230,13 @@ private static String getDisplayNameFromContentResolver(Uri uri, Context context
...
         if (mimeType != null) {
```
theScrabi, line 231: You could for example export this if clause.

theScrabi (changes requested):
LFTM I like it. Yet there are smaller things. Maybe you can have a look at them. Sometimes I just have questions.

owncloudApp/src/main/java/com/owncloud/android/utils/UriUtils.java
abelgardep: Sure, but it is not the purpose of this PR, i just moved some hardcoded strings into constants. Probably we will convert to kotlin and refactor this file in upcoming PRs.

owncloudData/src/test/java/com/owncloud/android/data/file/datasources/OCLocalFileDataSourceTest.kt
abelgardep: I use `ok` as success and `ko` as exception or failure.

I started writing names as `getFolderContentSuccess` or `getFolderContentException` but it is not as simple when you have several cases. So i decided to switch to a more readable way:

what are we testing - expected behaviour - additional info

owncloudData/src/test/java/com/owncloud/android/data/file/repository/OCFileRepositoryTest.kt
```diff
@@ -179,4 +179,28 @@ class OCFileRepositoryTest {
             localFileDataSource.getFileByRemotePath(OC_FILE.remotePath, OC_FILE.owner)
         }
     }
+
+    @Test
+    fun getFolderContentSuccess() {
+        every { localFileDataSource.getFolderContent(OC_FILE.parentId!!) } returns listOf(OC_FILE)
+
```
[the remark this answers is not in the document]
abelgardep: I like to keep those statements separated, i think it is easy to read, don't you think? Something like this:

mock

function to test

result assertion

calls verification

theScrabi (approved):

PR #2968, opened 2020-09-16
[Feature] Android Sharesheet

abelgardep:
Implements #2902 partially.

We cannot [exclude oC from sharing targets till API 24](https://developer.android.com/training/sharing/send#excluding-specific-targets-by-component), so we'll keep the previous behaviour for API <24, and use sharesheet for newer ones.

Old code will be removed when we upgrade min_sdk_version to api 24.

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/helpers/ShareSheetHelper.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/helpers/ShareSheetHelper.kt
@@ -1,0 +1,61 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Abel García de Prada
+ * Copyright (C) 2020 ownCloud GmbH.
+ * <p>
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ * <p>
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * <p>
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.ui.helpers
+
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.content.pm.ResolveInfo
+import android.os.Build
+import android.os.Parcelable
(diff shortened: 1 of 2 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/helpers/ShareSheetHelper.kt
```diff
@@ -0,0 +1,61 @@
...
+        packagesToExclude: Array<String>
+    ): Intent {
+
+        // Get excluding specific targets by component. We want to hide oC targets.
+        val resInfo: List<ResolveInfo> =
+            context.packageManager.queryIntentActivities(intent, PackageManager.MATCH_DEFAULT_ONLY)
+        val excludeLists = ArrayList<ComponentName>()
+        if (resInfo.isNotEmpty()) {
+            for (info in resInfo) {
+                val activityInfo = info.activityInfo
```
theScrabi, line 47: Maybe this block can be put into a function:
`getExcludeListForInfo(info: ResolveInfo, packagesToExcloude: Array<String>) : ArrayList<ComponentName`

theScrabi (approved):
LFTM

jesmrec:
### (1)

I've tested how the UI looks like with different versions. With Android 7 (API 24, it takes the new approach), no title is shown in the menu, so it seems not too clear. However, other newer versions (Android 9, Android 11), show a title :

|Android 11|Android 7|
|:-:|:-:|
|Screenshot 2020-10-06 at 12 32 20|Screenshot 2020-10-06 at 12 26 54|

is this something that can be fixed? or is it a system feature?

abelgardep:
Yes, you are right, i have added a title according to https://developer.android.com/training/sharing/send#adding-rich-content-previews

jesmrec:
### (2)

Just to keep compatibility and same wording as previous version, when the link is shared the proper string is `Send link to` (activity_chooser_title) instead of `share with`. What do you think?

jesmrec:
### (3)

1. Select a file
2. Select the option `Send`

Current: In the menu, we can see `Share` (Android11), or `Share with` (Android 7). 
Expected: `Send` 

is this feasible?

abelgardep:
> ### (2)
> Just to keep compatibility and same wording as previous version, when the link is shared the proper string is `Send link to` (activity_chooser_title) instead of `share with`. What do you think?

Ok, updated.


> ### (3)
> 1. Select a file
> 2. Select the option `Send`
> 
> Current: In the menu, we can see `Share` (Android11), or `Share with` (Android 7).
> Expected: `Send`
> 
> is this feasible?

Im not sure why it changes 🤔

jesmrec:
(3) affected only Android 7, since newer versions use their own wording "Share" 

PR Approved on my side, everything fixed

Thanks also @ddopik for your contribution!!

PR #2976, opened 2020-09-30
Update library reference

abelgardep:
Remove single cookie header in library. Updating the reference..

Needs https://github.com/owncloud/android-library/pull/344


theScrabi (approved):

jesmrec:
The `http.protocol.single-cookie-header` is already removed. Since we do not know how exactly can affect to the expected behaviour of the app, the following QA actions have been performed:

- All tests [here](https://github.com/owncloud/android-scenario-testing/tree/master/src/test/resources/io/cucumber) have been executed over two different devices (android 7 and 9)
- Tested with test redirect servers
- Tested with test OAuth2 server
- Tested with test oCIS server

i did not find any bad behaviour in the main actions after login (sharing, file handling, doc provider) by performing some exploratory testing.

Some upgrade tests will be performed in the regression stage.

From my side, this is OK to go

PR #2988, opened 2020-10-02
move to github ci

theScrabi:
This will introducd github CI, based on the submission by @hannesa2: https://github.com/owncloud/android/pull/2816
This will also add artifact upload and testing:

- [x] research
- [ ] implement testing
- [ ] implement artifact upload
- [ ] remove bitrise


theScrabi:
After further investigation I think bitrise so far is the best sollution for Android Instrumentation tests as github Actions for android is kind  of a "hack" right now, which apparently runs only in some cases. As it can be seen in the test results here. Espresso tests went through on Bitrise. They went through locally as well. However on Github they did not run properly which could be due to a slow emulator.

I'd recommend to move to the pubic version of bitrise, so other contributors can see what happens on bitrise CI, however for now I wouldn't not recommend moving over to github actions.

hannesa2:
That's sad

hannesa2:
After months you came to the same result as I, that on bitrise it works magically. 👏 
On my local machine I've the similar issues like Github has, but I can't verify bitrise settings.

My observations:

- Github has issues 
- local has issues
- bitrise seems to work, but I can't verify

theScrabi:
> After months you came to the same result

I work on this the first week. I haven't work with GithubCI before.

> After months you came to the same result as I, that on bitrise it works magically. 

Bitrise uses a dedicated setup for Instrumentation tests. GithubCI does not. The time we spend in trying to get GithubCI working we will need on other Projects and since bitrise works I don't see a reason to move.

>  On my local machine I've the similar issues like Github has, but I can't verify bitrise settings.

Currently bitrise and testresults are only visible to us. We need to change this. However I could not verify this, only one test fails which may be related to the fact that I didn't disable animations. However GithubCI fails more often. The reason it fails is because it could not launch Intents due to timeout. This looks to me like the machine the emulator runs on is to underpowered for what it should accomplish.

hannesa2:
> The reason it fails is because it could not launch Intents due to timeout. This looks to me like the machine the emulator runs on is to underpowered for what it should accomplish.

I've seen this in the past, so I introduced screenshots during tests. https://github.com/hannesa2/LiveEdgeDetection/commit/7360b34763c3592403f8f2fb7b7570059ed601a4
So I detect a stupid permission dialog on first usage of device, which blocks the test

Does bitrise recycle the device ?

theScrabi:
This could be an explanation however I haven't seen something like this locally.

JuancaG05:
Hi @hannesa2, we have in one of our BitRise workflows two lint jobs: KtLint and Android Lint, and that's where we see in CI the needed reports. Moreover, it is public right now, so anyone can see the results. For example, check https://app.bitrise.io/build/88a39c63-328d-431d-ba4b-eff4f820f4c0.

As commented above, if this is working nicely on BitRise, I don't see the reason to move to GitHub CI, which is not optimized for Android testing.

PR #3002, opened 2020-10-22
Remove cookie persistence

theScrabi:
This will resolve cookie persistence which should solve braking the cookiejar after restarting the app.

Related app PR: https://github.com/owncloud/android-library/pull/354

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/services/OperationsService.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/services/OperationsService.java
@@ -182,8 +182,6 @@ public class OperationsService extends Service {
     @Override
     public void onDestroy() {
         Timber.v("Destroying service");
-        // Saving cookies
-        SingleSessionManager.getDefaultSingleton().saveAllClients(this, MainApp.Companion.getAccountType());
 
         mUndispatchedFinishedOperations.clear();
 


owncloudApp/src/main/java/com/owncloud/android/services/OperationsService.java
```diff
@@ -183,7 +183,6 @@ public int onStartCommand(Intent intent, int flags, int startId) {
     public void onDestroy() {
         Timber.v("Destroying service");
         // Saving cookies
-        SingleSessionManager.getDefaultSingleton().saveAllClients(this, MainApp.Companion.getAccountType());
```
abelgardep, line 186: Remove the comment by the way

jesmrec:
This are my first insights about this PR:

- i do not see any cookie accumulation

- After attaching one account, cookies are renewed every time i open the app. But, after some times (no quick open/close, quiet actions), if i close and reopen the app, cookies are not renewed anymore

- Attaching two accounts and switching gives the same behaviour. Initial switches, cookies are renewed with no accumulation in both accounts. After a couple of switches, both accounts keep the cookies they had before

owncloudApp/src/main/java/com/owncloud/android/services/OperationsService.java
abelgardep, line 186: Remove the comment by the way.

abelgardep (changes requested):
Tiny change required @theScrabi

jesmrec:
Tested in two different cases

1) Only one account

- In account creation, cookies are received after checking `status.php` and then, sent in the requests. Capabilities' request does not send cookies, so the cookie `oc_sessionPassphrase` used in the session are the one received in the capabilities response. The cookie `oc...` (like `oc794f4ep19m`) is renewed frequently (`PROPFINDS`), sometimes responses return a new value even when the request contain the cookie recently received.  From client perspective this is correct, the cookie renewal is not in this side.

- In app closing/opening: same behaviour

2) More accounts

- Account switching (no closing): when switching to another account, no request is done to `status.php`, so same cookies are used as before. 

From my side:

- No cookie accumulation anymore ✅ 
- Cookies received from server are "reflected" in the following requests ✅ 
- Cookies are correctly switched when account switches ✅ 
- Cookies are not renewed when account switches ❓   (not bad, but... expected?)

theScrabi:
> Cookies are not renewed when account switches question (not bad, but... expected?)

No. This is bad behaviour. The library is saving cookies based on domain. If two accounts are running on the saim domain. Cookies will leak into other sessions.

The only real solution to all of the cookie issues. Is to make the okHttpClient not be a singleton any more.  Hacking around to make this work with the current situation feels like stitching everything together using bandages.

michaelstingl:
Multiple accounts on the same instance is a supported use case, and all date (credentials, cache, cookies etc) should be strictly separated. Needs 🍪 -jar per account for example…

Issue #3003, opened 2020-10-23
[BUG] Shared by link only shows files/folders from root


diocas:
### Actual behaviour
When opening the Share by Link page, I only see the public links that exist in the root/home folder. If this was consistent with the web ui, then I should see all my public links, including the ones inside other folders.

I'm not sure wether this is on purpose or not... The API only returns the data I'm seeing, because the app calls it with the `path=/` and `subfiles=true` parameters (which I might also argue that should include everything under the path `/`, but that's another question for the core repo..).

I also tried with the CERNBox, which return all my links, but the app still filters them and only shows the root ones. We might be doing something wrong, but I'm not seeing anything in the logs.

### Expected behaviour
The Shared by Links should be consistent with the web ui and show all my links.

### Environment data
Android version: 10

Device model: Emulator

Stock or customized system: Stock

ownCloud app version: 2.15.2

ownCloud server version: 10.5.0 (demo.oc.com)

abelgardep:
First of all, thanks for opening an issue. Yes it is intended that you only see the top shared folder. Same for Av. Offline files. Should we move to another behaviour? @michaelstingl

michaelstingl:
> Should we move to another behaviour? @michaelstingl

Should be similar to the web ui. It lists all links a user generated:

![CleanShot 2020-10-23 at 17 03 45@2x](https://user-images.githubusercontent.com/214010/97020413-d1b89000-1551-11eb-8e40-d252b382a30d.png)

https://demo.owncloud.com/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json&include_tags=true

```
{
"ocs": {
"meta": {
"status": "ok",
"statuscode": 100,
"message": null,
"totalitems": "",
"itemsperpage": ""
},
"data": [
{
"id": "9",
"share_type": 3,
"uid_owner": "test",
"displayname_owner": "test",
"permissions": 1,
"stime": 1603465020,
"parent": null,
"expiration": null,
"token": "E9dr4YwuaAJwfiU",
"uid_file_owner": "test",
"displayname_file_owner": "test",
"additional_info_owner": null,
"additional_info_file_owner": null,
"path": "/Parent/Folder1",
"item_type": "folder",
"mimetype": "httpd/unix-directory",
"storage_id": "home::test",
"storage": 4,
"item_source": 321,
"file_source": 321,
"file_parent": 320,
"file_target": "/Folder1",
"name": "Public link",
[...]

Issue #2955, opened 2020-09-10
[New Arch] Migrate OCFile to new database 


abelgardep:
To continue with the new architecture, we will need to migrate completely the old file database to the new one (Room).

This step will help us to develop the other synchronization issues faster, because we won't need to take care of two different databases and different ocfile models.

But this migration has different inconveniences since OCFile is present everywhere, and we will need to disable some functionalities until we finish completely the rearquitecture.

1. FileDataStorageManager handles every old database queries. So we will need to transform those operations to usecases. Some of them are currently planned, others may require new issues. Including:
- Get file by path, id.. #2957 
- Get folder content #2956 
- Get images from folder #2956 
- Save folder/file #2958
- Remove file/folder #2864 (includes local removal)
- Move file/folder #2865 (includes local move)
- Copy file/folder #2866 (includes local copy)
- Conflicts management
- Av. Offline management
- Shared by link management

2. Since the main idea is to release a version once we complete the rearquitecture, we could disable some features temporarely in the new_arch/synchronization branch, and work in different issues. For example:
- Camera Uploads
- Downloads #2872 
- Uploads
- Av. Offline synchronization
- DocumentsProvider integration

Once we have a compiling version, @jesmrec could start QA on each single functionality instead of waiting till the end of sync rearquitecture.

### Tasks

- [x] Transform RemoteFile to kotlin
- [x] Add old file attributes to new one. Needed to perform a proper migration
- [x] Use new create folder usecase instead of old database queries
- [x] Use new get file by id usecase instead of old database queries
- [x] Use new get file by path usecase instead of old database queries
- [x] Use new get folder content usecase instead of old database queries
- [x] Use new OCFile instead of old one on every file in the app
- [ ] Remove old OCFile [Deprecated for the moment. We may need to check its properties for pending new arch operations]
- [ ] Remove old database queries related to OCFile [Deprecated for the moment. We may need to check its properties for pending new arch operations]
- [ ] Try to perform attributes cleanup [Not possible at the moment. Let's check in next arch issues]
- [x] Add OCFile tests

### PR
- App #2995 
- Library (if needed) https://github.com/owncloud/android-library/pull/353

### Bugs and known issues after migration
- Av. Offline: Including get av.offline, toggle, sync, etc
- Uploads works partially: Files are uploaded properly to server but app shows an error
- Downloads works partially: Files are downloaded properly from server but app shows an error
- Shared by link
- Remove file
- Move local file
- Copy local file
- Conflicts

[...]

PR #3007, opened 2020-10-26
Align shared by link behaviour with other platforms

abelgardep:
Implements #3003 

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.16/3007-share_link_list_extended.md

Bugs & improvements:

- [X] (1) Links on non-root items stays in the list after being removed from Links view https://github.com/owncloud/android/pull/3007#issuecomment-719426005 [WONT FIX]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
@@ -235,7 +235,6 @@ class FileDataStorageManager {
     val sharedByLinkFilesFromCurrentAccount: Vector<OCFile>
         get() {
             val allSharedFiles = Vector<OCFile>()
-            val result = Vector<OCFile>()
             var cursorOnShared: Cursor? = null
             try {
                 cursorOnShared = contentResolver?.query(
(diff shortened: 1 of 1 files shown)


theScrabi (approved):

jesmrec:
### (1)

1. Create link on an item that is in root, and another one on an item that is not in root
2. Open "Links" view -> both items are there
3. From links view, open "Share" in every item and remove links created in 1.

Current: Item in root is not in the "Links" list anymore. Item in non-root stays in the list
Expected: Both items are not in the "Links" list anymore

Android11, Pixel2
commit `404c563`

abelgardep:
About [(1)](https://github.com/owncloud/android/pull/3007#issuecomment-719426005)

When we remove a share, we update shares database, but not files database. Files database is updated when we refresh a folder. So, at the moment, when we change something in the links view, that change is not visible until we refresh the folder that contains that file.

2 Options:
- Refresh the folder that contains that file after any change in links view.
- Update not only shares db but also files db after performing any change in links view.

jesmrec:
> Update not only shares db but also files db after performing any change in links view.

This one is the most desirable. But... how feasible/costly is it?

abelgardep:
> > Update not only shares db but also files db after performing any change in links view.
> 
> This one is the most desirable. But... how feasible/costly is it?

It would be pretty simple if we had both tables (files and shares) in room database. But it is not the case, so it requires some work.

michaelstingl:
Do we plan to have both tables (files and shares) in room database? Just wait?

abelgardep:
> Do we plan to have both tables (files and shares) in room database? Just wait?

Of course, we started working on it in #2955 . As soon as we have this in master, we could move this forward.

michaelstingl:
No real urgency here. We can wait a bit…

jesmrec:
Should we freeze this one till the database migration is merged into `master`?
or, should we merge this current approach, and fix (1) in the future when migration is merged into `master`?

I like the second option. What do you think?

jesmrec:
This is approved on my side. (1) will be improved in the future, so i will create new issue for that.

PR #3015, opened 2020-11-06
Fix crash when rotating create folder dialog

abelgardep:
Steps to reproduce this crash:

- Share an external file with ownCloud
- In the picker, click on create a new folder
- Rotate the device when create folder dialog appears


davigonz (approved):

jesmrec:
Fixed.

Issue #2927, opened 2020-08-10
Request previews for EVERY file type


michaelstingl:
I assume there's a hardcoded list of filetypes where the Android app requests thumbnails/previews. Depending on the server setup, previews for more file types can be provided. The only way to make sure the Android app shows previews for all files the server supports, is to request previews for EVERY file type. Android app will receive 404 in case there is no preview. (same approach as in new iOS app and in Phoenix web UI)

In the future, server-side could announce in the Capabilities what file types have previews: https://github.com/owncloud/core/issues/31267 (not implemented yet)

### PR Old
App #3017 

### PR New
App #3105

jesmrec:
Need to be re-enabled and re-checked in 2.17

Issue #2926, opened 2020-08-10
Use public preview API


jesmrec:
Currently (2.15), Android is using only thumbnails as part of the previews. (private API)
Request is  ´api/v1/thumbnail/448/448/<path>´ (private thumbnail API)

It could switch to the the public preview API:
`remote.php/dav/files/[user]/[file]?x=[width]&y=[height]&c=[eTag]&preview=1 `
(but there's no real urgency)

x and y parameters contains `120` in iOS platform (as reference)

- AC (keep legacy behaviour):

  - Thumbnails are correctly fetched, no matter if picture is downloaded or not 
  - Thumbnails are requested once browsing
  - Thumbnails are shown in the uploads view.
  - Thumbnails are shown in the folder picker (copy, move options)

### PR
App: #3017

abelgardep:
I was checking the api and the current implementation, and we can switch to the public preview api in a pretty simple way.

About the public preview api, i think that `Content-Length` header is missing.

> x and y parameters contains 120 in iOS platform (as reference)

About the thumbnails size, currently we retrieve a 448x448 which is much bigger than that reference. I have performed a quick test and for the same image: 
- 120x120 -> 4Kb
- 448x448 -> 35Kb

We have better thumbnails but maybe they are not as much important. Should we decrease thumbnails size? [Admins can limit their size anyway](https://doc.owncloud.com/server/admin_manual/configuration/files/previews_configuration.html#maximum-preview-size)

PR #3017, opened 2020-11-12
Use public preview api to retrieve thumbnails

abelgardep:
Implements #2926 and #2927 

____

### QA

Bugs & improvements:

- [x] (1) etag field is empty https://github.com/owncloud/android/pull/3017#issuecomment-729698876
- [X] (2) checking for updates is not only for images https://github.com/owncloud/android/pull/3017#issuecomment-732133576 [DONE]
- [X] (3) Txt thumbnail updates not reflected https://github.com/owncloud/android/pull/3017#issuecomment-732149912 [WONT FIX HERE]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -37,13 +37,11 @@ import android.widget.ImageView;
 import androidx.core.content.ContextCompat;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
-import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.lib.common.OwnCloudAccount;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.SingleSessionManager;
 import com.owncloud.android.lib.common.http.HttpConstants;
 import com.owncloud.android.lib.common.http.methods.nonwebdav.GetMethod;
-import com.owncloud.android.lib.resources.status.OwnCloudVersion;
 import com.owncloud.android.ui.adapter.DiskLruImageCache;
 import com.owncloud.android.utils.BitmapUtils;
 import timber.log.Timber;
@@ -287,8 +285,7 @@ public class ThumbnailsCacheManager {
 
                 } else {
                     // Download thumbnail from server
-                    OwnCloudVersion serverOCVersion = AccountUtils.getServerVersion(mAccount);
-                    if (mClient != null && serverOCVersion != null) {
+                    if (mClient != null) {
                         GetMethod get;
                         try {
                             String uri = getPreviewUrl(file, mAccount);
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -249,6 +249,12 @@ public class ThumbnailsCacheManager {
             return Math.round(r.getDimension(R.dimen.file_icon_size_grid));
         }
 
+        private String getPreviewUrl(OCFile ocFile, Account account) {
+            return mClient.getBaseUri() +
+                    "/remote.php/dav/files/" + account.name.split("@")[0] + Uri.encode(ocFile.getRemotePath(), "/")
+                    + "?" + "x=" + getThumbnailDimension() + "&y=" + getThumbnailDimension() + "&c=" + ocFile.getEtag() + "&preview=1";
+        }
+
         private Bitmap doOCFileInBackground() {
             OCFile file = (OCFile) mFile;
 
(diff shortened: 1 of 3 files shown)


owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -249,6 +247,12 @@ private int getThumbnailDimension() {
             return Math.round(r.getDimension(R.dimen.file_icon_size_grid));
         }
 
+        private String getPreviewUrl(OCFile ocFile, Account account) {
+            return mClient.getBaseUri() +
+                    "/remote.php/dav/files/" + account.name.split("@")[0] + Uri.encode(ocFile.getRemotePath(), "/")
```
theScrabi, line 252: Maybe this string can be put into a constant.

theScrabi: Maybe this whole line can be written as interpolatable string.

theScrabi (changes requested):

owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
abelgardep: I don't like it so much neither, but we can not use a constant. I could only use "/remote.php/dav/files/" as constant. But no the whole url since it is dynamic depending ocFile and account. About interpolatable string, well, i would like to, but this is java, not kotlin.

theScrabi: I updated my solution. Please have a look.

theScrabi (approved):

jesmrec:
Let's QA this one

jesmrec:
### (1)

I've checked out the requests and they have the following form:

`/remote.php/dav/files/user1/ownCloud%20Manual.pdf?x=336&y=336&c=&preview=1 `

- `c` parameter is empty. It use to contain the etag. Is that intended somehow?
- I lack of an API doc, so i checked the iOS client who actually implements that API. This is how an iOS request looks like:

`Archive.zip?y=120&preview=1&x=120&c=%22483f40d8166bcebaab085ea9cca57a79%22&a=1`

it contains also an `a` parameter. By checking the iOS code:

```
request.parameters = [NSMutableDictionary dictionaryWithObjectsAndKeys:
				@(size.width).stringValue, 	@"x",
				@(size.height).stringValue,	@"y",
				item.eTag, 			@"c",
				@"1",				@"a", // Request resize respecting aspect ratio 
				@"1", 				@"preview",
			nil];
```

should we include also the `a` to keep a correct ratio?

jesmrec:
### (2)

@abelgardep as you commented me, this part of the code:

```
if (!updatedLocalFile.isFolder() &&
                        remoteFile.isImage() &&
                        remoteFile.getModificationTimestamp() != localFile.getModificationTimestamp()) {
                    updatedLocalFile.setNeedsUpdateThumbnail(true);
```

should be changed, since not only images contain a thumbnail.

jesmrec:
### (3)

1. In an account, sync a folder that contains an txt file. Thumbnail is shown
2. In server, modify the txt file including some different text on the top (where the thumbnail pic is rendered)
3. Sync the folder / download the new version of the file

Current: Thumbnail does not change
Expected: Thumbnail change to show the new content

Same behaviour works fine with images

Pixel 2 Android 11
`b34f119`

jesmrec:
(1) -> etag is sent fro mthe moment it is stored in the DB.

abelgardep:
I think that `(2)` and `(3)` refers to the same problem and should be fixed now. We will ask for new thumbnails when something changes in the server even if it is not an image (Previous behaviour).

jesmrec:
About (2) and (3)

Thumbnails of txt files are correctly updated only if the file is not downloaded, and with a browsing back and forth. That means, with steps:

1. txt is not downloaded in the app
2. in web or other client, modify the txt
3. in the app, pull down to refresh the containing folder

Result: thumbnail not refreshed

4. in the app, browse to other folder and browse back to the containing folder

Result: thumbnail is now refreshed

5. tap on the txt file to download
6. close the file, so you are again in the list of files
7.  in web or other client, modify the txt
8. in the app pull down to refresh the containing folder

Result: thumbnail not refreshed

9. browse to another folder and then browse back to the containing folder

Result: thumbnail not refreshed

Expected: refresh should realise on the new version. Downloaded files must refresh also the thumbnail


Pixel 2 Android 11
`4ec916e`

abelgardep:
That's the behaviour since the beginning. If the file is available locally, we extract the thumbnail from the local file. The problem here is that the local thumbnail extraction seems not to work with txt files. So, the thumbnail is not updated.

jesmrec:
So, we have two different issues regarding previews:

- For txt files. We can not extract thumbnails from downloaded files.

- For images. If a downloaded image changes, thumbnail is not updated and you see the preview of the old version.

Any solution for this? getting rid of the local thumbnails and use only remote thumbnails, even when the file is locally downloaded? it will increase the number of request, but is any case we are going to request preview for every file. This will be addressed in a different ticket, for sure.

Which are the expectations @michaelstingl?

abelgardep:
> it will increase the number of request, but is any case we are going to request preview for every file

We only request thumbnails the first time and then if something changes in the server, but not always. If file is available locally we don't request for new thumbnails at the moment.

jesmrec:
> We only request thumbnails the first time and then if something changes in the server, but not always. If file is available locally we don't request for new thumbnails at the moment.

yes, for this reason, a posible solution would be requesting always the remote thumbnails, so that the preview is always updated. But first, we have to know the expectations for this feature.

theScrabi (approved):

jesmrec:
After discussions and calls, the final conclusion is that thumbnail and preview updates will need more elaboration. It will be moved to another issue:

Goals of the current PR:

- Replace the old endpoint with the new Preview API
- Request for every file. Pending: request also for downloaded. This will be moved to the new issue.

Issue #2997, opened 2020-10-15
use own okHttp client for AppAuth 


theScrabi:
~[AppAuthConnection](https://github.com/owncloud/android/blob/f7d720d/owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthConnectionBuilder.kt) still uses ApacheHttp, which we already replaced with OkHttp in other places. This should be changed in order not to have two network libraries. At the moment having two libraries causes several issues like having to handle TLS separately or having two different user agents.~

AppAuth uses plain java [HttpUrlConnection](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html). However in the java Runtime HttpUrlConnection is an abstract class that is implemented internally by the android build in [okHttp](https://android.googlesource.com/platform/external/okhttp/+/master/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java). This however is done in a way that it behaves as close as possible to the java implementation, this also means that an internal instance of the okHttp client instead of our instance is used.
In order to get our settings and interceptors to work with AppAuth we therefore have to create a HttpUrlConnection instance that uses our own okHttp client.

Tasks:

- research how to make Android HttpUrlConnection use our okHttp client

Need more research to know the impact/scope

*TL/DR:* We decided on modifying AppAuth, and add a wrapper to its network library. This way we can inject our own okHttpClient into AppAuth. The related AppAuth PR can be found [here](https://github.com/openid/AppAuth-Android/pull/621)

appauth: https://github.com/openid/AppAuth-Android/pull/621
app: https://github.com/owncloud/android/pull/3022
library: https://github.com/owncloud/android-library/pull/359

Related: https://github.com/owncloud/android/issues/3009

michaelstingl:
@abelgardep could this help with the unloggable requests in https://github.com/owncloud/android/issues/2887 ?

abelgardep:
Well, if we were able to use the same OkHttpClient to perform auth requests, it would help definitely. I'm not sure whether it is possible. Maybe @davigonz could clarify a little bit since he researched how to integrate AppAuth.

theScrabi:
[...]

PR #3022, opened 2020-11-25
Okhttp appauth

theScrabi:
issue: https://github.com/owncloud/android/issues/2997
library: https://github.com/owncloud/android-library/pull/359


owncloudApp/src/main/java/com/owncloud/android/authentication/AccountUtils.java
```diff
@@ -203,13 +203,6 @@ public static void updateAccountVersion(Context context) {
                                 accountMgr.getUserData(account, Constants.KEY_OC_VERSION)
                         );
 
-                        // copy cookies
-                        accountMgr.setUserData(
-                                newAccount,
-                                Constants.KEY_COOKIES,
-                                accountMgr.getUserData(account, Constants.KEY_COOKIES)
-                        );
-
```
abelgardep, line 212: I know that we are going to handle this in a different way in upcoming PR, but i would keep it for the moment.

owncloudApp/src/main/java/com/owncloud/android/services/OperationsService.java
```diff
@@ -182,8 +182,6 @@ public int onStartCommand(Intent intent, int flags, int startId) {
     @Override
     public void onDestroy() {
         Timber.v("Destroying service");
-        // Saving cookies
-        SingleSessionManager.getDefaultSingleton().saveAllClients(this, MainApp.Companion.getAccountType());
```
abelgardep, line 186: Same here

build.gradle
```diff
@@ -46,7 +46,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.0'
+        classpath 'com.android.tools.build:gradle:4.1.1'
```
abelgardep, line 49: I have not updated AS to v4.1.1 yet :( Maybe you can create a new PR for this update and update the library dependency too.

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthConnectionBuilder.kt
```diff
@@ -28,8 +30,9 @@ class OAuthConnectionBuilder(val context: Context) : ConnectionBuilder {
      */
 
     @Throws(IOException::class)
-    override fun openConnection(uri: Uri): HttpURLConnection {
+    override fun openConnection(uri: Uri): HttpConnection {
         val conn: HttpURLConnection
+        val conWrapper: HttpConnection
```
abelgardep, line 35: I don't see any use of conWrapper, is it needed?

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthUtils.kt
```diff
@@ -58,7 +61,7 @@ class OAuthUtils {
             AuthorizationServiceConfiguration.fetchFromUrl(
                 serviceDiscoveryLocation,
                 onGetAuthorizationServiceConfiguration,
-                OAuthConnectionBuilder(context)
+                OkConnectionBuilder(HttpClient.getOkHttpClient())
```
abelgardep, line 64: I see that you have replaced the old OAuthConnectionBuilder(context) here, and introduced the new one. But OAuthConnectionBuilder is used in some other places. Just curiosity, why do you replace it only here? By the way, this is the only appauth call that was logged. It may be related 🤔

abelgardep (changes requested):

owncloudApp/src/main/java/com/owncloud/android/authentication/oauth/OAuthConnectionBuilder.kt
theScrabi: OAuthConnectionBuilder is not required anymore the whole class should be deleted.

Issue #3024, opened 2020-11-25
[FEATURE REQUEST] Improvements in preview/thumbnails updates


jesmrec:
Current behaviour:

 - The preview of an txt file is updated only when it is not downloaded, and the containing folder is browsed back and forth. 

 - In case the file is downloaded, the preview is not updated because it is generated locally. That means, preview does not match with file content.

ACs

- When a file is updated in remote (web or other client), the preview will be updated in the following refresh.  (this behaviour is pretty similar to the av. offline files one)

- When a folder contains many files, requests should be scoped / reduced to avoid request-flooding in some way. -> Current behaviour: fetching previews of the files in the screen (pending to confirm)

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/improve_thumbnails
- [x] Development tasks
  - [x] Retrieve thumbnails even if the file is available locally
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/improve_thumbnails into master

### PR

- App #3035 
- Library (if needed)

jesmrec:
Important topic: https://github.com/owncloud/android/issues/2926#issuecomment-725927475

PR #3035, opened 2020-12-15
Improve thumbnails

abelgardep:
Implements #3024

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -274,24 +274,6 @@ public class ThumbnailsCacheManager {
 
                 int px = getThumbnailDimension();
 
-                if (file.isDown()) {
-                    Bitmap temp = BitmapUtils.decodeSampledBitmapFromFile(
-                            file.getStoragePath(), px, px);
-                    Bitmap bitmap = ThumbnailUtils.extractThumbnail(temp, px, px);
-
-                    if (bitmap != null) {
-                        // Handle PNG
-                        if (file.getMimetype().equalsIgnoreCase("image/png")) {
-                            bitmap = handlePNG(bitmap, px);
-                        }
-
-                        thumbnail = addThumbnailToCache(imageKey, bitmap, file.getStoragePath(), px);
-
-                        file.setNeedsUpdateThumbnail(false);
-                        mStorageManager.saveFile(file);
-                    }
-
-                } else {
                     // Download thumbnail from server
                     if (mClient != null) {
                         GetMethod get;
@@ -322,7 +304,6 @@ public class ThumbnailsCacheManager {
                         }
                     }
                 }
-            }
 
             return thumbnail;
 


owncloudApp/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -274,24 +274,6 @@ private Bitmap doOCFileInBackground() {
 
                 int px = getThumbnailDimension();
 
-                if (file.isDown()) {
```
davigonz, line 277: What happens when there's no connection? Are the already downloaded thumbnails not going to be shown in that case?

abelgardep: Steps to show a thumbnail:
1 Check if thumbnail is available in cache
2 a) If thumbnail is not available -> Ask preview API
2 b) If needsUpdateThumbnail flag is active* -> Ask preview API
2 c) thumbnail available and flag not active -> Show cached thumbnail
3 a) If preview API request is not successful, default image
3 b) If preview API request is successful, save thumbnail and show it

*needsUpdateThumbnail is active if modification time differs between remote and local one. And you need connection to set it up

CC/ @jesmrec

jesmrec: That's it. If there's no connection, thumbnail will not be fetched/updated (expected behaviour). It can cause a side-effect when image is locally edited and no connection is available. In that case, thumbnail and image will not be the same. As idea, it should be removed (not sure if feasible) waiting for connection to upload the new image version and fetch the new thumbnail. WDYT?

abelgardep: Ok, i see, at that point, i think we could add a new step when preview API request fails. At that point, instead of showing a default image, we could try to generate a thumbnail as we we doing before

jesmrec: Would it be replaced again in the next refresh?

davigonz (approved):

jesmrec:
Checks: 

Non dowloaded files

- [X] Modify remote an image using a different client -> thumbnail is updated
- [X] Modify remote a txt file using a different client -> thumbnail is updated
- [x] Modify remote an image using a different client with no connection -> thumbnail is updated after recovering connection
- [X] Modify remote a txt file using a different client with no connection -> thumbnail is updated after recovering connection

Downloaded files

- [x] Modify remote an image using a different client -> thumbnail is updated
- [x] Modify remote a txt file using a different client -> thumbnail is updated
- [X] Modify remote an image using a different client with no connection -> thumbnail is updated after recovering connection
- [X] Modify remote a txt file using a different client with no connection -> thumbnail is updated after recovering connection
- [x] Modify locally a txt file -> file uploaded and thumbnail updated after open it
- [x] Modify locally a av. offline txt file -> file uploaded and thumbnail updated 
- [x] Modify locally an image -> image uploaded and thumbnail updated after open it
- [x] Modify locally an av. offline image -> image uploaded and thumbnail updated

jesmrec:
New improvement: thumbnails are requested with same size so the shape does not usually match with the image shape. iOS app shows correctly thumbnails with no cuts. Any idea about it?

jesmrec:
Improvement above, addressed to a new issue #3050

jesmrec:
Approved

PR #3040, opened 2020-12-22
Use constants instead of hardcoded values for dir mimetypes

abelgardep:
Fixes a problem where file size was 0 in ocis due to hardcoded mimetypes.
Could fix a problem when removing a folder in ocis.

Needs: https://github.com/owncloud/android-library/pull/362

oc10 -> "DIR"
ocis -> "httpd/unix-directory"


individual-it:
is this actually a bug in oc10 or  ocis? shouldn't both APIs respond the same way?

theScrabi (approved):

jesmrec:
Now it always returns 4KB, no matter the size of the content inside. Is that the expected?

On the other hand, all stuff is dated as  `1/1/1970` (different problem due to different property, isnt's it?)

michaelstingl:
> On the other hand, all stuff is dated as `1/1/1970` (different problem due to different property, isnt's it?)

Yes, different problem: https://github.com/owncloud/ocis/issues/1154

michaelstingl:
> Now it always returns 4KB, no matter the size of the content inside. Is that the expected?

At least it's in sync with other platforms: https://github.com/owncloud/ocis/issues/1150#issuecomment-749424927

jesmrec:
Ok, the specific problem here is fixed. Let's wait for the other fixes mentioned above.

We can merge this.

Issue #3038, opened 2020-12-18
Replace AppAuth for custom implementation


jesmrec:
Actually, [AppAuth](https://github.com/openid/AppAuth-Android) library is the responsible to handle the OAuth2 - OIDC authentication flows inside the app. We have faced several issues with the library . The most important of them seems to be the unmaintenance of the repository (two and half years without releases). 

Issues like

https://github.com/owncloud/android/issues/2997
https://github.com/owncloud/android/issues/2929

are also related with that.

In order to be aligned with other client teams, we will go for a custom implementation of OAuth2 and OIDC that will allow us:

- to have control over the authentication feature
- to log everything we need
- not to be pending of external releases

For sure, having an external dependency is nothing bad (we keep several external dependencies), but in this case, the pros are less than the cons.

This is some related stuff, that can be helpful

- OAuth2 implementation PR: https://github.com/owncloud/android/issues/1724
- AppAuth spike: https://github.com/owncloud/android/issues/2755
- AppAuth integration: https://github.com/owncloud/android/pull/2800 (app)  https://github.com/owncloud/android-library/pull/300 (library)
- Chrome Custom Tab implementation: https://github.com/owncloud/android/pull/2157

It could be done in a two-step implementation (what do you think?):

1. Implement OAuth2 ( + automatic tests)
2. QA OAuth2
3. Implement OIDC
4. QA OIDC 
5. I would go for a beta stage 

ACs must be regressive. Keep same behaviour as before, adding as many logs as posible.

-----
### Implementation
app: https://github.com/owncloud/android/pull/3043
library: https://github.com/owncloud/android-library/pull/363

abelgardep:
### Steps to replace AppAuth
- [ ] Create usecase to retrieve oidc discovery information
- [ ] Create usecase to exchange authorization code for tokens
- [ ] Create usecase to refresh token
- [ ] Replace AppAuth UI and implement Custom Tabs

PR #3043, opened 2020-12-23
[Feature] OAuth2 - OIDC Custom implementation

abelgardep:
Implements #3038
Needs https://github.com/owncloud/android-library/pull/363

_____

## QA

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.17/RemoveAppAuth%20-%20Authentication.md

Bugs & improvements

- [X] (1) Redirections https://github.com/owncloud/android/pull/3043#issuecomment-766879561 [FIXED]
- [x] (2) Token renewal leads to 401 in following release https://github.com/owncloud/android/pull/3043#issuecomment-768932634 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -455,51 +407,74 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
             updateOAuthStatusIconAndText(authorizationException)
-            Timber.e(authorizationException, "OAuth request to get authorization code failed")
         }
     }
 
     /**
      * OAuth step 2: Exchange the received authorization code for access and refresh tokens
      */
-    private fun exchangeAuthorizationCodeForTokens(authorizationResponse: AuthorizationResponse) {
+    private fun exchangeAuthorizationCodeForTokens(authorizationCode: String) {
```
davigonz, line 429: It would be great to try to delegate this sort of responsibilities to the viewmodel... could part of this method be moved to the `LoginViewModel`, keeping here just the observe parts?

owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
```diff
@@ -34,6 +37,32 @@ class ClientManager(
...
 ) {
+    // This client will maintain cookies across the whole login process.
+    private var ownCloudClient: OwnCloudClient? = null
+
+    /**
+     * Returns a client for the login process.
+     * Helpful to keep the cookies from the status request to the final login and user info retrieval.
+     * For regular uses, use [getClientForAccount]
+     */
+    fun getClientForUnExistingAccount(
```
davigonz, line 48: Un**e**xisting ?

owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
```diff
@@ -34,6 +37,32 @@ class ClientManager(
...
+        path: String,
+        requiresNewClient: Boolean,
+        ownCloudCredentials: OwnCloudCredentials? = getAnonymousCredentials()
+    ): OwnCloudClient {
+        val safeClient = ownCloudClient
+
+        return if (requiresNewClient || safeClient == null) {
+            return OwnCloudClient(Uri.parse(path)).apply {
```
davigonz, line 56: Isn't this return redundant?

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCAccount.kt
```diff
@@ -24,3 +43,6 @@ const val OC_AUTH_TOKEN_TYPE = "owncloud.oauth2.access_token"
 const val OC_ACCESS_TOKEN = "Asqweh12p93yehd10eu"
 const val OC_REFRESH_TOKEN = "P3sd19DSsjdp1jwdd1"
 const val OC_SCOPE = "email"
+
+const val OC_TOKEN_ENDPOINT = "https://ocis.owncloud.works/konnect/v1/token"
```
davigonz, line 47: Should this endpoint be public? Can't we use a more dummy one?

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/oauth/OIDCServerConfiguration.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.testutil.oauth
+
+import com.owncloud.android.domain.authentication.oauth.model.OIDCServerConfiguration
+
+val OC_OIDC_SERVER_CONFIGURATION = OIDCServerConfiguration(
+    authorization_endpoint = "https://ocis.owncloud.works/signin/v1/identifier/_/authorize",
```
davigonz, line 24: And all these values? Use more dummy?

davigonz (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -369,82 +351,52 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
-                    Timber.e(
-                        exception,
-                        "OIDC failed. Code: ${exception.code} Error: ${exception.error} Error Description: ${exception.errorDescription} Error Uri: ${exception.errorUri} Type: ${exception.type}"
-                    )
-                    startNormalOauthorization()
-                } else if (serviceConfiguration != null) {
+        oauthViewModel.getOIDCServerConfiguration(serverBaseUrl)
+        oauthViewModel.oidcDiscovery.observe(this, {
+            when (it.peekContent()) {
+                is UIResult.Loading -> TODO()
```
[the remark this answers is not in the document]
abelgardep: No, in this case, we won't use loading, it will be transparent for the user. But I agree that instead of `TODO()`, I could use `{}`

owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -308,97 +307,102 @@ private void refreshToken(
...
-        AppAuthConfiguration.Builder appAuthConfigurationBuilder = new AppAuthConfiguration.Builder();
-        appAuthConfigurationBuilder.setConnectionBuilder(new OAuthConnectionBuilder(mContext));
-        AuthorizationService authService = new AuthorizationService(mContext, appAuthConfigurationBuilder.build());
+    private void handleSuccessfulRefreshToken(TokenResponse tokenResponse,
+                                              AccountAuthenticatorResponse accountAuthenticatorResponse,
+                                              Account account,
+                                              String authTokenType,
+                                              AccountManager accountManager,
+                                              String oldRefreshToken
+    ) {
```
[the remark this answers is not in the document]
abelgardep: Sure, it was formatted that way, so I left it that way. But I like this way a little bit more 👍

owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -308,97 +307,102 @@ private void refreshToken(
...
-                    AccountUtils.Constants.KEY_OC_BASE_URL
-            );
-            authorizationServiceConfiguration = new AuthorizationServiceConfiguration(
-                    Uri.parse(baseUrl + File.separator + mContext.getString(R.string.oauth2_url_endpoint_auth)), // auth endpoint
-                    Uri.parse(baseUrl + File.separator + mContext.getString(R.string.oauth2_url_endpoint_access)) // token endpoint
-            );
+        String baseUrl = accountManager.getUserData(account, AccountUtils.Constants.KEY_OC_BASE_URL);
+
+        // OIDC Discovery
+        @NotNull Lazy<OIDCDiscoveryUseCase> oidcDiscoveryUseCase = inject(OIDCDiscoveryUseCase.class);
```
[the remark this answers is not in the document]
abelgardep: Yes! Previously, there was another dependency, but then they included java support in the core one 🥇

owncloudApp/build.gradle
```diff
@@ -31,8 +31,8 @@ dependencies {
         exclude group: "com.android.support"
     }
 
-    // AppAuth
-    api 'net.openid:appauth:0.7.1'
+    // CustomTabs required for OAuth2 and OIDC.
+    implementation "androidx.browser:browser:1.3.0"
```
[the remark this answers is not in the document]
abelgardep: Sure, It was an option. But since it is only used in this module, I added it this way.

owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
abelgardep: Do you mean that there is a typo? 👀

owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
abelgardep: Yes, I think so. Thanks

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -369,82 +351,52 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+                is UIResult.Success -> {
+                    Timber.d("Service discovery: ${it.peekContent().getStoredData()}")
                     oidcSupported = true
-                    performGetAuthorizationCodeRequest(serviceConfiguration)
-                    authorizationServiceConfiguration = serviceConfiguration
+                    val oidcServerConfiguration = it.peekContent().getStoredData() ?: return@observe
+                    performGetAuthorizationCodeRequest(oidcServerConfiguration.authorization_endpoint.toUri())
```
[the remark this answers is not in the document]
abelgardep: Yes, basically, it's the same 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
abelgardep: Yes, I agree. I kept it the same way it was implemented, but I think I should move it. It would be easier to test.

owncloudApp/build.gradle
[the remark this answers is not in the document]
davigonz: In my opinion, having all the versions in a file, even if any of them is being used only in one module, is easier to maintain, you don't need to remember whether that version was hardcoded in a module build.gradle or not, you will see all the versions at a glance...

I've seen this approach in other projects but you can keep it as it is since there's other hardcoded versions in the app build.gradle

owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
davigonz: First I was going to suggest `AnExisting` because I thought it was a typo, then I search unexisting in Google and since it's a single word I would write Unexisting instead of UnExisting... 😄 not very important anyway

davigonz (approved):

jesmrec:
### (1)

Redirections with 301 are now working. The run into an infinite loop

This is the stacktrace:

````
16:02:20.383 14619-4706/com.owncloud.android.debug D/(RemoteOperationResult.java:301) .<init>(): RemoteOperationResult has processed UNHANDLED_HTTP_CODE: 301 HTTP 301 Moved Permanently```

abelgardep:
(1) It should be fixed now. Ready to test @jesmrec .

BTW, I have rebased with master.

jesmrec:
### (2)

A side effect i have realised regarding OAuth2

1. Open an OAuth2 session
2. Wait till token expires
3. Refresh root folder

After doing these steps, sometimes the "Sign In" message is displayed because the following request (usually, capabilities request) keeps using the old token. Session is running correctly but message could be confusing for the user.

Device: Nexus 6P, Android 7

jesmrec:
From my side, this is approved

Issue #2604, opened 2019-07-11
[BUG] Weird cookie handling


jvillafanez:
### Actual behaviour
New cookies sent by the server are ignored by the client. The client keeps sending the same cookie.
In addition, the cookies sent by the client seem to include some cookie flags that shouldn't be present

Below what is sent by the server

![Screen Shot 2019-07-11 at 14 19 03](https://user-images.githubusercontent.com/1477829/61050291-f9c8c200-a3e6-11e9-8878-809a2a77642c.png)

And below the client

![Screen Shot 2019-07-11 at 14 18 24](https://user-images.githubusercontent.com/1477829/61050332-1533cd00-a3e7-11e9-9b65-b2995a37ed64.png)

_____

Task checklist:

- [x] Check why is happening this
- [X] First refactor of cookie stuff to use in login requests the cookie retrieved from the server https://github.com/owncloud/android/pull/2970 @theScrabi 
- [X] Remove cookie persistence https://github.com/owncloud/android/pull/3002 @theScrabi 
- [x] Refactor library to avoid cookies leaking to other accounts  @theScrabi

davigonz:
Thanks @jvillafanez for reporting this, we'll have a look

michaelstingl:
We should have closer look after the 2.16 or earlier…

jesmrec:
Checked this problem in the content of the fix: https://github.com/owncloud/android/pull/2970.

I think the problem is not gone. These are the steps i performed:

1. Add a new account, everything goes OK. App receives some cookies and start to send them:

`oc794f4ep19m=o6adkiaict0gm5nqk17e7eskjd; cookie_test=test; oc_sessionPassphrase=WUf8q8glaFgv0HAjIziVDOyNVZ%2Foqw9UNRBXY3aooizWn9iiMDX75c1KJftAIBZqKl4koLWJvF7YJt8%2BlmVXkOsSWVm6%2BYv4SNYalv0jOtYdnjLlhJz8aCOFI2rKpRQ0
`

no new cookies are sent from the server during the request/response flow (doing some operations: copy, move, share etc etc...). So, the cookies above are repeatedly send in every request with no changes

2. Add another account and repeat same process, same result.

3. Switch again to the account added in 1. Cookies start to accumulate:

[...]

PR #3047, opened 2021-01-04
remove okhttp singleton

theScrabi:
issue: https://github.com/owncloud/android-library/issues/357
depends on: https://github.com/owncloud/android-library/pull/364


jesmrec:
Waiting for https://github.com/owncloud/android/issues/3096 before moving forward

abelgardep (changes requested):
Just a tiny question @theScrabi

abelgardep (approved):
Approved! Fix the conflicts before moving this forward 👍

build.gradle
```diff
@@ -29,6 +29,7 @@ buildscript {
         fragmentTestVersion = "1.1.0"
         uiAutomatorTestVersion = "2.2.0"
         annotationTestVersion = "1.2.0"
+        robolectricVersion = "4.4"
```
abelgardep, line 32: I would try not to add this version here. At the moment we use robolectric just in the library, so there is no need to add this version in the app Gradle from my point of view.

owncloudData/src/main/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSource.kt
```diff
@@ -41,9 +41,14 @@ class OCRemoteServerInfoDataSource(
...
-        val checkPathExistenceResult =
-            serverInfoService.checkPathExistence(path, isUserLogged = false, client = owncloudClient)
+        var checkPathExistenceResult =
+            serverInfoService.checkPathExistence(path, isUserLogged = false, owncloudClient)
+        var redirectionLocation = checkPathExistenceResult.redirectedLocation
+        while (!redirectionLocation.isNullOrEmpty()) {
+            checkPathExistenceResult =
+                serverInfoService.checkPathExistence(redirectionLocation, isUserLogged = false, owncloudClient)
+            redirectionLocation = checkPathExistenceResult.redirectedLocation
+        }
```
abelgardep, line 51: Is this change required? We removed this loop some PR ago. We should use the same client that requested the remote status, so it should follow redirections by itself. That's why we removed it.

Is it needed now after removing the singleton?

owncloudData/src/test/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSourceTest.kt
```diff
@@ -69,6 +69,39 @@ class OCRemoteServerInfoDataSourceTest {
...
+        } returns checkPathExistenceResultMocked
+
+        val authenticationMethod = ocRemoteServerInfoDatasource.getAuthenticationMethod(redirectedLocation)
+
+        assertNotNull(authenticationMethod)
+        assertEquals(AuthenticationMethod.BASIC_HTTP_AUTH, authenticationMethod)
+
+        verify { ocServerInfoService.checkPathExistence(OC_SERVER_INFO.baseUrl, false, ocClientMocked) }
+    }
+
```
abelgardep, line 104: Related to the previous comment. If previous one is required, then keep this test, if not, we can remove it 👍

abelgardep (changes requested):
Some tiny suggestions more @theScrabi

abelgardep (approved):
LGTM

owncloudData/src/main/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSource.kt
```diff
@@ -41,9 +41,8 @@ class OCRemoteServerInfoDataSource(
         // Use the same client across the whole login process to keep cookies updated.
         val owncloudClient = clientManager.getClientForAnonymousCredentials(path, false)
 
-        // Step 1: check whether the root folder exists.
-        val checkPathExistenceResult =
-            serverInfoService.checkPathExistence(path, isUserLogged = false, client = owncloudClient)
+        var checkPathExistenceResult =
+            serverInfoService.checkPathExistence(path, isUserLoggedIn = false, owncloudClient)
```
abelgardep, line 45: Do not remove the comment with step 1 if we keep the step 2 😉

BTW, I would keep the named params

abelgardep (changes requested):
@theScrabi Tiny suggestions here

abelgardep (approved):
Approved. Let's QA this @jesmrec

jesmrec:
This is working from my side. Since this change is transversal (affects everything that need some network operation), the way to do was checking same things as the network library switch.

- [X] Smoke test auth
- [X] Transfers (upload & download)
- [X] Automatic tests -> https://reports.cucumber.io/reports/275205ff-6bd7-475d-b311-6dc078f8106d
- [X] Additionaly, https://github.com/owncloud/android/issues/2604 was checked
- [X] Fixed a old problem about switching accounts from same server in same device

From my side this is  ready to go. Great job, @theScrabi!

Beta version should help to find any other corner case.

PR #3058, opened 2021-01-25
Introduce Dependabot

hannesa2:
This will help you to keep libraries up to date
Please see details here https://dependabot.com/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gradle" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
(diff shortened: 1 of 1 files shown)


abelgardep:
Thanks for the contribution, we will take a look. Does it update every module's dependencies? Or it needs additional configuration to do that?

michaelstingl:
Hm, we already used Dependabot, or at least I found traces. But it was most about the NodeJS docs stuff that also lives in this repo…(https://github.com/owncloud/android/pull/2951, https://github.com/owncloud/android/pull/2946)

hannesa2:
If it perform the submodules as well would surprise me, but I don't know.
Merge it and you will see. You can't destroy anything, because it only makes pull request

hannesa2:
> Hm, we already used Dependabot, or at least I found traces. But it was most about the NodeJS docs stuff that also lives in this repo…(#2951, #2946)

Interesting. It's triggered for NodeJS stuff without control file, which I introduce here.

Btw, an additional showcase about bitrise which waits till 3 days. But you are lucky with it, so there is no reason to care about
image

jesmrec:
need to move to public repo, the point is that you pushed from external repo (even when you are a member of the company)

hannesa2:
> need to move to public project, the point is that you pushed from external repo (even when you are a member of the Android orga)

I'm not sure what you mean A) bitrise sucking B) dependabot integration ?
Anyway, I hear the call to have the branch in upstream repo. With pleaseure I follow it, here you go #3067

jesmrec:
Thanks a lot

PR #3066, opened 2021-01-28
PreviewImageFragment to Kotlin

hannesa2:
This is a pre-task for an other change and I don't want to differ too much with upstream.
That's why it would be nice, if you accept it. There is no logic change at all.

It's just a Kotlin conversion


jesmrec:
Thanks. We'll review it

hannesa2:
I moved it to upstream repo #3068
so you can rebase and change what you want

PR #3068, opened 2021-01-29
PreviewImageFragment to Kotlin

hannesa2:
This is a pre-task for an other change and I don't want to differ too much with upstream.
That's why it would be nice, if you accept it. There is no logic change at all.

It's just a Kotlin conversion


owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
```diff
@@ -0,0 +1,389 @@
...
+
+                override fun onResourceReady(
+                    resource: Drawable?, model: Any, target: Target<Drawable?>,
+                    dataSource: DataSource, isFirstResource: Boolean
+                ): Boolean {
+                    Timber.d("Loading image %s", file.fileName)
+                    return false
+                }
+            })
+            .into(photo_view!!)
```
abelgardep, line 330: When we transform some java code to Kotlin, we should try to improve it. There are a lot of `!!` in this file that could be improved. It would be beneficial in the long-term from my point of view. What do you think?

hannesa2: Ok, I removed them. But it's always a question where to start and where to end. 
I only want to have it done

owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
```diff
@@ -285,23 +285,18 @@ class PreviewImageFragment : FileFragment() {
     }
 
     override fun onFileMetadataChanged() {
-        val storageManager = mContainerActivity.storageManager
-        if (storageManager != null) {
-            file = storageManager.getFileByPath(file.remotePath)
-        }
+        file = mContainerActivity.storageManager?.getFileByPath(file.remotePath)
```
abelgardep: In this case, do we want to set file to null when storage manager is null? Previously file was only updated when storage manager was not null.

hannesa2: It doesn't mean that before it was good. Show nothing, previous file.... no clue. All doesn't feel good.

Anywhay the `?` was pointless. I removed it

hannesa2: Btw the construct with `mContainerActivity` is questionable, but out of this topic

abelgardep: > It doesn't mean that before it was good.

I agree. I was just wondering because it had different behavior.

hannesa2:
Here bitrise since 4 h pending, I guess it will look like same in >10 days too

image

abelgardep (approved):
Approved on my side, ready to QA @jesmrec

jesmrec:
Checked Preview of images. Everything fine.

Go ahead with merging

hannesa2:
When I review your review, then I would say not all is fine. **Green** is yours

image

Especially the reintroduce this twice constant `const val EXTRA_FILE = "FILE"`

hannesa2:
It guess you rebased and pushed forced it without my last commit !

Please use always `git push --force-with-lease` to avoid this

hannesa2:
I added my last commit again

abelgardep:
Thanks for the suggestion. I rebased and pushed, yes. But that commit was never in this branch, I have checked it in Bitrise builds and there were no builds for that commit. So, I think that you pushed it to your repo, but not to this one.

jesmrec:
Everything perfect. 

Approved

PR #3069, opened 2021-01-29
Fix type and code format

hannesa2:
It's just a type and code format.

In my point of view, there is not much to test

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
@@ -353,7 +353,7 @@ class FileDataStorageManager {
             put(FILE_ETAG, file.etag)
             put(FILE_TREE_ETAG, file.treeEtag)
             put(FILE_SHARED_VIA_LINK, if (file.isSharedViaLink) 1 else 0)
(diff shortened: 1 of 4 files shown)


hannesa2:
Bitrise !

image

abelgardep:
Thanks for the contribution. However, when a file is shared with a group or other users, we call them "sharees", so there is no typo there.

BTW, bitrise fails because you updated the model but not the [unit test](https://github.com/owncloud/android/blob/master/owncloudApp/src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java)

hannesa2:
> BTW, bitrise fails because you updated the model but not the [unit test](https://github.com/owncloud/android/blob/master/owncloudApp/src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java)

I don't see this

hannesa2:
> However, when a file is shared with a group or other users, we call them "sharees", so there is no typo there.

@abelgardep 
Ok, "sharees" could be the wrong plural of "share" 🤨
But what are "sharee" ? I new word created by ownCloud ?  😄 Sorry, only insider will understand this
From where does the second **e** comes from ?

Anyway, all is ok, as long as someone knows the deeper sense of it 👍 I do understand it only somehow after this attempt to fix it

PR #3070, opened 2021-01-29
Android Studio and Gradle update

hannesa2:
The Gradle changes are the result of 

image

You already have a Github action to verify 

Library: https://github.com/owncloud/android-library/pull/369


hannesa2:
Bitrise !
image

abelgardep:
Thanks for the contribution @hannesa2. I have not updated AS to 4.1.2 yet. Let me update and check the Gradle changelog to assure it does not break anything 👍

abelgardep:
Why do you need to run that build wrapper now? I mean, if you run that task at master, it will update those files too. But it was not needed for the project compilation, was it? If we just update the Gradle version, it compiles as expected. So, I'm curious about this change. Could you help me to understand it @hannesa2 ?

hannesa2:
Try it out your self. Normally you only update version string and it works, but gradle is more then just gradle.properties.
With the Gradle task you update all corresponding files

hannesa2:
Here you can read more https://docs.gradle.org/6.8.1/release-notes.html#upgrade-instructions

abelgardep:
Ok, so it's not needed but it is good to update the wrapper from time to time. Could you update the library too?

hannesa2:
> the library

?

abelgardep:
I mean, if you can update AS and Gradle in https://github.com/owncloud/android-library

hannesa2:
This is an other repo and out of context of this one. But I can do

hannesa2:
Here you go https://github.com/owncloud/android-library/pull/369

abelgardep (approved):
Thanks, @hannesa2 LGTM

jesmrec:
Approved

Issue #3054, opened 2021-01-18
[FEATURE REQUEST] OIDC Dynamic Client Registration support


jesmrec:
This issue tracks support for OIDC Dynamic Client Registration support 

Info about it:

- https://tools.ietf.org/html/rfc7591
- https://openid.net/specs/openid-connect-registration-1_0.html

New endpoint `register` is available to get the `client_id` and `secret_id` that will be used in the token request. The location of the endpoint is available on `.well-known`. The `secret_id` has its own validity time. 

OIDC Dynamic Client Registration is available on `ocis.owncloud.works`.

Other clients' stuff:

iOS-sdk commit: https://github.com/owncloud/ios-sdk/commit/85e09cbc1fa9fbd5ddab961787d98130c0b2f1f2
Desktop client PR: https://github.com/owncloud/client/pull/8350
Web: https://github.com/owncloud/web/pull/4286

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/oidc_dynamic_client_registration
- [x] Development tasks
  - [x] Create a new remote operation to register a new client (library)
  - [x] Create a new use case to register a new client
  - [x] Call new use case when needed
  - [x] Handle token renewal
  - [x] Add unit tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/oidc_dynamic_client_registration into master

### PR

- App https://github.com/owncloud/android/pull/3073
- Library https://github.com/owncloud/android-library/pull/367

PR #3073, opened 2021-02-01
[FEATURE] OIDC Dynamic Client Registration

abelgardep:
Implements https://github.com/owncloud/android/issues/3054
Needs https://github.com/owncloud/android-library/pull/367

___

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.17/Dynamic%20Client%20Registration.md

Bugs & improvements:

- [X] (1) Register endpoint not called, with OIDC + oC10 https://github.com/owncloud/android/pull/3073#issuecomment-781195629 [WONT FIX HERE]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
@@ -38,9 +38,7 @@ import com.owncloud.android.R;
 import com.owncloud.android.authentication.oauth.OAuthUtils;
 import com.owncloud.android.domain.UseCaseResult;
 import com.owncloud.android.domain.authentication.oauth.OIDCDiscoveryUseCase;
-import com.owncloud.android.domain.authentication.oauth.RegisterClientUseCase;
 import com.owncloud.android.domain.authentication.oauth.RequestTokenUseCase;
-import com.owncloud.android.domain.authentication.oauth.model.ClientRegistrationInfo;
 import com.owncloud.android.domain.authentication.oauth.model.OIDCServerConfiguration;
 import com.owncloud.android.domain.authentication.oauth.model.TokenRequest;
 import com.owncloud.android.domain.authentication.oauth.model.TokenResponse;
@@ -54,6 +52,9 @@ import timber.log.Timber;
 
 import java.io.File;
 
+import static com.owncloud.android.data.authentication.AuthenticationConstantsKt.KEY_CLIENT_REGISTRATION_CLIENT_EXPIRATION_DATE;
+import static com.owncloud.android.data.authentication.AuthenticationConstantsKt.KEY_CLIENT_REGISTRATION_CLIENT_ID;
+import static com.owncloud.android.data.authentication.AuthenticationConstantsKt.KEY_CLIENT_REGISTRATION_CLIENT_SECRET;
 import static com.owncloud.android.data.authentication.AuthenticationConstantsKt.KEY_OAUTH2_REFRESH_TOKEN;
 import static com.owncloud.android.presentation.ui.authentication.AuthenticatorConstants.KEY_AUTH_TOKEN_TYPE;
 import static org.koin.java.KoinJavaComponent.inject;
@@ -183,7 +184,7 @@ public class AccountAuthenticator extends AbstractAccountAuthenticator {
             // OAuth, gets an auth token from the AccountManager's cache. If no auth token is cached for
             // this account, null will be returned
             accessToken = accountManager.peekAuthToken(account, authTokenType);
-            if (accessToken == null && canBeRefreshed(authTokenType)) {
+            if (accessToken == null && canBeRefreshed(authTokenType) && clientSecretIsValid(accountManager, account)) {
                 accessToken = refreshToken(account, authTokenType, accountManager);
             }
         }
@@ -198,6 +199,31 @@ public class AccountAuthenticator extends AbstractAccountAuthenticator {
 
         /// if not stored, return Intent to access the LoginActivity and UPDATE the token for the account
         return prepareBundleToAccessLoginActivity(accountAuthenticatorResponse, account, authTokenType, options);
+    }
+
+    /**
+     * Check if the client has expired or not.
+     * If the client has expired, we can not refresh the token and user needs to re-authenticate.
+     *
+     * @return true if the client is still valid
+     */
+    private boolean clientSecretIsValid(AccountManager accountManager, Account account) {
+        String clientSecretExpiration = accountManager.getUserData(account,
+                KEY_CLIENT_REGISTRATION_CLIENT_EXPIRATION_DATE);
+
+        Timber.d("Client secret expiration [" + clientSecretExpiration + "]");
+        if (clientSecretExpiration == null) {
+            return true;
+        }
+
+        long currentTimeStamp = System.currentTimeMillis() / 1000L;
+        int clientSecretExpirationInt = Integer.parseInt(clientSecretExpiration);
+        boolean clientSecretIsValid = clientSecretExpirationInt == 0 || clientSecretExpirationInt > currentTimeStamp;
+
+        Timber.d("Current time [" + currentTimeStamp + "]");
+        Timber.d("Client is valid [" + clientSecretIsValid + "]");
+
+        return clientSecretIsValid;
     }
 
     @Override
@@ -326,8 +352,18 @@ public class AccountAuthenticator extends AbstractAccountAuthenticator {
                 oidcDiscoveryUseCase.getValue().execute(oidcDiscoveryUseCaseParams);
 
         String tokenEndpoint;
-        String clientId = mContext.getString(R.string.oauth2_client_id);
-        String clientSecret =mContext.getString(R.string.oauth2_client_secret);
+
+        String clientId = accountManager.getUserData(account, KEY_CLIENT_REGISTRATION_CLIENT_ID);
+        String clientSecret = accountManager.getUserData(account, KEY_CLIENT_REGISTRATION_CLIENT_SECRET);
+
+        if (clientId == null) {
+            Timber.d("Client Id not stored. Let's use the hardcoded one");
+            clientId = mContext.getString(R.string.oauth2_client_id);
+        }
+        if (clientSecret == null) {
+            Timber.d("Client Secret not stored. Let's use the hardcoded one");
+            clientSecret = mContext.getString(R.string.oauth2_client_secret);
+        }
 
         if (oidcServerConfigurationUseCaseResult.isSuccess()) {
             Timber.d("OIDC Discovery success. Server discovery info: [ %s ]",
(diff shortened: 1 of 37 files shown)


owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -198,6 +201,31 @@ public Bundle getAuthToken(AccountAuthenticatorResponse accountAuthenticatorResp
         return prepareBundleToAccessLoginActivity(accountAuthenticatorResponse, account, authTokenType, options);
     }
 
+    /**
+     * Check if the client has expired or not.
+     * If the client has expired, we can not refresh the token and user needs to re-authenticate.
+     *
+     * @return true if the client is still valid
+     */
+    private boolean clientSecretIsValid(AccountManager accountManager, Account account) {
```
davigonz, line 210: Shouldn't the server take care of this?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -370,6 +375,35 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+                    it.peekContent().getStoredData()?.let { clientRegistrationInfo ->
+                        performGetAuthorizationCodeRequest(
+                            authorizationEndpoint,
+                            clientRegistrationInfo.clientId
+                        )
+                    }
+                }
+                is UIResult.Error -> {
+                    Timber.e(it.peekContent().getThrowableOrNull(), "Client registration failed.")
+                    Toast.makeText(this, it.peekContent().getThrowableOrNull().toString(), Toast.LENGTH_SHORT).show()
```
davigonz, line 401: What about creating an activity extension for this and just call `showError`?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -429,8 +466,15 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
      */
     private fun exchangeAuthorizationCodeForTokens(authorizationCode: String) {
         server_status_text.text = getString(R.string.auth_getting_authorization)
-        val clientAuth =
+
+        val clientRegistrationInfo = oauthViewModel.registerClient.value?.peekContent()?.getStoredData()
+
+        val clientAuth = if (clientRegistrationInfo?.clientId != null && clientRegistrationInfo.clientSecret != null) {
+            OAuthUtils.getClientAuth(clientRegistrationInfo.clientSecret as String, clientRegistrationInfo.clientId)
```
davigonz, line 473: About this `as String`, please have a look at https://kotlinlang.org/docs/reference/typecasts.html#unsafe-cast-operator, i think that it could throw an exception

Screenshot 2021-02-10 at 12 14 17

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/authentication/OCAuthenticationViewModelTest.kt
```diff
@@ -160,7 +161,13 @@ class OCAuthenticationViewModelTest : ViewModelTest() {
     fun loginOAuthOk() {
         every { loginOAuthAsyncUseCase.execute(any()) } returns UseCaseResult.Success(OC_BASIC_USERNAME)
         ocAuthenticationViewModel.loginOAuth(
-            OC_BASIC_USERNAME, OC_AUTH_TOKEN_TYPE, OC_ACCESS_TOKEN, OC_REFRESH_TOKEN, OC_SCOPE, null
+            username = OC_BASIC_USERNAME,
+            authTokenType = OC_AUTH_TOKEN_TYPE,
+            accessToken = OC_ACCESS_TOKEN,
+            refreshToken = OC_REFRESH_TOKEN,
+            scope = OC_SCOPE,
+            updateAccountWithUsername = null,
```
davigonz, line 169: Could we initialise `updateAccountWithUsername` to null in `loginOAuth` method and remove this line?

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/authentication/OCAuthenticationViewModelTest.kt
```diff
@@ -175,7 +182,13 @@ class OCAuthenticationViewModelTest : ViewModelTest() {
     fun loginOAuthException() {
         every { loginOAuthAsyncUseCase.execute(any()) } returns UseCaseResult.Error(commonException)
         ocAuthenticationViewModel.loginOAuth(
-            OC_BASIC_USERNAME, OC_AUTH_TOKEN_TYPE, OC_ACCESS_TOKEN, OC_REFRESH_TOKEN, OC_SCOPE, null
+            username = OC_BASIC_USERNAME,
+            authTokenType = OC_AUTH_TOKEN_TYPE,
+            accessToken = OC_ACCESS_TOKEN,
+            refreshToken = OC_REFRESH_TOKEN,
+            scope = OC_SCOPE,
+            updateAccountWithUsername = null,
```
davigonz, line 190: Same here

owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCLocalAuthenticationDataSource.kt
```diff
@@ -106,6 +111,18 @@ class OCLocalAuthenticationDataSource(
...
+            clientRegistrationInfo?.let { clientRegistrationInfo ->
+                accountManager.setUserData(it, KEY_CLIENT_REGISTRATION_CLIENT_ID, clientRegistrationInfo.clientId)
+                accountManager.setUserData(
+                    it, KEY_CLIENT_REGISTRATION_CLIENT_SECRET, clientRegistrationInfo.clientSecret
+                )
+                accountManager.setUserData(
+                    it,
+                    KEY_CLIENT_REGISTRATION_CLIENT_EXPIRATION_DATE,
+                    clientRegistrationInfo.clientSecretExpiration.toString()
+                )
```
davigonz, line 123:
```suggestion
accountManager.apply {
    setUserData
    setUserData
    setUserData
} 
```

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasource/impl/RemoteOAuthDataSourceImpl.kt
```diff
@@ -60,4 +65,24 @@ class RemoteOAuthDataSourceImpl(
...
+                    registrationEndpoint = clientRegistrationRequest.registrationEndpoint,
+                    clientName = clientRegistrationRequest.clientName,
+                    redirectUris = clientRegistrationRequest.redirectUris,
+                    tokenEndpointAuthMethod = clientRegistrationRequest.tokenEndpointAuthMethod,
+                    applicationType = clientRegistrationRequest.applicationType
+                )
+            )
+        }
+
+        return remoteClientRegistrationInfoMapper.toModel(remoteClientRegistrationInfo)!!
```
davigonz, line 86: !! 😮

owncloudData/src/main/java/com/owncloud/android/data/oauth/mapper/RemoteClientRegistrationInfoMapper.kt
```diff
@@ -0,0 +1,41 @@
...
+
+        return ClientRegistrationInfo(
+            clientId = remote.clientId,
+            clientSecret = remote.clientSecret,
+            clientIdIssuedAt = remote.clientIdIssuedAt,
+            clientSecretExpiration = remote.clientSecretExpiration
+        )
+    }
+
+    // Not needed
```
davigonz, line 38: This is a bit weird and it's probably mine 🤦 , you could also start using extensions to map from different objects instead of mappers. That way you create the extension you need and avoid empty inherited methods.

davigonz (changes requested):

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasource/impl/RemoteOAuthDataSourceImpl.kt
abelgardep: Mappers :( I don't like it either. `remoteClientRegistrationInfo` is not nullable, so `toModel` will return a non-nullable model. It is done that way since the beginning. I could create an issue to replace mappers (at least for remote) with extensions since they introduce `!!`

owncloudData/src/main/java/com/owncloud/android/data/oauth/mapper/RemoteClientRegistrationInfoMapper.kt
abelgardep: Yes, I will create a new issue to replace remote mappers. I think that extensions are better, at least, for one-way mapping. We will remove some unnecessary code.

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/authentication/OCAuthenticationViewModelTest.kt
abelgardep: Yes, thanks 🍻

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
abelgardep: We are checking that it is not null in the line above. So it is not null, thanks for it anyway 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
abelgardep: Ups, if that operation fails, we don't need to print the error, but follow the login flow.

owncloudApp/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
abelgardep: If we take care of it, we can save at least 2 network calls that will fail otherwise.

davigonz: Ok and what happens if the user changes the device time?

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasource/impl/RemoteOAuthDataSourceImpl.kt
davigonz: Perfect

davigonz (approved):

jesmrec:
### (1)

1. Set a server with OIDC + oC10 and DCR
2. Enter the URL and credentials

Current: 

Credentials are accepted but finally, the process finishes with no success. Checking the request flow, i realised that the `register` endpoint is not requested, so the final token request is done with the default `client_id` instead of the dynamic one.

Expected: 

Authentication correct

Google Pixel 2, Android11
Emulator Nexus 5X Android8
commit `12d336a6`

abelgardep:
About (1)

I have checked the problem and this is what's happening. Taking into account that you are using some kind of proxy.

1. Once you introduce this URL: `oc-10-6-0.oidc-2-0-0rc1-20210128.jw-qa.owncloud.works/` you are asked to accept the untrusted certificate. Then you accept and you can move forward.
2. Then we perform a discovery request, and we receive this registration endpoint: `https://konnect.oidc-2-0-0rc1-20210128.jw-qa.owncloud.works/konnect/v1/register`
3. We try to ask the registration endpoint, but it is not trusted, so a certificate exception is triggered, and the client is not registered.
4. Since client registration failed, we use the hardcoded client id, and we perform the token request to `https://konnect.oidc-2-0-0rc1-20210128.jw-qa.owncloud.works/signin/v1/identifier/_/authorize` but it fails again because the certificate is not trusted.

So, conclusion:
Checking it with trusted certificates, it won't fail and you are able to log in properly.
If you trust the initial server, and the other endpoints in the discovery request are in the same trusted server (for example `ocis.owncloud.works`), then you are able to log in.

owncloudData/src/main/java/com/owncloud/android/data/oauth/mapper/RemoteClientRegistrationInfoMapper.kt
abelgardep: Issue: https://github.com/owncloud/android/issues/3091

jesmrec:
thanks for the explanation @abelgardep. That means that (1) is not a problem of DCR itself, and also not a problem in this branch, so it will be addressed and prioritized in a separate issue.

jesmrec:
About this issue, several regards.

- As https://github.com/owncloud/openidconnect/issues/142, `client_id` expires and does not renew. This leads to a non-authorized status after the first token renewal since the `client_id` has expired. Known issue to be fixed in the idP side.

- If the idP is hosted under a non-secured machine or certificate, user will not be warned to accept the cert (as the initial URL does) and finally, the default `client_id` and  `secret_id` will be used, pointing to error because the expected ones are the dynamic ones. This is a collateral-effect, that will be managed in https://github.com/owncloud/android/issues/3109. (Report (1) above )

- This known problem: https://github.com/owncloud/android/issues/3107 was detected and reproduced while testing this feature. 

- The main feature is checking the discovery endpoint, and in case register is included, get the `client_id` and `secret_id` from them. This is correctly achieved

Taking these issues in account, we can move this forward

PR #2970, opened 2020-09-23
[Fix] Handle redirections with cookies in OAuth

abelgardep:
Needs: https://github.com/owncloud/android-library/pull/343

_____

BUGS & IMPROVEMENTS

- [x] (1) App crashes after validating OIDC URL https://github.com/owncloud/android/pull/2970#issuecomment-705020422 [FIXED]
- [x] (2) Cookie handling in regular oC server https://github.com/owncloud/android/pull/2970#issuecomment-708319259

PR #3074, opened 2021-02-02
[Fix] Cookie handling

abelgardep:
Continuation https://github.com/owncloud/android/pull/2970

Needs owncloud/android-library#368

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountUtils.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountUtils.java
@@ -203,13 +203,6 @@ public class AccountUtils {
                                 accountMgr.getUserData(account, Constants.KEY_OC_VERSION)
                         );
 
-                        // copy cookies
-                        accountMgr.setUserData(
-                                newAccount,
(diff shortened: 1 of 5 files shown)


theScrabi (approved):

abelgardep:
Ready to test @jesmrec

jesmrec:
QA checks:

- [X] Redirections 301: working fine. It will be needed an improvement to address in a separate issue: users must be notified when the redirections happen, something similar as in iOS app:

Screenshot 2021-03-03 at 13 12 17

The current behaviour is correct.

- [X] Cookie handling. Problem reproduced [here](https://github.com/owncloud/android/pull/2970#issuecomment-712952345) is not reproducible anymore. No cookie accumulation when accounts are switched. 

I detected some cookie leaks between accounts in the same server, but i think that this is something that will be solved in https://github.com/owncloud/android/pull/3047. Is that correct @theScrabi ??

jesmrec:
Let's move forward this. It contains some specific-customer content, so, we will await for some feedback. 

Redirections are working fine. 

Approved on my side

Issue #2847, opened 2020-04-15
[FEATURE REQUEST] New search bar


abelgardep:
Continuation of #2742

First mock up:

![newUi_toolbar](https://user-images.githubusercontent.com/47524927/79344588-5043cd80-7f30-11ea-909e-fda2c7ccb3fd.png)

ACs:

- New search bar will be automatically shown without pressing lens icon (root)
- Lens icon is gone (root)
- Searching means filtering: only in the current folder (same behaviour as current)
- ownCloud word in the hint must be brandable
- If folder is not root, same UI (lens and name of the folder)
- If folder is empty, bar is disabled (N2H)
- Extended top bar at first and then a regular bar. Extended top bar will show the oC icon on the top
- Search bar in all tabs but in uploads 
- Refresh account must be removed from three-dot-menu on the file list to account manager (check if it does not hurt)
- In pickers/share with oC view, Toolbar will show the ownCloud label, the lens, but the avatar 

![image](https://user-images.githubusercontent.com/60343863/100236634-35a5ee00-2f2e-11eb-95a8-e53eaa4d574d.png)

Figma Mockup
https://www.figma.com/file/EVQQcpITjVPiF4yT3eWKuf/Android-Proposals?node-id=0%3A1

### TASKS

- [x] Research (if needed)
- [x] Create branch new_design/search_bar
- [x] Development tasks
  - [x] Implement new toolbar design
  - [x] Connect new toolbar search view with current listeners
  - [X] Move refresh account option to account management
  - [x] Check icons color in account management view
  - [x] Show avatar in root toolbar
  - [x] Use new design in root view and old one in rest of the app
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch new_design/search_bar into master

### PR

- App
- Library (if needed)

abelgardep:
### AC to discuss:

1. Hint root folder -> Search in `app_name` (brandable)
2. Hint other folders -> Search in `folder_name`
3. What happens with uploads toolbar (search is not available there at the moment)?
4. What happens with Av offline and share by link toolbar?

michaelstingl:
@abelgardep examples in other apps you’d recommend to look at?

abelgardep:
> @abelgardep examples in other apps you’d recommend to look at?

You can check some google apps: Drive, GPlay, Gmail for example.

jesmrec:
> 1. Hint root folder -> Search in app_name (brandable)

[...]

Issue #2596, opened 2019-07-01
[BUG] clear all finished in upload view clears failed alse


mstojanovic:
### Actual behaviour
-when clicking on clear all finished in the upload view all streams are cleared

### Expected behaviour
-only uploaded streams should be cleared from the view
 
### Steps to reproduce
1. upload a few files
2. make some fail and let some upload
3. go to the upload view
4. select clear all finished from the dot menu

Can this problem be reproduced with the official owncloud server?
did not try, but not relevant

### Environment data
Android version: 9

Device model: samsung s9

Stock or customized system: stock

ownCloud app version: 2.11.1 release f5c5096a7

ownCloud server version: not relevant

#### Web server error log
not relevant

#### ownCloud log (data/owncloud.log)
not present on at given location

jesmrec:
Thanks for reporting!

My suggestion is: the options to clear should match the sections in the list. So, the following options should be available:

- Clear successful
- Clear failed
- Clear pending 
- Clear all (current will not be removed)
- Retry failed

clear all finished leads to confusion, i'd get rid of it. Open to new ideas.

mstojanovic:
i could live with that as long as the actions are what they say

jesmrec:
Next release 2.17, a new option to clear will be included for every list:

https://github.com/owncloud/android/issues/2847#issuecomment-749696480

So, this ticked is solved with such fix. 

Thanks for reporting and fell free to reopen in case something is missing.

PR #3087, opened 2021-02-05
[New design] Toolbar

abelgardep:
Implements #2847 

It may fix an old bug https://github.com/owncloud/android/issues/2596

____

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.17/3087-new_toolbar_design.md

Bugs & improvements:

- [x] (1) Black hamburger menu https://github.com/owncloud/android/pull/3087#issuecomment-784008229 [FIXED]
- [x] (2) Back arrow in images preview https://github.com/owncloud/android/pull/3087#issuecomment-784050267 [FIXED]
- [x] (3) Lens icon in uploads view https://github.com/owncloud/android/pull/3087#issuecomment-784069418 [FIXED]
- [x] (4) Avatar size https://github.com/owncloud/android/pull/3087#issuecomment-784140145 [DONE]


owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1151,20 +1124,26 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
...
+                    FileListOption.AV_OFFLINE -> resources.getString(R.string.drawer_item_only_available_offline)
+                    FileListOption.SHARED_BY_LINK -> resources.getString(R.string.drawer_item_shared_by_link_files)
+                    FileListOption.ALL_FILES -> resources.getString(R.string.default_display_name_for_root_folder)
                 }
-            updateActionBarTitleAndHomeButtonByString(title)
+            setupToolbar(ToolbarConfig.ToolbarRoot(title, true))
+            listOfFilesFragment?.setSearchListener(findViewById(R.id.root_toolbar_search_view))
+        } else {
+            setupToolbar(ToolbarConfig.ToolbarStandard(
+                title = chosenFile?.fileName!!,
```
davigonz, line 1143: !! 😮

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
```diff
@@ -99,8 +99,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
         );
 
-        setupToolbar();
-        updateActionBarTitleAndHomeButtonByString(getResources().getString(R.string.prefs_manage_accounts));
+        ToolbarConfig.ToolbarStandard toolbarConfig = new ToolbarConfig.ToolbarStandard(
+                getResources().getString(R.string.prefs_manage_accounts),
+                true,
+                true
+        );
+        setupToolbar(toolbarConfig);
```
davigonz, line 107: Instead of setting up the toolbar in so many activities, why don't you create some parent activities like `ToolbarStandardActivity`? Is it possible?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/PrivacyPolicyActivity.java
```diff
@@ -100,12 +101,10 @@ public void onReceivedError(WebView view, int errorCode, String description, Str
...
-        switch (item.getItemId()) {
-            case android.R.id.home:
-                finish();
-                break;
-            default:
-                retval = super.onOptionsItemSelected(item);
+        if (item.getItemId() == android.R.id.home) {
+            finish();
+        } else {
+            retval = super.onOptionsItemSelected(item);
```
davigonz, line 107: I know that this name comes from time ago but what does `retval` mean?  Could we use a more descriptive name?

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/UploadListFragment.java
```diff
@@ -104,6 +106,31 @@ public boolean onChildClick(ExpandableListView parent, View v, int groupPosition
 
     }
 
+    public enum OptionsInUploadList {
+        CLEAR_FAILED, CLEAR_SUCCESSFUL, RETRY_FAILED;
+    }
+
+    @Override
+    public void onClick(OptionsInUploadList option) {
```
davigonz, line 114: This is not toolbar related, isn't it?

owncloudApp/src/main/res/layout/owncloud_toolbar.xml
```diff
@@ -0,0 +1,97 @@
...
+            android:textSize="@dimen/toolbar_title_text_size"
+            app:drawableEndCompat="@drawable/ic_search"
+            app:drawableTint="@color/white"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toStartOf="@id/root_toolbar_avatar"
+            app:layout_constraintStart_toEndOf="@id/root_toolbar_left_icon"
+            app:layout_constraintTop_toTopOf="parent"
+            tools:text="ownCloud" />
+
+        <com.google.android.material.imageview.ShapeableImageView
```
davigonz, line 76: Why `ShapeableImageView`? To be protected against design changes, right? 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/toolbar/ToolbarActivity.kt
```diff
@@ -0,0 +1,120 @@
...
+        with(searchView) {
+            isVisible = false
+            setOnCloseListener {
+                searchView.visibility = View.GONE
+                toolbarTitle.visibility = VISIBLE
+                false
+            }
+        }
+
+        with(avatarView) {
```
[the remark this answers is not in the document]
abelgardep: Yes, it is needed to set the onClickListener.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
abelgardep: All of them extends from ToolbarActivity. The problem here is that we need to set a title and if we need to set the back arrow. I tried to keep a similar behavior, but I considered some different options too. For example, create a new component OwncloudToolbar, and set in the xml the title, back navigation, etc using attrs. But it was too complex.

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/UploadListFragment.java
abelgardep: Well, it is related because we need to remove the three-dot menu from the uploads fragment to set the avatar. So, we have moved those options to the fragment.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/PrivacyPolicyActivity.java
abelgardep: TBH, I don't know what it means...

owncloudApp/src/main/res/layout/owncloud_toolbar.xml
abelgardep: Nothing special, I have changed it to AppCompatImageView. We will move to material in the future 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/toolbar/ToolbarActivity.kt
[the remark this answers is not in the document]
davigonz: I mean this:

Screenshot_2021-02-16_at_11_08_32

You are using apply but later writing `avatarView` again instead of `this`

Besides, `AccountUtils.getCurrentOwnCloudAccount(context) ?: return` doesn't use `avatarView` so I would call it out of the `apply`.

I would not add `AvatarUtils().loadAvatarForAccount(` call to the `apply` either, because we are not operating on the members of the receiver object, i.e. configuring the `avatarView` with that call.

```
AccountUtils.getCurrentOwnCloudAccount(context) ?: return

with(avatarView) {
   AvatarUtils().loadAvatarForAccount(
        avatarView,
        AccountUtils.getCurrentOwnCloudAccount(context),
        true,
        context.resources.getDimension(R.dimen.toolbar_avatar_radius)
    )
}

avatarView.setOnClickListener {
        startActivity(Intent(context, ManageAccountsActivity::class.java))
    }
}
```

Do you see the difference?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
davigonz: Ok, I've asked this in order to avoid some logic in activities like https://github.com/owncloud/android/pull/3087/files#diff-0d32bb6cd423d0567750a9f31037db8e2edadc51b3f583685d85a3aa865d37fcR1132 . Maybe creating something like

```
class Toolbar(
    val title: String?,
    val displayHomeAsUpEnabled: Boolean,
    val homeButtonEnabled: Boolean,
    val enableSearch: Boolean
)
```

And a

```
class ToolbarCreator() {
    fun getToolbar(
        remotePath: String
    ): Toolbar {
        if (remotePath == "whatever"...) {
              return Toolbar (whatever)
        }
    }
}
```

From a viewmodel calls:

toolbarCreator.getToolbar( )

And you set the toolbar to the fragment via a liveData so the fragment doesn't know anything about the logic.

It's just an idea if you want to do something like that in the future.

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/UploadListFragment.java
davigonz: We don't have a viewmodel here, right? To move the `TransferRequester` and `UploadsStorageManager` dependencies there and call viewmodel methods here.

abelgardep: Not yet, unfortunately.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
abelgardep: Fixed in a nice way from my point of view.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
abelgardep: Hard to do at the moment. Let's wait a little bit more till the new arch is finished.

davigonz (approved):

jesmrec:
Let's QA this

jesmrec:
### (1)

1. Browse to a non-root folder
2. Turn orientation to landscape

Current: Back arrow turns to black hamburger:

Screenshot 2021-02-23 at 09 38 10

By turning to portrait again, the black hamburger persists

Expected: white back arrow instead

Pixel2 Android11
`c295f080`

jesmrec:
### (2)

Probably related with (1)

1. Open an image file

Current: Back arrow is gone at the top left
Expected: Back arrow is there to browse back to the file list

No matter portrait or landscape

Pixel2 Android11
`c295f080`

jesmrec:
### (3)

1. Open Uploads view

Current: In the uploads view, the lens icon is visible and does nothing

Expected: 
 - The easiest one: hide the lens icon
 - The complex one: make the lens icon work, searching by pattern in the list

both are valid for me. Probably the second one makes more sense in a separate issue.

Pixel2 Android11
`c295f080`

jesmrec:
### (4)

This is only a question, open to discuss. Don't you think the avatar is a bit small?:

Screenshot 2021-02-23 at 12 35 41

Would yo make it a bit bigger? is this feasible?

@abelgardep @theScrabi @JuancaG05 @michaelstingl

JuancaG05:
> ### (4)
> This is only a question, open to discuss. Don't you think the avatar is a bit small?:
> 
> Screenshot 2021-02-23 at 12 35 41
> 
> Would yo make it a bit bigger? is this feasible?
> 
> I've compared it with other apps (like Google ones) and there seems to be slightly bigger
> 
> @abelgardep @theScrabi @JuancaG05 @michaelstingl

Yes, maybe it's a bit small compared to other apps' profile picture views, if it could be a bit bigger it would be great from my point of view

abelgardep:
**(3)** Should be fixed now
**(4)** Size increased, but now it is bigger than the search lens, what do you think about that?

jesmrec:
(3) is fixed

> (4) Size increased, but now it is bigger than the search lens, what do you think about that? 

for me, this is not a problem. Lens and avatar are different things. Better now.

abelgardep:
**(1)** and **(2)** should be fixed now. Ready to test again @jesmrec

jesmrec:
QA finished. Approved. Nice work here!!!

Issue #3041, opened 2020-12-22
[BUG] ReceiveExternalFilesActivity randomly scrolls to top


theScrabi:
### Actual behaviour
When scrolling down a large list of directories in the [`ReceiveExternalFilesActivity`](https://github.com/owncloud/android/blob/f75ad5844e31e1aba17ee484fa26a72cafe25c66/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java#L109), the scroll view randomly jumps to the top of the list.

### Expected behaviour
Smooth scrolling all the way to the bottom without the list randomly jumping to the top.
 
### Steps to reproduce
1. Crate a directory with a large list of sub directories (>~ 300).
2. Go to another app and share a file
3. Share with ownCloud
4. Select the directory with the large amount of subdirectories
5. scroll to the bottom.

### Environment data
Android version: API 29

ownCloud app version: 2.15.3

theScrabi:
I was not able to find a single reason/bug why ListView behaves like this. However ListView is not the right tool for this purpose either. For lists with lots of items it is highly recommended to use [RecyclerView](https://developer.android.com/guide/topics/ui/layout/recyclerview). Imo this is simple to do for this specific Activity, however requires some time to be done.

shashankdaima:
Is it solved or not?

jesmrec:
Not solved yet

theScrabi:
Fix here: https://github.com/owncloud/android/pull/3117

theScrabi:
@jesmrec I yet found another issue while trying to fix this. It is regarding search in the ReceiveExternalFilesActivity.
Here is a description of what happens on current Master:

#### How to reproduce
1. Share a file with the android app
2. Search for a folder which is not the first one in the list
3. enter it
#### Expected
Selected folder is entered
#### Actual
Folder that would be first in the list without the search filter being applied is entered.

[This](https://github.com/owncloud/android/pull/3117/commits/01696863066fd0b69b19f9dd375f4ee11b05c084) commit fixes it.

PR #3117, opened 2021-02-25
fix scroll jump on share to directory

theScrabi:
fix for https://github.com/owncloud/android/issues/3041

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -350,7 +350,11 @@ public class ReceiveExternalFilesActivity extends FileActivity
     public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
         // click on folder in the list
         Timber.d("on item click");
-        Vector<OCFile> tmpfiles = getStorageManager().getFolderContent(mFile);
+        // get current representation of files:
+        // This does not necessarily mean this is the content of the current folder.
+        // If the user searches for a folder mAdapter.getFiles() returnes only the folders/files
+        // that match the currently entered search query.
+        Vector<OCFile> tmpfiles = mAdapter.getFiles();
         tmpfiles = sortFileList(tmpfiles);
 
         if (tmpfiles.size() <= 0) {
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -10,7 +10,7 @@
  * @author Abel García de Prada
  * @author John Kalimeris
  * Copyright (C) 2012  Bartek Przybylski
- * Copyright (C) 2020 ownCloud GmbH.
+ * Copyright (C) 2021 ownCloud GmbH.
  * <p>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -128,8 +128,10 @@ public class ReceiveExternalFilesActivity extends FileActivity
     private LocalBroadcastManager mLocalBroadcastManager;
     private SyncBroadcastReceiver mSyncBroadcastReceiver;
     private UploadBroadcastReceiver mUploadBroadcastReceiver;
-    private ReceiveExternalFilesAdapter mAdapter;
-    private Menu mMainMenu;
+    // this is inited laizly, when an acocunt is selected. If no account is selected but an instance of this would
+    // be crated it would result in an null pointer exception.
+    private ReceiveExternalFilesAdapter mAdapter = null;
+    private ListView mListView;
     private boolean mSyncInProgress = false;
     private boolean mAccountSelected;
     private boolean mAccountSelectionShowing;
@@ -149,7 +151,7 @@ public class ReceiveExternalFilesActivity extends FileActivity
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
-        prepareStreamsToUpload();
+        prepareStreamsToUpload(); //streams need to be prepared before super.onCreate() is called
 
         if (savedInstanceState == null) {
             mParents = new Stack<>();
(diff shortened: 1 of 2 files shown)


abelgardep:
I have tested it and it does not display any file. Could you recheck it?

Pixel 3XL Android 11

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -190,6 +191,9 @@ protected void onCreate(Bundle savedInstanceState) {
             fm.beginTransaction()
                     .add(taskRetainerFragment, TaskRetainerFragment.FTAG_TASK_RETAINER_FRAGMENT).commit();
         }   // else, Fragment already created and retained across configuration change
+
+        mAdapter = new ReceiveExternalFilesAdapter(
+                this, mFiles, getStorageManager(), getAccount());
```
abelgardep, line 196: At this point `getAccount()` could be null. You need to initialize it in `onAccountSet()` to ensure that an account is set.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -431,10 +435,9 @@ private void populateDirectoryList() {
         mFile = getStorageManager().getFileByPath(full_path);
         if (mFile != null) {
             Vector<OCFile> files = getStorageManager().getFolderContent(mFile);
-            files = sortFileList(files);
+            mFiles = sortFileList(files);
```
abelgardep, line 438: I think that you need to update directly the adapter's vector. And after updating, call `notifyDataSetChanged()`. `mFiles` is not needed in this file.

JuancaG05 (commented):
The code structure seems correct, but I couldn't test it, is there a way to create quickly 300 subdirectories?

abelgardep:
I have tested with multi-account and it crashes :(

Steps to reproduce it:
1 - Add two accounts
2 - Share with ownCloud

Expected:
1 - Add two accounts
2 - Share with ownCloud
3 - Choose the account you want to share with

@theScrabi

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -394,58 +414,44 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-                    this, files, getStorageManager(), getAccount());
-            mListView.setAdapter(mAdapter);
+            if(mAdapter == null) {
+                mAdapter = new ReceiveExternalFilesAdapter(
+                        this, getStorageManager(), getAccount());
+                mListView.setAdapter(mAdapter);
+            }
+            Vector<OCFile> files = sortFileList(getStorageManager().getFolderContent(mFile));
+            mAdapter.setNewItemVector(files);
+            mAdapter.notifyDataSetChanged();
```
abelgardep, line 436: I have tried to search for a file, and it does not work as expected. It shows nothing when searching.

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/ReceiveExternalFilesAdapter.java
```diff
@@ -89,6 +86,10 @@ public Object getItem(int position) {
         }
     }
 
+    public void setNewItemVector(Vector<OCFile> newItemVector) {
+        mFiles = newItemVector;
+    }
```
abelgardep, line 91: I would suggest moving `notifyDataSetChanged()` here, and update `mImmutableFilesList` too

owncloudApp/src/main/java/com/owncloud/android/ui/adapter/ReceiveExternalFilesAdapter.java
```diff
@@ -51,25 +51,22 @@
...
     private OnSearchQueryUpdateListener mOnSearchQueryUpdateListener;
 
     public ReceiveExternalFilesAdapter(Context context,
-                                       Vector<OCFile> files,
                                        FileDataStorageManager storageManager,
-                                       Account account
-    ) {
-        mFiles = files;
+                                       Account account) {
         mImmutableFilesList = (Vector<OCFile>) mFiles.clone();
```
abelgardep, line 64: `mImmutableFilesList` will be always empty since `mFiles` is always empty at this point, isn't it?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -394,58 +414,44 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-            files = sortFileList(files);
-
-            mAdapter = new ReceiveExternalFilesAdapter(
-                    this, files, getStorageManager(), getAccount());
-            mListView.setAdapter(mAdapter);
+            if(mAdapter == null) {
+                mAdapter = new ReceiveExternalFilesAdapter(
+                        this, getStorageManager(), getAccount());
+                mListView.setAdapter(mAdapter);
+            }
```
abelgardep, line 433: Just curiosity, when will `mAdapter` be null?

abelgardep (changes requested):

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
theScrabi: When `onCreate()` is called we can not yet set the adapter as this would lead the application to crash if thumbnails need to be rendered. So the Adapter is created lazily. However we do not want to override the adapter but only create it once. Therefore the null check.

abelgardep (approved):
LGTM. Ready to test @jesmrec

jesmrec:
https://github.com/owncloud/android/issues/3041 -> FIXED

https://github.com/owncloud/android/issues/3041#issuecomment-796686779 -> FIXED

Issue #3099, opened 2021-02-12
[Android X Preference] Security section


jesmrec:
### AC

- It is placed in the same place, just below `Camera Uploads`

- Four cells, all of them always visible
  - Passcode: exclusive with Pattern
  - Pattern:  exclusive with Passcode
  - Biometric: only if one of the above is enabled. If not, greyed out
  - Touches: independent form the previous ones. Only a checkbox
  
- Names and definition keep, no needed new ones

- No new features, except for the removal of snackbars when passcode and pattern locks are set correctly. At the end it must work as before.

### TASKS

- [x] Research a little bit how settings are implemented and how we can include androidX ones.
- [x] Create branch settings/security
- [x] Development tasks
  - [x] Add [new dependency](https://developer.android.com/jetpack/androidx/releases/preference#declaring_dependencies) to owncloudApp build gradle.
  - [x] Create a new activity, a new fragment, and a new preferences XML. [Inflate them](https://developer.android.com/guide/topics/ui/settings#inflate_the_hierarchy)
  - [x] Create a new preference category for the security section
  - [x] Include 4 new checkbox preferences in the new preference category
  - [x] [Retrieve current values](https://developer.android.com/guide/topics/ui/settings/use-saved-values) to check or not those checkboxes
  - [x] Support navigation and back when passcode and pattern are clicked (onActivityResult)
  - [x] Disable biometric till pattern or passcode are enabled
  - [x] Remove snackbars when passcode and pattern locks are set correctly and add new ones when there is an error setting these locks
  - [x] Include UI tests / Unit tests
- [x] Code review and apply changes requested
- [x] Merge branch settings/security into settings/androidx

### PR

- App #3120 
- Library (if needed)

JuancaG05:
New task added: "Remove snackbars when passcode and pattern locks are set correctly and add new ones when there is an error setting these locks"

PR #3120, opened 2021-02-26
[Settings] Security

JuancaG05:
Implements #3099

### DIFF
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
@@ -132,7 +132,6 @@ class OCSettingsSecurityTest {
         intending(hasAction(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_passcode)).perform(click())
         assertTrue(mPrefPasscode.isChecked)
-        //onView(withText(R.string.pass_code_stored)).check(matches(isDisplayed()))
     }
 
     @Test
@@ -143,7 +142,6 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_pattern)).perform(click())
         assertTrue(mPrefPattern.isChecked)
-        //onView(withText(R.string.pattern_stored)).check(matches(isDisplayed()))
     }
 
     @Test
@@ -181,7 +179,6 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_CHECK_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_passcode)).perform(click())
         assertFalse(mPrefPasscode.isChecked)
-        //onView(withText(R.string.pass_code_removed)).check(matches(isEnabled()))
         onView(withText(R.string.prefs_biometric)).check(matches(not(isEnabled())))
     }
 
@@ -194,7 +191,6 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_CHECK_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_pattern)).perform(click())
         assertFalse(mPrefPattern.isChecked)
-        //onView(withText(R.string.pattern_removed)).check(matches(isEnabled()))
         onView(withText(R.string.prefs_biometric)).check(matches(not(isEnabled())))
     }
 
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
@@ -132,7 +132,7 @@ class OCSettingsSecurityTest {
         intending(hasAction(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_passcode)).perform(click())
         assertTrue(mPrefPasscode.isChecked)
-        onView(withText(R.string.pass_code_stored)).check(matches(isDisplayed()))
+        //onView(withText(R.string.pass_code_stored)).check(matches(isDisplayed()))
     }
 
     @Test
@@ -143,7 +143,7 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_pattern)).perform(click())
         assertTrue(mPrefPattern.isChecked)
-        onView(withText(R.string.pattern_stored)).check(matches(isDisplayed()))
+        //onView(withText(R.string.pattern_stored)).check(matches(isDisplayed()))
     }
 
     @Test
@@ -181,7 +181,7 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_CHECK_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_passcode)).perform(click())
         assertFalse(mPrefPasscode.isChecked)
-        onView(withText(R.string.pass_code_removed)).check(matches(isEnabled()))
+        //onView(withText(R.string.pass_code_removed)).check(matches(isEnabled()))
         onView(withText(R.string.prefs_biometric)).check(matches(not(isEnabled())))
     }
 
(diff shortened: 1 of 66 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,227 @@
...
+
+    private fun enableBiometric() {
+        prefBiometric?.isEnabled = true
+        prefBiometric?.summary = null
+    }
+
+    private fun disableBiometric(summary: String) {
+        if (prefBiometric?.isChecked == true) {
+            prefBiometric?.isChecked = false
+        }
```
abelgardep, line 217: What if you remove this check?
```suggestion
            prefBiometric?.isChecked = false
```

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/settings/SettingsViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    fun handleEnablePattern(data: Intent?): UIResult<Unit> {
+        val pattern = data?.getStringExtra(PatternLockActivity.KEY_PATTERN)
+        pattern?.let {
+            preferencesProvider.putString(PatternLockActivity.KEY_PATTERN, it)
+            preferencesProvider.putBoolean(PatternLockActivity.PREFERENCE_SET_PATTERN, true)
+            return UIResult.Success()
+        }
+        return UIResult.Error()
+    }
```
abelgardep, line 66: You can use the same idea:
```suggestion
        val pattern = data?.getStringExtra(PatternLockActivity.KEY_PATTERN) ?: return UIResult.Error()
           
        preferencesProvider.putString(PatternLockActivity.KEY_PATTERN, pattern)
        preferencesProvider.putBoolean(PatternLockActivity.PREFERENCE_SET_PATTERN, true)
        return UIResult.Success()
    }
```

abelgardep (changes requested):
Some changes requested @JuancaG05 But GREAT JOB!!

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
```diff
@@ -132,7 +132,7 @@ class OCSettingsSecurityTest {
         intending(hasAction(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_passcode)).perform(click())
         assertTrue(mPrefPasscode.isChecked)
-        onView(withText(R.string.pass_code_stored)).check(matches(isDisplayed()))
+        //onView(withText(R.string.pass_code_stored)).check(matches(isDisplayed()))
```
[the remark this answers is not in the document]
theScrabi: I think we should remove lines before outcommend them. You could still lookup the line in the git history if you would need it.

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
```diff
@@ -143,7 +143,7 @@ class OCSettingsSecurityTest {
         intending(hasAction(PatternLockActivity.ACTION_REQUEST_WITH_RESULT)).respondWith(intentResult)
         onView(withText(R.string.prefs_pattern)).perform(click())
         assertTrue(mPrefPattern.isChecked)
-        onView(withText(R.string.pattern_stored)).check(matches(isDisplayed()))
+        //onView(withText(R.string.pattern_stored)).check(matches(isDisplayed()))
```
theScrabi, line 146: Same here

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsFragmentTest.kt
```diff
@@ -0,0 +1,443 @@
...
+            context.getString(R.string.prefs_touches_with_other_visible_windows_summary),
+            prefTouchesWithOtherVisibleWindows.summary
+        )
+        assertTrue(prefTouchesWithOtherVisibleWindows.isVisible)
+        assertTrue(prefTouchesWithOtherVisibleWindows.isEnabled)
+        assertFalse(prefTouchesWithOtherVisibleWindows.isChecked)
+    }
+
+    @Test
+    fun passcodeOpen() {
```
theScrabi, line 174: @abelgardep I thought we had some sort of naming convention for tests like the `ok - ko` thing or that test cases always start with the word `test` in front of them? However looking through the code I can see we do not have a consistent way of doing so? What is the right way to do it then?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -751,7 +751,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
                 }
                 mPreferencesProvider.putBoolean(PassCodeActivity.PREFERENCE_SET_PASSCODE, true);
 
-                showSnackMessage(R.string.pass_code_stored);
+                //showSnackMessage(R.string.pass_code_stored);
```
theScrabi, line 754: I'd recommend removing out commented lines.

theScrabi (changes requested):

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
abelgardep: The point here is that we could delete the whole file. The security section is now tested in a different way. But if we want to remove the old implementation in the last step, I agree with you that we should remove the line.

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsFragmentTest.kt
abelgardep: Yes, that's it. We need to have a consistent way of naming for tests. But the `ok - ko` [is not supported by the Android runtime](https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods) (at least for the moment). So, we will use it when possible: Unit tests. As you can see, we use the `ok - ko` in the `SettingsViewModelTest`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsFragment.kt
```diff
@@ -0,0 +1,227 @@
...
+
+        prefSecurityCategory = findPreference(PREFERENCE_SECURITY_CATEGORY)
+        prefPasscode = findPreference(PassCodeActivity.PREFERENCE_SET_PASSCODE)
+        prefPattern = findPreference(PatternLockActivity.PREFERENCE_SET_PATTERN)
+        prefBiometric = findPreference(BiometricActivity.PREFERENCE_SET_BIOMETRIC)
+        prefTouchesWithOtherVisibleWindows = findPreference(PREFERENCE_TOUCHES_WITH_OTHER_VISIBLE_WINDOWS)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            biometricManager = BiometricManager.getBiometricManager(activity)
+        }
```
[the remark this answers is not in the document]
JuancaG05: Yeah you are right, I'll do it, thanks!

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsFragment.kt
JuancaG05: Not worth it to check if the preference is checked everytime you want to disable it, I agree

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsSecurityTest.kt
JuancaG05: Let's keep the class until the last PR and when we have all the preferences updated to AndroidX, we can delete these unused files. I'll remove the out-commented lines until then, thx!

owncloudApp/src/main/java/com/owncloud/android/ui/activity/Preferences.java
JuancaG05: Same as before, unused class, but I'll try to make it cleaner from now on by removing lines instead of comment them, thanks!

JuancaG05:
There you have the requested changes! @abelgardep @theScrabi

abelgardep (approved):

abelgardep:
Approved on my side. Great work @JuancaG05 👍

abelgardep:
It will be tested when all sections are migrated. We can merge it into `settings/androidx`

Issue #3101, opened 2021-02-12
[Android X Preference] Logs section


abelgardep:
### AC

- The category is placed in the same place, below `security`, but now the preferences are displayed in another screen

- Two cells, visible only when logs are enabled.
  - HTTP Logs
  - Logs view
  
- Names and definition keep, no needed new ones

- New feature: 
  - DEBUG logic is removed: logs are not enabled by default although you use the debug version

### TASKS

- [x] Research a little bit how settings are implemented and how we can include androidX ones.
- [x] Create branch settings/logs
- [x] Development tasks(https://developer.android.com/guide/topics/ui/settings#inflate_the_hierarchy)
  - [x] Create a new preference screen for the logs section
  - [x] Include a new switch preference, a checkbox preference and a simple preference in the new preference screen
  - [x] [Retrieve current values](https://developer.android.com/guide/topics/ui/settings/use-saved-values) to check or not enable logging and HTTP logs.
  - [x] Support navigation and back when logs view is clicked (onActivityResult)
  - [x] Show logs settings only after enabling logging
  - [x] Remove DEBUG logic in the app (logs will start if you enable them from settings, no matter if you use the debug version or release version)
  - [x] Adapt LogHistoryActivity to the new toolbar and move it to presentation package
  - [x] Include UI tests / Unit tests
- [x] Code review and apply changes requested
- [x] Merge branch settings/logs into settings/androidx

### PR

- App
- Library (if needed)

JuancaG05:
Hey, I have a proposal for this issue.
Looking forward to improve the section, I suggest to stop using the 5 clicks to enable the Logs section -maybe it's not as intuitive as it seems-, and instead of that, do something similar to what we are going to do in the [Camera uploads section](https://github.com/owncloud/android/issues/3130#issuecomment-789665455). The result might be something like this:

| Current implementation | Suggestion |
| ---------------------------------------------- | -------------------------------------------- |
| <img width="210" alt ="device-2021-03-11-122924" src="https://user-images.githubusercontent.com/57049315/110783339-969ccc00-8268-11eb-8ecc-bf241a66846b.png"> | <img width="210" alt="Captura de pantalla 2021-03-11 a las 12 43 33" src="https://user-images.githubusercontent.com/57049315/110783374-a87e6f00-8268-11eb-8673-e68c4e340086.png"> |

[...]

PR #3150, opened 2021-03-16
[Settings] Logs

JuancaG05:
Implements https://github.com/owncloud/android/issues/3101


abelgardep (changes requested):
Some changes requested @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/SettingsActivity.kt
```diff
@@ -58,12 +63,40 @@ class SettingsActivity : AppCompatActivity() {
...
+            tag = "logs"
+            titleId = R.string.actionbar_logger
+        }
+        supportFragmentManager.beginTransaction()
+            .replace(R.id.settings_container, fragment, tag)
+            .addToBackStack(null)
+            .commit()
+        supportActionBar?.setTitle(titleId!!)
+        return true
+    }
```
abelgardep, line 96: I understand it, but from my point of view, it is not a good solution. Tags are not needed at all to handle this, and the action bar title should be handled in a different way. We could add an `addOnBackStackChangedListener` and check there the fragment and update the action bar depending on it.

```suggestion
    override fun onPreferenceStartFragment(caller: PreferenceFragmentCompat, pref: Preference): Boolean {
        val fragment = supportFragmentManager.fragmentFactory.instantiate(classLoader, pref.fragment)

        supportFragmentManager.beginTransaction()
            .replace(R.id.settings_container, fragment)
            .addToBackStack(null)
            .commit()

        return true
    }
```

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/SettingsActivity.kt
```diff
@@ -58,12 +63,40 @@ class SettingsActivity : AppCompatActivity() {
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
             android.R.id.home -> {
-                intent = Intent(this, FileDisplayActivity::class.java).apply {
-                    addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+                if (supportFragmentManager.findFragmentByTag("logs") != null) {
+                    supportFragmentManager.popBackStack()
+                    supportActionBar?.setTitle(R.string.actionbar_settings)
```
abelgardep, line 68: We can check directly if we are on root asking `backStackEntryCount` instead of checking tags 🍻 
```suggestion
                if (supportFragmentManager.backStackEntryCount > 0) {
                    supportFragmentManager.popBackStack()
                } else {
```

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/SettingsActivity.kt
```diff
@@ -58,12 +63,40 @@ class SettingsActivity : AppCompatActivity() {
...
+            .addToBackStack(null)
+            .commit()
+        supportActionBar?.setTitle(titleId!!)
+        return true
+    }
+
+    override fun onBackPressed() {
+        super.onBackPressed()
+        supportActionBar?.setTitle(R.string.actionbar_settings)
+    }
```
abelgardep, line 101: I would remove this if we add an `addOnBackStackChangedListener` and updating the toolbar in consecuence:

`        supportFragmentManager.addOnBackStackChangedListener {
            updateToolbarTitle()
        }`

And updateToolbarTitle() could be something like this:
```   
private fun updateToolbarTitle() {
        val titleId = when (supportFragmentManager.fragments.lastOrNull()) {
            is SettingsLogsFragment -> R.string.actionbar_logger
            else -> R.string.actionbar_settings
        }

        supportActionBar?.setTitle(titleId)
    }
```
And remember that if the device rotates, the activity is recreated and you won't need to add the SettingsFragment again because it will remember 👍 
```
        if (savedInstanceState != null) return

        supportFragmentManager
            .beginTransaction()
            .replace(R.id.settings_container, SettingsFragment())
            .commit()
```

🍻 🍻

owncloudApp/src/main/res/xml/settings.xml
```diff
@@ -42,6 +43,19 @@
...
 
+    <PreferenceCategory
+        app:iconSpaceReserved="false"
+        app:key="logs_category"
+        app:title="@string/actionbar_logger">
+
+        <Preference
+            app:icon="@drawable/ic_baseline_description_24"
+            app:key="logs_subsection"
+            app:title="@string/prefs_enable_logging"
```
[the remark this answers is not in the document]
JuancaG05: Well it was a first approach, not sure about it either, but I chose it because it shares title with the first preference in the new screen `settings_logs.xml`. We'll let QA decide 😬 .

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsLogsFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+        setPreferencesFromResource(R.xml.settings_logs, rootKey)
+
+        prefEnableLogging = findPreference(PREFERENCE_ENABLE_LOGGING)
+        prefHttpLogs = findPreference(PREFERENCE_LOG_HTTP)
+        prefLogsView = findPreference(PREFERENCE_LOGGER)
+
+        if (!settingsViewModel.isEnableLoggingOn()) {
+            prefHttpLogs?.isVisible = false
+            prefLogsView?.isVisible = false
+        }
```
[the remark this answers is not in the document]
JuancaG05: I chose this name because maybe the one you propose could create confusion since "enabled" is also the name of the state of a UI view (enabled/disabled), but in this case I refer to the internal state of the preference and consequently the state (checked/not checked) of the SwitchPreference. What's your opinion on this?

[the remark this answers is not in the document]
abelgardep: Ok, I see. I think that it should not be a problem at all, since this is a ViewModel call and you know that the UI view is not handled there. But, as you consider. There is no problem for me 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -96,7 +96,7 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
         super.onCreate(savedInstanceState)
 
         // Protection against screen recording
-        if (!MainApp.isDeveloper) {
+        if (!MainApp.enabledLogging) {
```
abelgardep, line 99: I think that this is not the same. In this case, we should not disable the protection when logs are enabled, am I right @jesmrec ?

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
```diff
@@ -158,21 +156,12 @@ class MainApp : Application() {
...
+                .getBoolean(SettingsLogsFragment.PREFERENCE_ENABLE_LOGGING, false)
 
-        if (isDeveloper) {
-            val dataFolder = dataFolder
-
-            // Set folder for store logs
-            LoggingHelper.startLogging(
-                File(Environment.getExternalStorageDirectory().absolutePath + File.separator + dataFolder), dataFolder
-            )
-            Timber.d("${BuildConfig.BUILD_TYPE} start logging ${BuildConfig.VERSION_NAME} ${BuildConfig.COMMIT_SHA1}")
```
hannesa2, line 173: It's not smart to remove it. 
With it you see in logfile, when the was started and it makes absolute clear which version you are using

hannesa2: Sorry, I see now it's here https://github.com/owncloud/android/pull/3150/files?file-filters%5B%5D=.java&file-filters%5B%5D=.kt&file-filters%5B%5D=.xml&hide-deleted-files=true#diff-67893595fc2d9801bbc5624857ff2527b1ffc6c7a6ad30c763d054590bfd1a73R44

Please forget my comment

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
jesmrec: You are right. Screen protection should be active only in case it is a release version. For developer/debug/etc... it does not.

abelgardep (approved):
Ok, let's remove those ignore tests in the upcoming PRs. Good job @JuancaG05

Issue #3153, opened 2021-03-18
[FEATURE REQUEST] Organize settings by subsections


JuancaG05:
### EPIC: https://github.com/owncloud/android/issues/2867

Based on other apps' design, we propose some changes to the whole Settings menu. All settings are currently being updated to AndroidX (https://github.com/owncloud/android/issues/2867), so it would be suitable to address this issue now.

The idea is to group the different preferences by subsections that drives the user to other screens/fragments. This would make navigation through settings easier, since the users would find whatever they're looking for in a simpler and hierarchical way. Moreover, it makes it easily expansible if we want to add more settings in the future. Code would look cleaner and tests would be easier too.

Here's a prototype of the final result:

![device-2021-03-18-132323](https://user-images.githubusercontent.com/57049315/111625491-302a2780-87ed-11eb-956a-ad48b089eb7e.png)

Then, each item would go to a screen with its respective preferences, respecting the original ones (such in this issue https://github.com/owncloud/android/issues/3101#issuecomment-796688422, we move preference items to a new screen).

It's an open discussion, you can give your opinion and please discuss the following points:
- Would you place a divider line between "Picture uploads" and "Video uploads"? Currently it's not placed as they belong to the same category, but it may create confusion. Other option is to delete all those lines, please let me know what do you think about it.
- The "Enable logging" (now always visible) could maybe be moved to the end, just below "More", as it is something that only a certain segment of users will use. What do you think?

### TASKS

- [x] Create branch settings/subsections
- [x] Development tasks
  - [x] Make "Enable logging" section items disable instead of hide when the switch is off 
  - [x] Move "Security" section preferences to a new fragment
  - [x] Move "More" section preferences to a new fragment
  - [x] Move "App version" item to the general settings menu, and make it copy-able by a simple tap
  - [x] Try to unify wording with the iOS app
  - [x] Adapt existent UI tests (now they belong to different fragments)
  - [x] Create UI tests for the general settings screen
- [x] Code review and apply changes requested
- [x] Merge branch settings/subsections into settings/androidx

### PR

- App
- Library (if needed)

JuancaG05:
@abelgardep @jesmrec @theScrabi @michaelstingl @tbsbdr opinion on this?

abelgardep:
I like it. I would try to unify some common wording with iOS by the way. 👍

michaelstingl:
> @abelgardep @jesmrec @theScrabi @michaelstingl @tbsbdr opinion on this?

[...]

PR #3168, opened 2021-03-24
[Settings] Organize settings by subsections

JuancaG05:
Implements https://github.com/owncloud/android/issues/3153


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/SettingsActivity.kt
```diff
@@ -65,7 +65,9 @@ class SettingsActivity : AppCompatActivity() {
 
     private fun updateToolbarTitle() {
         val titleId = when (supportFragmentManager.fragments.lastOrNull()) {
-            is SettingsLogsFragment -> R.string.actionbar_logger
+            is SettingsSecurityFragment -> R.string.prefs_subsection_security
+            is SettingsLogsFragment -> R.string.prefs_subsection_logging
+            is SettingsMoreFragment -> R.string.prefs_subsection_more
```
abelgardep, line 70: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsLogsFragment.kt
```diff
@@ -38,26 +38,26 @@ class SettingsLogsFragment : PreferenceFragmentCompat() {
 
     private var prefEnableLogging: SwitchPreferenceCompat? = null
     private var prefHttpLogs: CheckBoxPreference? = null
-    private var prefLogsView: Preference? = null
+    private var prefBrowse: Preference? = null
```
abelgardep, line 41: As I suggested in the test, I think that this naming is not as good as it was before. Here our idea is to open the logs view and check the latest logs directly.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
```diff
@@ -0,0 +1,218 @@
...
+        // Biometric lock
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            screenSecurity?.removePreference(prefBiometric)
+        } else if (prefBiometric != null) {
+            biometricManager = BiometricManager.getBiometricManager(activity)
+
+            // Disable biometric lock if Passcode or Pattern locks are disabled
+            if (prefPasscode?.isChecked == false && prefPattern?.isChecked == false) {
+                disableBiometric(getString(R.string.prefs_biometric_summary))
+            }
```
abelgardep, line 161: Is this relevant? Is there a scenario where the biometric could be enabled, and passcode and pattern not? Just wondering 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
```diff
@@ -0,0 +1,218 @@
...
+                    disablePatternLauncher.launch(intent)
+                }
+            }
+            false
+        }
+
+        // Biometric lock
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            screenSecurity?.removePreference(prefBiometric)
+        } else if (prefBiometric != null) {
```
abelgardep, line 155: Is the `prefBiometric != null` needed? It should not be null at this point. You have some secure `?.` calls later anyway 🤔

owncloudApp/src/main/res/drawable/ic_baseline_lock_24.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
abelgardep, line 1: I would rename the icon to do it more descriptive. For example `ic_lock`. And curiously, if you check it, we already have a lock icon 👍  Same for other icons, the more descriptive, the better, we will find them easily.

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -607,6 +610,7 @@
     <string name="prefs_logs_summary">Logs may contain sensitive information. Sharing logs with others is sole user responsibility</string>
 
     <string name="synchronizing_account">Synchronizing account</string>
-    <string name="prefs_enable_logging">Enable logging</string>
-    <string name="prefs_enable_logging_summary">When activated, logs may impact performance and include sensitive information. However the logs are not subject to automatic submission to ownCloud servers. Sharing logs with others is sole user responsibility.</string>
+    <string name="prefs_subsection_security_summary">Set locks to access the app and allow touches with other visible windows</string>
+    <string name="prefs_subsection_logging_summary">Keep track of logs produced by the execution of the app</string>
+    <string name="prefs_subsection_more_summary">App info, feedback and others</string>
```
abelgardep, line 615: Group new `prefs_subsection` with the other `prefs` keys. I know this is a little messy, but the cleaner we leave it, the less work we will need to do in the future.

abelgardep (changes requested):
Some suggestions here

owncloudApp/src/androidTest/java/com/owncloud/android/settings/logs/SettingsLogsFragmentTest.kt
```diff
@@ -60,7 +60,7 @@ class SettingsLogsFragmentTest {
 
     private lateinit var prefEnableLogging: SwitchPreferenceCompat
     private lateinit var prefHttpLogs: CheckBoxPreference
-    private lateinit var prefLogsView: Preference
+    private lateinit var prefBrowse: Preference
```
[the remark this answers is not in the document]
JuancaG05: Yes, it was a confusion, I thought the "Browse" button had exactly the same function than the "Open logs view" button. I'll return to the previous naming ⬅️

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
JuancaG05: This is only executed when the fragment is inflated. It's not that biometric is enabled/checked, but we have to disable it as by default, the initial state of the `CheckBoxPreference` is enabled. If we had chosen to set it to disabled by default in the XML, we would have to check the opposite: if passcode or pattern are checked, then enable biometric. I think it's easier this way, but let me know if you think the other way is better

JuancaG05: Btw, I took advantage of this and revised the `disableBiometric` function. Now it doesn't receive any parameter, it was unnecessary 😁

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
JuancaG05: Well, it can be null if `findPreference` doesn't find it. It's true that we have some secure calls `?.` inside this else if, but this way, we save the `biometricManager = BiometricManager.getBiometricManager(activity)` instruction in case prefBiometric is null. Anyway, would it be worth it to transform the else if into else?

abelgardep: Ok, I think that it will be null only when the `if condition is true`, but I'm not sure. If you think we need it, keep it, no problem 👍

abelgardep (approved):
Approved, good job @JuancaG05

Issue #3121, opened 2021-02-26
Blank view in music player


jesmrec:
While the music is reproduced, something should be shown to replace the white background: artwork, cover, a placeholder image...

As first improvement, setting an music icon on the background is enough. Then, we will iterate to improve.

![device-2021-02-23-132712](https://user-images.githubusercontent.com/14894746/108843443-fcd3ee80-75da-11eb-8b66-b9b4d43e3ff5.png)

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/feature_name
- [x] Development tasks
  - [x] Implement icon design
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

PR: https://github.com/owncloud/android/pull/3182

theScrabi:
1, 2, 3 what will it be?

| 1 | 2 | 3|
|-|-|-|
|![1](https://user-images.githubusercontent.com/1891273/110804632-a6c0a580-8280-11eb-8f4e-b99d97f06900.png) |![2](https://user-images.githubusercontent.com/1891273/110804636-a7593c00-8280-11eb-8d96-c28013315212.png) |![3](https://user-images.githubusercontent.com/1891273/110804639-a7593c00-8280-11eb-87ff-b02e09083fd0.png) |

michaelstingl:
@tbsbdr ⬆️

jesmrec:
i like the 2

theScrabi:
Second it is then :)

theScrabi:
Ok so I found out we can already handle coverart. Though it looks a bit weird. The web frontend tries to make coverart rectangular while we just show what we get.
With our permission I would change this to.

| current state | my suggestion |
| ---- | ---- |
| ![Bild_fix](https://user-images.githubusercontent.com/1891273/113008291-34542e80-9177-11eb-8f43-9a086f37cd42.png) |  ![cover_small](https://user-images.githubusercontent.com/1891273/113008323-3a4a0f80-9177-11eb-8521-35e340ffc19d.png) |

What web frontend shows: 
![grafik](https://user-images.githubusercontent.com/1891273/113008534-66fe2700-9177-11eb-8831-a16cf6f5f1d5.png)

theScrabi:
Just a design peak for a possible future solution. This could be implemented with simple draw able overlay and gradients:
[...]

PR #3182, opened 2021-03-30
improve music player design

theScrabi:
issue: https://github.com/owncloud/android/issues/3121


owncloudApp/src/main/res/layout/preview_audio_fragment.xml
```diff
@@ -14,47 +14,55 @@
...
-        android:layout_width="match_parent"
-        android:layout_height="0dp"
         android:layout_alignParentTop="true"
+        android:layout_centerHorizontal="true"
         android:layout_above="@+id/media_controller"
-        >
+        android:layout_gravity="center"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:padding="40dp">
```
abelgardep, line 39: I'm not a huge fan of this. Adding a constraint layout just to include an image view may not be the best option. If you need some constraints, please move the whole layout to constraint layout and get rid of the relative one 👍

owncloudApp/src/main/res/layout-land/preview_audio_fragment.xml
```diff
@@ -0,0 +1,67 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/top"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:gravity="center"
+    tools:context=".ui.preview.PreviewAudioFragment"
+    android:filterTouchesWhenObscured="true">
+
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
abelgardep, line 31: I think that we could get the same behaviour using the proper constraints in the portrait layout. That way, we could get rid of this second layout that we will need to maintain. What do you think?

owncloudApp/src/main/res/drawable/ic_music_note_eighth.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="200dp"
+    android:height="200dp"
+    android:viewportWidth="200"
+    android:viewportHeight="200">
+  <path
+      android:pathData="M60,8V113.5C54.1,110.1 47.3,108 40,108 17.9,108 0,125.9 0,148c0,22.1 17.9,40 40,40 22.1,0 40,-17.9 40,-40V48c33.102,0 66.203,0 99.305,0v67.332c-5.9,-3.4 -12.7,-5.5 -20,-5.5 -22.1,0 -40,17.9 -40,40 0,22.1 17.9,40 40,40 22.1,0 40,-17.9 40,-40V8C152.87,8 106.435,8 60,8Z"
+      android:strokeWidth="10"
+      android:fillColor="#cccccc"/>
```
abelgardep, line 9: How did you choose that color? What do you think about using the app one? Could it be interesting to make it brandable?

abelgardep (changes requested):
Some questions here @theScrabi.

BTW, do not forget to add the changelog item as described [here](https://github.com/owncloud/android/tree/master/changelog)

owncloudApp/src/main/res/drawable/ic_music_note_eighth.xml
theScrabi: Using our App colorpallet would make sense, however I don't think branding the placeholder for the music cover is very useful.

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
theScrabi: done.

owncloudApp/src/main/res/layout-land/preview_audio_fragment.xml
theScrabi: I don't think so. We need a different parameter for making the image view rectangular.

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
```diff
@@ -29,32 +29,51 @@
 
     <FrameLayout
         android:id="@+id/visual_area"
-        android:layout_width="match_parent"
+        android:layout_width="0dp"
         android:layout_height="0dp"
-        android:layout_alignParentTop="true"
         android:layout_above="@+id/media_controller"
-        >
+        android:layout_alignParentTop="true"
```
abelgardep, line 35: `layout_above` and `layout_alignParentTop` are not needed anymore, are they? We moved to ConstraintLayout 🍻

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
```diff
@@ -29,32 +29,51 @@
...
             android:id="@+id/image_preview"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:layout_margin="@dimen/standard_margin"
-            android:layout_gravity="center"
             android:contentDescription="@string/preview_image_description"
-            android:src="@drawable/logo" />
+            android:scaleType="centerCrop"
+            android:src="@drawable/ic_place_holder_music_cover_art"
+            app:layout_constraintBottom_toBottomOf="parent" />
```
abelgardep, line 54: The parent is not a Constraint layout, we don't need this constraint here, am I right?

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
```diff
@@ -29,32 +29,51 @@
...
+    <!-- last to draw, will be top in Z axis -->
     <com.owncloud.android.media.MediaControlView
         android:id="@id/media_controller"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_alignParentBottom="true"
         android:layout_margin="16dp"
-        />
+        app:layout_constraintBottom_toBottomOf="parent"
+        tools:layout_editor_absoluteX="16dp" />
```
abelgardep, line 75: Do we need that `tools:layout_editor_absoluteX`?

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
```diff
@@ -29,32 +29,51 @@
...
-        android:layout_width="match_parent"
+        android:layout_width="0dp"
         android:layout_height="0dp"
-        android:layout_alignParentTop="true"
         android:layout_above="@+id/media_controller"
-        >
+        android:layout_alignParentTop="true"
+        android:layout_centerHorizontal="true"
+        android:padding="40dp"
+        app:layout_constraintDimensionRatio="H,1:1"
```
abelgardep, line 38: If we reduce the padding to 16dp (Same as media_controller) and we move that `layout_constraintDimensionRatio="H,1:1"` to `layout_constraintDimensionRatio="1:1"` we would have the same behavior in landscape and we could remove the specific `preview_audio_fragment` for landscape. What do you think?

abelgardep (changes requested):
Some questions more @theScrabi

owncloudApp/src/main/res/layout/preview_audio_fragment.xml
theScrabi: I didn't know I could leave out the `H` parameter.
However I don't like `16dp` on portrait, as the cover looks to huge:
|16dp|40dp|
|---|---|
|bla|asdf|

theScrabi: All right I've settled with 30dp for land/portrait. :)

abelgardep (approved):
Approved on my side, Ready to QA @jesmrec

jesmrec:
Checked with Android 7 & Android 11, new background image correctly displayed in both and both orientations.

Not much to test here

Approved

Issue #3139, opened 2021-03-08
[BUG] Some glitches in media players when device rotates


jesmrec:
### Steps to reproduce

1. Play a music or video file 
2. Rotate the device

### Actual behaviour

The bottom bar is visible, but the buttons do nothing but the "Uploads" one. 

![device-2021-03-08-130712](https://user-images.githubusercontent.com/14894746/110319545-5858a000-800f-11eb-98fd-63dfe0bf4f87.png)

New rotations don't make the app to hide the bottom bar

### Expected behaviour

As reference, the image preview hides the bottom bar. It'd be a congruent behaviour across all previews
 
NOTE: also reproduced with txt files.

### Environment data
Android version: 11

Device model:  Pixel 2

Stock or customized system: stock

ownCloud app version: 2.16

ownCloud server version: 10.6

### PR
App: #3184

PR #3184, opened 2021-04-01
fix navbar is vissible in filepreview after rotation

theScrabi:
issue: https://github.com/owncloud/android/issues/3139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,20 @@ ownCloud admins and users.
 Summary
 -------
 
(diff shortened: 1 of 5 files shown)


abelgardep (changes requested):
Code looks good to me 👍 , but changelog item is missing.

Please take a look at [calens](https://github.com/owncloud/android/tree/master/changelog) and add a new item [here](https://github.com/owncloud/android/tree/master/changelog/unreleased). That way the changelog will be generated automatically. @theScrabi

abelgardep (approved):
Approved on my side. Ready to QA @jesmrec

jesmrec:
Fixed.

PR #3189, opened 2021-04-06
Format manifest and fix empty body

hannesa2:
It formats manifest and fixes 

image

There are no logic changes at all


abelgardep (approved):
LGTM

PR #3193, opened 2021-04-07
test bitrise

jesmrec:
## Related Issues
App:

Library PR (if needed):

_____


owncloudApp/src/main/res/values/setup.xml
```diff
@@ -60,6 +60,10 @@
 
     <!-- Help, imprint and feedback -->
     <bool name="help_enabled">true</bool>
+
+
+
```
hannesa2, line 65: There is no need to change anything. This is enough

`git commit -m "something" --allow-empty`

PR #3194, opened 2021-04-07
update fdroid badge url

theScrabi:
This will update the url of the fdroid badge icon in the README.md. The old url was broken.


abelgardep (approved):
Approved. The old URL redirects to the new one, so let's update it.

PR #3202, opened 2021-04-14
Make some fields not mandatory in discovery response

abelgardep:
`registration_endpoint` is mandatory at the moment but it should not according to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

It should fix a problem that happens when `registration_endpoint` is not retrieved in the discovery.

## Related Issues

Library PR (if needed): https://github.com/owncloud/android-library/pull/392

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Summary
 -------
 
 * Bugfix - Fix navbar is visible in file preview screen after rotation: [#3184](https://github.com/owncloud/android/pull/3184)
+* Bugfix - Fix a bug when some fields where not retrieved from OIDC Discovery: [#3202](https://github.com/owncloud/android/pull/3202)
 * Enhancement - Replace blank view in music player with cover art: [#3121](https://github.com/owncloud/android/issues/3121)
 * Enhancement - Support for apk files: [#2691](https://github.com/owncloud/android/issues/2691)
 * Enhancement - Align previews actions: [#3155](https://github.com/owncloud/android/issues/3155)
@@ -23,6 +24,15 @@ Details
 
    https://github.com/owncloud/android/issues/3139
    https://github.com/owncloud/android/pull/3184
+
+* Bugfix - Fix a bug when some fields where not retrieved from OIDC Discovery: [#3202](https://github.com/owncloud/android/pull/3202)
+
+   Problem when requesting the OIDC discovery was fixed. Some fields were handled as mandatory,
+   but they are recommended according to the docs. It prevented from a proper login. Now it is
+   possible to login as expected when some fields are not retrieved.
+
+   https://github.com/owncloud/android/pull/3202
+   https://github.com/owncloud/android-library/pull/392
 
 * Enhancement - Replace blank view in music player with cover art: [#3121](https://github.com/owncloud/android/issues/3121)
 
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
@@ -360,10 +360,15 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
                     Timber.d("Service discovery: ${it.peekContent().getStoredData()}")
                     oidcSupported = true
                     it.peekContent().getStoredData()?.let { oidcServerConfiguration ->
-                        registerClient(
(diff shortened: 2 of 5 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -360,10 +360,15 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
-                        )
+                        val registrationEndpoint = oidcServerConfiguration.registrationEndpoint
+                        if (registrationEndpoint != null) {
+                            registerClient(
+                                oidcServerConfiguration.authorizationEndpoint.toUri(),
+                                registrationEndpoint
+                            )
+                        } else {
+                            performGetAuthorizationCodeRequest(oidcServerConfiguration.authorizationEndpoint.toUri())
+                        }
```
JuancaG05, line 371: How about using `let`and `?:` to kotlinize it even more? Just a suggestion, no important changes
```suggestion
                        registrationEndpoint?.let {
                            registerClient(
                                oidcServerConfiguration.authorizationEndpoint.toUri(),
                                registrationEndpoint
                            ) ?: performGetAuthorizationCodeRequest(oidcServerConfiguration.authorizationEndpoint.toUri())
```

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/oauth/ClientRegistrationRequest.kt
```diff
@@ -22,7 +22,7 @@ import com.owncloud.android.domain.authentication.oauth.model.ClientRegistration
 import com.owncloud.android.testutil.OC_REDIRECT_URI
 
 val OC_CLIENT_REGISTRATION_REQUEST = ClientRegistrationRequest(
-    registrationEndpoint = OC_OIDC_SERVER_CONFIGURATION.registrationEndpoint,
+    registrationEndpoint = OC_OIDC_SERVER_CONFIGURATION.registrationEndpoint!!,
     clientName = "Android Client v2.17",
```
JuancaG05, line 26: Could this String not being hardcoded? Just asking, in order to to make it easier when we release new versions

JuancaG05 (commented):
Some comments, not explicit approval but no requested changes, only looking forward to know your opinion

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/oauth/ClientRegistrationRequest.kt
abelgardep: It is used for testing proposals, nothing relevant. This is in the owncloudTestUtil module. We could remove the `v2.17` but it won't affect at all.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
abelgardep: Yes, of course, we could do it that way. I used the `if else` because from my point of view it is more clear than `let {} ?:` but it is basically the same.

JuancaG05 (approved):
Ok, LGTM, we can move this forward

CHANGELOG.md
```diff
@@ -24,6 +25,15 @@ Details
    https://github.com/owncloud/android/issues/3139
    https://github.com/owncloud/android/pull/3184
 
+* Bugfix - Fix a bug when some fields where not retrieved from OIDC Discovery: [#3202](https://github.com/owncloud/android/pull/3202)
+
+   We've fixed a bug when requesting the OIDC discovery. Some fields were handled as mandatory,
```
jesmrec, line 30: One proposal: i think we should not use the first person to describe the changelog features. Instead of `We've fixed...` i'd set something like: `Problem requesting OIDC discovery was fixed`. What do you think? @abelgardep @JuancaG05 @theScrabi

abelgardep: Ok, I can agree with that. But we need to update current tickets too: https://github.com/owncloud/android/blob/master/CHANGELOG.md

JuancaG05: I agree, better in a passive voice. I was checking some apps in the Play Store and some of them use this new proposed way and others use first person, so I guess both of them are common. Personally I prefer the one @jesmrec said

jesmrec:
Calens updated. Feature is approved on my side. Please, take a look to the calens file modification as last check.

Issue #3028, opened 2020-12-02
Replace kotlin-android-extensions


abelgardep:
With latest update to kotlin 1.4.20, `kotlin-android-extensions` are deprecated. The most important thing here, is that we need to replace synthetics. ViewBinding? We need to use also the new `kotlin-parcelize` plugin

### TASKS

- [x] Research (if needed)
- [x] Create branch replace_kotlin_android_extensions
- [x] Development tasks
  - [x] Remove synthetics from DrawerActivity
  - [x] Remove synthetics from PreviewImageFragment
  - [x] Remove synthetics from ToolbarActivity
  - [x] Remove synthetics from FileDisplayActivity
  - [x] Remove synthetics from ShareFileFragment
  - [x] Remove synthetics from PublicShareDialogFragment
  - [x] Remove synthetics from SearchShareesFragment
  - [x] Remove synthetics from LoginActivity
  - [x] Remove synthetics from SharePublicLinkAdapter
  - [x] Remove synthetics from SortOptionsView
  - [x] Remove synthetics from EditPrivateShareFragment
  - [x] Remove synthetics from BottomSheetFragmentView
  - [x] Use kotlin-parcelize instead of old kotlin android extension
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA - UI Regressions
- [ ] Merge branch replace_kotlin_android_extensions into master

### PR

- App #3203 
- Library (if needed)

jesmrec:
It will need to be changed medium - term.

Used in login view and shares.

PR #3203, opened 2021-04-19
Replace kotlin android extensions

abelgardep:
Let's move to ViewBinding.

## Related Issues
App: #3028 

_____


JuancaG05 (approved):

jesmrec:
Some exploratory testing performed here. Automatic test passing, and no glitches detected in:

- Login view
- Drawer
- Share view
- Public share view
- Private share view

So, we can go ahead with this. If problems'd happen, they'd be taken as regressions.

Issue #2722, opened 2019-11-14
[BUG] Snackbar in passcode view is not displayed


jesmrec:
#### Case 1
1. Open Settings > Passcode lock
2. Enter passcode
3. In the 2nd try, enter a different number

Update: two more cases where this bug appeared were fixed (the solution was the same)

#### Case 2
1. Register a passcode correctly
2. Close the app
3. Open the app and enter an incorrect passcode

#### Case 3
1. Register a passcode correctly
2. Go to Passcode lock to try to remove the current passcode
3. Enter an incorrect passcode

### Actual behaviour
Snackbar that shows the error is hidden between the soft keyboard

### Expected behaviour
Error message is shown just below the passcode input, only when the input error has just happened and the user hasn't introduced another passcode yet, and snackbar is not shown anymore

### TASKS

- [x] Create branch fix/bug_snackbar_passcode
- [x] Development tasks
  - [x] Add a new TextView to the XML file
  - [x] Replace Snackbar creation by printing a message in the TextView
  - [x] Adapt UI tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch fix/bug_snackbar_passcode into master

JuancaG05:
I found myself with this bug. Maybe a good approach, instead of hiding the keyboard for the snackbar to be visible (a bit uncomfortable) or displaying the snackbar above the keyboard, is to add a little TextView just above "Enter your passcode" and make it visible only when the passcode confirmation is not correct?

jesmrec:
That makes sense, and interesting improvement we can include. Thanks for the feedback!

theScrabi:
How about we adjust the layout according to visibility of the virtual keyboard:
https://developer.android.com/training/keyboard-input/visibility#Respond

JuancaG05:
> I have another suggestion. We could adjust the layout according to the visibility of the softkeyboard:
> https://developer.android.com/training/keyboard-input/visibility#Respond

The softkeyboard usually fills more or less half of the fragment height (with the device in vertical position), maybe doing what you suggest requires scrolling, what makes it more uncomfortable. Also, if we displayed the snackbar over the keyboard, maybe in small devices it could cover some displayed views such as the button. In this case I think the best way is to simply add a TextView warning about the error, but let's collect more opinions @abelgardep @jesmrec

jesmrec:
[...]

PR #3210, opened 2021-04-22
[FIX] Snackbar in passcode view is not displayed

JuancaG05:
## Related Issues
App: fixes https://github.com/owncloud/android/issues/2722

_____

## QA

Checks: https://github.com/owncloud/android/pull/3210#issuecomment-829191968

### DIFF
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsPasscodeTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsPasscodeTest.kt
@@ -37,6 +37,8 @@ import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.ActivityTestRule
 import com.owncloud.android.R
 import com.owncloud.android.ui.activity.PassCodeActivity
+import com.owncloud.android.utils.matchers.isDisplayed
+import com.owncloud.android.utils.matchers.withText
 import org.junit.After
 import org.junit.Assert.assertThat
 import org.junit.Assert.assertTrue
@@ -70,15 +72,22 @@ class OCSettingsPasscodeTest {
         //Open Activity in passcode creation mode
         openPasscodeActivity(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)
 
-        onView(withId(R.id.header)).check(matches(isDisplayed()))
-        onView(withId(R.id.explanation)).check(matches(isDisplayed()))
-        onView(withText(R.string.pass_code_configure_your_pass_code)).check(matches(isDisplayed()))
-        onView(withText(R.string.pass_code_configure_your_pass_code_explanation)).check(matches(isDisplayed()))
+        with(R.id.header) {
+            isDisplayed(true)
+            withText(R.string.pass_code_configure_your_pass_code)
+        }
+        with(R.id.explanation) {
+            isDisplayed(true)
+            withText(R.string.pass_code_configure_your_pass_code_explanation)
+        }
         onView(withId(R.id.txt0)).check(matches(isDisplayed()))
         onView(withId(R.id.txt1)).check(matches(isDisplayed()))
         onView(withId(R.id.txt2)).check(matches(isDisplayed()))
         onView(withId(R.id.txt3)).check(matches(isDisplayed()))
-        onView(withId(R.id.cancel)).check(matches(isDisplayed()))
+        with(R.id.cancel) {
+            isDisplayed(true)
+            withText(android.R.string.cancel)
+        }
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
```diff
@@ -342,9 +342,9 @@ private void processFullPassCode() {
                 savePassCodeAndExit();
 
             } else {
-                showErrorAndRestart(
-                        R.string.pass_code_mismatch, R.string.pass_code_configure_your_pass_code, View.VISIBLE
-                );
+                mPassCodeHdr.setText(R.string.pass_code_mismatch);
+                mPassCodeHdrExplanation.setVisibility(View.VISIBLE);
+                clearBoxes();
```
abelgardep, line 347: `showErrorAndRestart` is used somewhere else. That means that some errors will be shown in a snack bar and we will have the same problem after this fix. Could we try to show the potential errors in the same way? What do you think?

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -232,7 +232,7 @@
     <string name="pass_code_configure_your_pass_code_explanation">The passcode will be requested every time the app is started</string>
     <string name="pass_code_reenter_your_pass_code">Please reenter your passcode</string>
     <string name="pass_code_remove_your_pass_code">Remove your passcode</string>
-    <string name="pass_code_mismatch">The passcodes are not the same</string>
+    <string name="pass_code_mismatch">The passcodes are not the same\n Enter your password</string>
```
abelgardep, line 235: I would keep the error and title in different strings. I would not mix them. Showing the error in a different field would do the job. What do you think?

abelgardep (changes requested):
Just two little suggestions. BTW, what do you think about this approach https://github.com/owncloud/android/issues/2722#issuecomment-826552591 ? @JuancaG05

owncloudApp/src/main/res/values/strings.xml
JuancaG05: Ok, I will take the approach you suggested in the comment so that these two requested changes are applied too 🍻

JuancaG05:
@abelgardep new approach you suggested applied over here 🍻

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsPasscodeTest.kt
```diff
@@ -192,6 +190,7 @@ class OCSettingsPasscodeTest {
         typePasscode(wrongPassCode)
 
         onView(withText(R.string.pass_code_enter_pass_code)).check(matches(isDisplayed()))
+        onView(withText(R.string.pass_code_wrong)).check(matches(isDisplayed()))
```
abelgardep, line 193: Since we are showing this text in the layout, I would try to use `withId` instead of `withText`. That way, we will check that the text is shown in that specific TextView. Otherwise, we will check that that text is shown somewhere and is not as accurate.

Let's give it a try 🍻

owncloudApp/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
```diff
@@ -352,13 +356,8 @@ private void processFullPassCode() {
...
         Arrays.fill(mPassCodeDigits, null);
-        CharSequence errorSeq = getString(errorMessage);
-        Snackbar snackbar = Snackbar.make(
-                findViewById(android.R.id.content),
-                errorSeq,
-                Snackbar.LENGTH_LONG
-        );
-        snackbar.show();
+        mPassCodeError.setText(errorMessage);
+        mPassCodeError.setVisibility(View.VISIBLE);
```
abelgardep, line 360: 👍

abelgardep (changes requested):
Tiny suggestion and it will be ready to QA 👍  @JuancaG05

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsPasscodeTest.kt
JuancaG05: Ok, I adapted the rest of tests also

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/OCSettingsPasscodeTest.kt
```diff
@@ -70,15 +72,22 @@ class OCSettingsPasscodeTest {
...
-        onView(withText(R.string.pass_code_configure_your_pass_code)).check(matches(isDisplayed()))
-        onView(withText(R.string.pass_code_configure_your_pass_code_explanation)).check(matches(isDisplayed()))
+        with(R.id.header) {
+            isDisplayed(true)
+            withText(R.string.pass_code_configure_your_pass_code)
+        }
+        with(R.id.explanation) {
+            isDisplayed(true)
+            withText(R.string.pass_code_configure_your_pass_code_explanation)
+        }
```
abelgardep, line 82: 🚀

abelgardep (approved):
Approved! Good job. Ready to QA @jesmrec

jesmrec:
QA checks

- [X] Create passcode: correct 2nd try -> passcode created 
- [X] Create passcode: incorrect 2nd try -> new error message
- [X] Enter correct passcode -> enter the app
- [X] Enter incorrect passcode -> new error message

Issue #2867, opened 2020-04-27
[FEATURE REQUEST] Move to AndroidX Preference and new structure for settings


jesmrec:
Current implementation of Settings is deprecated. So, it is recommended to move to AndroidX Preferences.

Move to [Android X implementation](https://developer.android.com/guide/topics/ui/settings), fix flaky tests and check viability #2651 

This issue is an epic, that will have several individual issues attached to move every section of the Settings to the new implementation.

Every implemented section will be merged to a branch, that will be merged into `master` when all the work is finished, to avoid bad intermediate statuses.

**Additional context**
- https://developer.android.com/guide/topics/ui/settings
- https://medium.com/google-developer-experts/exploring-android-jetpack-preferences-8bcb0b7bdd14
- https://blog.mindorks.com/implementing-android-jetpack-preferences

### Issues list
- Camera uploads section #3130 -> DONE #3183
- Security section #3099 -> DONE #3120 
- Logs section #3101 -> DONE #3150 & #3159
- More section #3100 -> DONE #3136
- Kotlinize PrivacyPolicyActivity #3129  -> DONE #3142
- Refactor tests for More and Security settings sections #3148 -> DONE #3157
- Organize settings by subsections #3153 -> DONE #3168
- Divide SettingsViewModel into several ViewModels #3175 -> DONE #3178

PR: 

- App #3143
- Library (if needed)

JuancaG05:
Due to a Google's intern bug, lint is not passing with Fragment version 1.3.1, showing the following error:
`/bitrise/src/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsFragment.kt:72: Error: Upgrade Fragment version to at least 1.3.0. [InvalidFragmentVersionForActivityResult]
          registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
          ^`

To temporary fix it, we have added a new line to `lint.xml`:
`<issue id="InvalidFragmentVersionForActivityResult" severity="ignore" />`

The issue will be fixed in the next release, so we MUST remember to remove that line before merging with master, checking whether the new Activity release took place

Link to Google's issue: https://issuetracker.google.com/issues/182388985

JuancaG05:
[...]

PR #3213, opened 2021-04-28
[Feature] AndroidX settings and new structure changelog

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/2867 (adding the changelog for this epic)

### DIFF
--- a/changelog/unreleased/3143
+++ b/changelog/unreleased/3143
@@ -1,0 +1,8 @@
+Enhancement: Move to AndroidX Preference and new structure for settings
+
+Settings have been updated to use the current Android's recommendation, AndroidX framework.
+In addition, they have been reorganized into subsections for a better understanding and
+navigation structure. Also, new features have been added: now, source path and behaviour
+in auto uploads can be chosen differently for pictures and videos.
+
+https://github.com/owncloud/android/issues/2867 https://github.com/owncloud/android/pull/3143
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Summary
 
 * Bugfix - Fix navbar is visible in file preview screen after rotation: [#3184](https://github.com/owncloud/android/pull/3184)
 * Enhancement - Replace blank view in music player with cover art: [#3121](https://github.com/owncloud/android/issues/3121)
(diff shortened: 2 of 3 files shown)


changelog/unreleased/3143
```diff
@@ -0,0 +1,8 @@
+Enhancement: Move to AndroidX Preference and new structure for settings
```
jesmrec, line 1: I'd set is as `Feature`. Enhancement is like a scoped improvement in a feature. What do you think?

changelog/unreleased/3143
```diff
@@ -0,0 +1,8 @@
...
+
+Settings have been updated to use the current Android's recommendation, AndroidX framework.
+In addition, they have been reorganized into subsections for a better understanding and
+navigation structure. Also, new features have been added: now, source path and behaviour
+in auto uploads can be chosen differently for pictures and videos.
+
+https://github.com/owncloud/android/issues/2867 https://github.com/owncloud/android/pull/3143
```
jesmrec, line 8: Should the links be placed in different lines?

changelog/unreleased/3143
JuancaG05: I agree and though the same, but I wrote `Enhancement` as in the [README](https://github.com/owncloud/android/blob/feature/settings_androidx/changelog/README.md) it says that the four types of changes are: `Bugfix, Change, Enhancement, Security`. Enhancement is the one which fits the most from my point of view, let me know your opinion

changelog/unreleased/3143
JuancaG05: I think it would be better, but taking a look at the [template](https://github.com/owncloud/android/blob/feature/settings_androidx/changelog/TEMPLATE.md) and some other files in the [unreleased folder](https://github.com/owncloud/android/tree/feature/settings_androidx/changelog/unreleased) I found that most of them follow this structure

changelog/unreleased/3143
jesmrec: You're right, it fits the best.. `feature` is only for the branch name. My fault!

changelog/unreleased/3143
jesmrec: ok, imho it will be more structured in different lines, but, as you said, we have to change more items... will do in a separate PR in that case

jesmrec (approved):

Issue #3138, opened 2021-03-08
[FEATURE REQUEST] Use new extensions in the whole project


JuancaG05:
New Activity extensions were added in https://github.com/owncloud/android/pull/3136 to avoid code repetition. We realized these extensions could be used in some other parts of the project and that way get a more consistent code, so that's the aim of this issue.

### TASKS

- [ ] Research (if needed)
- [ ] Create new branch
- [ ] Development tasks
  - [ ] Search for replaceable code with these extensions, and adapt it
- [ ] Make sure all the tests keep passing
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

JuancaG05:
The objective here is to replace in the whole code some Intents creations by a call to these new extensions, so that everything is more unified. We might have to add some other simple extensions, depending on the way Intents are created and which properties are set, but doesn't seem too complex at first sight

jesmrec:
QA -> open web views in browsers. open mail app ...

jesmrec:
QA checks

- [x] OAuth2 auth
- [x] OIDC auth
- [x] Settings > More > Help
- [x] Settings > More > Davx5
- [x] Settings > More > Feedback
- [x] Settings > More > Recommend
- [x] Settings > More > Privacy policy

jesmrec:
Approved on my side

PR #3224, opened 2021-05-04
[Feature] Use new extensions in the whole project

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3138
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -55,11 +55,12 @@ fun Activity.showErrorInToast(
 
 fun Activity.goToUrl(
     url: String,
-    flags: Int = 0) {
+    flags: Int? = null
+) {
     if (url.isNotEmpty()) {
         val uriUrl = Uri.parse(url)
         val intent = Intent(Intent.ACTION_VIEW, uriUrl)
-        if (flags != 0) intent.addFlags(flags)
+        if (flags != null) intent.addFlags(flags)
         startActivity(intent)
     }
 }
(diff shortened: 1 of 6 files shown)


JuancaG05:
Apart from the Activity extensions mentioned in the issue (`goToUrl` and `sendEmail`), the extension `showMessageInSnackbar` has also been used throughout all the code where possible

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -53,10 +53,13 @@ fun Activity.showErrorInToast(
...
 
-fun Activity.goToUrl(url: String) {
+fun Activity.goToUrl(
+    url: String,
+    flags: Int = 0) {
     if (url.isNotEmpty()) {
         val uriUrl = Uri.parse(url)
         val intent = Intent(Intent.ACTION_VIEW, uriUrl)
+        if (flags != 0) intent.addFlags(flags)
         startActivity(intent)
```
abelgardep, line 63: Reformat this code, and I think that instead of using 0 as the default value, we could make it nullable and add flags only when we receive actual flags. What do you think?

abelgardep (changes requested):
Just a tiny suggestion @JuancaG05

abelgardep (changes requested):
@JuancaG05 Latest thing before moving this forward 👍

JuancaG05:
@abelgardep ready for final check and approve 👌

jesmrec:
QA checks

- [x] OAuth2 auth
- [x] OIDC auth
- [x] Settings > More > Help
- [x] Settings > More > Davx5
- [x] Settings > More > Feedback
- [x] Settings > More > Recommend
- [x] Settings > More > Privacy policy

Issue #2714, opened 2019-11-13
AsyncTask to Kotlin Coroutines


abelgardep:
Asynctask will be [deprecated](https://www.xda-developers.com/asynctask-deprecate-android-11/), so it could be a good option to start moving current asynctasks to Kotlin Coroutines.

farhazulMullick:
I installed this project on my local machine but Terminal window prompt "> Could not resolve project :owncloud-android-library:owncloudComLibrary."

 Doesn't say anything about com.android.build.api.attributes.BuildTypeAttr (required 'release')
                  - Doesn't say anything about its usage (required an API)
                  - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'androidJvm')

michaelstingl:
> What can i do for a clean build

@farhazulMullick please use search to find previous issues:
https://github.com/owncloud/android/issues?q=is%3Aissue+%22could+not+resolve%22

You comment is unrelated to this issue. I'll hide it.

abelgardep:
> I installed this project on my local machine but Terminal window prompt "> Could not resolve project :owncloud-android-library:owncloudComLibrary."
> 
> Doesn't say anything about com.android.build.api.attributes.BuildTypeAttr (required 'release')
> - Doesn't say anything about its usage (required an API)
> - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'androidJvm')
> What can i do for a clean build

Hi @farhazulMullick, check this issue: https://github.com/owncloud/android/issues/3196

Probably you missed the `--recursive` clause when cloning the repo.

Follow the [setup guide](https://github.com/owncloud/android/blob/master/SETUP.md#1-fork-and-download-the-owncloudandroid-repository) and tell us if something is still wrong.

abelgardep:
Related to this issue about Asynctasks and kotlin coroutines, we are replacing async tasks with coroutines along the way. So I think that we can close this issue.

farhazulMullick:
@ablegardep are you working on this
issue if not i can work on this

Issue #2351, opened 2018-11-12
Welcome new Android components [WIP]


davigonz:
Android is getting more mature and modern and it would be great to start using some of the next technologies: 
- [Android architecture components](https://developer.android.com/topic/libraries/architecture/)
- [Kotlin](https://developer.android.com/kotlin/)

To do that, we first need to define some tasks:
- Update Gradle + support for Android Studio 3.2.1: #2327 
- Remove Eclipse structure: #2044 
- Architecture research: 
   - SOLID https://stackify.com/solid-design-principles/
   - Read documentation and try some basic examples with new android components and Kotlin integration. 
   - Define ownCloud use cases.
   - Check how different parts of the current arch will be added to the new one. Some diagrams to clarify.
- Clean up + refactors, keep working on https://github.com/owncloud/android/issues/2153
- Define lint policies and publish them for community
- Implement first use case: get server info, shares, rename fine, not defined yet. This first use case should include tests (unit, integration, UI)
- Use current `RemoteOperationResult` objects or opt for a new exceptions mechanism.

davigonz:
Let's start with an "easy" usecase like **Get public shares for a file.** To do that, let's follow the next architecture:

![owncloud android app architecture](https://user-images.githubusercontent.com/15261363/49936572-8f918280-fed4-11e8-9a86-103169b897bd.png)

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are
closing it as outdated. If this is still relevant on a current ownCloud release,
please **reopen** (or open a fresh issue) with up-to-date details — thank you\!

PR #3236, opened 2021-05-15
Async task/async task to coroutines

farhazulMullick:
## Related Issues
App: 
AsyncTask_to_Coroutines #2714

Library PR (if needed):

-. Added Coroutines Support to all AsyncTask class
_____

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ buildscript {
         sqliteVersion = "2.1.0"
         lifecycleLiveData = "2.3.1"
         biometricVersion="1.0.1"
+        coroutinesVersion= "1.4.2"
 
         // Kotlin
         kotlinVersion = "1.4.32"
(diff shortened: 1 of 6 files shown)


farhazulMullick:
@ChrisEdS 
would you like to review my pull request

abelgardep:
Hi @farhazulMullick, this is not exactly what we expect. We are replacing AsyncTasks with coroutines but in a different way. We are moving to [MVVM architecture](https://github.com/owncloud/android/issues/2351) and this logic may belong to a ViewModel or another module, so we are replacing them step by step. BTW we are adding unit tests to assure they work the way they are supposed.

That's why I closed #2714 

Anyway, thanks for your interest.

hannesa2:
We are **_replacing_** ... is there an open pull request or a branch without pull request ? 

You can anyway accept this pull request and continue your hidden **_replacing_**

It's not smart to close an un-merged pull request. This causes mostly to see this as last pull request from this person and you loose one additional of your few contributors.

abelgardep:
Maybe "We are replacing" is not the best expression.

I meant that we have already replaced some async tasks during the new architecture process. For example `AuthenticatorAsyncTask` was replaced in https://github.com/owncloud/android/pull/2826 and that logic was moved to domain and data module with their corresponding unit tests.

That's why I closed the issue. We are doing this along the way 👍

Issue #3239, opened 2021-05-17
[BUG] Crash when sharing with ownCloud


jesmrec:
extracted from Google Play console:

```
java.lang.IllegalStateException: 
  at androidx.fragment.app.FragmentManager.checkStateLoss (FragmentManager.java:1844)
  at androidx.fragment.app.FragmentManager.enqueueAction (FragmentManager.java:1884)
  at androidx.fragment.app.BackStackRecord.commitInternal (BackStackRecord.java:329)
  at androidx.fragment.app.BackStackRecord.commit (BackStackRecord.java:294)
  at androidx.fragment.app.DialogFragment.dismissInternal (DialogFragment.java:355)
  at androidx.fragment.app.DialogFragment.dismiss (DialogFragment.java:307)
  at com.owncloud.android.ui.activity.BaseActivity.dismissLoadingDialog (BaseActivity.java:315)
  at com.owncloud.android.ui.activity.ReceiveExternalFilesActivity.onTmpFilesCopied (ReceiveExternalFilesActivity.java:850)
  at com.owncloud.android.ui.asynctasks.CopyAndUploadContentUrisTask.onPostExecute (CopyAndUploadContentUrisTask.java:240)
  at com.owncloud.android.ui.asynctasks.CopyAndUploadContentUrisTask.onPostExecute (CopyAndUploadContentUrisTask.java:47)
  at android.os.AsyncTask.finish (AsyncTask.java:755)
  at android.os.AsyncTask.access$900 (AsyncTask.java:192)
  at android.os.AsyncTask$InternalHandler.handleMessage (AsyncTask.java:772)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  at android.os.Looper.loop (Looper.java:213)
  at android.app.ActivityThread.main (ActivityThread.java:8178)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:513)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1101)
```

it occurs with Android10 the most (57%) but also with other older versions

abelgardep:
I was able to reproduce it. You can try these steps:
- Share several files with ownCloud
- When loading dialog appears, move the app to the background
- It will try to dismiss the dialog and it will crash

Should be fixed via #3240

jesmrec:
Fixed!

PR #3240, opened 2021-05-17
Fix crash when sharing with ownCloud

abelgardep:
## Related Issues
App: https://github.com/owncloud/android/issues/3239
_____


jesmrec:
Approved on my side. Crash is not reproducible anymore with the detailed steps.

jesmrec (approved):

PR #3250, opened 2021-05-20
Replace mappers with kotlin extensions

abelgardep:
Mapper implementation added unneeded complexity and nullability. With Kotlin extensions, we can get rid of some ugly `!!`


owncloudData/src/androidTest/java/com/owncloud/android/data/sharing/shares/db/OCShareDaoTest.kt
```diff
@@ -77,33 +76,27 @@ class OCShareDaoTest {
     fun insertSharesFromDifferentFilesAndRead() {
         ocShareDao.insert(
```
davigonz, line 77: You could move this kind of calls to a `givenListOfShares` method to make the test more readable, like part of the `given, that, then` in tests... 🍻

owncloudData/src/androidTest/java/com/owncloud/android/data/user/UserDaoTest.kt
```diff
@@ -57,23 +57,23 @@ class UserDaoTest {
 
     @Test
     fun insertQuotaForAccount() {
-        userDao.insert(userQuotaMapper.toEntity(OC_USER_QUOTA)!!.copy(accountName = OC_ACCOUNT_NAME))
+        userDao.insert(OC_USER_QUOTA.toEntity(OC_ACCOUNT_NAME))
```
davigonz, line 60: You could also have `OC_USER_QUOTA_ENTITY` somewhere and call `OC_USER_QUOTA.toEntity` from there 🍻

owncloudData/src/main/java/com/owncloud/android/data/capabilities/datasources/implementation/OCLocalCapabilitiesDataSource.kt
```diff
@@ -19,29 +19,95 @@
...
+            OCCapability(
+                id = id,
+                accountName = accountName,
+                versionMayor = versionMayor,
+                versionMinor = versionMinor,
+                versionMicro = versionMicro,
+                versionString = versionString,
+                versionEdition = versionEdition,
+                corePollInterval = corePollInterval,
+                davChunkingVersion = davChunkingVersion,
```
davigonz, line 61: Maybe you could remove the named params here since are using the same name

owncloudData/src/main/java/com/owncloud/android/data/capabilities/datasources/implementation/OCLocalCapabilitiesDataSource.kt
```diff
@@ -19,29 +19,95 @@
...
+                filesSharingPublicExpireDateEnforced = CapabilityBooleanType.fromValue(filesSharingPublicExpireDateEnforced),
+                filesSharingPublicUpload = CapabilityBooleanType.fromValue(filesSharingPublicUpload),
+                filesSharingPublicMultiple = CapabilityBooleanType.fromValue(filesSharingPublicMultiple),
+                filesSharingPublicSupportsUploadOnly = CapabilityBooleanType.fromValue(filesSharingPublicSupportsUploadOnly),
+                filesSharingResharing = CapabilityBooleanType.fromValue(filesSharingResharing),
+                filesSharingFederationOutgoing = CapabilityBooleanType.fromValue(filesSharingFederationOutgoing),
+                filesSharingFederationIncoming = CapabilityBooleanType.fromValue(filesSharingFederationIncoming),
+                filesBigFileChunking = CapabilityBooleanType.fromValue(filesBigFileChunking),
+                filesUndelete = CapabilityBooleanType.fromValue(filesUndelete),
+                filesVersioning = CapabilityBooleanType.fromValue(filesVersioning)
```
davigonz, line 79: We have a bunch of properties here, maybe splitting them up in some new classes? 🍻

owncloudData/src/androidTest/java/com/owncloud/android/data/sharing/shares/db/OCShareDaoTest.kt
abelgardep: Well, probably that is out of this PR scope, here I have just replaced the mappers. If we need to refactor some test naming, I think that we could do it in another PR.

owncloudData/src/main/java/com/owncloud/android/data/capabilities/datasources/implementation/OCLocalCapabilitiesDataSource.kt
abelgardep: Well, I like the named params even if we use the same name. And there is no counterpart of using them. So, I would keep them 👍

owncloudData/src/main/java/com/owncloud/android/data/capabilities/datasources/implementation/OCLocalCapabilitiesDataSource.kt
abelgardep: In this PR, I just wanted to remove the mappers. Splitting this class into smaller ones will need probably database changes and migrations and so on. So, I would keep it at least for the moment 👍

owncloudData/src/androidTest/java/com/owncloud/android/data/user/UserDaoTest.kt
abelgardep: Ok, applied 👍

davigonz (approved):

Issue #2899, opened 2020-06-11
Replace the ownCloud file picker with the Android one


abelgardep:
As a continuation of #2436, where we replaced the ownCloud file picker with the Android one when uploading files, oC app should replace completely the file picker with the android one.

This oC file picker is still used to select the camera folder available for Camera Uploads in Preferences. Steps:
1- Open settings
2- Enable picture uploads
3- Choose camera folder
| oC Custom file picker | Storage Access Framework |
| ------------- | ------------- |
| ![custom-file-picker](https://user-images.githubusercontent.com/47524927/84506126-29e7a780-acbf-11ea-8d77-21938a096e74.jpg) | ![saf](https://user-images.githubusercontent.com/47524927/84506303-7632e780-acbf-11ea-95c6-27d134358add.jpg)| 

With this replacement, we will perform a big cleanup and it will be ready for Scoped Storage.
### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/camera_uploads
- [x] Development tasks
  - [x] Use SAF file picker
  - [x] decide weather to use SAF for accessing photos
  - [x] if yes: move accessing local storage to SAF.
  - [x] Create a new model in domain module for camera uploads config
  - [x] Add a new worker to perform camera uploads
  - [x] Add a new usecase to retrieve camera uploads configuration
  - [x] Add a new worker to perform single uploads with a content uri
  - [x] Add a constraint to perform uploads only when wifi is enabled
  - [x] Remove local file if behavior for camera uploads is move
  - [x] Show the uploads in the uploads list view
  - [x] Cancel uploads when removing the account
  - [x] Cancel uploads individually
  - [x] Notifications: Syncing camera uploads and uploading file individually (Check if needed)
  - [x] Improve error handling: Add new errors cases
  - [x] Check problems when upgrading: Previous uri is not valid
  - [x] Clean legacy code
  - [x] Fix unit tests
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch new_arch/camera_uploads into master

### PR

- App #3293
- Library (if needed) https://github.com/owncloud/android-library/pull/414

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.18.1/Camera%20uploads%20%2B%20picker.md

Bugs / improvements

jesmrec:
Add another case: share with ownCloud

Case 1:

1. Open "Photos". 
2. Select one or more pictures.
[...]

PR #3293, opened 2021-07-02
[New arch] Camera uploads

abelgardep:
## Related Issues
Implements #2899 

Library PR (if needed): https://github.com/owncloud/android-library/pull/414

## Screenshots
| Timestamp field added | Native picker to select camera folder |
| ----- | ----- |
| new_field| camera_uploads_picker |

| Camera upload succeeded notification | Camera upload failed notification |
| ----- | ----- |
| notification_success | notification_fail |

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.18.1/Camera%20uploads%20%2B%20picker.md

- [X] (1) Notification of bad source path https://github.com/owncloud/android/pull/3293#issuecomment-874754395 (Ready to test)
- [X] (2) Notification with 0 uploads https://github.com/owncloud/android/pull/3293#issuecomment-874822188 (Ready to test)
- [X] (3) String Camera folder https://github.com/owncloud/android/pull/3293#issuecomment-874845675 [WONT DO]
- [x] (4) Repetitions in uploads https://github.com/owncloud/android/pull/3293#issuecomment-875373972 [FIXED]
- [X] (5) Upload location changes unexpectedly https://github.com/owncloud/android/pull/3293#issuecomment-875559535
- [x] (6) Move option behaves like keep https://github.com/owncloud/android/pull/3293#issuecomment-876312428 (Ready to test) [FIXED]


owncloudApp/build.gradle
```diff
@@ -36,9 +36,11 @@ dependencies {
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0'
```
jabarros, line 39: Don't you define this versions on common file?

owncloudApp/src/main/java/com/owncloud/android/workers/CameraUploadsWorker.kt
```diff
@@ -0,0 +1,291 @@
...
+            timeOut = null
+        )
+    }
+
+    private fun updateTimestamp(folderBackUpConfiguration: FolderBackUpConfiguration, syncType: SyncType) {
+        val currentTimestamp = System.currentTimeMillis()
+
+        when (syncType) {
+            SyncType.PICTURE_UPLOADS -> {
+                val savePictureUploadsConfigurationUseCase: SavePictureUploadsConfigurationUseCase by inject()
```
jabarros, line 204: cool

owncloudApp/src/main/java/com/owncloud/android/workers/CameraUploadsWorker.kt
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+    }
+
+    private fun getFilesReadyToUpload(
+        syncType: SyncType,
+        sourcePath: String,
+        lastSyncTimestamp: Long,
+    ): List<DocumentFile> {
+        val simpleDateFormat = SimpleDateFormat("yyyy/MM/dd HH:mm:ss", Locale.getDefault())
```
jabarros, line 223: Maybe it is better to place it as a constant

owncloudDomain/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
 
     // Kotlin
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0'
```
jabarros, line 42: Same here with version

abelgardep: You are absolutely right!!

owncloudApp/src/main/java/com/owncloud/android/workers/CameraUploadsWorker.kt
abelgardep: I will remove it directly, we can use the Date().toString() 💯

jabarros (approved):
LGTM

jesmrec:
### (1)

1. Install from scratch
2. Enable instant uploads 

Current: notification with `Picture uploads source path is not valid anymore`.  By clicking on the notification, user is redirected to the `Uploads` view

Expected: Notification should point the user to the proper section in `Settings`.  Also (as another idea), i'd add it in the notification message with something like: `Click to set up correctly`

Pixel 2
Android 11
`5bea766f`

jesmrec:
### (2)

Every time the camera folder is scanned, a new notification arises with the number of uploads to be done (replacement of the progress bar).

But, notification is also shown when there is no uploads to do. Would it be posible not to show it when the number is 0?  if no pictures/videos were taken nothing new has happened to notify.

Pixel 2
Android 11
`5bea766f`

jesmrec:
### (3)

This is more a question than report

The camera folder is shown as `/tree/primary:DCIM/xxxxx`

is this string handleable? in case it is, i'd get rid of the first chunk (`/tree/primary`) and let the "real" path for the user (`DCIM/...`).

jesmrec:
> I agree that the notification should point to the proper section in 'Settings'. Related to the message, you mean to add that text to the current one? In that case, yes, we could add it, but I would not replace the current message.

yes, adding the message to the current one. So, users will know that they have to do something, apart of being notified.

> Of course!! We can remove the notification when nothing new is found. Up to you, I think that previously the "Syncing camera uploads" were shown independently of the files found. That's why I kept it this way.

probably, in testing stage i enter/leave `Settings` many times, so, i get a bunch of notifications with `0 uploads`, that could be annoying. In real cases won't happen this... but i'd prefer to use notifications for cases in which new uploads are incoming. 

> Yes, it is completely handleable. Removing the/tree/primary path could be confusing for SD card paths. But, it is completely fine to modify the String.

I did not check with SD card yet... let me check before making a final decision about (3)

jesmrec:
### (4)

I am experiencing repetitions in uploads in different devices. That means:

1. Take 2 pictures
2. Go to Settings and go back (the way to force triggering uploads)
3. After a couple of minutes, every pic is uploaded twice

I can see in the logs:

```
2021-07-07 09:33:01.143 8957-11393/com.owncloud.android.debug V/(UploadsStorageManager.java:102) .storeUpload(): Inserting content://com.android.externalstorage.documents/tree/primary%3ADCIM%2FCamera/document/primary%3ADCIM%2FCamera%2FPXL_20210707_073145894.jpg with status=UPLOAD_IN_PROGRESS
...
2021-07-07 09:33:02.633 8957-11394/com.owncloud.android.debug V/(UploadsStorageManager.java:138) .updateUpload(): Updating content://com.android.externalstorage.documents/tree/primary%3ADCIM%2FCamera/document/primary%3ADCIM%2FCamera%2FPXL_20210707_073145894.jpg with status=UPLOAD_SUCCEEDED
```

and some seconds later:

````
2021-07-07 09:33:20.282 8957-11851/com.owncloud.android.debug V/(UploadsStorageManager.java:102) .storeUpload(): Inserting content://com.android.externalstorage.documents/tree/primary%3ADCIM%2FCamera/document/primary%3ADCIM%2FCamera%2FPXL_20210707_073145894.jpg with status=UPLOAD_IN_PROGRESS
...
2021-07-07 09:33:21.432 8957-8994/com.owncloud.android.debug D/(UploadsStorageManager.java:155) .updateUpload(): updateCameraUploadSync returns with: 1 for file: content://com.android.externalstorage.documents/tree/primary%3ADCIM%2FCamera/document/primary%3ADCIM%2FCamera%2FPXL_20210707_073145894.jpg

same picture is uploaded again

could it be a matter of a non-updated last sync date?

Pixel 2
Android 11
`5bea766f`

jesmrec:
about (3), i'd keep the whole string, indicating whether SDcard or internal storage is used.

catching files from SD card seems to work fine.

jesmrec:
### (5)

2. Enable Camera and Video uploads, both with same set up (same account, folder and camera folder)
3. Wait till `WorkManager` passes at least once for photos and once for videos (here, the notification asked to be removed in (2) helps a lot to know when happens)
4. Open Settings -> Video uploads
5. Change `Video upload path` for a different location 
6. Wait again for `WorkManager` for the videos
7. Open Settings -> Video uploads

Current: Location is the one selected in step 2
Expected: Location is the one selected in step 5

Pixel 2 Android 11
Nexus 6P Android 7
`5bea766f`

abelgardep:
I see that there are some problems that could be fixed with a little tweak.

At the moment, each time we leave the picture uploads section or video uploads section, we enqueue the work to perform the camera uploads replacing the current one. ([More info about it](https://developer.android.com/topic/libraries/architecture/workmanager/how-to/managing-work#conflict-resolution))

This means that we cancel the current one and enqueue a new one, so it could lead to some errors when replacing the worker if the timestamp is not updated yet. The new worker could get the same timestamp as the previous one.

This behavior is pretty useful during development, so we don't need to wait till the next auto-trigger, at the moment 15 minutes. But it can lead to errors, so I will change to [KEEP](https://developer.android.com/reference/androidx/work/ExistingWorkPolicy#KEEP) policy. I think that this tweak could fix [(4)](https://github.com/owncloud/android/pull/3293#issuecomment-875373972) and [(5)](https://github.com/owncloud/android/pull/3293#issuecomment-875559535)

> probably, in testing stage I enter/leave Settings many times, so, I get a bunch of notifications with 0 uploads, that could be annoying. In real cases won't happen this... but I'd prefer to use notifications for cases in which new uploads are incoming.

With this change, the notification with 0 uploads won't be shown every time we enter settings sections. But we can hide them anyway.

@jesmrec

abelgardep:
(1) Ready to test. I didn't change the message since it is too long and it won't be readable on most devices.
(2) Done, No notification when there are no files to upload.
(4) and (5) According to [this comment](https://github.com/owncloud/android/pull/3293#issuecomment-875580225) I have changed the policy from REPLACE to KEEP, so they should be ready to test again

jesmrec:
### (6)

1. Enable camera uploads for pictures or videos and select correct camera folder
2. In `Original file will be`, select `move` 
3. Take pictures and wait till they are uploaded 

Current: Pictures are uploaded to oC, but the original file keeps in the device

Expected: Pictures are uploaded to oC and the original file is removed from device

Pixel 2 Android 11
090df6e88

jesmrec:
Everything passed. Nice new feature that will help a lot in the future!

Approved

Issue #3272, opened 2021-06-01
Note icon in music player to be branded


jesmrec:
Related to https://github.com/owncloud/android/issues/3121

The note icon is tinted with the main color of the app. Conceptually, this is correct.

Tecnically, the color is set [in this way](https://github.com/owncloud/android/tree/master/owncloudApp/src/main/res/drawable/ic_place_holder_music_cover_art.xml), using `@color/primary`

[The list of colors](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/res/values/colors.xml) is not brandable itself, it should be used as a parameter inside `setup.xml` afaik. i guess the affected key is: `primary_button_background_color`, corresponding to `Primary color` in internal branding.

PR #3297, opened 2021-07-02
Update the music icon with the branded toolbar color

abelgardep:
## Related Issues
App: #3272 

Let's use the same color as the toolbar, I think it is the more consistent way.

_____


jesmrec:
Cool! much better using the toolbar color. Looks much better than the primary color.

Approved.

jesmrec (approved):

PR #3302, opened 2021-07-08
Fix build with aarch64 Studio on Apple Silicon M1

hannesa2:
It's about to build with an aarch64 Android Studio https://stackoverflow.com/a/68285501/1079990


abelgardep:
I see, here is the issue: https://issuetracker.google.com/issues/174695268

hannesa2:
> At the moment I don't have an M1, so I cannot test it. We use room also in the ownCloudData module, isn't it affected by this problem?

It's not affected. With this fix I can build native, otherwise I've to use much slower x86

jesmrec:
Will merge. Once Room 2.4 release is out, the dependency will be removed.

abelgardep (approved):
Approved. Temporal fix, let's wait till Room 2.4 stable version, thanks @hannesa2

Issue #125, opened 2013-04-16
Files remain if deleting account


svenclemens:
### Expected behaviour

If deleting an account the files on the device associated to the account should get deleted.
### Actual behaviour

The account get deleted but the files will remain.
### Steps to reproduce
1. Download some files
2. Deleting your account (in owncloud app)
3. View downloaded files in file manager
### Environment data

Android version: 4.1.2

Device model: Samsung Galaxy S2

Stock or customized system: Stock

ownCloud app version: 1.3.20

ownCloud server version: 4.5.7

_____

EDIT (July 2021)

Check https://github.com/owncloud/android/issues/125#issuecomment-188351540

- App removal -> to do here

By removing account from oC app, remove the account folder from  app storage.

- Device Settings removal -> future feature (worker)

tobiasKaminsky:
What about a dialog "Do you want to remove all files on device?" 
Easy and simple ;)

AndyScherzinger:
I second @tobiasKaminsky suggestion, when deleting an account on the device the app should simply ask if the accounts data should be deleted on the device. :)

jochenwezel:
The suggested dialog depends on the *user* - but this behaviour must be controlled by the organization/Company who "owns" the data and who has to ensure that some people can't access that data any more.

tobiasKaminsky:
@jochenwezel 
I see your point but how can this be achieved?
The app then must verify that the account was deleted and that the option "force delete files on remote devices" is enabled and then delete the complete folder with all files?

@davivel Do you know how this is being handled by the desktop sync?

jochenwezel:
@tobiasKaminsky yes, this would match approx. to the requirement. Only 1 additional point: "force delete files on remote devices" should be default setting (if you like, a General Switch for changing the Default to "disabled" could be provided in administration area)

joeplus:
Just out of interest: How is it possible for an organization to administer the remote owncloud app and to ensure that when the user account is deleted (e.g. when the employee leaves the company) all files are deleted as well?

jochenwezel:
1. the Client app must still sync with the Server "with a valid Login"  (for this, the Server must still allow (at least for a few weeks) to Login with the current Login data)
2. after successful Login: Client app recieves delete notifications for all files and Folders
3. cleanup done on 1 device - maybe other devices/sync Clients follow as soon as they sync next time

joeplus:
[...]

PR #3336, opened 2021-07-29
clean old user directories

theScrabi:
Related issue: https://github.com/owncloud/android/issues/125

### QA

Bugs & reports

- [x] (1) NPE after installing https://github.com/owncloud/android/pull/3336#issuecomment-895120451 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
@@ -21,6 +21,8 @@
  */
 package com.owncloud.android
 
+import android.accounts.AccountManager
+import android.accounts.OnAccountsUpdateListener
 import android.app.Activity
 import android.app.Application
 import android.app.NotificationManager.IMPORTANCE_LOW
@@ -29,7 +31,9 @@ import android.content.pm.PackageInfo
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
+import android.os.Handler
 import android.view.WindowManager
+import com.owncloud.android.authentication.AccountUtils
 import com.owncloud.android.authentication.BiometricManager
 import com.owncloud.android.authentication.PassCodeManager
 import com.owncloud.android.authentication.PatternManager
@@ -53,6 +57,7 @@ import com.owncloud.android.ui.activity.WhatsNewActivity
 import com.owncloud.android.utils.DOWNLOAD_NOTIFICATION_CHANNEL_ID
 import com.owncloud.android.utils.FILE_SYNC_CONFLICT_CHANNEL_ID
 import com.owncloud.android.utils.FILE_SYNC_NOTIFICATION_CHANNEL_ID
+import com.owncloud.android.utils.FileStorageUtils
 import com.owncloud.android.utils.MEDIA_SERVICE_NOTIFICATION_CHANNEL_ID
 import com.owncloud.android.utils.UPLOAD_NOTIFICATION_CHANNEL_ID
 import org.koin.android.ext.koin.androidContext
@@ -83,6 +88,8 @@ class MainApp : Application() {
 
         // initialise thumbnails cache on background thread
         ThumbnailsCacheManager.InitDiskCacheTask().execute()
+
+        cleanupUnusedAccountDirectories()
 
         // register global protection with pass code, pattern lock and biometric lock
         registerActivityLifecycleCallbacks(object : ActivityLifecycleCallbacks {
@@ -205,6 +212,14 @@ class MainApp : Application() {
             description = getString(R.string.file_sync_notification_channel_description),
             importance = IMPORTANCE_LOW
         )
+    }
+
+    private fun cleanupUnusedAccountDirectories() {
+        val accountManager = AccountManager.get(this)
+        accountManager.addOnAccountsUpdatedListener(OnAccountsUpdateListener {
+            val accounts = AccountUtils.getAccounts(this)
+            FileStorageUtils.deleteUnusedUserDirs(accounts)
+        }, Handler(), false)
     }
 
     companion object {
--- a/owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
@@ -23,6 +23,7 @@
  */
 package com.owncloud.android.data.storage
 
+import android.accounts.Account
 import android.annotation.SuppressLint
 import android.net.Uri
 import android.os.Environment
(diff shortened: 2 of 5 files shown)


abelgardep:
UI tests are failing, please take a look @theScrabi

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
```diff
@@ -84,6 +85,9 @@ class MainApp : Application() {
         // initialise thumbnails cache on background thread
         ThumbnailsCacheManager.InitDiskCacheTask().execute()
 
+        // check for dangling account data and delete it
+        FileStorageUtils.performCleanup(this)
```
abelgardep, line 89: Maybe trigger this each time the app is opened is too much. Maybe triggering it once a day or something like that would be enough. We could do this on the background using a worker as this PR: https://github.com/owncloud/android/pull/3337

BTW, I think that we could find a better name for this function. With a more specific name, the comment could be omitted 👀

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
```diff
@@ -86,6 +89,52 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
...
+                }
+            }
+            if (!dir_is_ok) {
+               danglingDirs.add(dir)
+            }
+        }
+        return danglingDirs
+    }
+
+    private fun recursiveDeleteDir(dir: File) {
```
abelgardep, line 117: I think that there are some kotlin extensions to do this, so we don't need to do it by ourselfs

try `dir.deleteRecursively()`

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
```diff
@@ -86,6 +89,52 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
...
 
+    /**
+     * Checks if there is user data which does not have a corresponding account in the Account manager.
+     */
+    private fun getDanglingAccountDirs(context: Context?): List<File> {
+        val am = AccountManager.get(context)
+        val rootFolder = File(getRootFolderPath())
+        val danglingDirs = mutableListOf<File>()
+        for (dir in rootFolder.listFiles()) {
+            var dir_is_ok = false
```
abelgardep, line 100: In Kotlin we try to use underscores only for constants -> https://kotlinlang.org/docs/coding-conventions.html#property-names

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
```diff
@@ -86,6 +89,52 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
     @SuppressLint("UsableSpace")
     fun getUsableSpace(): Long = getPrimaryStorageDirectory().usableSpace
 
+    /**
+     * Checks if there is user data which does not have a corresponding account in the Account manager.
+     */
+    private fun getDanglingAccountDirs(context: Context?): List<File> {
+        val am = AccountManager.get(context)
```
abelgardep, line 96: Do not save some letters 👀   we should try to name it `accountManager` so we know exactly what is going on 👍

abelgardep: BTW, we have some AccountUtils to get logged accounts, we can reuse that code to retrieve the accounts.

abelgardep (dismissed):
Some comments here @theScrabi

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
theScrabi: I don't know about triggering a cleanup every day. This seems a bit to complicated to me. Especially because the check is not really computational expensive.

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
theScrabi: thx

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
theScrabi: thx

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
theScrabi: Sure. Should I use AccountUtils as a Singleton or should I use it via proper dependency injection?

theScrabi: Also why is AccountUtils existing twice? Once [here](https://github.com/owncloud/android/blob/2a23807f4ea79f6e7fa2683af59242892993b690/owncloudApp/src/main/java/com/owncloud/android/authentication/AccountUtils.java) and once [here](https://github.com/owncloud/android-library/blob/351682cc7f5d91c1c6d21d7a8debd541d7c45674/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java#L115)

theScrabi:
I basically started from the beginning again. There is a callback that is triggered when an account is deleted. For this callback to work the app is launched shortly when an account is deleted. This callback is triggered regardless of where the account is removed from: The app or Android settings. This also prevents periodic checking.

jesmrec:
QA checks to do here:

- [ ] Remove account in the app -> account folder is deleted (legacy storage)
- [ ] Remove account in device settings -> account folder is deleted (legacy storage)
- [ ] Remove account in the app -> account folder is deleted (scoped storage)
- [ ] Remove account in device settings -> account folder is deleted (scoped storage)

jabarros:
👍

jabarros (approved):
👍

jesmrec:
### (1) 

2. Add an account

Current: 

Android 7 -> app crashes (NPE below), after opening the list of files is displayed, but every item is listed twice. 
Android 9 -> NPE, but no crash
Android12b -> NPE, and infinite loop in list of files. Items not displayed

This is the NPE:

```
2021-08-09 12:23:10.526 8503-8503/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.owncloud.android.debug, PID: 8503
    java.lang.NullPointerException: rootFolder.listFiles() must not be null
        at com.owncloud.android.data.storage.LocalStorageProvider.getDanglingAccountDirs(LocalStorageProvider.kt:96)
        at com.owncloud.android.data.storage.LocalStorageProvider.deleteUnusedUserDirs(LocalStorageProvider.kt:115)
        at com.owncloud.android.utils.FileStorageUtils.deleteUnusedUserDirs(FileStorageUtils.java:244)
        at com.owncloud.android.MainApp.cleanupUnusedAccountDirectories$lambda-0(MainApp.kt:221)
        at com.owncloud.android.MainApp.lambda$SQ1UeDHSy5SdfZbu-bBboRMT-2I(Unknown Source:0)
        at com.owncloud.android.-$$Lambda$MainApp$SQ1UeDHSy5SdfZbu-bBboRMT-2I.onAccountsUpdated(Unknown Source:2)
        at android.accounts.AccountManager$18.run(AccountManager.java:2324)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
        at android.os.Looper.loop(Looper.java:288)
        at android.app.ActivityThread.main(ActivityThread.java:7829)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:982)
```

Nexus 6P v7
Pixel5 v12b
`4221c7a`

@theScrabi

jesmrec:
This is approved on my side, ready to merge once CI passes. 

Some testings checks are pending to do, when the current branch is merged with Scoped Storage. Added to 2.19 release checklist.

Issue #3328, opened 2021-07-23
[FEATURE REQUEST] Logs removal


jesmrec:
**Is your feature request related to a problem? Please describe.**

Scoped Storage will enclose all files in a SandBox. We must remove logs files to avoid exhausting device memory as time goes by

**Describe the solution you'd like**

a Worker will remove the old logs every 10 days

**Describe alternatives you've considered**

**Additional context**

AC:

- A worker will remove the logs in the background
- It will work every 10 days (parameter)
- Worker enqueued. Not network required.
- Create "Logs" folder and store logs inside, instead of the ownCloud root folder inside the ownCloud folder

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/logs_collector
- [x] Development tasks
  - [x] Create logs inside "Logs" folder instead of the ownCloud root one.
  - [x] Create a new worker to enqueue the logs collector
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/logs_collector into master

### PR

- App #3337 
- Library (if needed)

abelgardep:
Logs collector worker will be triggered every 7 days, and it will delete logs older than 7 days.

A new logs folder has been created to store the logs files: "Logs"

When moving to scoped storage, we just need to change the LocalStorageProvider from `LegacyStorageProvider` to `ScopedStorageProvider`. Logs stored in Legacy storage will be removed in the migration to ScopedStorage, so we don't need to worry about them.

jesmrec:
Question here... worker is executed no matter if the `Logs` switch is on/off?

PR #3337, opened 2021-07-30
[Feature] Old logs collector

abelgardep:
## Related Issues
App: #3328 

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
@@ -23,7 +23,6 @@ import androidx.work.CoroutineWorker
 import androidx.work.WorkerParameters
 import com.owncloud.android.MainApp
 import com.owncloud.android.data.storage.LocalStorageProvider
-import org.koin.core.KoinComponent
 import timber.log.Timber
 import java.io.File
 import java.io.IOException
@@ -36,7 +35,7 @@ class OldLogsCollectorWorker(
 ) : CoroutineWorker(
     appContext,
     workerParameters
-), KoinComponent {
+) {
 
     override suspend fun doWork(): Result {
         val logsDirectory = getLogsDirectory()
--- a/owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
@@ -26,6 +26,7 @@ import com.owncloud.android.data.storage.LocalStorageProvider
 import org.koin.core.KoinComponent
 import timber.log.Timber
 import java.io.File
+import java.io.IOException
 import java.util.Calendar
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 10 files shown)


owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
```diff
@@ -0,0 +1,86 @@
...
+) {
+
+    override suspend fun doWork(): Result {
+        val logsDirectory = getLogsDirectory()
+        val logsFiles = getLogsFiles(logsDirectory)
+
+        return try {
+            removeOldLogs(logsFiles)
+            Result.success()
+        } catch (ioException: IOException) {
```
theScrabi, line 47: If IOException and SecurityException are the only two exceptions you could combine both cases to a simple
``` kotlin
catch(exception: Exception) {
   Result.failure()
}
```

owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private fun getLogsDirectory(): File {
+        val localStorageProvider = LocalStorageProvider.LegacyStorageProvider(MainApp.dataFolder)
+        val logsPath = localStorageProvider.getLogsPath()
+        return File(logsPath)
+    }
+
+    private fun getLogsFiles(logsFolder: File): List<File> {
+        return logsFolder.listFiles()?.toList() ?: listOf()
```
theScrabi, line 61: Also you might want to catch the NPE exception which could be triggered here.

theScrabi (approved):

owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
abelgardep: No NPE can be triggered here. Kotlin syntax 👀   This would be the analog code, maybe it is easier to understand:

```
if (logsFolder.listFiles() != null) {
      return logsFolder.listFiles().toList()
} else {
      return listOf()
}
```

owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt
abelgardep: I agree in part. 

Catching them separately lets us handle each error in a different way. And we know exactly which are the potential errors.

In this case, that we don't do anything specific for each error, we could unify them, but I would keep them separately anyway.

jesmrec:
In order to test this, i will cheat the following value: 

https://github.com/owncloud/android/blob/0276e4e1e26350d9fbd2ed5eb0c9bad8ecf83ef1/owncloudApp/src/main/java/com/owncloud/android/workers/OldLogsCollectorWorker.kt#L82

setting a lower frequency for the worker

Just checking that after the correct time the correct files are deleted should be enough. No corner cases.

jesmrec:
These were the specific tests:

cheated code:

```
    companion object {
        const val OLD_LOGS_COLLECTOR_WORKER = "OLD_LOGS_COLLECTOR_WORKER"
        const val repeatInterval: Long = 2L
        val repeatIntervalTimeUnit: TimeUnit = TimeUnit.HOURS
        private const val numberOfDaysToKeepLogs = 2
    }
```

**Test 1:** 

1. Enable `Logging` in the app
2. In Device settings' , set manually a date 5 days earlier than today
3. Open the app and do something to generate logs 
4. Repeat 2. and 3., setting dates 4, 3, 2, 1 day earlier. In this way, we generate 5 files with different dates.
5. Set correct date in device settings'

After 2h, all log files were removed but the last two (newer than 2 days). ✅ 

**Test 2:** 

1. Enable `Logging` in the app
2. In Device settings' , set manually a date 5 days earlier than today
3. Open the app and do something to generate logs 
4. Repeat 2. and 3., setting dates 4, 3, 2, 1 day earlier. In this way, we generate 5 files with different dates.
5. In Device settings', set a date 4 days in the future.

After 2h, all log files were removed  ✅ 


**Test 3:** 

1. Enable `Logging` in the app
2. In Device settings' , set manually a date 5 days earlier than today
3. Open the app and do something to generate logs 
4. Repeat 2. and 3., setting dates 4, 3, 2, 1 day earlier. In this way, we generate 5 files with different dates.
5. In Device settings', set a date 3 days in the past.

After 2h, no files were removed ✅

jesmrec:
this is approved on my side.

Issue #3355, opened 2021-08-19
[New arch] Shared by link 


abelgardep:
At the moment, shared by link tab shows an empty list when there are current shared by link files.
We have the shared by link attribute stored in the database. We can see that the function to show every single shared by link file from this account is [commented](https://github.com/owncloud/android/blob/new_arch%2Fsynchronization/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt#L218) on at the moment. This should be fixed within this issue. The shared by link tab should show the shared files within this account.

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/shared_by_link
- [x] Development tasks
  - [x] Create Query to retrieve data from DB
  - [x] Create new method in Local Data Source
  - [x] Create new method in File Repository
  - [x] Create UseCase
  - [x] Implement Testing
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch new_arch/shared_by_link into new_arch/synchronization

### PR

- App:  #3359

abelgardep:
Done, shared by link files are now displayed in the Links tab.

PR #3359, opened 2021-08-23
[New arch] Shared by link

fesave:
## Related Issues
App: https://github.com/owncloud/android/issues/3355
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
@@ -221,7 +221,7 @@ class FileDataStorageManager : KoinComponent {
 
         val result = withContext(CoroutineScope(CoroutinesDispatcherProvider().io).coroutineContext) {
             getFilesSharedByLinkUseCase.execute(GetFilesSharedByLinkUseCase.Params(account.name))
-        }.getDataOrNull()
+        }.getDataOrNull() ?: emptyList()
         result
     }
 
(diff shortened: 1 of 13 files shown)


owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
```diff
@@ -215,50 +216,14 @@ class FileDataStorageManager : KoinComponent {
...
-//            result.sort()
-//            return result
-        }
+    fun sharedByLinkFilesFromCurrentAccount(): List<OCFile>? = runBlocking(CoroutinesDispatcherProvider().io) {
+        val getFilesSharedByLinkUseCase: GetFilesSharedByLinkUseCase by inject()
+
+        val result = withContext(CoroutineScope(CoroutinesDispatcherProvider().io).coroutineContext) {
+            getFilesSharedByLinkUseCase.execute(GetFilesSharedByLinkUseCase.Params(account.name))
+        }.getDataOrNull()
+        result
```
abelgardep, line 225: I think that we should return an empty list in case there are no shared by link files. What do you think?

abelgardep (changes requested):
Just a tiny suggestion here @fesave 👍

owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
fesave: I agree with you.

abelgardep (approved):
Approved on my side 👍

jesmrec:
Approved on my side as well.

Issue #3354, opened 2021-08-19
[New arch] Passcode activity


theScrabi:
[Passcode Activity](https://github.com/owncloud/android/blob/e6194afd35e533e08adecbd78a5aff3d66fe4871/owncloudApp/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java#L50) is still written in the old architecture. Recently I changed the [the code of this activity](https://github.com/owncloud/enterprise/issues/4696), however I was not able to write proper tests for it since the code is not in MVVM architecture yet.
In order to be able to test the activity properly and align it with our new architecture we should convert it to MVVM.

### TASKS

- [x] Create branch new_arch/passcode
- [x] Development tasks
  - [x] Create new PassCodeViewModel
  - [x] Move logic from PassCodeActivity to PassCodeViewModel
  - [x] Save and remove passcode in PassCodeActivity instead of SettingsSecurityFragment
  - [x] Include unit tests
  - [x] Incude UI tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch new_arch/passcode into master

PR #3361, opened 2021-08-26
[New arch] Passcode activity

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3354
_____

## Already merged
#3365
#3372 

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.19/Passcode%20-%20New%20Arch.md

BUGS & REPORTS:

- [X] (1) Soft keyboard vanishes between 1st and 2nd try https://github.com/owncloud/android/pull/3361#issuecomment-919845895 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
```diff
@@ -56,28 +56,22 @@ class SettingsSecurityFragment : PreferenceFragmentCompat() {
     private val enablePasscodeLauncher =
         registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
             if (result.resultCode != Activity.RESULT_OK) return@registerForActivityResult
-            val passcodeEnableResult = securityViewModel.handleEnablePasscode(result.data)
-            if (passcodeEnableResult.isSuccess) {
+            else {
```
abelgardep, line 59: The `else` clause could be omitted, couldn't it? In case `result.resultCode != Activity.RESULT_OK`, it will return, so it won't reach this code. Anyway, I think that it adds readability, so it's fine 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val preferencesProvider: SharedPreferencesProvider
+) : ViewModel() {
+
+    fun setPassCode(passcode: String) {
+        preferencesProvider.putString(PassCodeActivity.PREFERENCE_PASSCODE, passcode)
+        preferencesProvider.putBoolean(PassCodeActivity.PREFERENCE_SET_PASSCODE, true)
+    }
+
+    fun removePassCode() {
+        preferencesProvider.putBoolean(PassCodeActivity.PREFERENCE_SET_PASSCODE, false)
```
abelgardep, line 37: Don't we remove the current passcode? I mean, we keep the `PREFERENCE_PASSCODE` stored after removing the passcode, but should we delete it? 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fun checkPassCodeIsValid(passCodeDigits: Array<String?>): Boolean {
+        val passcodeString = preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE, loadPinFromOldFormatIfPossible())
+        var isValid = true
+        var i = 0
+        while (i < passCodeDigits.size && isValid) {
+            val originalDigit = passcodeString!![i].toString()
```
abelgardep, line 45: Try to avoid using `!!` here. In case `passcodeString` is `null`, we can return `isValid = false` directly, can't we? Same when `PREFERENCE_PASSCODE` is `null` and `loadPinFromOldFormatIfPossible` returns an empty string

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModelTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    @Before
+    fun setUp() {
+        preferencesProvider = mockk(relaxUnitFun = true)
+        passCodeViewModel = PassCodeViewModel(preferencesProvider)
+    }
+
+    @After
+    override fun tearDown() {
+        super.tearDown()
```
abelgardep, line 49: `tearDown` could be omitted since it is the same as its super

abelgardep (changes requested):
Some questions here @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
JuancaG05: Yes, it's not necessary, but as you said, I added it because of readability, I think it's easier this way

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
JuancaG05: In the current version it is not removed. We could delete it, but anyways the passcode is only read when `PREFERENCE_SET_PASSCODE` is true, and when it is put to true again (the moment when you register a new passcode) `PREFERENCE_PASSCODE` is overwritten, so there shouldn't be problem (the app has been working like this until now). Maybe removing it or setting it to null is safer, what do you think?

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
JuancaG05: Ok, I added a pair of tests to check these situations

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
abelgardep: I think that removing it would be the best option 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    fun removePassCode() {
+        preferencesProvider.removePreference(PassCodeActivity.PREFERENCE_PASSCODE)
+        preferencesProvider.putBoolean(PassCodeActivity.PREFERENCE_SET_PASSCODE, false)
+    }
+
+    fun checkPassCodeIsValid(passCodeDigits: Array<String?>): Boolean {
+        val passCodeString = preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE, loadPinFromOldFormatIfPossible()) ?: return false
+        if (passCodeString == "") return false
```
abelgardep, line 43: We can save a return, and that's a good improvement 👍 
```suggestion
        val passCodeString = preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE, loadPinFromOldFormatIfPossible())
        if (passCodeString.isNullOrEmpty()) return false
```

JuancaG05: Didn't know about this function, good one! 💯

abelgardep (approved):
LGTM

jesmrec:
### (1)

1. Enable passcode in `Settings` 
2. Enter first try

Current: two different behaviours in different devices:

- With Pixel2 (Android11): between the first try and the second try the soft keyboard vanishes, so you have to tap in the gap to make it visible again. 
- With Huawei (Android9): same issue
-  With Nexus 6P (Android 7): works fine

is there any difference in terms of code? 

Expected: Soft keyboard does not vanish 
 
`62a706ea`

JuancaG05:
**(1)** should be fixed now! @jesmrec

jesmrec:
Approved on my side

1Star21 (approved):

PR #3368, opened 2021-09-01
shorten length of edit text on longher pin code input

theScrabi:
## Related Issues
App:

Library PR (if needed):

_____

### DIFF
--- a/owncloudApp/src/main/res/layout/passcode_edit_text.xml
+++ b/owncloudApp/src/main/res/layout/passcode_edit_text.xml
@@ -4,4 +4,5 @@
     android:id="@+id/passCodeEditText"
     style="@style/PassCodeStyle"
     android:cursorVisible="true"
+    android:layout_width="32dp"
     android:imeOptions="flagNoExtractUi"/>


owncloudApp/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.kt
```diff
@@ -117,10 +117,18 @@ class PassCodeActivity : BaseActivity() {
         setTextListeners()
     }
 
+    fun dpToPx(dp: Int): Int {
+        return (dp * resources.displayMetrics.density).toInt()
```
abelgardep, line 121: I think that we could just set a fixed width of 32dp in the `passcode_edit_text.xml` instead of doing it programmatically. I think that it looks good when we have 4 or 6 digits

abelgardep (approved):
LGTM

abelgardep:
I have tested in some devices with different screen sizes: Nexus S, Nexus 5X, Pixel 3XL, and Nexus 10.

It looks good in every one of them.

Issue #3358, opened 2021-08-20
[FEATURE REQUEST] Passcode digits brandable


jesmrec:
- add int to `setup.xml` to switch between different number of digits. Branding will allow to choose 4 or 6 ftm.
- make 4 digits default and minimum
- Migration 4 -> 6 (iOS PR: https://github.com/owncloud/ios-app/pull/1000)

### TASKS

- [x] Create branch feature/brandable_passcode
- [x] Development tasks
  - [x] Implement 4 to 6 digits migration
  - [x] Implement (ViewModel) tests (may require to mock resources)
- [ ] Trigger ownBrander devs to adopt the new flag
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/brandable_passcode into master

### PR

- App #3369 

### QA

Test plan:

Reports

- [X] (1) Minimum is not 4 https://github.com/owncloud/android/issues/3358#issuecomment-924643645 [FIXED]
- [X] (2) (enhancement) Title in migration https://github.com/owncloud/android/issues/3358#issuecomment-924707826 [DONE]
- [X] (3) New passcode delayed to biometrical dismisshttps://github.com/owncloud/android/issues/3358#issuecomment-924751859 [DONE]
- [x] (4) New passcode view is cancellable https://github.com/owncloud/android/issues/3358#issuecomment-927790900 [DONE]
- [x] (5) Killing app in migration deletes the passcode https://github.com/owncloud/android/issues/3358#issuecomment-927889847 [DONE]

jesmrec:
### (1)

4 digits is not the minimum

Current:

By setting `<integer name="passcode_digits">2</integer>`,  passcode is asked with only 2 digits. Also, if `0` or negative value is set, app crashes

Expected:

if the value is >= 4, that value applies
If the value is <4, 4 digits are asked

Pixel2 Android11
`54e070a4`

jesmrec:
### (2)

1. Enable passcode with 4 digits
2. In `setup.xml` set `passcode_digits` = 8
3. Close the app and open again
4. Enter 4-digit passcode to unlock

Current: 8-digit passcode is asked, but the title message is `Enter a new 6-digit passcode`

Expected: 8-digit passcode is asked, and the title message is `Enter a new 8-digit passcode`. Number of digits should be correctly shown in the title message depending on the setup value.

Pixel2 Android11
`54e070a4`

jesmrec:
### (3) (to discuss)

[...]

PR #3369, opened 2021-09-02
[FEATURE REQUEST] Passcode digits brandable

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3358
_____

### DIFF
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
@@ -46,6 +46,8 @@ import org.junit.Test
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import com.owncloud.android.presentation.viewmodels.security.PassCodeViewModel
+import com.owncloud.android.testutil.security.EXAMPLE_PASSCODE_4_DIGITS
+import com.owncloud.android.testutil.security.EXAMPLE_PASSCODE_6_DIGITS
 import io.mockk.every
 import io.mockk.mockk
 import nthChildOf
@@ -68,7 +70,6 @@ class PassCodeActivityTest {
 
     private val defaultPassCode = arrayOf('1', '1', '1', '1', '1', '1')
     private val wrongPassCode = arrayOf('1', '1', '1', '2', '2', '2')
-    private val passCodeToSave = "111111"
 
     private lateinit var passCodeViewModel: PassCodeViewModel
 
@@ -89,7 +90,7 @@ class PassCodeActivityTest {
             )
         }
 
-        every { passCodeViewModel.getPassCode() } returns "1111"
+        every { passCodeViewModel.getPassCode() } returns EXAMPLE_PASSCODE_4_DIGITS
         every { passCodeViewModel.getNumberOfPassCodeDigits() } returns 4
     }
 
@@ -228,7 +229,7 @@ class PassCodeActivityTest {
         every { passCodeViewModel.checkPassCodeIsValid(any()) } returns true
 
         //Save a passcode in Preferences
-        storePasscode(passCodeToSave)
+        storePasscode(EXAMPLE_PASSCODE_6_DIGITS)
 
         //Open Activity in passcode deletion mode
         openPasscodeActivity(PassCodeActivity.ACTION_CHECK_WITH_RESULT)
@@ -244,7 +245,7 @@ class PassCodeActivityTest {
         every { passCodeViewModel.checkPassCodeIsValid(any()) } returns false
 
         //Save a passcode in Preferences
-        storePasscode(passCodeToSave)
+        storePasscode(EXAMPLE_PASSCODE_6_DIGITS)
 
         //Open Activity in passcode deletion mode
         openPasscodeActivity(PassCodeActivity.ACTION_CHECK_WITH_RESULT)
@@ -272,7 +273,7 @@ class PassCodeActivityTest {
             onView(nthChildOf(withId(R.id.passCodeTxtLayout), i)).perform(replaceText(digits[i].toString()))
     }
 
-    private fun storePasscode(passcode: String = passCodeToSave) {
+    private fun storePasscode(passcode: String = EXAMPLE_PASSCODE_6_DIGITS) {
         val appPrefs = PreferenceManager.getDefaultSharedPreferences(context).edit()
 
         appPrefs.putString(PassCodeActivity.PREFERENCE_PASSCODE, passcode.substring(0, passCodeViewModel.getNumberOfPassCodeDigits()))
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
@@ -56,7 +56,7 @@ import org.koin.core.context.stopKoin
 import org.koin.dsl.module
 import withChildViewCount
 
-class OCSettingsPasscodeTest {
+class PassCodeActivityTest {
 
     @Rule
     @JvmField
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
@@ -46,8 +46,8 @@ import org.junit.Test
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import com.owncloud.android.presentation.viewmodels.security.PassCodeViewModel
-import com.owncloud.android.testutil.security.EXAMPLE_PASSCODE_4_DIGITS
-import com.owncloud.android.testutil.security.EXAMPLE_PASSCODE_6_DIGITS
+import com.owncloud.android.testutil.security.OC_PASSCODE_4_DIGITS
+import com.owncloud.android.testutil.security.OC_PASSCODE_6_DIGITS
 import io.mockk.every
 import io.mockk.mockk
 import nthChildOf
@@ -90,7 +90,7 @@ class PassCodeActivityTest {
             )
         }
 
-        every { passCodeViewModel.getPassCode() } returns EXAMPLE_PASSCODE_4_DIGITS
+        every { passCodeViewModel.getPassCode() } returns OC_PASSCODE_4_DIGITS
         every { passCodeViewModel.getNumberOfPassCodeDigits() } returns 4
     }
 
@@ -229,7 +229,7 @@ class PassCodeActivityTest {
         every { passCodeViewModel.checkPassCodeIsValid(any()) } returns true
 
         //Save a passcode in Preferences
-        storePasscode(EXAMPLE_PASSCODE_6_DIGITS)
+        storePasscode(OC_PASSCODE_6_DIGITS)
 
         //Open Activity in passcode deletion mode
         openPasscodeActivity(PassCodeActivity.ACTION_CHECK_WITH_RESULT)
@@ -245,7 +245,7 @@ class PassCodeActivityTest {
         every { passCodeViewModel.checkPassCodeIsValid(any()) } returns false
 
         //Save a passcode in Preferences
-        storePasscode(EXAMPLE_PASSCODE_6_DIGITS)
+        storePasscode(OC_PASSCODE_6_DIGITS)
 
         //Open Activity in passcode deletion mode
         openPasscodeActivity(PassCodeActivity.ACTION_CHECK_WITH_RESULT)
@@ -273,7 +273,7 @@ class PassCodeActivityTest {
             onView(nthChildOf(withId(R.id.passCodeTxtLayout), i)).perform(replaceText(digits[i].toString()))
     }
 
-    private fun storePasscode(passcode: String = EXAMPLE_PASSCODE_6_DIGITS) {
+    private fun storePasscode(passcode: String = OC_PASSCODE_6_DIGITS) {
         val appPrefs = PreferenceManager.getDefaultSharedPreferences(context).edit()
 
         appPrefs.putString(PassCodeActivity.PREFERENCE_PASSCODE, passcode.substring(0, passCodeViewModel.getNumberOfPassCodeDigits()))
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
@@ -51,6 +51,7 @@ import com.owncloud.android.testutil.security.OC_PASSCODE_6_DIGITS
 import io.mockk.every
 import io.mockk.mockk
 import nthChildOf
+import org.junit.Assert.assertEquals
 import org.junit.Before
 import org.koin.androidx.viewmodel.dsl.viewModel
 import org.koin.core.context.startKoin
@@ -122,6 +123,16 @@ class PassCodeActivityTest {
             isDisplayed(true)
             withText(android.R.string.cancel)
         }
+    }
+
+    @Test
+    fun passcodeViewCancelButton() {
+        //Open Activity in passcode creation mode
+        openPasscodeActivity(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)
+
+        onView(withId(R.id.cancel)).perform(click())
+
+        assertEquals(activityRule.activityResult.resultCode, Activity.RESULT_CANCELED)
     }
 
     @Test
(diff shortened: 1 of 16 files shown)


owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
```diff
@@ -88,6 +88,9 @@ class OCSettingsPasscodeTest {
                 }
             )
         }
+
+        every { passCodeViewModel.getPassCode() } returns "1111"
```
fesave, line 92: Well, I think it would be nice if we put these examples in the corresponding class located in the owncloudTestUtil module.

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
```diff
@@ -73,6 +74,7 @@ class SettingsSecurityFragmentTest {
     private lateinit var biometricManager: BiometricManager
 
     private lateinit var securityViewModel: SettingsSecurityViewModel
+    private lateinit var passCodeViewModel: PassCodeViewModel
     private lateinit var context: Context
 
     private val patternValue = "1234"
```
fesave, line 80: Same here

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
```diff
@@ -95,10 +98,16 @@ class SettingsSecurityFragmentTest {
...
                         securityViewModel
                     }
+                    viewModel {
+                        passCodeViewModel
+                    }
                 }
             )
         }
 
+        every { passCodeViewModel.getPassCode() } returns "1111"
```
fesave, line 108: Same here

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -129,7 +136,9 @@ class PassCodeActivity : BaseActivity() {
 
     private fun inflatePasscodeTxtLine() {
         val passcodeTxtLayout = findViewById<LinearLayout>(R.id.passCodeTxtLayout)
-        for (i in 0 until numberOfPassInputs) {
+        println("PACOOOOO1 " + passCodeViewModel.getPassCode()?.length)
```
fesave, line 139: Is it necessary to print this data? Maybe it would be nice to print it through Timber? Also, if you use:
`"PACOOOOO1 ${passCodeViewModel.getPassCode ()?.length} "` is more Kotlin friendly 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -204,6 +214,13 @@ class PassCodeActivity : BaseActivity() {
                 /// pass code accepted in request, user is allowed to access the app
                 passCodeError.visibility = View.INVISIBLE
                 hideSoftKeyboard()
+                if (passCodeViewModel.getPassCode()!!.length < passCodeViewModel.getNumberOfPassCodeDigits()) {
```
fesave, line 217: If `getPassCode()` is null, the application will crash. Be careful with the operator `!!`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -204,6 +214,13 @@ class PassCodeActivity : BaseActivity() {
...
+                    passCodeViewModel.removePassCode()
+                    val intent = Intent(baseContext, PassCodeActivity::class.java)
```
fesave, line 219: This would be more Kotlin's friendly:
`intent.apply{
action = ACTION_REQUEST_WITH_RESULT
putExtra(EXTRAS_MIGRATION, true)
}`

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModelTest.kt
```diff
@@ -146,4 +166,19 @@ class PassCodeViewModelTest : ViewModelTest() {
         }
     }
 
+    @Test
+    fun `get number of passcode digits - ok`() {
+        val numberDigits = 4
```
fesave, line 171: Same here

fesave (changes requested):

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
JuancaG05: Ok, good idea, thanks! 😃

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
JuancaG05: No, it wasn't necessary, only used with testing purposes but I forgot to remove it, thanks for reporting it and for the tip! 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
JuancaG05: Well, that's because the `getString` method of `SharedPreferencesProvider` returns `String?`, but semantically the idea is that execution only arrives here (`processFullPassCode` with `ACTION_CHECK`) when we have to check the already saved passcode, so it shouldn't return null. What do you think? Do you purpose another solution? Maybe checking `!= null` before the if?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
JuancaG05: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
fesave: Yes, is a good solution.

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModelTest.kt
JuancaG05: Well in this case I don't think it's worth it to move this 4 to the owncloudTestUtil module, but I made some changes in the rest of the tests which used some other examples :)

fesave (approved):
LGTM

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/security/PassCodeUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.testutil.security
+
+const val EXAMPLE_PASSCODE_4_DIGITS = "1111"
+const val EXAMPLE_PASSCODE_6_DIGITS = "111111"
```
abelgardep, line 24: To keep consistency, I would rename those constants to `OC_PASSCODE_X_DIGITS`

BTW, we use the ownCloudTestUtil module for constants or objects that are used in several modules to not duplicate several times the same constant. At the moment, since we only use the passcode or pattern in the ownCloudApp module, we can include this file directly in the ownCloudApp module.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ManageSpaceActivity.java
```diff
@@ -108,10 +107,10 @@ protected Boolean doInBackground(Void... params) {
 
 
             final String passcodeString = appPrefs.getString(PREFERENCE_PASSCODE, null);
-            final String[] passCodeDigits = new String[PassCodeActivity.numberOfPassInputs];
+            final String[] passCodeDigits = new String[getBaseContext().getResources().getInteger(R.integer.passcode_digits)];
```
abelgardep, line 110: You could create a new variable, for example `numberOfPasscodeInputs`, and initialize it to `getBaseContext().getResources().getInteger(R.integer.passcode_digits)` so you don't need to call the `getResources` several times in the same method.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -129,7 +136,7 @@ class PassCodeActivity : BaseActivity() {
 
     private fun inflatePasscodeTxtLine() {
         val passcodeTxtLayout = findViewById<LinearLayout>(R.id.passCodeTxtLayout)
-        for (i in 0 until numberOfPassInputs) {
+        for (i in 0 until (passCodeViewModel.getPassCode()?.length ?: passCodeViewModel.getNumberOfPassCodeDigits())) {
```
abelgardep, line 139: I would go with the approach you use in the  `setTextListeners` method, 

```
        val numberOfPasscodeDigits = (passCodeViewModel.getPassCode()?.length ?: passCodeViewModel.getNumberOfPassCodeDigits())
        for (i in 0 until numberOfPasscodeDigits) {
```

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -51,11 +56,14 @@ class PassCodeViewModel(
...
+    fun getNumberOfPassCodeDigits() = contextProvider.getInt(R.integer.passcode_digits)
 
-        return pinString
+    private fun loadPinFromOldFormatIfPossible(): String? {
+        var pinString = ""
+        for (i in 1..4) {
+            val pinChar = preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE_D + i, null)
+            pinChar?.let { pinString += pinChar }
+        }
+        return if (pinString == "") null else pinString
```
abelgardep, line 67: `isEmpty()` is our friend 👍 
```suggestion
        return if (pinString.isEmpty()) null else pinString
```

abelgardep (changes requested):
Some suggestions

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/security/PassCodeUtils.kt
JuancaG05: Ok, I changed the name. Related to the module, I've seen there are some other constants that are only used in one module and they are defined also in the ownCloudTestUtil module. I think it's more clear to keep all of them in this utils module, what do you think?

abelgardep: Ok, keep them in ownCloudTestUtils

abelgardep (approved):
LGTM

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -51,11 +56,20 @@ class PassCodeViewModel(
         return isValid
     }
 
-    private fun loadPinFromOldFormatIfPossible(): String {
-        var pinString = ""
-        for (i in 1..4)
-            pinString += preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE_D + i, null)
+    fun getNumberOfPassCodeDigits(): Int {
+        val numberOfPassCodeDigits = contextProvider.getInt(R.integer.passcode_digits)
+        return if (numberOfPassCodeDigits >= 4) numberOfPassCodeDigits else 4
```
abelgardep, line 61: Let's create a new constant `PASSCODE_MIN_LENGTH` instead of having a hardcoded 4 👍

abelgardep (changes requested):
Just some tiny things @JuancaG05

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/PassCodeActivityTest.kt
```diff
@@ -112,14 +117,24 @@ class OCSettingsPasscodeTest {
...
     }
 
+    @Test
+    fun passcodeViewCancelButton() {
+        //Open Activity in passcode creation mode
+        openPasscodeActivity(PassCodeActivity.ACTION_REQUEST_WITH_RESULT)
+
+        onView(withId(R.id.cancel)).perform(click())
+
+        assertEquals(activityRule.activityResult.resultCode, Activity.RESULT_CANCELED)
```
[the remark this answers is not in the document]
JuancaG05: Ok, I will keep the `assertEquals` way as I just realized that `assertThat` is deprecated 😁

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -397,12 +422,14 @@ class PassCodeActivity : BaseActivity() {
...
+        const val PREFERENCE_MIGRATION_REQUIRED = "PrefMigrationRequired"
 
         // NOTE: This is required to read the legacy pin code format
         const val PREFERENCE_PASSCODE_D = "PrefPinCode"
 
-        const val numberOfPassInputs = 4
         private const val KEY_PASSCODE_DIGITS = "PASSCODE_DIGITS"
         private const val KEY_CONFIRMING_PASSCODE = "CONFIRMING_PASSCODE"
+
+        const val EXTRAS_MIGRATION = "MIGRATION"
```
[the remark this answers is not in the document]
JuancaG05: 😬 😬 😬

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
```diff
@@ -51,11 +56,20 @@ class PassCodeViewModel(
         return isValid
     }
 
-    private fun loadPinFromOldFormatIfPossible(): String {
-        var pinString = ""
-        for (i in 1..4)
-            pinString += preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE_D + i, null)
+    fun getNumberOfPassCodeDigits(): Int {
+        val numberOfPassCodeDigits = contextProvider.getInt(R.integer.passcode_digits)
+        return if (numberOfPassCodeDigits >= PassCodeActivity.PASSCODE_MIN_LENGTH) numberOfPassCodeDigits else PassCodeActivity.PASSCODE_MIN_LENGTH
```
abelgardep, line 61: We could get rid of some verbosity using `maxOf`, what do you think?
```suggestion
        return maxOf(numberOfPassCodeDigits, PassCodeActivity.PASSCODE_MIN_LENGTH)
```

owncloudApp/src/main/java/com/owncloud/android/ui/activity/BiometricActivity.java
```diff
@@ -125,6 +129,19 @@ public void onAuthenticationError(int errorCode, @NonNull CharSequence errString
             @Override
             public void onAuthenticationSucceeded(@NonNull BiometricPrompt.AuthenticationResult result) {
                 super.onAuthenticationSucceeded(result);
+                SharedPreferencesProvider preferencesProvider = new SharedPreferencesProviderImpl(getApplicationContext());
+                // In this line, null is only provisional until the rearchitecture of BiometricActivity
+                String passCode = preferencesProvider.getString(PassCodeActivity.PREFERENCE_PASSCODE, null);
+                int passCodeDigits = getApplicationContext().getResources().getInteger(R.integer.passcode_digits);
+                if (passCodeDigits < 4) passCodeDigits = 4;
```
abelgardep, line 136: We can use the constant we created: `PASSCODE_MIN_LENGTH`, right?

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/security/PatternUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.testutil.security
+
+const val OC_EXAMPLE_PATTERN = "1234"
```
abelgardep, line 23: `OC_` prefix is enough, we can get rid of `EXAMPLE` one

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/security/PassCodeViewModel.kt
JuancaG05: Great, I didn't know about that one, thanks for the tip!

abelgardep (approved):

jesmrec:
QA reports were done by mistake in [issue](https://github.com/owncloud/android/issues/3358)

Everything passed.

Approved

Issue #3344, opened 2021-08-11
[FEATURE REQUEST] Lock delay for app


jesmrec:
Actually, if the app is protected with passcode/biometrical, it is not posible to access the accounts using doc provider apps:

![Screenshot 2021-08-11 at 09 07 33](https://user-images.githubusercontent.com/14894746/128984909-5f89c61a-db0e-4008-a655-ae403bf7fae8.png)

This is an expected behaviour after the first iteration of document provider implementation. 

The idea is researching/implementing a solution in order not to make users to decide between document provider or passcode/pattern protection.

### PROPOSED FIX

As in the iOS app, a lock delay will be implemented with its respective preference:

![CleanShot 2021-09-07 at 13 46 56@2x](https://user-images.githubusercontent.com/57049315/132375835-407ec893-5bbc-4ee4-a738-dd6c8e5a00b0.png)

As we're implementing here the lock delay but it won't solve the document provider's problem (https://github.com/owncloud/android/issues/3344#issuecomment-918152437), it will be addressed in another issue: https://github.com/owncloud/android/issues/3379

### TASKS

- [x] Research about the best option on how to implement the lock delay
- [x] Create branch feature/lock_delay
- [x] Development tasks
  - [x] Implement lock delay with a default time in app's lock with passcode
  - [x] Implement lock delay with a default time in app's lock with pattern
  - [x] Implement lock delay with a default time in app's lock with biometric
  - [x] Implement preference of lock delay time chooser
  - [x] Update UI tests
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/lock_delay into master

### PR

- App #3375
- Library (if needed)

michaelstingl:
Lock delay a simple approach on Android?

jesmrec:
> Lock delay a simple approach on Android?

yes, it could be an approach (like iOS before).

abelgardep:
Lock delay could be an option. 

IMHO, it could be easier for the user to include a new security option. It would be permanent, if the user wants to disallow, he needs to uncheck it again. 

New checkbox -> "Allow access from DocumentsProvider" Disabled by default

michaelstingl:
> It would be permanent, if the user wants to disallow, he needs to uncheck it again.

Probably not what big organisations want. Users can forget to lock again.

System shows "Lock with pattern". Is there a method to trigger unlock UI there? (no requirement for first implementation)

abelgardep:
[...]

PR #3375, opened 2021-09-10
[FEATURE REQUEST] Lock delay for app

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3344

_____

## QA

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.20/3375-lock_delaymd.md

Reports:

- [X] (1) Default value https://github.com/owncloud/android/pull/3375#issuecomment-964893863 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -155,6 +166,7 @@ class PassCodeActivity : BaseActivity() {
             if (i > 0) {
                 passCodeEditTexts[i]?.setOnKeyListener { v: View, keyCode: Int, _: KeyEvent? ->
                     if (keyCode == KeyEvent.KEYCODE_DEL && bChange) {  // TODO WIP: event should be used to control what's exactly happening with DEL, not any custom field...
+                        passCodeEditTexts[i - 1]?.isEnabled = true
```
fesave, line 169: To avoid repeating the same thing, you can do it like this:
`passCodeEditTexts[i - 1]?.apply{
.isEnabled = true
.setText("")
.requestFocus()
}`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -295,6 +290,7 @@ class PassCodeActivity : BaseActivity() {
      */
     protected fun clearBoxes() {
         for (passCodeEditText in passCodeEditTexts) {
+            passCodeEditText?.isEnabled = true
```
fesave, line 293: To avoid repeating the same thing, you can do it like this:
`passCodeEditText?.apply{
.isEnabled=true .setText()}`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -327,32 +323,29 @@ class PassCodeActivity : BaseActivity() {
...
+     *
+     * @param index         Position in the pass code of the input field that will be bound to
+     * this watcher.
+     * @param lastOne       'True' means that watcher corresponds to the last position of the
+     * pass code.
+     */
+    private inner class PassCodeDigitTextWatcher(private val index: Int, private val lastOne: Boolean) : TextWatcher {
         private operator fun next(): Int {
-            return if (mLastOne) 0 else mIndex + 1
+            return if (lastOne) 0 else index + 1
```
fesave, line 348: More Kotlin friendly: `return if (lastOne) 0 else index.plus(1)`

fesave (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
JuancaG05: Thanks 😃

fesave (approved):
LGTM

owncloudApp/src/main/java/com/owncloud/android/authentication/BiometricManager.java
```diff
@@ -145,6 +136,11 @@ public boolean hasEnrolledBiometric() {
      * USE WITH CARE
      */
     public void bayPassUnlockOnce() {
-        setUnlockTimestamp();
+        int timeout = LockTimeout.Companion.fromStringToMilliseconds(preferencesProvider.getString(LOCK_TIMEOUT, LockTimeout.IMMEDIATELY.name()));
+        long lastUnlockTimestamp = preferencesProvider.getLong(LAST_UNLOCK_TIMESTAMP, 0);
+        if (System.currentTimeMillis() - lastUnlockTimestamp > timeout) {
```
abelgardep, line 141: I have seen those 3 lines somewhere else 🤔  Could we use a common function to check this? Maybe with a descriptive name 👍

owncloudApp/src/main/java/com/owncloud/android/authentication/PatternManager.java
```diff
@@ -99,28 +109,23 @@ private void checkPattern(Activity activity) {
...
      * again.
-     * <p>
+     *
      * USE WITH CARE
      */
     public void bayPassUnlockOnce() {
-        setUnlockTimestamp();
+        int timeout = LockTimeout.Companion.fromStringToMilliseconds(preferencesProvider.getString(LOCK_TIMEOUT, LockTimeout.IMMEDIATELY.name()));
+        long lastUnlockTimestamp = preferencesProvider.getLong(LAST_UNLOCK_TIMESTAMP, 0);
+        if (System.currentTimeMillis() - lastUnlockTimestamp > timeout) {
```
abelgardep, line 126: I'm pretty sure I have seen this several times 😄  Check the previous review comment 😉

owncloudApp/src/main/java/com/owncloud/android/authentication/PatternManager.java
```diff
@@ -99,28 +109,23 @@ private void checkPattern(Activity activity) {
...
-        }
-        return false;
-    }
-
-    public boolean isPatternEnabled() {
-        SharedPreferences appPrefs = PreferenceManager.getDefaultSharedPreferences(MainApp.Companion.getAppContext());
-        return appPrefs.getBoolean(PatternLockActivity.PREFERENCE_SET_PATTERN, false);
+    public void onBiometricCancelled(Activity activity) {
+        // Ask user for pattern
+        checkPattern(activity);
```
abelgardep, line 114: Not very important and I know that it was there, but sometimes this kind of commentary can be removed by using a descriptive function name f.e. `askUserForPattern(activity)`

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/SecurityUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+enum class LockTimeout {
+    IMMEDIATELY, ONE_MINUTE, FIVE_MINUTES, THIRTY_MINUTES;
+
+    companion object {
+        fun fromStringToMilliseconds(string: String?): Int {
+            return when {
+                string.equals("IMMEDIATELY", ignoreCase = true) -> 0
+                string.equals("ONE_MINUTE", ignoreCase = true) -> 60000
+                string.equals("FIVE_MINUTES", ignoreCase = true) -> 300000
+                else -> 1800000
```
abelgardep, line 35: Several things here:
1. Let's use [`_` to write big numbers:](https://kotlinlang.org/docs/basic-types.html#literal-constants) 300_000, 60_000 etc. It's easier to read 👍  
2. In case the `string` is null or a not expected value, the app will be unprotected for 30 minutes, right? Should this be the expected behavior?
3. This function takes care of checking the LockTimeout, and then transform it to millis. The first part, is basically the enum function `valueOf()` right? [Let's check the docs about enums in kotlin](https://kotlinlang.org/docs/enum-classes.html#working-with-enum-constants). Maybe we could write this in a different and better way 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
```diff
@@ -70,8 +74,9 @@ class SettingsSecurityFragment : PreferenceFragmentCompat() {
             else {
                 prefPasscode?.isChecked = false
 
-                // Do not allow to use biometric lock since Passcode lock has been disabled
+                // Do not allow to use biometric lock nor lock delay since Passcode lock has been disabled
                 disableBiometric()
+                prefLockApplication?.isEnabled = false
```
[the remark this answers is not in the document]
JuancaG05: Yeah I know I didn't use the same approach, but I tried 😄 . The thing is that I cannot include the line of disabling the "Lock application" preference inside `disableBiometric`, I need to disable it also when the device has no biometrics hardware, and this is the only way I can do it. I don't think moving it to a function is worth it, as I think the line is self-defining.

However, I DO can include the enabling of the preference in the `enableBiometric` function. That's life! 😆

owncloudApp/src/main/java/com/owncloud/android/authentication/BiometricManager.java
JuancaG05: Well in fact not only those 3 lines, the whole function 😄 . So, I don't have to think about a new descriptive name, `bayPassUnlockOnce` has been moved to `SecurityUtils.kt` 😛

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/SecurityUtils.kt
JuancaG05: Ok, I changed the logic following your advice, you tell me 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -516,11 +513,7 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
      */
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-            BiometricManager.getBiometricManager(this).bayPassUnlockOnce()
-        }
-        PassCodeManager.bayPassUnlockOnce()
-        PatternManager.getPatternManager().bayPassUnlockOnce()
+        bayPassUnlockOnce()
```
JuancaG05, line 516: This is a change I didn't plan but with the change of the logic of `bayPassUnlockOnce` and knowing now it's the same function, I think we can save all these lines! I tested it and it seems to work 😄

JuancaG05:
@abelgardep changes performed, ready to re-review!

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/SecurityUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+
+enum class LockTimeout {
+    IMMEDIATELY, ONE_MINUTE, FIVE_MINUTES, THIRTY_MINUTES;
+
+    fun toMilliseconds(): Int {
+        return when (this) {
+            ONE_MINUTE -> 60_000
+            FIVE_MINUTES -> 300_000
+            THIRTY_MINUTES -> 1_800_000
+            else -> 0
```
abelgardep, line 37: Instead of `else`, we could use `IMMEDIATELY` and this `when` would be exhaustive too.

abelgardep (approved):

owncloudApp/src/main/java/com/owncloud/android/ui/activity/BiometricActivity.java
```diff
@@ -125,6 +129,8 @@ public void onAuthenticationError(int errorCode, @NonNull CharSequence errString
             @Override
             public void onAuthenticationSucceeded(@NonNull BiometricPrompt.AuthenticationResult result) {
                 super.onAuthenticationSucceeded(result);
+                SharedPreferencesProvider preferencesProvider = new SharedPreferencesProviderImpl(getApplicationContext());
```
abelgardep, line 132: getContext should be enough, anyway, as we are going to rearchitect this biometric stuff, we can move it forward.
```suggestion
                SharedPreferencesProvider preferencesProvider = new SharedPreferencesProviderImpl(getContext());
```

jesmrec:
### (1)

A question: why is the default value `Not Set` when the behaviour is the same as `Immediately`? Does it make sense to set as default `Immediately`?

jesmrec:
Approved on my side

PR #3269, opened 2021-05-28
Scoped Storage

abelgardep:
## Related Issues
App: #2877 #2920 

Library PR (if needed):

- [ ] Added changelog files for the fixed issues in folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.19/Scoped%20Storage.md

Reports:

- [X] (1) Logs https://github.com/owncloud/android/pull/3269#issuecomment-888461682  [FIXED]

PR #3378, opened 2021-09-13
Api30

hannesa2:
Update used build api

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,9 @@
 buildscript {
     ext {
         // SDK
-        sdkCompileVersion = 29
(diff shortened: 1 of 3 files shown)


abelgardep:
We will upgrade the compileSdkVersion using the scoped storage PR: https://github.com/owncloud/android/pull/3269 since it is mandatory if we target 30. The lint fixes were applied there too.

Thanks for the contribution anyway @hannesa2

hannesa2:
I've also something where it's mandatory.
Why not splitting it and avoid huge pull requests ?

Issue #3340, opened 2021-08-06
[BUG] Account removed from "Settings" is not removed from the drawer


jesmrec:
### Steps to reproduce

1.  Add two accounts to the oC app
2. Go to device Settings -> Accounts
3. Remove one of the accounts
4. Open the app

### Actual behaviour
The drawer shows both accounts, whereas the `Manage accounts` menu shows correctly the remaining one. By closing the app and open again, removed account is finally gone.

### Expected behaviour
The drawer shows only one account, same as `Manage accounts` menu

jesmrec:
Related to #3336??? @theScrabi

jesmrec:
Fixed! Ready to 2.19

PR #3381, opened 2021-09-14
[Fix] Account removed from "Settings" is not removed from the drawer

fesave:
## Related Issues
App: #3340 

_____

### DIFF
--- a/changelog/unreleased/3381
+++ b/changelog/unreleased/3381
@@ -1,0 +1,7 @@
+Bugfix: Account removed is not removed from the drawer.
+
+When an account was deleted from the phone settings, in the accounts section, it was not deleted from the Navigation Drawer,
+now when deleting an account from there, the Navigation Drawer refreshes and it is deleted correctly.
+
+https://github.com/owncloud/android/issues/3340
+https://github.com/owncloud/android/pull/3381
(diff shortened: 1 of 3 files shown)


changelog/unreleased/3381
```diff
@@ -0,0 +1,7 @@
+Bugfix: Account removed is not removed from the drawer
+
+When an account was deleted from the phone settings, in the accounts section, it was not deleted from the Navigation Drawer,
+now when deleting an account from there, the Navigation Drawer refreshes and it is deleted correctly.
```
JuancaG05, line 4: Just a tiny change I suggest, but up to you 😉 
P.S.: I like the passive voice you used 👍 
```suggestion
When an account was deleted from the app settings, in the accounts section, it was not removed from the Navigation Drawer.
Now, when deleting an account from there, the Navigation Drawer is refreshed and the removed account is no more shown.
```

JuancaG05 (changes requested):
Just a suggestion, but it's up to you 👍

changelog/unreleased/3381
fesave: Oops, you're wrong my friend 😄 , the problem appears when the user deletes his account from the phone settings (not from the app settings)

Thanks for the other suggestions, no problem to change them.

JuancaG05: 😱  My fault!! Sorry, thanks for the information

JuancaG05 (approved):

jesmrec:
Fixed.

Issue #3315, opened 2021-07-19
[FEATURE REQUEST] Trigger camera uploads with different events


abelgardep:
**Describe the solution you'd like**
Trigger camera uploads only when the device is charging. A new check should be added to the camera uploads view, so I can select if camera uploads should be triggered when the device is charging. Pretty similar to the current option: `Upload only when connected to Wifi`. We should take into account that when multiple constraints are specified, the work will run only when all the constraints are met.

**Additional context**
Camera uploads are enqueued via WorkManager, and we can select different constraints. At the moment we only use network type to select that uploads should be triggered when connected to wifi, but we can add any of them:

Constraint | Description
-- | --
NetworkType | Constrains the type of network required for your work to run. For example, Wi-Fi (UNMETERED).
BatteryNotLow | When set to true, your work will not run if the device is in low battery mode.
RequiresCharging | When set to true, your work will only run when the device is charging.
DeviceIdle | When set to true, this requires the user’s device to be idle before the work will run. This can be useful for running batched operations that might otherwise have a negative performance impact on other apps running actively on the user’s device.
StorageNotLow | When set to true, your work will not run if the user’s storage space on the device is too low.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

abelgardep:
This is a duplicate of https://github.com/owncloud/android/issues/465 but I added some context here, and it should not be so complex to add thanks to workmanager.

PR #3385, opened 2021-09-16
[FEATURE] Instant upload only when charging

fesave:
## Related Issues
App: #465 #3315 

_____


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsPictureUploadsFragment.kt
```diff
@@ -117,6 +120,10 @@ class SettingsPictureUploadsFragment : PreferenceFragmentCompat() {
                 prefPictureUploadsPath?.summary = DisplayUtils.getPathWithoutLastSlash(it.uploadPath)
                 prefPictureUploadsSourcePath?.summary = DisplayUtils.getPathWithoutLastSlash(it.sourcePath.toUri().path)
                 prefPictureUploadsOnWifi?.isChecked = it.wifiOnly
+                prefPictureUploadsOnCharging?.apply {
+                    isVisible = it.wifiOnly
```
JuancaG05, line 124: When I implemented the auto upload pictures and videos sections, we had a quick meeting and we decided to keep items not **enabled** instead of not **visible** when they cannot be activated yet, so that user perceives the app's value and all its possibilities since the beginning and not by discovering them

JuancaG05, line 124: When I implemented the auto upload pictures and videos sections, we had a quick meeting and we decided to keep items not **enabled** instead of not **visible** when they cannot be activated yet, so that user perceives the app's value and all its possibilities since the beginning and not by discovering them when they activate other options.

On the other hand, it might be a bit weird in this case, as everything is disabled when the switch button is off, and then when turning it on, this preference would continue being disabled.

What do you think about this? Should we keep the preference not visible or not enabled?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsVideoUploadsFragment.kt
```diff
@@ -117,6 +120,10 @@ class SettingsVideoUploadsFragment : PreferenceFragmentCompat() {
                 prefVideoUploadsPath?.summary = DisplayUtils.getPathWithoutLastSlash(it.uploadPath)
                 prefVideoUploadsSourcePath?.summary = DisplayUtils.getPathWithoutLastSlash(it.sourcePath.toUri().path)
                 prefVideoUploadsOnWifi?.isChecked = it.wifiOnly
+                prefVideoUploadsOnCharging?.apply {
+                    isVisible = it.wifiOnly
```
JuancaG05, line 124: Same as before. Whatever you decide in the other case, apply it here too 😉

owncloudApp/src/main/res/xml/settings_video_uploads.xml
```diff
@@ -45,7 +52,6 @@
         app:defaultValue="NOTHING"
         app:dialogTitle="@string/prefs_camera_upload_behaviour_dialog_title"
         app:iconSpaceReserved="false"
-        app:negativeButtonText=""
```
JuancaG05, line 48: This is necessary! Otherwise, an ugly "Cancel" button will appear in the AlertDialog. I think it looks better without it, like in Camera uploads screen 😄

JuancaG05 (changes requested):
Some comments @fesave 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsPictureUploadsFragment.kt
```diff
@@ -121,7 +121,9 @@ class SettingsPictureUploadsFragment : PreferenceFragmentCompat() {
                 prefPictureUploadsSourcePath?.summary = DisplayUtils.getPathWithoutLastSlash(it.sourcePath.toUri().path)
                 prefPictureUploadsOnWifi?.isChecked = it.wifiOnly
                 prefPictureUploadsOnCharging?.apply {
-                    isVisible = it.wifiOnly
+                    isEnabled = it.wifiOnly
+                    summary =
+                        if (!it.wifiOnly) getString(R.string.prefs_camera_upload_on_charging_summary) else getString(R.string.prefs_camera_upload_on_charging_summary_disable)
```
JuancaG05, line 126: I wouldn't show a summary when the preference is enabled, as the user knows that it's been enabled only when activating the Wi-Fi preference. Similarly to "Biometric lock", that way we keep consistency in settings 😃

JuancaG05 (changes requested):
Just one more detail @fesave 😄

JuancaG05 (approved):
LGTM 💯

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/settings/SettingsPictureUploadsViewModel.kt
```diff
@@ -84,10 +84,26 @@ class SettingsPictureUploadsViewModel(
...
+            savePictureUploadsConfigurationUseCase.execute(
+                if (!wifiOnly) {
+                    SavePictureUploadsConfigurationUseCase.Params(
+                        composePictureUploadsConfiguration(wifiOnly = wifiOnly, chargingOnly = false)
+                    )
+                } else {
+                    SavePictureUploadsConfigurationUseCase.Params(
+                        composePictureUploadsConfiguration(wifiOnly = wifiOnly)
+                    )
+                }
```
abelgardep, line 97:
```suggestion
                    SavePictureUploadsConfigurationUseCase.Params(
                        composePictureUploadsConfiguration(wifiOnly = wifiOnly, chargingOnly = false)
                    )
```

If we disable wifiOnly, chargingOnly should be disabled too. That's ok, but if we enable it, it means that chargingOnly was already disabled, so, we don't need this `if else`, right?

Anyway, I would try to use 

`if (condition) { } else { }` instead of `if(!condition) { } else { }`

owncloudApp/src/main/java/com/owncloud/android/usecases/UploadFileFromContentUriUseCase.kt
```diff
@@ -66,6 +67,7 @@ class UploadFileFromContentUriUseCase(
         val behavior: String,
         val uploadPath: String,
         val uploadIdInStorageManager: Long,
-        val wifiOnly: Boolean
+        val wifiOnly: Boolean,
+        val chargingOnly: Boolean
```
abelgardep, line 71: Let's add [trailing commas ](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas)when possible to keep commits a little bit cleaner 👍

owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_35.kt
```diff
@@ -0,0 +1,44 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *
+ *
+ *
```
abelgardep, line 21: Too many `*` here

owncloudData/src/main/java/com/owncloud/android/data/migrations/Migration_35.kt
```diff
@@ -0,0 +1,44 @@
...
+        database.run {
+            execSQL(
+                "CREATE TABLE IF NOT EXISTS `${FOLDER_BACKUP_TABLE_NAME}2` (`accountName` TEXT NOT NULL, `behavior` TEXT NOT NULL, `sourcePath` TEXT NOT NULL, `uploadPath` TEXT NOT NULL, `wifiOnly` INTEGER NOT NULL, `chargingOnly` INTEGER NOT NULL, `name` TEXT NOT NULL, `lastSyncTimestamp` INTEGER NOT NULL, `id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL)"
+            )
+            execSQL("ALTER TABLE $FOLDER_BACKUP_TABLE_NAME ADD COLUMN chargingOnly INTEGER NOT NULL DEFAULT '0'")
+            execSQL(
+                "INSERT INTO `${FOLDER_BACKUP_TABLE_NAME}2` SELECT accountName, behavior, sourcePath, uploadPath, wifiOnly, IFNULL(chargingOnly, '0'), name, lastSyncTimeStamp, id  FROM $FOLDER_BACKUP_TABLE_NAME"
+            )
+            execSQL("DROP TABLE $FOLDER_BACKUP_TABLE_NAME")
+            execSQL("ALTER TABLE ${FOLDER_BACKUP_TABLE_NAME}2 RENAME TO $FOLDER_BACKUP_TABLE_NAME")
```
abelgardep, line 41: Let's add a test to ensure that migration is done properly 👍

abelgardep (approved):
LGTM

jesmrec:
### (1) (improvement)

I noticed that you included a comment below the new option when it is disabled, letting users know how it works. That's right

AFAIK, uploads will be only triggered when both "only wifi" and "charging" are ticked in. As we commented in the issue, this fact could be confusing for users who expect an OR behaviour. 

A comment like `Uploads will be done only when device is under wifi AND charging` could be useful to understand and avoiding confusions. I'd make it visible when both options are enabled

what do you think?

abelgardep:
Related to [**(1)**](https://github.com/owncloud/android/pull/3385#issuecomment-925711738) We could use a new category to make it more understandable. The wording could change if needed. 

CC @fesave

jesmrec:
Approved on my side. Clean feature. Great job!

Rajakashi420:
> ## Related Issues
> App: #465 #3315
> 
> * [x]  Added changelog files for the fixed issues in folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
> 
> ## QA
> Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.19/Upload%20when%20charging.md
> 
> Reports:
> 
> * [x]  (1) Explanation message when both options are enabled [[FEATURE] Instant upload only when charging #3385 (comment)](https://github.com/owncloud/android/pull/3385#issuecomment-925711738) [DONE]

Issue #3373, opened 2021-09-03
[New arch] Pattern activity


JuancaG05:
Pattern activity is written in Java in the old architecture. The aim of this issue is to move the pattern activity and all its related stuff to `presentation.ui.security` package and rewrite the class in Kotlin.

### TASKS

- [x] Create branch new_arch/pattern
- [x] Development tasks
  - [x] Move all the classes to `presentation.ui.security` package
  - [x] Kotlinize PatternActivity
  - [x] Create new PatternViewModel
  - [x] Move logic from PatternActivity to PatternViewModel
  - [x] Save and remove pattern in PatternActivity instead of SettingsSecurityFragment
  - [x] Kotlinize PatternManager
  - [x] Include unit tests
  - [x] Include UI tests
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch new_arch/pattern into master

PR #3389, opened 2021-09-21
[New arch] Pattern activity

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3373
_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.20/3389-pattern_lock_architecture.md

Reports:


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PatternActivity.kt
```diff
@@ -0,0 +1,322 @@
...
+        super.onSaveInstanceState(outState)
+        outState.apply {
+            putBoolean(KEY_CONFIRMING_PATTERN, confirmingPattern)
+            putString(KEY_PATTERN_STRING, patternValue)
+            putString(PATTERN_HEADER_VIEW_TEXT, patternHeader.text.toString())
+            if (patternExplanation.visibility == View.VISIBLE) {
+                putBoolean(PATTERN_EXP_VIEW_STATE, true)
+            } else {
+                putBoolean(PATTERN_EXP_VIEW_STATE, false)
+            }
```
abelgardep, line 286: You can use `isVisible` in those cases to check that the `visibility == VISIBLE`
```suggestion
            putBoolean(PATTERN_EXP_VIEW_STATE, patternExplanation.isVisible)
```

What do you think? Maybe you can use this on other places 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PatternActivity.kt
```diff
@@ -0,0 +1,322 @@
...
+            //This block is executed when the user is setting the pattern lock (i.e enabling the pattern lock)
+            if (!confirmingPattern) {
+                patternError.visibility = View.INVISIBLE
+                requestPatternConfirmation()
+            } else if (confirmPattern()) {
+                savePatternAndExit()
+            } else {
+                showErrorAndRestart(
+                    R.string.pattern_not_same_pattern,
+                    R.string.pattern_enter_pattern, View.VISIBLE
```
abelgardep, line 236: When we have 3 or more parameters, I would recommend using named params. It will make the code easier to read. For example, at the moment I don't know what `showErrorAndRestart` does, but if it had named params, I could imagine 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/BiometricActivity.kt
```diff
@@ -0,0 +1,128 @@
...
+import timber.log.Timber
+import java.util.concurrent.Executor
+
+class BiometricActivity : AppCompatActivity() {
+
+    // ViewModel
+    private val biometricViewModel by viewModel<BiometricViewModel>()
+
+    private lateinit var cryptoObject: BiometricPrompt.CryptoObject
+    private lateinit var activity: BiometricActivity
```
abelgardep, line 45: 😮  I'm sure that you can remove this 😉

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -692,7 +693,7 @@ void loadStateFrom(Bundle savedInstanceState) {
     }
 
     private void clearLocalSearchView() {
-        ((FileActivity) requireActivity()).hideSoftKeyboard();
+        FragmentExtKt.hideSoftKeyboard(this);
```
abelgardep, line 696: You could use the ActivityExtension here and save that FragmentExtension if it is only used here.

```suggestion
        ActivityExtKt.hideSoftKeyboard(requireActivity());
```

abelgardep (changes requested):
Fantastic job @JuancaG05 ! 👍 

I added some suggestions here

jesmrec:
Approved.

PR #3393, opened 2021-09-28
Remove old way to retrieve the default camera folder path

abelgardep:
Clean some old code related to camera uploads.

This code was used previously to select the default camera folder but it is not working anymore after moving to the native way.
_____


fesave (approved):
LGTM

jesmrec:
Move this forward. This is just a cleaning up that could cause regressions, to be tested in the final regression test. No worthy to execute a sanity over camera uploads (every test will take ~15 mins) to be repeated in the short term.

Thanks!

Issue #3327, opened 2021-07-23
[FEATURE REQUEST] Scoped Storage Migration from legacy storage


jesmrec:
**Is your feature request related to a problem? Please describe.**

What to do when users will upgrade to Scoped Storage. 

**Describe the solution you'd like**

A Wizard will guide them through the process.
Finally, they will decide if they want to remove the old folder by themselves, or the app will do it.
Pending uploads to be migrated

**Describe alternatives you've considered**

**Additional context**

- When user upgrades to 2.19, stuff to migrate:
  - Pending uploads migration: Sharing with oC, manual uploads, Document provider (over temporal folder).
  - Check one for one pending uploads, whether they are on external storage. 
  - Move files to ScSt and queue them (are they directly uploaded?) 

Create Wizard with some slides. NO SKIP OPTION
   - What ScSt is, quick description for users to understand what is it about
   - Advantages of Scoped Storage
   - Warning: all av offline stuff will be directly downloaded in some seconds. Not sure if tell user this... he/she can do nothing.
   - Migration

Check when and where, the options will be displayed to let user decide.

Ask user whether legacy storage is deleted or not. Be clear with the consequences of every option
   - If "yes" -> owncloud legacy removed from device
   - If "not"  -> owncloud legacy folder remains. Let him know what happens there...

### TASKS

- [x] Research (if needed)
- [x] Create branch scoped_storage_migration
- [ ] Development tasks
  - [x] Implement the designs for the migration
  - [x] Create logic to show or not the migration wizard
  - [x] Implement the viewmodel logic
  - [x] Implement the navigation inside the wizard
  - [x] Show how many files and Gb are going to be migrated and a time estimation
  - [x] Migrate files from legacy storage to scoped storage
  - [ ] Update database with the new path for already downloaded files
- [ ] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch scoped_storage_migration into master

### PR

- App
- Library (if needed)

jesmrec:
Suggestion for the slides:

1. A new way for storing your files in Android
2. All your files locally stored will be locked up and no longer available from outside the app.
3. They will be only accessed by the Android system, no by other apps.
4. You will be asked what to do with the legacy storage
5. You can choose: remove automatically (we will take care), or do it by yourself
6. This new feature will add more security to your files locally stored.

jesmrec:
QA checks

[...]

PR #3396, opened 2021-09-29
Scoped storage migration

abelgardep:
## Related Issues
App: #3327 

Library PR (if needed): https://github.com/owncloud/android-library/pull/436

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.19/Scoped%20Storage%20Migration.md

Reports:

- [x] (1) Learn more link in 1st slide https://github.com/owncloud/android/pull/3396#issuecomment-934191587 [WONT FIX FTM]
- [X] (2) Remind me later missing (conclusion) https://github.com/owncloud/android/pull/3396#issuecomment-934201178 [WONT FIX]
- [X] (3) Power off device + migration https://github.com/owncloud/android/pull/3396#issuecomment-934341047 [not always reproducible WONT FIX as corner case]
- [x] (4) Av. offline enqueued after migration https://github.com/owncloud/android/pull/3396#issuecomment-934578584 [FIXED]
- [x] (5) Issues when not enough space on disk https://github.com/owncloud/android/pull/3396#issuecomment-935696030
- [x] (6) Logs not migrated https://github.com/owncloud/android/pull/3396#issuecomment-936595150 [FIXED]
- [x] (7) Small typo https://github.com/owncloud/android/pull/3396#issuecomment-936631598 [FIXED]
- [x] (8) Finished uploads not migrated https://github.com/owncloud/android/pull/3396#issuecomment-936662867
- [x] (9) `owncloud` folder stays in legacy storage https://github.com/owncloud/android/pull/3396#issuecomment-949554419  [FIXED]
- [x] (10) Crash after upgrading https://github.com/owncloud/android/pull/3396#issuecomment-950809084  [FIXED]
- [x] (11) Repetitions in multiaccount https://github.com/owncloud/android/pull/3396#issuecomment-954558608 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/migration/StorageMigrationActivity.kt
```diff
@@ -0,0 +1,107 @@
...
+            .commit()
+    }
+
+    companion object {
+
+        const val PREFERENCE_ALREADY_MIGRATED_TO_SCOPED_STORAGE = "MIGRATED_TO_SCOPED_STORAGE"
+
+        private fun hasDataInLegacyStorage(): Boolean {
+            val legacyStorageProvider = LegacyStorageProvider(MainApp.dataFolder)
+            val legacyStorageFolder = File(legacyStorageProvider.getRootFolderPath())
```
JuancaG05, line 78: These 2 variables are used in 2 different methods, what do you think about defining them outside of these methods as `private val`, inside the companion object?

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/migration/MigrationViewModel.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private val _migrationState = MediatorLiveData<Event<MigrationState>>()
+    val migrationState: LiveData<Event<MigrationState>> = _migrationState
+
+    init {
+        _migrationState.postValue(Event(MigrationState.MigrationIntroState))
+    }
+
+    private fun getLegacyStorageSizeInBytes(): Long {
+        val legacyStorageDirectory = File(LegacyStorageProvider(rootFolder).getRootFolderPath())
```
JuancaG05, line 57: `legacyStorageProvider` is also used in some other functions here. Just asking, couldn't it be in a `private val` property inside the ViewModel? Or even received as parameter in the constructor and injected via Koin?

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -627,4 +627,15 @@
...
 
     <string name="synchronizing_account">Synchronizing account</string>
+
+    <!-- Scoped storage migration wizard -->
+    <string name="scoped_storage_wizard_explanation">Found %1$s of data on your external storage. They will be moved to the safe storage on your device. Remaining files should be cleaned from your external storage after the migration to avoid duplicates and vulnerability.</string>
+    <string name="scoped_storage_wizard_intro">Your app has been updated. This app update migrates your files to a more secure location on your device. This one-time data migration is needed and will take a few minutes. Learn more</string>
+    <string name="scoped_storage_wizard_intro_button">Let\'s get started</string>
+    <string name="scoped_storage_wizard_migrate_and_clean_button">Migrate and clean (Recommended)</string>
+    <string name="scoped_storage_wizard_migrate_and_keep_button">Migrate but keep old files</string>
+    <string name="scoped_storage_wizard_migration_completed">Migration completed successfully. Your files are now saver than ever before.</string>
```
JuancaG05, line 637: I would say "safer", is that what you meant? 😁

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
```diff
@@ -128,7 +112,63 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
...
+    }
+
+    fun sizeOfDirectory(dir: File): Long {
+        if (dir.exists()) {
+            var result: Long = 0
+            val fileList = dir.listFiles() ?: arrayOf()
+            fileList.forEach { file ->
+                // Recursive call if it's a directory
+                result += if (file.isDirectory) {
+                    sizeOfDirectory(file)
```
JuancaG05, line 159: How brave implementing recursion 😨 🏅 💯

owncloudData/src/main/java/com/owncloud/android/data/storage/LegacyStorageProvider.kt
```diff
@@ -0,0 +1,30 @@
+/**
+ * ownCloud Android client application
+ *
```
JuancaG05, line 3: Who is the author? A ghost? 👻 . In `LocalStorageProvider.kt` we did wrote our names hehe

owncloudData/src/main/java/com/owncloud/android/data/storage/ScopedStorageProvider.kt
```diff
@@ -0,0 +1,30 @@
+/**
+ * ownCloud Android client application
+ *
```
JuancaG05, line 3: The ghost again!!! 👻

JuancaG05 (changes requested):
Some tiny suggestions @abelgardep, great job! 💯

owncloudApp/src/main/res/values/strings.xml
abelgardep: Blame https://github.com/owncloud/android/issues/3327#issuecomment-909460377 🙈

fesave (approved):
Great job!!! Congratulations!!! 🥇

JuancaG05 (approved):
💯

jesmrec:
### (1)

1. Display the wizard (by upgrading from prev version)

Current: in 1st slide, the `Learn more` points nothing

Expected: `Learn more` should point somewhere. If there is no a suitable way to point (complex or very technical information, or something not very friendly), i think we should remove it.

Pixel 2 Android 11
`7253e2ae`

jesmrec:
### (2)

How important is the `Remind me later` link in 1st slide? it is missing. Check [here](https://github.com/owncloud/android/issues/3327#issuecomment-909460377) the mock up. Which is the expected behaviour? As @tbsbdr:

` For those people, who really need the legacy storage-option for a longer transition period (until all apps migrate to the scoped storage) we should provide an option for them to postpone the migration to a convenient later time. I think we should be polite with our wizard and show it eg. only every 7 days (and also have the opportunity to start it directly form the settings). So the basic idea is: With postponing the migration a little bit, we buy some time for those people who still rely on legacy storage and eg. wait for soped-storage-upades for their eg. fancy pdf viewer.`

How complex is this to implement? Is it technically correct keep using legacy storage when the app points to Android 11 SDK?

The "longer transition period" should not be very long, because all apps must do the update. Therefore, if we have to implement a complex logic to delay the migration, it would not be worthy because they will be updated soon. 

More ideas welcome.

abelgardep:
Related to (2), if we have a `remind me later` it adds more potential error cases.

Imagine that you do want to perform the migration later, new files will be downloaded to the scoped storage. Maybe there are some with the same hierarchy in the legacy one. After that, if we try to migrate, those files to scoped storage, some conflicts could appear. 

We won't have full control over the files if we add a remind me later option

jesmrec:
> Related to (2), if we have a remind me later it adds more potential error cases.

> Imagine that you do want to perform the migration later, new files will be downloaded to the scoped storage. Maybe there are some with the same hierarchy in the legacy one. After that, if we try to migrate, those files to scoped storage, some conflicts could appear.

> We won't have full control over the files if we add a remind me later option

Is this ok for you to skip the "Remind later" option and perform migration just in the moment of the app upgrade? @theScrabi @michaelstingl @tbsbdr @JuancaG05 @fesave  

Pros: easier technically to implement (already done). Don't letting user to decide is straightforward and avoid confusions. Only one direction.

Contras: More work to do, closer to the deadline. Users will have to wait for other apps to implement SAF inside (from November on, without Scoped Storage and SAF, not posible to update in Google Play) and could forget to trigger the migration. Conflicts can happen, because we will not maintain the legacy data anymore, causing a bigger problem.

michaelstingl:
> Pros of removing "Remind me later": easier technically to implement (already done). Don't letting user to decide is straightforward and avoid confusions. Only one direction.

I also prefer the simplicity of this approach.

jesmrec:
### (3)

Check this out:

When the wizard is displayed after upgrading:

- If the app is killed, when it is opened again, the wizard is again there. ✅ 
- If the device is switched off/rebooted, by opening again the app, migration is not done and wizard not displayed. New downloads are done correctly, but downloaded stuff before is lost.

As this is a corner case, it could happen only in case the device is running out of battery (we can guess the user will not power off the device at this moment). From my side, it does not make sense to develop a mechanism to prevent such scenario, but it could be interesting to add in the first slide a recommendation like "please don't switch your device off. If your battery is low, please plug it on before continuing the process"

what do you think?

jesmrec:
### (4)

1. Install previous version and add one account
2. Set some stuff as av. offline (not needed a big amount)
3. Migrate to new version (no matter if cleaning or keeping)

Current:

When the migration finishes, all av. offline files and folder are queued in uploads view to upload to the server

Expected: 

Nothing changed, nothing should be uploaded

Pixel2 Android11
`7253e2aee`

abelgardep:
> Question -> Are log files migrated? @abelgardep

Yes, they should, right? The point here is that in 2.18.1, logs are not stored inside a logs folder, they were directly in the root folder and they are migrated to the root folder instead of a logs folder and it may break the logs collector 🤔

abelgardep:
> ### (4)
> 1. Install previous version and add one account
> 2. Set some stuff as av. offline (not needed a big amount)
> 3. Migrate to new version (no matter if cleaning or keeping)
> 
> Current:
> 
> When the migration finishes, all av. offline files and folder are queued in uploads view to upload to the server
> 
> Expected:
> 
> Nothing changed, nothing should be uploaded
> 
> Pixel2 Android11 `7253e2aee`

Migrating the files from legacy to scoped storage, at the moment, does not preserve copied file attributes such as creation/modification date, permissions, etc. 

This is a problem for the AvOffline service since it compares the last modification timestamp in the local file with the last time we get data from the server. It means that it detects that something changed in the local storage and it enqueues a new upload.

There are two potential ways to fix this.
1. Find a way to migrate the files keeping the local attributes.
2. Alter the current AvOffline service so it won't need to enqueue new uploads when a local file changes.

The second one could be a good option since we will store the files in the scoped storage and other apps won't be able to modify them. The only way to modify files within scoped storage will be using the documents provider and it already performs a synchronization.

jesmrec:
### (6) 

as discussed above

> Yes, they should, right? The point here is that in 2.18.1, logs are not stored inside a logs folder, they were directly in the root folder and they are migrated to the root folder instead of a logs folder and they won't be detected by the logs collector 🤔

we'd go for an approach in which log files (out of the logs folder) will not be migrated. Log files are removed after a week and the content could be reproduced again. Otherwise, old logs will not be removed by the logs collector.

jesmrec:
### (7)

Small typo?:

Screenshot 2021-10-06 at 18 28 55

safer would sound better. ;)

jesmrec:
### (8)

Finished uploads are lost in the uploads view after migrating, that means:

1. Install previous version
2. Perform some correct uploads -> they will appear as finished in the uploads view
3. Upgrade to this branch 

Current:

Uploads view is empty

Expected:

Finished uploads in the list

Is there any regard in the migration? afaik, the list of finished uploads is stored in DB.

Nexus 6P v7
`7253e2ae`

jesmrec:
Regarding (5)

With the latest commit `135e682` i have performed the following steps:

9GB available in device
Download a 5GB file and two 1GB files 

So, we have only ~2GB available

Migration

After migration:

5GB file is downloaded (green mark)
1GB files are not downloaded (no green mark)

But, checking the Device explorer:

1GB files are complete 
5GB file is not complete (1.8GB)

is there any way to remove pieces of file that were not totally migrated?

jesmrec:
Everything fixed, ready to go!

PR #3406, opened 2021-10-05
[Cleanup] Legacy error fragment

abelgardep:
Perform a little cleanup around some legacy code.

Apparently, it was introduced to migrate some files in 1.3.16 but at the moment it is not used anymore.

This was one of the strings: 
>As of version 1.3.16, files uploaded from this device are copied into the local %1$s folder to prevent data loss when a single file is synced with multiple accounts.\n\nDue to this change, all files uploaded in previous versions of this app were copied into the %2$s folder. However, an error prevented the completion of this operation during account synchronization. You may either leave the file(s) as is and remove the link to %3$s, or move the file(s) into the %1$s folder and retain the link to %4$s.\n\nListed below are the local file(s), and the remote file(s) in %5$s they were linked to.
_____


abelgardep:
`mForgottenLocalFiles` was a HashMap introduced to migrate some files in the past. But at the moment, it is not used anymore, it is always empty. Because of that, `ErrorsWhileCopyingHandlerActivity` is never called.

We can remove that flow completely

Issue #3415, opened 2021-10-21
[MDM] Integrate Android Enterprise in the app


JuancaG05:
[EPIC] -> https://github.com/owncloud/android/issues/3398
[SPIKE] -> https://github.com/owncloud/android/issues/3399

The very first step to be able to manage the app via MDM is to introduce Android Enterprise in the app.
In this video playlist they show how to begin with Android Enterprise: https://www.youtube.com/playlist?list=PLWz5rJ2EKKc-nnQcMIVqAM9_CNmbGu12O
Guide about managed configurations in Android documentation: https://developer.android.com/work/managed-configurations

In our case, it'll be necessary to manage in some way the code that will have to be executed ONLY in the branded apps which use MDM. We discussed 2 different approaches:
- Create a flag variable in `setup.xml` to indicate whether that instance of the app is going to be managed via MDM
- Create a new product flavor so that this code is only executed if the app is built for that variant

For the moment, we are taking the second approach, but more opinions and suggestions are welcome.

### TASKS

- [x] Research
- [x] Create branch feature/mdm
- [x] Development tasks
  - [x] Create new product flavors "mdmDebug" and "mdmRelease"
  - [x] Include the `APP_RESTRICTIONS` meta-data node in the Android manifest 
  - [x] Create a new `managed_configurations.xml` resource file with a `restriction` node for test.
  - [x] Instantiate a `RestrictionsManager` and retrieve the restriction set previously **on app launch**
  - [x] Use the restriction value to do whatever it implies
  - [x] Test it with Test DPC
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Fdroid integration
- [ ] Merge branch feature/mdm into master
- [ ] Adapt BitRise workflow for tests

### CHANGES IN BITRISE WORKFLOWS
Since we introduced new flavors in the "owncloudApp" module in this issue, it is necessary to adapt workflows in BitRise so that they continue working after merging this into master. The changes performed are summarised in the next points:
- The old workflows "primary" (UI and unit tests) are now "**new_arch**". As new architecture is being developed in a separate branch, it will be necessary to keep these workflows until a rebase against master is done.
- The new "**primary**" workflow is a copy of the old "primary", but with some adaptations:
  - In the UI tests workflow, now the apk is assembled for the "originalDebug" flavor
[...]

PR #3419, opened 2021-10-26
[MDM] Integrate Android Enterprise in the app

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3415

_____


owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
```diff
@@ -207,8 +214,28 @@ class MainApp : Application() {
...
+
+    private fun cacheRestrictions(activity: Activity, restrictions: Bundle) {
+        val preferencesProvider = SharedPreferencesProviderImpl(activity)
+        if (restrictions.containsKey(CONFIGURATION_SERVER_URL)) {
+            val managedString = restrictions.getString(CONFIGURATION_SERVER_URL)
+            managedString?.let { preferencesProvider.putString(CONFIGURATION_SERVER_URL, it) }
+        }
+        if (restrictions.containsKey(CONFIGURATION_SERVER_URL_INPUT_VISIBILITY)) {
+            val managedBool = restrictions.getBoolean(CONFIGURATION_SERVER_URL_INPUT_VISIBILITY)
+            preferencesProvider.putBoolean(CONFIGURATION_SERVER_URL_INPUT_VISIBILITY, managedBool)
```
fesave, line 231: A tiny question, Could `managedBool` be null? as `managedString` restriction? In this case, you should manage it.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -634,16 +637,17 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
 
     private fun initBrandableOptionsUI() {
-        if (!contextProvider.getBoolean(R.bool.show_server_url_input)) {
-            binding.hostUrlFrame.isVisible = false
-            binding.centeredRefreshButton.run {
-                isVisible = true
-                setOnClickListener { checkOcServer() }
-            }
+        val preferencesProvider = SharedPreferencesProviderImpl(this)
+        val showInput = if (BuildConfig.FLAVOR == "mdm") preferencesProvider.getBoolean(CONFIGURATION_SERVER_URL_INPUT_VISIBILITY, true) else contextProvider.getBoolean(R.bool.show_server_url_input)
```
fesave, line 641: I've seen you use the string `"mdm"` several times, maybe it could be a variable you can call. What if the flavor name changes?

fesave (changes requested):
Good work! 👍🏻 , but I post some comments, check them out!

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
JuancaG05: No, it can't hehe. In this case `getBoolean` assures that the value returned is not nullable, and that's why I check it in the previous line with `containsKey`.

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
JuancaG05: Ok, good point, I will move it to a variable 👍

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
fesave: Perfect 👍🏻

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
```diff
@@ -207,8 +214,28 @@ class MainApp : Application() {
...
+        }
+    }
+
     companion object {
         private const val BETA_VERSION = "beta"
+        const val CONFIGURATION_SERVER_URL = "server_url_configuration"
+        const val CONFIGURATION_SERVER_URL_INPUT_VISIBILITY = "server_url_input_visibility_configuration"
```
abelgardep, line 238: Now we have 2 constants and it could seem a good place to add those constants, but how will this companion object look like after 30 configurations?

Let's create a new file, maybe `MdmConfigurations.kt`, and add there those constants. It will be easier to maintain and MainApp will be cleaner 👍

owncloudApp/src/main/res/xml/managed_configurations.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<restrictions xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <restriction
+        android:key="server_url_configuration"
+        android:restrictionType="string"
+        android:title="@string/server_url_configuration_title"
+        android:description="@string/server_url_configuration_description"
+        android:defaultValue="demo.owncloud.com" />
```
abelgardep, line 9: How do default values affect the app?

For example, let's build a new OEM apk. Someone introduce in oB as server Url: `dummy.server.url`.

Let's distribute that APK without any MDM solution. Will that app show `dummy.server.url` or `demo.owncloud.com`?

Just curious 🤔

abelgardep (changes requested):
Some questions here @JuancaG05

owncloudApp/src/main/res/xml/managed_configurations.xml
JuancaG05: If you distribute the app built with the MDM flavor, the MDM configurations will be prior to the brandable configurations (the ones in `setup.xml`).
If you distribute the app with the original flavor (no MDM there), the `managed_configurations.xml` will be "invisible" because no code related to retrieving data from that file will be executed, so, the brandable configuration will be the only one and prior (the previous behaviour).

Answering your question, in that case the app will show `dummy.server.url` because `demo.owncloud.com` is in the `managed_configurations.xml` file, which is irrelevant in the original flavor (the one you proposed). For the moment that's the priority policy I applied and the one I consider is more logical, but it can be changed if we decide so.

Let me know if I answered your question completely.

abelgardep: My question is related to the MDM flavor. I already know that if we build with the original flavor, MDM is transparent.

But what about mdm flavor with the sample I posted?

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -620,4 +620,10 @@
     <string name="prefs_logs_summary">Logs may contain sensitive information. Sharing logs with others is sole user responsibility</string>
 
     <string name="synchronizing_account">Synchronizing account</string>
+
+    <!-- MDM configurations -->
+    <string name="server_url_configuration_title">Server URL</string>
+    <string name="server_url_configuration_description">The server URL that will appear in the login screen by default</string>
+    <string name="server_url_input_visibility_configuration_title">Server URL visibility</string>
+    <string name="server_url_input_visibility_configuration_description">Indicates whether the input field for the server URL in the login screen is visible or not</string>
```
abelgardep, line 628: Can those strings be shown within the app? Do we really need to add them to translatable strings?

owncloudApp/src/main/res/xml/managed_configurations.xml
```diff
@@ -0,0 +1,17 @@
...
+    <restriction
+        android:key="server_url_input_visibility_configuration"
+        android:restrictionType="bool"
+        android:title="@string/server_url_input_visibility_configuration_title"
+        android:description="@string/server_url_input_visibility_configuration_description"
```
abelgardep, line 14: What happens when we upload these restrictions to a MDM? Does it recognize `@string/server_url_input_visibility_configuration_description`?

abelgardep (changes requested):
Some more questions here @JuancaG05

owncloudApp/src/main/res/xml/managed_configurations.xml
JuancaG05: Ok, I get you, now code is adapted to have a more coherent behaviour when some MDM configs don't exist 😀

owncloudApp/src/main/res/values/strings.xml
JuancaG05: No, they aren't shown in the app, they're only kind of meta-data for MDM managed configurations, so I will move them to `strings__not_to_translate.xml`.

owncloudApp/src/main/res/xml/managed_configurations.xml
JuancaG05: It depends on the platform. For example, in IBM MaaS360 you have to upload the `managed_configurations.xml` file by itself, so that it doesn't recognise the value of the strings. However, in Blackberry platform, the file is taken directly from the APK you upload, so they're recognised. Anyway, you're not allowed to build the app if you hardcode the strings, it results in an error of the type:

`AAPT: error: 'Hardcoded description' is incompatible with attribute description (attr) reference.`

owncloudApp/src/main/res/values/strings.xml
JuancaG05: They aren't shown in the app, but documentation about managed configurations say that this should be in localized resources: https://developer.android.com/work/managed-configurations#define-configuration

fesave (approved):
LGTM 🥇

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -640,16 +644,17 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
     }
 
     private fun initBrandableOptionsUI() {
-        if (!contextProvider.getBoolean(R.bool.show_server_url_input)) {
-            binding.hostUrlFrame.isVisible = false
-            binding.centeredRefreshButton.run {
-                isVisible = true
-                setOnClickListener { checkOcServer() }
-            }
+        val preferencesProvider = SharedPreferencesProviderImpl(this)
```
abelgardep, line 647: We should avoid using an implementation directly. Let's inject the sharedPreferenceProvider the same way we inject the context provider. That way, we could test this in an easier way. 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -640,16 +644,17 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+        val showInput = if (BuildConfig.FLAVOR == MDM_FLAVOR) preferencesProvider.getBoolean(CONFIGURATION_SERVER_URL_INPUT_VISIBILITY, contextProvider.getBoolean(R.bool.show_server_url_input)) else contextProvider.getBoolean(R.bool.show_server_url_input)
+        binding.hostUrlFrame.isVisible = showInput
+        binding.centeredRefreshButton.isVisible = !showInput
+        if (!showInput) {
+            binding.centeredRefreshButton.setOnClickListener { checkOcServer() }
         }
 
-        if (contextProvider.getString(R.string.server_url).isNotEmpty()) {
-            binding.hostUrlInput.setText(contextProvider.getString(R.string.server_url))
+        val url = if (BuildConfig.FLAVOR == MDM_FLAVOR) preferencesProvider.getString(CONFIGURATION_SERVER_URL, contextProvider.getString(R.string.server_url)) else contextProvider.getString(R.string.server_url)
```
abelgardep, line 655: This logic shouldn't belong to the activity, but we can keep it at least for the moment till we add more mdm parameters 👍

abelgardep (changes requested):
Last tiny change and we will be ready to go

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
JuancaG05: 👍

abelgardep (approved):
First steps with MDM implemented, Well done @JuancaG05

PR #3429, opened 2021-11-02
Inlined value for parameter "key"

rahul6975:
## Related Issues
App:

Library PR (if needed):

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -139,7 +139,7 @@ public abstract class PreferenceManager {
      * @param context Caller {@link Context}, used to access to shared preferences manager.
      */
     public static void setLastUploadPath(String path, Context context) {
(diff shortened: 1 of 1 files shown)


owncloudApp/src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -208,9 +208,9 @@ private static void saveBooleanPreference(String key, boolean value, Context con
         appPreferences.apply();
     }
 
-    private static void saveStringPreference(String key, String value, Context context) {
+    private static void saveStringPreference(String value, Context context) {
```
abelgardep, line 211: Thanks for contributing to ownCloud @rahul6975. It's highly appreciated.

Unfortunately, this change would reduce the reusability of this method. When we call `saveStringPreference` we expect to save a string. There are more methods in this class like `saveIntPreference` or `saveBooleanPreference` that work the same way. So I would keep the same behavior.

If we remove the key from the parameters, we won't be able to use it to save other strings. So I would not merge it.

Anyway, thanks for your effort. Feel free to send any other contribution 👍

Issue #3400, opened 2021-09-30
[FEATURE REQUEST] Security enforced


jesmrec:
This is a feature that could be used in branding/MDM.

ACs

- App is always covered by one of the locking methods: passcode or pattern.
- Only affected passcode and pattern, not biometrical. 
- New branding/MDM option called `passcode_enforced` or similar. Boolean.
   - If the value is true, in the next opening of the app (or first opening, if the app was just installed), user is asked if prefers passcode or pattern lock
   - If the value is false, nothing happens
- After entering the passcode/pattern correctly twice, those fields will be hidden (and enabled) in Settings view. Only available the biometrical lock feature.
- If the `passcode_enforced` turns on after app is already installed
  - Next opening of the app will ask for the lock 
- If the `passcode_enforced` turns off after app is already installed
  - Lock fields become visible in Settings, so that user can decide to keep or disable.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/security_enforced
- [x] Development tasks
  - [x] Create a temporal setup variable
  - [x] Check if dialog should be visible
  - [x] Hide passcode/pattern in settings
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/security_enforced into master

### PR

- App #3434

PR #3434, opened 2021-11-04
[FEATURE REQUEST] Security enforced

fesave:
## Related Issues
App: #3400

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.20/3434-security_enforced.md

Reports:

- [X] (1) Cancel button should be hidden https://github.com/owncloud/android/pull/3434#issuecomment-972790583 [FIXED]
- [X] (2) Passcode skipped with the round button https://github.com/owncloud/android/pull/3434#issuecomment-972795091 [FIXED]
- [X] (3) Security skipped with triangle button https://github.com/owncloud/android/pull/3434#issuecomment-972797375 [FIXED]
- [X] (4) Biometrical locked https://github.com/owncloud/android/pull/3434#issuecomment-972803014 [FIXED]

### DIFF
--- a/changelog/unreleased/3434
+++ b/changelog/unreleased/3434
@@ -1,8 +1,7 @@
 Enhancement: Security enforced
 
-When a new branding/MDM option is true,
-a dialog appears and the user must select an option
-to lock the app.
+A new branding/MDM option has been added to make app
+lock via passcode or pattern compulsory, whichever the user chooses.
 
 https://github.com/owncloud/android/pull/3434
 https://github.com/owncloud/android/issues/3400
--- a/changelog/unreleased/3434
+++ b/changelog/unreleased/3434
@@ -1,0 +1,8 @@
+Enhancement: Security enforced
+
+When a new branding/MDM option is true,
+a dialog appears and the user must select an option
+to lock the app.
+
+https://github.com/owncloud/android/pull/3434
+https://github.com/owncloud/android/issues/3400
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
@@ -37,17 +37,17 @@ import androidx.test.espresso.matcher.ViewMatchers.isEnabled
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import androidx.test.platform.app.InstrumentationRegistry
 import com.owncloud.android.R
+import com.owncloud.android.presentation.ui.security.BiometricActivity
 import com.owncloud.android.presentation.ui.security.BiometricManager
 import com.owncloud.android.presentation.ui.security.PREFERENCE_LOCK_TIMEOUT
-import com.owncloud.android.presentation.ui.settings.fragments.SettingsSecurityFragment
-import com.owncloud.android.presentation.viewmodels.settings.SettingsSecurityViewModel
-import com.owncloud.android.presentation.ui.security.BiometricActivity
 import com.owncloud.android.presentation.ui.security.PassCodeActivity
+import com.owncloud.android.presentation.ui.security.PatternActivity
+import com.owncloud.android.presentation.ui.settings.fragments.SettingsSecurityFragment
+import com.owncloud.android.presentation.ui.settings.fragments.SettingsSecurityFragment.Companion.PREFERENCE_ACCESS_FROM_DOCUMENT_PROVIDER
 import com.owncloud.android.presentation.viewmodels.security.PassCodeViewModel
+import com.owncloud.android.presentation.viewmodels.settings.SettingsSecurityViewModel
 import com.owncloud.android.testutil.security.OC_PASSCODE_4_DIGITS
 import com.owncloud.android.testutil.security.OC_PATTERN
-import com.owncloud.android.presentation.ui.security.PatternActivity
-import com.owncloud.android.presentation.ui.settings.fragments.SettingsSecurityFragment.Companion.PREFERENCE_ACCESS_FROM_DOCUMENT_PROVIDER
 import com.owncloud.android.utils.matchers.verifyPreference
 import com.owncloud.android.utils.mockIntent
 import io.mockk.every
@@ -456,6 +456,34 @@ class SettingsSecurityFragmentTest {
         onView(withText(R.string.common_yes)).perform(click())
         onView(withText(R.string.prefs_touches_with_other_visible_windows)).perform(click())
         assertFalse(prefTouchesWithOtherVisibleWindows.isChecked)
+    }
+
+    @Test
+    fun passcodeLockNotVisible() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns true
+        launchTest()
+        assertFalse(prefPasscode.isVisible)
+    }
+
+    @Test
+    fun patternLockNotVisible() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns true
+        launchTest()
+        assertFalse(prefPattern.isVisible)
+    }
+
+    @Test
+    fun passcodeLockVisible() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
+        launchTest()
+        assertTrue(prefPasscode.isVisible)
+    }
+
+    @Test
+    fun patternLockVisible() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
+        launchTest()
+        assertTrue(prefPattern.isVisible)
     }
 
     private fun firstEnablePasscode() {
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
@@ -474,14 +474,12 @@ class SettingsSecurityFragmentTest {
 
     @Test
     fun passcodeLockVisible() {
-        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
         launchTest()
         assertTrue(prefPasscode.isVisible)
     }
 
     @Test
     fun patternLockVisible() {
-        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
         launchTest()
         assertTrue(prefPattern.isVisible)
     }
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
@@ -106,6 +106,7 @@ class SettingsSecurityFragmentTest {
 
         every { passCodeViewModel.getPassCode() } returns OC_PASSCODE_4_DIGITS
         every { passCodeViewModel.getNumberOfPassCodeDigits() } returns 4
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
 
         Intents.init()
     }
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -271,7 +271,7 @@ fun Activity.checkPasscodeEnforced(securityEnforced: ISecurityEnforced) {
     }
 }
 
-fun Activity.manageOptionLockSelected(type:LockType){
+fun Activity.manageOptionLockSelected(type: LockType) {
     when (type) {
         LockType.PASSCODE -> startActivity(Intent(this, PassCodeActivity::class.java).apply {
             action = PassCodeActivity.ACTION_REQUEST_WITH_RESULT
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -42,6 +42,7 @@ import com.owncloud.android.interfaces.ISecurityEnforced
 import com.owncloud.android.interfaces.LockType
 import com.owncloud.android.lib.common.network.WebdavUtils
 import com.owncloud.android.presentation.ui.security.PassCodeActivity
+import com.owncloud.android.presentation.ui.security.PassCodeActivity.Companion.EXTRAS_PASSCODE_ENFORCED
 import com.owncloud.android.presentation.ui.security.PatternActivity
 import com.owncloud.android.ui.dialog.ShareLinkToDialog
 import com.owncloud.android.ui.helpers.ShareSheetHelper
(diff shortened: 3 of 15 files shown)


owncloudApp/src/main/res/values/setup.xml
```diff
@@ -108,4 +108,8 @@
 
     <!-- Number of passcode digits -->
     <integer name="passcode_digits">4</integer>
-</resources>
\ No newline at end of file
+
+    <!--Security enforced -->
+    <bool name="passcode_enforced">true</bool>
```
jesmrec, line 113: false by default!

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -225,3 +231,42 @@ private fun makeIntent(file: File?, context: Context): Intent {
...
+fun Activity.manageOptionLockSelected(type: LockType) {
+    when (type) {
+        LockType.PASSCODE -> startActivity(Intent(this, PassCodeActivity::class.java).apply {
+            action = PassCodeActivity.ACTION_REQUEST_WITH_RESULT
+        })
+        LockType.PATTERN -> startActivity(Intent(this, PatternLockActivity::class.java).apply {
+            action = PatternLockActivity.ACTION_REQUEST_WITH_RESULT
+        })
+    }
+}
```
JuancaG05, line 271: This function is only called from here (indirectly, because `optionLockSelected` calls `manageOptionLockSelected`), so maybe it's not worth it to keep it as an extension. This would imply removing `optionLockSelected` from the interface and its implementations, and also making this a private function here (such as `makeIntent`), but I think it simplifies the code and reduces coupling in some way. Do you agree?

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt
```diff
@@ -705,4 +711,8 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
         }
         super.finish()
     }
+
+    override fun optionLockSelected(type: LockType) {
+        manageOptionLockSelected(type)
+    }
```
JuancaG05, line 717: If you perform the change I suggested, this method will be removed.

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModel.kt
```diff
@@ -52,4 +54,6 @@ class SettingsSecurityViewModel(
 
     fun setPrefTouchesWithOtherVisibleWindows(value: Boolean) =
         preferencesProvider.putBoolean(SettingsSecurityFragment.PREFERENCE_TOUCHES_WITH_OTHER_VISIBLE_WINDOWS, value)
+
+    fun isSecurityEnforcedEnabled(context: Context) = context.resources.getBoolean(R.bool.passcode_enforced)
```
JuancaG05, line 58: OMG, don't pass `Context` as parameter to VM methods! 😁  Inject contextProvider in the ViewModel, just like in `SettingsMoreViewModel`, for example, and get the boolean from there. Otherwise, it will be almost impossible to test. By the way, talking about tests, I don't see unit tests, have you forgotten them? 😈

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1693,6 +1699,10 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
         else -> R.id.nav_all_files
     }
 
+    override fun optionLockSelected(type: LockType) {
+        manageOptionLockSelected(type)
+    }
```
JuancaG05, line 1704: This method will be deleted if you apply the changes I suggested before, just like in LoginActivity.

JuancaG05 (changes requested):
Some changes requested here, good job @fesave 👍

changelog/unreleased/3434
```diff
@@ -0,0 +1,8 @@
+Enhancement: Security enforced
+
+When a new branding/MDM option is true,
+a dialog appears and the user must select an option
+to lock the app.
```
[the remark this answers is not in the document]
fesave: I love it! Thanks!

owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModel.kt
fesave: Oops! Good catch! Thank u 😉

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
fesave: In other scenarios I have had problems with dialogs and UI usability, so, I prefer to do it through an Interface.

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
```diff
@@ -425,6 +426,34 @@ class SettingsSecurityFragmentTest {
...
+        launchTest()
+        assertTrue(prefPasscode.isVisible)
+    }
+
+    @Test
+    fun patternLockVisible() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
+        launchTest()
+        assertTrue(prefPattern.isVisible)
+    }
```
JuancaG05, line 455: Great that you decided to add UI tests! 👏  But in these 2 tests, the `every` clauses are not necessary as you already set that this mocked method returns `false` in `setUp()`. In the other 2 tests you added, it IS necessary, as the value changes from what you specified at the beginning 😀

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
```diff
@@ -200,4 +204,18 @@ class SettingsSecurityViewModelTest : ViewModelTest() {
...
+        assertTrue(contextProvider.getBoolean(R.bool.passcode_enforced))
+    }
+
+    @Test
+    fun `set pref is security enforced enabled - ok - false`() {
+        every { securityViewModel.isSecurityEnforcedEnabled() } returns false
+
+        assertFalse(contextProvider.getBoolean(R.bool.passcode_enforced))
+    }
+
```
JuancaG05, line 220: Hm 🤔 , I don't think these unit tests are valid. Here, you are not actually testing the method `isSecurityEnforcedEnabled()`. You don't have to mock it for the tests to work, it is not like in UI tests. You should mock the necessary stuff of the content of the method, and check whether what the method returns is what you expect. Therefore, you mustn't mock the method itself, otherwise there's no testing and no coverage here! Tip: think or check the coverage the method provides 👍 
Also, would be cool if you added a `verify` clause to check if the method executes the lines you expect.
You can find an example of all of this in the `is help enabled - ok - true()` method in `SettingsMoreViewModelTest`.

JuancaG05 (changes requested):
Some other changes @fesave 😄

owncloudApp/src/androidTest/java/com/owncloud/android/settings/security/SettingsSecurityFragmentTest.kt
fesave: Good catch! Thanks 😉

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
fesave: The tests are new to me, sorry for the mistakes, it's a bit complicated at the beginning. thanks for teaching me!

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
```diff
@@ -200,4 +205,30 @@ class SettingsSecurityViewModelTest : ViewModelTest() {
         }
     }
 
+    @Test
+    fun `set pref is security enforced enabled - ok - true`() {
+        every { contextProvider.getBoolean(any()) } returns true
+
+        securityViewModel.isSecurityEnforcedEnabled().apply {
+            assertEquals(true, this)
+        }
```
JuancaG05, line 214: Two things here:
1) Do you think is it necessary to use `apply` here? Being only one line, I think it makes legibility worse. But that's only an opinion, and therefore it's discussable. You can keep it if you like it like that 👍 
2) When checking is true or false, try to use `assertTrue` or `assertFalse` instead of `assertEquals` with a boolean

JuancaG05 (changes requested):
Some last changes here @fesave!

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
fesave: Fixed!

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
JuancaG05: You don't have to be sorry 😄

JuancaG05 (approved):
Great job @fesave! Congrats 💯

jesmrec:
### (1)

1. Set passcode_enforced = true
2. Open app and select Passcode -> Passcode view displayed
3. Click `Cancel` 

Current: Security skipped
Expected: `Cancel` button should not be visible in this view, so user will not be able to skip the locking method

NOTE: Same behaviour if the selected method is pattern lock

Pixel2, Android 11
`c6a1b846`

jesmrec:
### (2)

1. Set passcode_enforced = true
2. Open app and select Passcode -> Passcode view displayed
3. Click the Android round button on the bottom to minimize the app
4. Open app again

Current: Passcode view skipped
Expected: Passcode view not skipped

NOTE: Pattern lock works fine in this case

Pixel2, Android 11
c6a1b846

jesmrec:
### (3)

1. Set passcode_enforced = true
2. Open app and select Passcode -> Passcode view displayed
3. Click the Android triangle button on the bottom to go back

Current: Passcode view skipped
Expected: Passcode view not skipped

NOTE: Same behaviour if the selected method is pattern lock

Pixel2, Android 11
c6a1b84

jesmrec:
### (4)

1. Set Pattern as method after installing
2. Enter correct pattern twice -> app unlocked
3. Open Settings -> Security

Current: Biometrical is greyed out
Expected: Biometrical is unlocked and can be ticked by the user

Pixel2, Android 11
`c6a1b84`

jesmrec:
This feature is approved on my side.

During the testing, the following problem was detected -> https://github.com/owncloud/android/issues/3458

it is not a matter of this branch, because is also reproducible in `master`, but not in the last released version. This is a regression to be fixed as p2 in the current release.

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -237,3 +244,44 @@ fun Activity.hideSoftKeyboard() {
         )
     }
 }
+
+fun Activity.checkPasscodeEnforced(securityEnforced: ISecurityEnforced) {
+    val sharedPreferencesProvider = SharedPreferencesProviderImpl(this)
+
+    val passcodeEnforced = this.resources.getBoolean(R.bool.passcode_enforced)
```
JuancaG05, line 251: If you are using this variable as a flag for the security lock enforcement, maybe you can call it `lockEnforced` instead of `passcodeEnforced`, it can lead to confusion (I thought there may be another variable for pattern enforcement)

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -237,3 +244,44 @@ fun Activity.hideSoftKeyboard() {
...
+
+fun Activity.manageOptionLockSelected(type: LockType) {
+    when (type) {
+        LockType.PASSCODE -> startActivity(Intent(this, PassCodeActivity::class.java).apply {
+            action = PassCodeActivity.ACTION_REQUEST_WITH_RESULT
+            putExtra(EXTRAS_PASSCODE_ENFORCED, true)
+        })
+        LockType.PATTERN -> startActivity(Intent(this, PatternActivity::class.java).apply {
+            action = PatternActivity.ACTION_REQUEST_WITH_RESULT
+            putExtra(EXTRAS_PASSCODE_ENFORCED, true)
```
JuancaG05, line 283: Same as the previous comment. It may sound strange that inside the pattern branch you activate the `EXTRAS_PASSCODE_ENFORCED` flag 😁

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
```diff
@@ -466,5 +470,6 @@ class PassCodeActivity : AppCompatActivity() {
 
         const val EXTRAS_MIGRATION = "PASSCODE_MIGRATION"
         const val PASSCODE_MIN_LENGTH = 4
+        const val EXTRAS_PASSCODE_ENFORCED = "EXTRAS_PASSCODE_ENFORCED"
```
JuancaG05, line 473: If you apply the changes I suggested in the previous comments, maybe you can move this constant to a more general class, such as `SettingsSecurityFragment`. It may increase coupling a bit, but makes more sense as it is analogous for both classes in the same way. What do you think? 🤔

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
```diff
@@ -38,11 +40,13 @@ import org.junit.Test
 class SettingsSecurityViewModelTest : ViewModelTest() {
     private lateinit var securityViewModel: SettingsSecurityViewModel
     private lateinit var preferencesProvider: SharedPreferencesProvider
+    private lateinit var contextProvider: ContextProvider
 
     @Before
     fun setUp() {
         preferencesProvider = mockk(relaxUnitFun = true)
-        securityViewModel = SettingsSecurityViewModel(preferencesProvider)
+        contextProvider = mockk()
```
JuancaG05, line 48: Maybe in this case it is not necessary, but just a tip: when you mock an object with MockK, it's usually very helpful (and sometimes it avoids headaches) to pass `relaxUnitFun = true` as parameter. This mocks automatically the methods of the object which return Unit (which sometimes are called internally, not directly by us in our code), preventing you of having to mock all of them in the same way 😀

JuancaG05 (changes requested):
Some questions here @fesave 😀

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
fesave: Good catch! 👍🏻

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/PassCodeActivity.kt
fesave: You saved my day! ❤️   I was looking for a general class. But I didn't find it.

owncloudApp/src/test/java/com/owncloud/android/presentation/viewmodels/settings/SettingsSecurityViewModelTest.kt
fesave: Okey, thanks!

JuancaG05 (approved):
Great job here @fesave 👍

jesmrec:
after the last changes in the CR, approved again.

Great job!!

Issue #3325, opened 2021-07-23
[RELEASE] 2.19


jesmrec:
### TASKS:

 - [x] [GIT] Create branch release/1.0.13 in owncloud/android-library from master
 - [x] [GIT] Create branch release/2.19 in owncloud/android from master
 - [x] [DEV] Update version number and name in build.gradle in owncloudComLibrary module
 - [x] [DEV] Update version number and name in build.gradle in owncloudApp module
 - [x] [DIS] Create a folder for the new version like 2.19.0_YYYY-MM-DD
 - [x] [DIS] Move all changelog files from the unreleased folder to the new version folder
 - [x] [DIS] Update screenshots, if needed, in README.md
 - [x] [QA] Design Test plan
 - [x] [QA] Regression Test plan
 - [x] [QA] Test pending stuff with Scoped Storage -> https://github.com/owncloud/android/pull/3336#issuecomment-894075798
 - [x] [DIS] Generate test bundle from `release/2.19` branch in owncloud/android
 - [x] [GIT] Create and sign tag 'oc-android-2.19' in HEAD commit of release branch, in owncloud/android
 - [x] [GIT] Create and sign tag '1.0.13' in HEAD commit of release branch, in owncloud/android-library
 - [x] [DIS] Generate final bundle from signed commit in owncloud/android
 - [ ] [COM] Inform `#release_coordination` and `#marketing` in internal chat
 - [ ] [GIT] Merge branch `release/1.0.13` in owncloud/android-library, into master
 - [ ] [GIT] Merge branch `release/2.19` in owncloud/android, into master
 - [x] [DIS] Upload & publish release bundle and changelog in Play Store
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [GIT] Publish a new release in owncloud/android
 - [x] [DIS] Upload release APK and bundle to internal owncloud instance
 - [ ] [GIT] Merge master branch into stable, in owncloud/android-library
 - [ ] [GIT] Merge master branch into stable, in owncloud/android
 - [ ] [DOC] Update owncloud.org/download version numbers (notify rocketchat #marketing)

_____

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.19/Regression%202.19.md

BUGS & IMPROVEMENTS:

- [x] https://github.com/owncloud/android/issues/3362 (regression)
[...]

PR #3435, opened 2021-11-05
[FIX] Scoped Storage Wizard Text 

fesave:
## Related Issues
App: https://github.com/owncloud/android/issues/3325#issuecomment-960656573

_____

### DIFF
--- a/owncloudApp/src/main/res/layout/fragment_migration_completed.xml
+++ b/owncloudApp/src/main/res/layout/fragment_migration_completed.xml
@@ -51,7 +51,7 @@
         android:layout_height="wrap_content"
         android:layout_margin="@dimen/standard_margin"
         android:background="@color/action_mode_background"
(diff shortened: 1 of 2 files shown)


changelog/unreleased/3435
```diff
@@ -0,0 +1,7 @@
+Bugfix: Scoped Storage Wizard Text
```
abelgardep, line 1: From my point of view, we don't need the calens item here. This is just an update about a text that was not released in a stable channel. So it's not needed to add a calens item for it.

abelgardep (changes requested):
Just a tiny change about calens

changelog/unreleased/3435
fesave: Okay, perfect!

abelgardep (approved):

PR #3456, opened 2021-11-21
Connection validator

theScrabi:
This is connection Validator.

Library: https://github.com/owncloud/android-library/pull/450

Test Plan (regression): https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.21/ConnectionValidator.md

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
@@ -60,7 +60,7 @@ import java.io.File
 
 class UploadFileFromContentUriWorker(
     private val appContext: Context,
-    private val workerParameters: WorkerParameters
+    private val workerParameters: WorkerParameters,
 ) : CoroutineWorker(
     appContext,
     workerParameters
@@ -231,7 +231,7 @@ class UploadFileFromContentUriWorker(
     }
 
     private fun getClientForThisUpload(): OwnCloudClient = SingleSessionManager.getDefaultSingleton()
-        .getClientFor(OwnCloudAccount(AccountUtils.getOwnCloudAccountByName(appContext, account.name), appContext), appContext)
+        .getClientFor(OwnCloudAccount(AccountUtils.getOwnCloudAccountByName(appContext, account.name), appContext), appContext, SingleSessionManager.getConnectionValidator())
 
     companion object {
         const val TRANSFER_TAG_CAMERA_UPLOAD = "TRANSFER_TAG_CAMERA_UPLOAD"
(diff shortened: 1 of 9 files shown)


owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt
```diff
@@ -37,24 +38,32 @@ class ClientManager(
...
     val context: Context,
     val accountType: String,
+    private val connectionValidator: ConnectionValidator
 ) {
     // This client will maintain cookies across the whole login process.
     private var ownCloudClient: OwnCloudClient? = null
 
+    init {
+        SingleSessionManager.setConnectionValidator(connectionValidator)
+    }
```
abelgardep, line 48: We inject the client manager via koin and I'm not sure if we can skip this `init` block in some way because it could be redundant. What do you think?

owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCRemoteAuthenticationDataSource.kt
```diff
@@ -42,16 +42,15 @@ class OCRemoteAuthenticationDataSource(
...
             ).apply { credentials = ownCloudCredentials }
 
-        val checkPathExistenceRemoteOperation = CheckPathExistenceRemoteOperation("/", true)
-        executeRemoteOperation { checkPathExistenceRemoteOperation.execute(client) }
+        val getBaseUrlRemoteOperation = GetBaseUrlRemoteOperation()
+        val rawBaseUrl = executeRemoteOperation { getBaseUrlRemoteOperation.execute(client) }
 
-        val userBaseUri =
-            checkPathExistenceRemoteOperation.redirectionPath?.lastPermanentLocation?.trimEnd(*WEBDAV_FILES_PATH_4_0.toCharArray())
+        val userBaseUri = rawBaseUrl?.replace(WEBDAV_FILES_PATH_4_0, "")
```
abelgardep, line 53: Instead of replace with "", we could try to trim it directly, right?

abelgardep (changes requested):

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
```diff
@@ -60,7 +61,8 @@ import java.io.File
 
 class UploadFileFromContentUriWorker(
     private val appContext: Context,
-    private val workerParameters: WorkerParameters
+    private val workerParameters: WorkerParameters,
+    private val connectionValidator: ConnectionValidator
```
abelgardep, line 65: This approach won't work. We do not inject anything via parameters in the workers. We need to find another way to do it.

1. We could inject it later as we do sometimes in other workers, check [camera uploads worker](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/workers/CameraUploadsWorker.kt#L71)
2. If we need to inject it, we could follow [koin guidelines for workmanager](https://insert-koin.io/docs/reference/koin-android/workmanager/)
3. We could try to inject the connectionValidator directly in the ClientManager if possible

abelgardep (changes requested):
Last change and we will be ready @theScrabi

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
abelgardep: A rollback to the [previous approach](https://github.com/owncloud/android/pull/3456/files/add5b5dd399540df81fe1a87e6fcd59b60af5188#diff-5332e04954815da4d22dc0c04385920f2b1e357b1e8d6c29758ca769bd17c833R234) should work too

abelgardep (approved):
LGTM

PR #3461, opened 2021-11-25
Retrieve data from DB as LiveData has been implemented.

fesave:
## Related Issues
App: #2869

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
@@ -19,6 +19,8 @@
 
 package com.owncloud.android.presentation
 
+import com.owncloud.android.domain.UseCaseResult
+
 sealed class UIResult<out T> {
     data class Loading<out T>(val data: T? = null) : UIResult<T>()
     data class Success<out T>(val data: T? = null) : UIResult<T>()
@@ -41,6 +43,15 @@ sealed class UIResult<out T> {
             is Error -> error
             else -> null
         }
+
+    companion object {
+        fun <T> fromResult(result: UseCaseResult<T>): UIResult<T> =
+            if (result.isSuccess) {
+                Success(result.getDataOrNull())
+            } else {
+                Error(result.getThrowableOrNull())
+            }
+    }
 }
 
 fun <T> UIResult<T>.onLoading(action: (data: T?) -> Unit): UIResult<T> {
@@ -69,3 +80,5 @@ fun <T> UIResult<T>.fold(
         is UIResult.Error -> onFailure(error)
     }
 }
+
+fun <T> UseCaseResult<T>.toUIResult() = UIResult.fromResult(this)
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
@@ -19,8 +19,6 @@
 
 package com.owncloud.android.presentation
 
-import com.owncloud.android.domain.UseCaseResult
-
 sealed class UIResult<out T> {
     data class Loading<out T>(val data: T? = null) : UIResult<T>()
     data class Success<out T>(val data: T? = null) : UIResult<T>()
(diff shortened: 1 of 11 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
```diff
@@ -41,6 +43,15 @@ sealed class UIResult<out T> {
...
         }
+
+    companion object {
+        fun <T> fromResult(result: UseCaseResult<T>): UIResult<T> =
+            if (result.isSuccess) {
+                Success(result.getDataOrNull())
+            } else {
+                Error(result.getThrowableOrNull())
+            }
+    }
```
abelgardep, line 54: Why do we have a companion function and an extension that do exactly the same? I think that they could be merged into one. What do you think?

owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
```diff
@@ -41,6 +43,15 @@ sealed class UIResult<out T> {
             is Error -> error
             else -> null
         }
+
+    companion object {
+        fun <T> fromResult(result: UseCaseResult<T>): UIResult<T> =
+            if (result.isSuccess) {
+                Success(result.getDataOrNull())
```
abelgardep, line 50: If we use `result is Success` instead of `result.isSuccess`, the cast inside will be smart 😉 

```suggestion
            if (result is Success) {
                Success(result.data)
```

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/files/filelist/MainFileListViewModel.kt
```diff
@@ -25,18 +25,19 @@ import androidx.lifecycle.MutableLiveData
...
 import com.owncloud.android.providers.ContextProvider
-import kotlinx.coroutines.Dispatchers
+import com.owncloud.android.providers.CoroutinesDispatcherProvider
 import kotlinx.coroutines.launch
 
 class MainFileListViewModel(
-    private val getFolderContentUseCase: GetFolderContentUseCase,
+    private val getFolderContentAsFlowUseCase: GetFolderContentUseCase,
     private val contextProvider: ContextProvider,
+    private val coroutinesDispatcherProvider: CoroutinesDispatcherProvider,
```
abelgardep, line 40: 🥇

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/files/filelist/MainFileListViewModel.kt
```diff
@@ -25,18 +25,19 @@ import androidx.lifecycle.MutableLiveData
...
 import com.owncloud.android.presentation.UIResult
+import com.owncloud.android.presentation.toUIResult
 import com.owncloud.android.providers.ContextProvider
-import kotlinx.coroutines.Dispatchers
+import com.owncloud.android.providers.CoroutinesDispatcherProvider
 import kotlinx.coroutines.launch
 
 class MainFileListViewModel(
-    private val getFolderContentUseCase: GetFolderContentUseCase,
+    private val getFolderContentAsFlowUseCase: GetFolderContentUseCase,
```
abelgardep, line 38: Not flow as far as I see, let's rename it.

By the way, I see that we keep using the `GetFolderContentUseCase`, shouldn't we use `GetFolderContentAsLiveDataUseCase` instead?

abelgardep (changes requested):
Some suggestions here @fesave

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/files/filelist/MainFileListViewModel.kt
fesave: Oh, I forgot to change it after some testing. Thank you.

owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
fesave: I just removed it.

owncloudApp/src/main/java/com/owncloud/android/presentation/UIResult.kt
fesave: Same here.

abelgardep (approved):
LGTM

PR #3474, opened 2021-12-15
Test CI with external repository 

jesmrec:
To check if that problem could be fixed.

Fool PR.


hannesa2:
There is no need to change anything, just do a 
`git commit -m "Empty" --allow-empty`

Issue #3458, opened 2021-11-23
[BUG] Rotation triggers security lock


jesmrec:
### Steps to reproduce

1.  Enable pattern/passcode
2. Open any view inside the app
3. Rotate device

### Actual behaviour
Security asked 

### Expected behaviour
Security not asked. It should be only asked to resume/reopen the app

This is not reproducible in 2.19 released version, is a regression to fix for 2.20.

Pixel2 Android11
Nexus6P Android7

jesmrec:
if i remember correctly, to avoid such effect a small delay was included in the activity

https://github.com/owncloud/android/blob/26b246abdd92fc1cf92ce9f84fe3fc7496e4d701/owncloudApp/src/main/java/com/owncloud/android/authentication/PatternManager.java#L104

it it helps...

jesmrec:
QA checks

Rotate device inside the app with 

- [X] Immediate + Passcode
- [X] Immediate + Pattern
- [X] Immediate + Biometrical
- [X] Non-immediate + Passcode
- [x] Non-immediate + Pattern
- [X] Non-immediate + Biometrical

Tested with Pixel2 Android11, Nexus 6P v7

Approved on my side

PR #3479, opened 2021-12-20
[FIX] Rotation triggers security lock

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3458
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/SecurityUtils.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/security/SecurityUtils.kt
@@ -32,7 +32,7 @@ enum class LockTimeout {
 
     fun toMilliseconds(): Int {
         return when (this) {
-            IMMEDIATELY -> 0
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -25,6 +26,14 @@ Details
    https://github.com/owncloud/android/issues/3060
    https://github.com/owncloud/android/pull/3437
 
+* Bugfix - Rotation triggers security lock: [#3458](https://github.com/owncloud/android/issues/3458)
```
abelgardep, line 29: Since this is a fix for an issue that has not been released yet, I think that we do not need to add a calens item for it. What do you think?

abelgardep (changes requested):
Just a tiny suggestion here @JuancaG05

CHANGELOG.md
JuancaG05: Ok, I thought about it but finally I decided to add it just to reflect the work done. As you want, I'll remove it 👍

abelgardep (approved):
LGTM 👍  REady to test @jesmrec

PR #3512, opened 2022-01-24
Api 31

hannesa2:
close https://github.com/owncloud/android/issues/3485

includes #3581

Issue #3485, opened 2021-12-23
Bump target SDK to Android 12 (31)


abelgardep:
We should bump the compileSdk and targetSdk to 31 since there are some dependencies that require those versions.

To do it, we need to verify that the ownCloud App supports each point in https://developer.android.com/about/versions/12/behavior-changes-12

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/feature_name
- [x] Research which Android12 features must be checked inside the app, due to feasible changes.
- [x] Development tasks
  - [x] Develop changes
  - [x] ...
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Check if in is installable from Internal testing track in Android 12
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

hannesa2:
This is my attempt https://github.com/owncloud/android/pull/3512

fesave:
## Apps targeting API 31+

[Behaviour changes](https://developer.android.com/about/versions/12/behavior-changes-12)

### [User experience](https://developer.android.com/about/versions/12/behavior-changes-12#ux)

- [ ] Custom notifications
- [ ] Android App Links verification changes
- [ ] PiP Mode
- [ ] Toast redesign

### [Security and privacy](https://developer.android.com/about/versions/12/behavior-changes-12#security-privacy)

- [ ] Approximate location
- [ ] Modern SameSite cookies in WebView
  - [ ] Test SameSite behaviors in your app
- [ ] Motion sensors are rate-limited
- [ ] App hibernation
- [ ] Attribution declaration in data access auditing
- [ ] ADB backup restriction
- [ ] Safer component exporting
- [ ] Messages in Android Studio
- [ ] Pending intents mutability
  - [ ]  Test the pending intent mutability change
  - [ ] Unsafe intent launches

### [Performance](https://developer.android.com/about/versions/12/behavior-changes-12#performance)

- [ ] Foreground service launch restrictions
- [ ] Exact alarm permission
- [ ] Disable the behavior change
- [ ] Notification trampoline restrictions
   - [ ] Identify which app components act as notification trampolines
   - [ ] Update your app
   - [ ] Toggle the behavior

### [Backup and restore](https://developer.android.com/about/versions/12/behavior-changes-12#backup-restore)

- [ ] D2D transfer functionality changes
- [ ] New include and exclude format
- [ ] XML format changes
- [ ] Manifest flag for apps

[...]

PR #3513, opened 2022-01-24
Empty commit to test CI

hannesa2:
At least local master is broken 

image


hannesa2:
I found the reason for broken local master builds. It's Room version 2.3.0. It has build issues on Apple Silicon M1.
Room 2.4.0 need api31 https://github.com/owncloud/android/pull/3512 which comes with an issue tool. -lockdown-

And as far as I remember I already suggest to use Room >= `2.4.0-alpha04` but you decline it.
That means I'm out of the game till you bump Room >= `2.4.0-alpha04`

abelgardep:
We merged some time ago this PR https://github.com/owncloud/android/pull/3302 that should make the work for M1 users.

A lot of new bumps require API 31, so we need to take care of them soon. We have an issue to track it https://github.com/owncloud/android/issues/3485

hannesa2:
Ahh ok. My bad, I've forgotten this.
It looks like, now it's too less. I tested locally with `2.4.0-alpha04` and it builds properly.

hannesa2:
> We merged some time ago this PR #3302 that should make the work for M1 users.

It's no more needed with api31 https://github.com/owncloud/android/pull/3512/files#r791376141

PR #3518, opened 2022-01-25
Restructured debug manifest

hannesa2:
This was the issue 

image


jesmrec:
Need to check here why the UI tests are not passing.

PR #3522, opened 2022-01-26
Update private link text to be identical to web UI

abelgardep:
Little update to use the same wording as web UI.
_____


jesmrec:
Nothing to test here. Go ahead with this.

jesmrec (approved):

PR #3530, opened 2022-01-30
Kotlin conversion PrepareVideoPlayerAsyncTask

hannesa2:
Kotlin conversion of `PrepareVideoPlayerAsyncTask.java`
There is no logic change.


theScrabi:
@hannesa2 thank you for your contribution. I think it was good if we not only convert from Java to Kotlin, but also clean up the code during that process. This way we ensure that the kotlinized code is truely modernized. It would be awesome if we could also change the logic of the code, as AsyncTask is deprecated. @hannesa2 what do you think would you also like to take care about getting rid of AsyncTask?

theScrabi:
@hannesa2 as described above, that we think its better to also modernize the code once we convert it to kotlin i think it's better to close it for now.

PR #3533, opened 2022-01-31
Fix lint error

hannesa2:
It's about 
image

You will see lint error with Android BumbleBee. 
Only string `uploader_error_message_read_permission_not_granted` comes with string format

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
@@ -48,31 +48,30 @@ public class CopyAndUploadContentUrisTask extends AsyncTask<Object, Void, Result
 
     /**
      * Helper method building a correct array of parameters to be passed to {@link #execute(Object[])} )}
-     *
+     * <p>
      * Just packages the received parameters in correct order, doesn't check anything about them.
      *
-     * @param   account             OC account to upload the shared files.
-     * @param   sourceUris          Array of "content://" URIs to the files to be uploaded.
-     * @param   remotePaths         Array of absolute paths in the OC account to set to the uploaded files.
-     * @param   behaviour           Indicates what to do with the local file once uploaded.
-     * @param   contentResolver     {@link ContentResolver} instance with appropriate permissions to open the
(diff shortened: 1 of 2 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
```diff
@@ -257,10 +256,7 @@ protected void onPostExecute(ResultCode result) {
                     default:
                         messageId = R.string.common_error_unknown;
                 }
-                String message = String.format(
-                        mAppContext.getString(messageId),
-                        mAppContext.getString(R.string.app_name)
-                );
+                String message = mAppContext.getString(messageId);
```
abelgardep, line 259: If we fix the lint error this way, we will need to update every translation of this key: `uploader_error_message_read_permission_not_granted` and it is translated to several languages.

We could,  for example, retrieve the string message inside the switch, so that, we won't need to update every single translation. What do you think @hannesa2 ?

abelgardep (changes requested):
Just a little suggestion, let me know what do you think about it 👍

owncloudApp/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
hannesa2: The root cause is to use several translated strings, where one has a placeholder for "Owncloud".
What's the point to have the (dynamic ?) app name there ?

To be honest, I would solve this root cause (this pull request) and change translated string in that way to remove this placeholder.
This would be a nice, but not a mandatory step.

Anyway, you can transform this pull request in any way you want (or later in an own pull request). The branch is in origin repository

abelgardep: I know which is the root cause. I agree that it may be unnecessary to have that placeholder, but this PR does not fix the issue completely. Some translations will expect an argument for that key and we are not providing it.

Thanks anyway to report the lint error 👍

abelgardep:
Will be fixed via #3541

Thanks for the contribution anyway 👍

PR #3537, opened 2022-02-01
Prevent a crash when logging in

abelgardep:
Prevent a crash when logging in. This crash does not happen in the 2.19 version

_____


JuancaG05 (approved):
LGTM

jesmrec:
Nothing to test here. Feel free to move forward.

Issue #3520, opened 2022-01-26
[FEATURE REQUEST] Make document provider visibility independent from locking methods


jesmrec:
Actually, the feature is tied to the passcode/pattern.

Option will be independent from locking/methods with the same behaviour

### TASKS

- [x] Research
- [x] Create branch feature/doc_provider_visibility
- [x] Development tasks
  - [x] Change visibility condition in document provider
  - [x] Make preference available even with no lock method enabled
  - [x] Adapt unit tests
  - [x] Adapt UI test
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/doc_provider_visibility into master

Issue #3379, opened 2021-09-14
[FEATURE REQUEST] Allow access from document provider preference


JuancaG05:
[QA pending - #3344]

Actually, if the app is protected with passcode/biometrical, it is not posible to access the accounts using doc provider apps:

![Screenshot 2021-08-11 at 09 07 33](https://user-images.githubusercontent.com/14894746/128984909-5f89c61a-db0e-4008-a655-ae403bf7fae8.png)

The idea is to implement a new preference in "Security" section which lets the user choose between allowing access from document provider or not, disabled by default so that it is more secure.

### TASKS

- [x] Create branch feature/access_from_document_provider
- [x] Development tasks
  - [x] Implement preference "Allow access from document provider"
  - [x] Put some security warnings for the user when enabling this preference
  - [x] Update unit tests
  - [x] Update UI tests
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/access_from_document_provider into master

### PR

- App
- Library (if needed)

hannesa2:
I strongly vote for this feature. With this restriction the app is almost useless for my purpose

JuancaG05:
@hannesa2 which feature do you refer to? This was already implemented and closed.

hannesa2:
sorry, I wrote to wrong issue

3oooo0:
> [QA pending - [#3344](https://github.com/owncloud/android/issues/3344)]
> 
> Actually, if the app is protected with passcode/biometrical, it is not posible to access the accounts using doc provider apps:
> 
> ![Screenshot 2021-08-11 at 09 07 33](https://user-images.githubusercontent.com/14894746/128984909-5f89c61a-db0e-4008-a655-ae403bf7fae8.png)
> 
> The idea is to implement a new preference in "Security" section which lets the user choose between allowing access from document provider or not, disabled by default so that it is more secure.
> 
> ### TASKS
> * [x]  Create branch feature/access_from_document_provider* [x]  Development tasks
>     
>     * [x]  Implement preference "Allow access from document provider"* [x]  Put some security warnings for the user when enabling this preference* [x]  Update unit tests* [x]  Update UI tests[x]  Code review and apply changes requested[x]  Design test plan[x]  QA[x]  Merge branch feature/access_from_document_provider into master
> 
> ### PR
> * App
> * Library (if needed)

PR #3538, opened 2022-02-01
[FEATURE REQUEST] Make document provider visibility independent from locking methods

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3520

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.20/3538-document_provider_visibility.md

- [x] (1) Remove dialog https://github.com/owncloud/android/pull/3538#issuecomment-1027926345 [FIXED]
- [x] (2) owncloud listed in doc. prov when no accounts are there https://github.com/owncloud/android/pull/3538#issuecomment-1027933138
- [x] (3) Update changelog https://github.com/owncloud/android/pull/3538#issuecomment-1031376556 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
```diff
@@ -190,25 +188,24 @@ class SettingsSecurityFragment : PreferenceFragmentCompat() {
...
             prefLockApplication?.isEnabled = false
-            disableAccessFromDocumentProvider()
         }
 
-        // Access from document provider
-        prefAccessDocumentProvider?.setOnPreferenceChangeListener { preference: Preference?, newValue: Any ->
-            if (newValue as Boolean) {
+        // Lock access from document provider
+        prefLockAccessDocumentProvider?.setOnPreferenceChangeListener { _: Preference?, newValue: Any ->
+            if (!(newValue as Boolean)) {
```
abelgardep, line 198: I don't know if this is the expected behavior 🤔  

Not sure if we should show a dialog to the user when he wants to enable the documents provider(which is the expected behavior) from a UX point of view

abelgardep (commented):
Just a question related to UX, but the code looks good! Good job @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/presentation/ui/settings/fragments/SettingsSecurityFragment.kt
jesmrec: IMHO, the dialog does not make sense here since access to document provider is open by default. Warning users when the `Lock` option is switched off would be the same as warning when the `Security` section is opened. I'd remove the dialog because the feature meaning is opposite as the previous one.

JuancaG05: Ok, I will remove it then. Thank you!

jesmrec:
### (1)

this one: https://github.com/owncloud/android/pull/3538#discussion_r797581091

jesmrec:
### (2)

I noticed a tiny difference with the old behaviour

1. Install app 
2. Go to Settings without adding an account)
3. Enable `Lock access from document provider`
4. Open any app that implements document provider

Current:  `owncloud` is shown as Locked

Expected (before): `owncloud` is hidden because no accounts were attached yet

Good to fix, but not a blocker

Pixel2 V11
`08b6cf94`

jesmrec:
### (3)

Please, don't forget removing the following entry from the Changelog:

```
Enhancement - Allow access from document provider preference: [#3379](https://github.com/owncloud/android/issues/3379)
```

it does not make sense anymore, since the current one covers totally the final result.

fesave:
> ### (3)
> Please, don't forget removing the following entry from the Changelog:
> 
> ```
> Enhancement - Allow access from document provider preference: [#3379](https://github.com/owncloud/android/issues/3379)
> ```
> 
> it does not make sense anymore, since the current one covers totally the final result.

Removed 😉

fesave:
> ### (3)
> Please, don't forget removing the following entry from the Changelog:
> 
> ```
> Enhancement - Allow access from document provider preference: [#3379](https://github.com/owncloud/android/issues/3379)
> ```
> 
> it does not make sense anymore, since the current one covers totally the final result.

Removed 😉

fesave:
> hidden because no accounts were attached yet

I think it's fixed, if you can test it thoroughly to make sure I'd appreciate it, anything you see, let me know.

jesmrec:
Everything is correct!! approved on my side and right to 2.20

abelgardep (approved):

Issue #3468, opened 2021-12-09
[BUG] Security flags for screen recording


JuancaG05:
### Actual behaviour
Currently, if we set the passcode or the pattern as lock for the app, the screen is black from that screen on, in programs such as Vysor, or if we try to record the screen in some way.

### Expected behaviour
The passcode or pattern unlock screen should appear in black when recording, but not the rest of the screens.
 
### Steps to reproduce
1. Set passcode or pattern
2. Close the app
3. Open the app while recording it or watching it with Vysor
4. Unlock the app
5. From that moment, every screen appear in black

The code responsible for this is located in `MainApp.kt`:
```
                if (!enabledLogging) {
                    // To enable biometric you need to enable passCode or pattern, so no need to add check to if
                    if (passCodeEnabled || patternCodeEnabled) {
                        activity.window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
                    } else {
                        activity.window.clearFlags(WindowManager.LayoutParams.FLAG_SECURE)
                    }
                } // else, let it go, or taking screenshots & testing will not be possible
```
We should check if this makes sense and if so, how to fix this unexpected behaviour.

### Environment data

ownCloud app version: 2.19

jesmrec:
Howto

Debug:

- All clear

Release:

- Passcode / Pattern hidden
- Login hidden

michaelstingl:
> but not the rest of the screens.

Maybe other content is also worth protecting it? Branding/MDM option for full-protection vs. login-protection?

jesmrec:
How could this be useful for external users/customers? our idea is helping ourselves with that... any hidden use case?

PR #3540, opened 2022-02-02
Remove pointless debug condition

hannesa2:
This

`if (MainApp.Companion.getEnabledLogging()) {`

makes not that much sense here


abelgardep (approved):
Thanks for the update. I agree, `enabled logging` is already checked to allow timber to log, so it makes no sense anymore.

We will remove the other uses of it within https://github.com/owncloud/android/issues/3468

PR #3560, opened 2022-02-18
[FIX] Security flags for screen recording

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3468

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.20/3560-security_flags.md

Reports:

- [x] (1) Biometrical view is black in debug mode https://github.com/owncloud/android/pull/3560#issuecomment-1047992231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,21 @@ ownCloud admins and users.
 Summary
 -------
 
+* Bugfix - Security flags for recording screen: [#3468](https://github.com/owncloud/android/issues/3468)
 * Enhancement - Lock delay enforced: [#3440](https://github.com/owncloud/android/issues/3440)
 * Enhancement - Extended security enforced: [#3543](https://github.com/owncloud/android/issues/3543)
 
 Details
 -------
+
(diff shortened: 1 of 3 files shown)


abelgardep (approved):
LGTM

jesmrec:
### (1)

1. Install debug variant
2. Enable passcode/pattern
3. Enable biometrical
4. Quit the app and enter again

Current: biometrical asked with black screen in screen mirror (vysor used)
Expected: biometrical is mirrored as well as passcode and pattern

Pixel 2 Android11
`f9b173c8`

JuancaG05:
> Regarding (1), i've checked with more devices and the only one in which this is reproducible is the Pixel2. Nexus6P or Huawei don't reproduce the problem either.
> 
> So, maybe something related with that device itself...

I also checked with Samsung Galaxy S9 (Android 9) and it is not reproducible 🤔

jesmrec:
Ok, we can assume that the report is something specific of one device. Let's move forward this then

fesave (approved):
LGTM

Issue #3553, opened 2022-02-11
[MDM] Making code more scalable


jesmrec:
[EPIC] -> https://github.com/owncloud/android/issues/3398

Currently, every MDM configuration is handled in the place the code is written for. 

For example, in the [current MDM branch](https://github.com/owncloud/android/pull/3419):

https://github.com/owncloud/android/blob/0f5e4cba3875f32d321758850bb104344679ccfe/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt#L656

The idea is creating a specific provider to handle all stuff requiring MDM, like setting priorities together with `setup.xml` and provide the value that the code needs in every place.
 
Other ideas to improve are welcome.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/mdm_scalable
- [x] Development tasks
  - [x] Create a new provider for MDM
  - [x] Retrieve the MDM restrictions from the new provider
  - [x] Cache the MDM restrictions in the new provider
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/mdm_scalable into feature/mdm

abelgardep:
Already merged into https://github.com/owncloud/android/pull/3419

PR #3563, opened 2022-02-22
[MDM] Making code more scalable

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3553
_____

## QA

Check list: https://github.com/owncloud/android/pull/3563#issuecomment-1072224933

1st test matrix: https://github.com/owncloud/android/pull/3563#issuecomment-1069350769
2nd test matrix: https://github.com/owncloud/android/pull/3563#issuecomment-1081716764

Reports:

- [X] (1) Deleted field https://github.com/owncloud/android/pull/3563#issuecomment-1073568001 [DONE]
- [X] (2) Empty field https://github.com/owncloud/android/pull/3563#issuecomment-1073585795 [DONE]
- [X] (3) UI test MDM flow becomes flaky https://github.com/owncloud/android/pull/3563#issuecomment-1073865807 [DONE]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Summary
 * Bugfix - Security flags for recording screen: [#3468](https://github.com/owncloud/android/issues/3468)
 * Enhancement - First steps in Android Enterprise integration: [#3415](https://github.com/owncloud/android/issues/3415)
 * Enhancement - Lock delay enforced: [#3440](https://github.com/owncloud/android/issues/3440)
+* Enhancement - Provide app feedback to MDM admins: [#3420](https://github.com/owncloud/android/issues/3420)
 * Enhancement - Extended security enforced: [#3543](https://github.com/owncloud/android/issues/3543)
 * Enhancement - Improvements for the UI in the file details screen: [#3557](https://github.com/owncloud/android/issues/3557)
 
@@ -42,6 +43,14 @@ Details
 
    https://github.com/owncloud/android/issues/3440
    https://github.com/owncloud/android/pull/3547
+
+* Enhancement - Provide app feedback to MDM admins: [#3420](https://github.com/owncloud/android/issues/3420)
+
+   Now, when a MDM configuration is applied for the first time or changed by an IT administrator,
+   the app sends feedback that will be shown in the EMM console.
+
+   https://github.com/owncloud/android/issues/3420
+   https://github.com/owncloud/android/pull/3480
 
 * Enhancement - Extended security enforced: [#3543](https://github.com/owncloud/android/issues/3543)
 
--- a/changelog/unreleased/3480
+++ b/changelog/unreleased/3480
@@ -1,0 +1,8 @@
+Enhancement: Provide app feedback to MDM admins
+
+Now, when a MDM configuration is applied for the first time
+or changed by an IT administrator, the app sends feedback that
+will be shown in the EMM console.
+
+https://github.com/owncloud/android/issues/3420
+https://github.com/owncloud/android/pull/3480
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -171,7 +171,6 @@ android {
         }
         mdm {
             dimension "management"
-            applicationIdSuffix ".mdm"
         }
     }
 
--- a/owncloudApp/build.gradle
+++ b/owncloudApp/build.gradle
@@ -30,7 +30,7 @@ dependencies {
     // CustomTabs required for OAuth2 and OIDC.
     implementation "androidx.browser:browser:1.3.0"
 
-    //Zooming Android ImageView.
+    // Zooming Android ImageView.
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion"
@@ -58,6 +58,9 @@ dependencies {
 
     // Preferences
     implementation 'androidx.preference:preference-ktx:1.1.1'
+
+    // MDM feedback
+    implementation 'androidx.enterprise:enterprise-feedback:1.1.0'
 
     // Tests
     testImplementation project(':owncloudTestUtil')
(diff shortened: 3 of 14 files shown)


fesave (approved):
Excellent work @abelgardep! Not forgetting @JuancaG05's work either! Congratulations to both of you, for me it's perfect. LGTM

jesmrec:
Checks before finishing:

- Check option in TestDPC to use the default values
- Check if the issues above are OK
- Check if the MDM BitRise flow works over this branch and the parent branch
- Check how the branding scripts are working

jesmrec:
Let's check individually the issues mentioned above

jesmrec:
### (1)

1. Install oC app in work profile
2. Open TestDPC
3. Load Manifest restrictions & save -> default values are applied in the app
4. Remove one of the configurations from TestDPC (f. ex `server_url_configuration`) & save

Current: Nothing happens in the app, the old version stays

Expected: Here, some options:

- Current behaviour is OK. As commented above, we could assume this behaviour, so that field are not intended to be removed but to be changed
- If (string) field is removed from configurations, it is set as empty
- If field is removed from configurations setup.xml is set (default priority)

any other options or ideas?

jesmrec:
### (2)

This is an open question

1. Install oC app in work profile
2. Open TestDPC
3. Load Manifest restrictions & save -> default values are applied in the app
4. In TestDPC, open `server_url_configuration` and set it as empty value & save
5. Kill and open oC app

Current: server URL field is empty in oC app

As field is empty after performing the steps, the current behaviour is not a bad one. My question is, for string fields, empty field will mean empty in the oC app as well? so, the `setup.xml` value will be set only in case the configuration field is not there. Does this make sense?

jesmrec:
### (3)

Unit tests new MDM flow  ✅  -> https://app.bitrise.io/build/d7c20289-2d48-4026-9846-29e3fdefa6cf

UI tests new MDM flow -> app builds correctly but there is a non-passing test -> `checkServerInfo_isSuccess_None` . Check the following builds: 

https://app.bitrise.io/build/5d5583e1-ebda-4cb4-92ca-1aef73fb5212
https://app.bitrise.io/build/e68c6b8c-982a-4c32-986a-73cf537f667a

jesmrec:
Checks over the building scripts are blocked by https://github.com/owncloud/enterprise/issues/5083

abelgardep:
**About (1)**
Not sure what is the expected behavior here. Is removing an already set field a common/expected case? If I want the enrolled apps to use a different field, I could just update that field to the new value instead of removing it to use the setup one. Am I missing something? Any special case? 

**About (2)**
I understand it, but not sure what we should expect in those cases. I know that some configurations could make the app a "brick". For example: empty serverUrl + hidden URL. But, not sure if we should protect the app against those cases. It's an admin task, this case could happen when building the setup XML too.

**About (3)**
Ok, I'll take a look and will send a fix asap.

@michaelstingl We need some clarification for [report 1](https://github.com/owncloud/android/pull/3563#issuecomment-1073568001) and [report 2](https://github.com/owncloud/android/pull/3563#issuecomment-1073585795) related to expected behaviors for MDM configurations.

jesmrec:
Thanks all for your inputs. Let's propose something and discuss/iterate

- If a MDM configuration is not included in the MDM file, it's supposed that sysadmin is not interested anymore in the MDM configuration for such value. So, `setup.xml` value should be set instead. Now, this is not happening and will need a fix (not detected when restriction is removed as described in (1)) . feasible? 

- If a MDM configuration is empty, is a valid configuration for strings (some user could be interested in empty string as proper configuration). For boolean or integer, i's needed a fallback value so, i'd set a default one (0 for integer, false for boolean).

With these policies:

- MDM parameters will be always prior than oB parameters
- If MDM parameter is removed from configuration, oB will replace after the next app opening
- If MDM parameter is an empty string, it will be set as is
- If MDM parameter is an empty int/bool, we will set defaults. No fallback to oB

is this ok??

jesmrec:
> (1) Should be fixed now. If the admin removes the mdm configuration, we will start using the setup value again.

:rocket:  will do another testing round

> (2) We already use false as default value for booleans. Not sure if this would be the expected behavior. No integers in MDM configurations atm. We also need to remove the demo server from default value in the configurations btw. What should we write there?

i'd set empty value . URLs do not have default values and any value we set as default is not valid for all users.

> (3) Should be fixed now, tests are green again: https://app.bitrise.io/build/5873efeb-e0f6-4837-9173-3320a21bbde8
Scripts should be fixed now and will be merged in next release

:rocket::rocket:

jesmrec:
Let's give another try to the matrix with the last fixes. `MDM Default` removed because it's only useful to load a default set of properties, not for the decision about which one has to be pushed to the app

| `server_url_configuration`<br>(MDM config)| `server_url` (setup.xml) | Outcome | Result | 
|:-:|:-:|:-:|:-:|
| myOCserver1  | mySetupServer | myOCserver1 | ✅   | 
| myOCserver1  | empty | myOCserver1 |  ✅  |
| empty  | mySetupServer | empty | ✅   |
| empty  | empty | empty |  ✅   |
| non existing key  | mySetupServer | mySetupServer | ✅    |
| non existing key   | empty | empty |  ✅  |

| `server_url_input_visibility_configuration`<br>(MDM config)| `show_server_url_input`<br> (setup.xml) | Outcome | Result | 
|:-:|:-:|:-:|:-:|
| true  | true | URL visible | ✅  | 
| true  | false | URL visible | ✅   |
| false  | true | URL hidden |  ✅  |
| false  | false | URL hidden |  ✅   |
| non existing key  | true | URL visible | ✅   |
| non existing key   | false | URL hidden |  ✅   |

*it's assumed that `setup.xml` will always contain a true/false value for `show_server_url_input`

so, (1) and (2) work as expected now

jesmrec:
About uploads, these were the performed steps:

1. Install `oc-android.2.20_oem` in device (no new variants there, i got the `debug` one), in work profile.
2. Add an account
3. Checkout to the current branch
4. Upgrade using the `mdmDebug` variant
5. In Login view, apply the `server_url_configuration` and `server_url_input_visibility_configuration`

Same steps done with `release` and `mdmRelese` variants

Result: OK

abelgardep:
One last task pending: Check whether f-droid builds the app properly with the new flavors

PR #3568, opened 2022-02-24
Feature/add stetho

theScrabi:
Library PR (if needed): https://github.com/owncloud/android-library/pull/469

This will add facebook stetho and leak canary permanently to the debug version of the app. This makes debuging network requests easier as requests can be viewed all the time without having to use mitm proxy and therefore explicitly change the phone configuration.
These two things will not be pressent in the release build.


abelgardep (approved):
LGTM

PR #3570, opened 2022-02-25
Multiselection

fesave:
Everything related to multiselection has been done in this branch.


fesave (commented):
LGTM!! But I can't approve it because I'm the author.

PR #3575, opened 2022-03-03
[Cleanup] Polish some code

abelgardep:
Polish some code: Making `when` exhaustive and some more
_____


theScrabi (approved):
According to bitrise there are unused imports. Other than this LFTM.

Issue #3001, opened 2020-10-22
add lint / checkscript


theScrabi:
Add a `lint.sh` or `check.sh` script to the android and android-library project, that will check all the small and nasty things one tends to forget when creating a new Pull request.
This script could check
- that the license header was added to all `*.kt` and `*.java` files
- code formation is correct in all `*.kt` and `*.java` files
- lines do not exceed a certain length in all `*.kt` and `*.java` files
- check that author is added
- copyright year got updated

This script should be executed by the developers before committing/creating a PR and it must also be executed by the CI.

PR #3576, opened 2022-03-03
Feature/check script

theScrabi:
## Related Issues
App: https://github.com/owncloud/android/issues/3001

Library PR (if needed): https://github.com/owncloud/android-library/pull/472
_____

This is a little check script that should be runned before moving a pull request to CR state. This way some obviouse issues can be resolved before performing the actual CR.

Things it does handle right now:
- Checking for correct license header
- Formating Kotlin code

### DIFF
--- a/check_script.sh
+++ b/check_script.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+check_directory() {
+    if ! head -n 20 $FILE | grep -q "This program is free software: you can redistribute it and/or modify"
+    then
+        if ! head -n 20 $FILE | grep -q "Licensed under the Apache License, Version 2.0"
+        then
+            echo "$FILE does not contain a current copyright header"
+        fi
+    fi
+}
+
+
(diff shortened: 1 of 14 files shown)


check_script.sh
```diff
@@ -0,0 +1,26 @@
...
+    do
+        check_directory
+    done
+done
+
+./gradlew ktlintFormat
+
+echo "========= Check Library =========="
+cd owncloud-android-library
+./check_script.sh
```
abelgardep, line 26: Really interesting script to not forget to add the licenses header. 💯 

I think we should move it out of the root folder though. In both repositories, android and android-library. What do u think?

abelgardep (changes requested):
Little suggestion here @theScrabi but interesting stuff 🥇

check_script.sh
jesmrec: And also, give it another name. `check_script` is not very descriptive. Something like `check_license_headers` or similar. About the location, not a problem for me to keep it in root, but also no strong opinion.

abelgardep (approved):

jesmrec (approved):

Issue #3571, opened 2022-03-03
[BUG] Crash when changing orientation in Details view after installing


jesmrec:
### Steps to reproduce

1.  Install the app and add one account
2. Select a file and open three dot button -> `Details`
3. Change orientation to landscape

### Actual behaviour

App crashes. Crash happens only the first time, new orientation changes on the same file or other does not make the app crash. I realised that is not reproducible with any file. But the same file causes the crash in different devices and servers. This is a file name i reproduce the problem with, if it helps: `rr"·$%()=?€~\^+*´¨çÇ{}-.,äôéì|–…„.JPG`

Stacktrace:

```
2022-03-03 10:56:34.666 12564-12564/com.owncloud.android.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.owncloud.android.debug, PID: 12564
    java.lang.RuntimeException: Unable to start activity ComponentInfo{com.owncloud.android.debug/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.IllegalArgumentException: storageManager must not be NULL
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2666)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2727)
        at android.app.ActivityThread.handleRelaunchActivity(ActivityThread.java:4520)
        at android.app.ActivityThread.-wrap19(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1484)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6121)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
     Caused by: java.lang.IllegalArgumentException: storageManager must not be NULL
        at com.owncloud.android.datamodel.ThumbnailsCacheManager$ThumbnailGenerationTask.<init>(ThumbnailsCacheManager.java:157)
        at com.owncloud.android.ui.fragment.FileDetailFragment.setFiletype(FileDetailFragment.java:445)
        at com.owncloud.android.ui.fragment.FileDetailFragment.updateFileDetails(FileDetailFragment.java:366)
[...]

PR #3589, opened 2022-03-24
Crash due to throw exception was fixed

fesave:
## Related Issues
App: #3571


abelgardep (approved):
LGTM

davcres (approved):
LGTM

jesmrec:
Crash not reproducible after checking with the files i was able to before the fix. 

Let's move this forward

Issue #3585, opened 2022-03-15
[BUG] Crash reported in Google Play - NPE in RefreshFolderOperation 


jesmrec:
```
java.lang.NullPointerException: 
  at com.owncloud.android.operations.RefreshFolderOperation.run (RefreshFolderOperation.java:128)
  at com.owncloud.android.lib.common.operations.RemoteOperation.runOperation (RemoteOperation.java:240)
  at com.owncloud.android.lib.common.operations.RemoteOperation.run (RemoteOperation.java:259)
  at java.lang.Thread.run (Thread.java:929)
```

<img width="1123" alt="Screenshot 2022-03-15 at 17 23 26" src="https://user-images.githubusercontent.com/14894746/158424548-5a43d945-1102-431d-a70f-fc22d6ced27f.png">

PR #3590, opened 2022-03-28
NPE in RefreshFolderOperation - Fixed

fesave:
## Related Issues
App: #3585 
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -122,7 +122,15 @@ public class RefreshFolderOperation extends SyncOperation<ArrayList<RemoteFile>>
         OwnCloudVersion serverVersion = null;
 
         // get 'fresh data' from the database
-        mLocalFolder = getStorageManager().getFileByPath(mLocalFolder.getRemotePath());
+        OCFile tempFile = getStorageManager().getFileByPath(mLocalFolder.getRemotePath());
+
+        if (tempFile != null) {
+            mLocalFolder = tempFile;
+        } else {
+            Timber.w("File with path: " + mLocalFolder.getRemotePath() + " and account " + mAccount.name
+                    + " could not be retrieved from database " + "for account : "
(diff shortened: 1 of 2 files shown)


owncloudApp/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -122,7 +122,13 @@ public RefreshFolderOperation(OCFile folder,
         OwnCloudVersion serverVersion = null;
 
         // get 'fresh data' from the database
-        mLocalFolder = getStorageManager().getFileByPath(mLocalFolder.getRemotePath());
+        OCFile tempFile = getStorageManager().getFileByPath(mLocalFolder.getRemotePath());
+
+        if (tempFile != null) {
+            mLocalFolder = tempFile;
+        } else {
+            Timber.e("Error, file not received because mLocalFolder is null");
```
abelgardep, line 130: Let's try to make the log a little bit more useful just in case we reproduce the issue in the future. What do u think? 
```suggestion
            Timber.w("File with path: " + mLocalFolder.getRemotePath() + " and account " +  mAccount.name + " could not be retrieved from database for account : " + getStorageManager().getAccount().name + " Let's try to synchronize it anyway");
```

abelgardep (changes requested):
Tiny suggestion here @fesave 👍

owncloudApp/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
fesave: Excellent! Thank u for the suggestion 😉

abelgardep (approved):
LGTM, Let's check if it fixes the issue

jesmrec:
any clue about how to reproduce this?

fesave:
> any clue about how to reproduce this?

Sorry, but we were unable to reproduce it 😢

jesmrec:
In that case, nothing to test from my side unless you recommend me something :)

We should move this forward and let the issue open till releasing in order to check if the reports are gone from Google Play.

PR #3598, opened 2022-04-05
Open deep links in app, items already discovered (1st round)

### LINKED ISSUE #3573: [FEATURE REQUEST] Open deeplinks in app, items in root folder
Private links generated by the server, must be opened inside the oC Android app.

This is a first approach with specific ACs:

- App matching the host to be opened when private link is clicked. Not needed to brand this. Static.
- Working in only one account mode
- If the link points to a file, the file will be downloaded/displayed
- If the link points to a folder, the folder content will be displayed
- If the file/folder is not reachable, app is opened and error is shown (toast/snackbar/...)
- If no accounts are attached in the app, app is opened and error is shown (toast/snackbar/...) (feasible?)

Additional checks:

- Lack of URL in the scheme (or *) are valid as well.

### TASKS

- [ ] Research (if needed)
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Design test plan
- [ ] QA

### PR

- App
- Library (if needed)

fesave:
## Related Issues
App: #3573 

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_2.21/3598-Universal%20Link%20I.md

Reports:

- [x] (1) Browsing folders not working https://github.com/owncloud/android/pull/3598#issuecomment-1095182438 [FIXED]
- [X] (2) Openability backwards after opening the link https://github.com/owncloud/android/pull/3598#issuecomment-1095278047 [WONT FIX] -> new arch
- [x] (3) Manifest errors https://github.com/owncloud/android/pull/3598#issuecomment-1101167054 [FIXED]
- [x] (4) Links with security https://github.com/owncloud/android/pull/3598#issuecomment-1101337301 [FIXED]
- [X] (5) Links with extra chunks do not work https://github.com/owncloud/android/pull/3598#issuecomment-1102218421 [FIXED]
- [x] (6) Crash when no accounts are attached https://github.com/owncloud/android/pull/3598#issuecomment-1112963889 [FIXED]


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -647,4 +647,8 @@
     <string name="security_enforced_first_option">Passcode lock</string>
     <string name="security_enforced_second_option">Pattern lock</string>
 
+    <!--Deep Links-->
+    <string name="no_account_configured">No account configured previously</string>
+    <string name="no_file_found">No file found</string>
```
abelgardep, line 652: I would more descriptive strings to advertise to the user the error in a more detailed way. Probably @jesmrec can help us with the error messages.

BTW, I would add the `deep_link` prefix to the string keys

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1647,6 +1657,36 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
...
+                if (OCFile != null) {
+                    manageItem(OCFile)
+                } else {
+                    showMessageInToast(getString(R.string.no_file_found))
+                }
+            }
+        }
+    }
+
+    private fun isFileDiscovered(uri: Uri?): OCFile? = storageManager.getFileByPrivateLink(uri.toString())
```
abelgardep, line 1674: `isSomethingSomehow()` functions are expected to return booleans. If we are returning a specific object, I would use a different name.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -225,6 +227,9 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
             createMinFragments()
         }
 
+        val dataIntent: Uri? = intent.data
+        manageDataIntent(dataIntent)
```
abelgardep, line 231: In `manageDataIntent` we have several checks to ensure that the URI received is not null that could be removed in an easy way.

```suggestion
        dataIntent?.let { manageDataIntent(dataIntent) }
```

BTW, `manageDataIntent` is way too uncertain. Something like `handleDeepLink` could be more accurate, what do u think?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -294,7 +299,7 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
     private fun initFragmentsWithFile() {
         if (account != null && file != null) {
             /// First fragment
-            listOfFilesFragment?.listDirectory(currentDir)
+            listOfFilesFragment?.listDirectory(file)
```
abelgardep, line 302: Is this a needed change? `currentDir` will use `file` if it's a folder anyway, right? Just curious

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1647,6 +1657,36 @@ class FileDisplayActivity : FileActivity(), FileFragment.ContainerActivity, OnEn
         manageOptionLockSelected(type)
     }
 
+    private fun manageDataIntent(uri: Uri?) {
+        if (uri != null && AccountUtils.getAccounts(applicationContext).isEmpty()) {
+            showMessageInToast(getString(R.string.no_account_configured))
+        } else if (uri != null && AccountUtils.getAccounts(applicationContext).size == 1) {
+            isFileDiscovered(uri).let { OCFile ->
```
abelgardep, line 1664: I would use lowerCamelCase for the name inside the `let` as we do with variable names

abelgardep (changes requested):
First iteration 💯  Some suggestions here @fesave

owncloudApp/src/main/res/values/setup.xml
```diff
@@ -116,4 +116,7 @@
     <!-- Apm settings -->
     <bool name="clear_cookies_on_validation">false</bool>
 
+    <!--Deep Links -->
+    <string name="url_link_host">*</string>
```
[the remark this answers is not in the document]
fesave: Fixed

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -70,6 +70,14 @@ fun Activity.showMessageInSnackbar(
...
 }
 
+fun Activity.showMessageInToast(
+    message: CharSequence,
+    duration: Int = Toast.LENGTH_LONG
+) {
+    Toast.makeText(this, message, duration).show()
+}
+
+
```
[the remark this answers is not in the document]
fesave: Fixed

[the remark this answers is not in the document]
fesave: Repeated?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -73,6 +73,8 @@
     public static final String EXTRA_ACCOUNT = "com.owncloud.android.ui.activity.ACCOUNT";
     public static final String EXTRA_FROM_NOTIFICATION =
             "com.owncloud.android.ui.activity.FROM_NOTIFICATION";
+    public static final String ALREADY_HANDLED_DEEP_LINK =
```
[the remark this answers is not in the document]
fesave: Fixed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
fesave: You're right! Fixed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
fesave: Fixed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
fesave: Fixed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
fesave: it's no needed!

abelgardep (approved):
LGTM!

jesmrec:
### (1)

1. Install app and add account
2. Browse inside any folder in root, in order to discover it.
3. Get the link of the folder browsed in 2.
4. Kill the app, so that it will be closed when link is clicked
5. Click on link over the folder

Current: Folder not opened, it is shown the root folder. Check the video, the link corresponds with the `Documents` folder. `Documents`appears in the top bar but the content is the root:

https://user-images.githubusercontent.com/14894746/162770848-8fc3b976-958a-483e-a64d-35f9dd534df5.mp4

Nexus 6P Android7
Huawei P20 Android 9
`214061ad`

jesmrec:
### (2)

Prerequisite:

1. Create a folder structure like /A/B/C
2. Inside C, upload an openable file (f. ex, an image), and a non-openable file (f. ex a PDF)

Two scenarios:

1. 
Open via link the image -> image is opened
Browse back with the arrow -> details view
Browse back from details view -> "C" folder content is displayed

2.
Open via link the pdf -> details view is opened
Browse back with the arrow -> root folder

that means, depending on the content's openability, browsing works in a different way. Is this expected somehow? Could if be fixed?

Nexus 6P Android7
`214061ad`

fesave:
> that means, depending on the content's operability, browsing works in a different way. Is this expected somehow? Could it be fixed?

I think that with the new architecture the current behavior can be improved, but for the moment it is the first version.

jesmrec:
### (3)

Not posible to build the branch after the latest changes. Some manifest error happens:

```
/Documents/android/owncloudApp/src/debug/AndroidManifest.xml:86:9-106:20 Error:
	android:exported needs to be explicitly specified for element <activity#com.owncloud.android.ui.activity.FileDisplayActivity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
```

BitRise build also fails with same error: https://app.bitrise.io/build/f4c6256e-f684-4d9f-a22c-467179da0ddf

jesmrec:
### (4)

1. Enable passcode/pattern
2. Open deeplink from other app

Current: First time, passcode/pattern is asked. If app is killed (no lock delay), next times the deeplink is opened, pattern/passcode is not asked.

Expected: Every time a link is opened, pattern/passcode is asked

Nexus 6P Android 7
`6b4b0303`

jesmrec:
(2) will need a new issue with `New Architecture` label

jesmrec:
### (5)

Checking the feature in some servers, i realised that the private link does not always is "plain". That means, it could contain some extra chunks. For example, in cloud.damken.com the private links are like `https://cloud.damken.com/index.php/f/28728865` with an extra `index.php`. The oC internal instance works in that way too.

With Android 7, the link is opened in the browser.  With Android 12, it can not be manually added as supported. 

I also tried adding `index.php` , `*`  and some combinations to `setup.xml` with no success in any of the Android versions. 

Is there anything we could do with that?

fesave:
> ### (3)
> Not posible to build the branch after the latest changes. Some manifest error happens:
> 
> ```
> /Documents/android/owncloudApp/src/debug/AndroidManifest.xml:86:9-106:20 Error:
> 	android:exported needs to be explicitly specified for element <activity#com.owncloud.android.ui.activity.FileDisplayActivity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
> ```
> 
> BitRise build also fails with same error: https://app.bitrise.io/build/f4c6256e-f684-4d9f-a22c-467179da0ddf

fixed :)

jesmrec:
Regarding (5). Problem comes from the `android:pathPrefix` setup. We should add the extra chunks. For the example above

`https://cloud.damken.com/index.php/f/28728865`

```
android:host="cloud.damken.com"
android:pathPrefix="/index.php/f/"
```
Checked, this setup works.

So, one option is creating a new parameter in `setup.xml` together with `deep_link_host` (something like  `deep_link_suffix`) in which those links containing subdomains or other chunks will add this part there:

If `deep_link_suffix` is empty ->  `android:pathPrefix="/f/"` as before
If `deep_link_suffix` is not empty -> `android:pathPrefix="/" + deep_link_suffix + /f/"`

(logic not completed, so the leading slashes should be handled in the same line since no ifs are allowed in Manifest files)

is this feasible @fesave?

fesave:
> Regarding (5). Problem comes from the `android:pathPrefix` setup. We should add the extra chunks. For the example above
> 
> `https://cloud.damken.com/index.php/f/28728865`
> 
> ```
> android:host="cloud.damken.com"
> android:pathPrefix="/index.php/f/"
> ```
> 
> Checked, this setup works.
> 
> So, one option is creating a new parameter in `setup.xml` together with `deep_link_host` (something like `deep_link_suffix`) in which those links containing subdomains or other chunks will add this part there:
> 
> If `deep_link_suffix` is empty -> `android:pathPrefix="/f/"` as before If `deep_link_suffix` is not empty -> `android:pathPrefix="/" + deep_link_suffix + /f/"`
> 
> (logic not completed, so the leading slashes should be handled in the same line since no ifs are allowed in Manifest files)
> 
> is this feasible @fesave?

AFAIK, the `AndroidManifest.xml` file is a configuration file where we can apply the basic settings of our application. Therefore, it is not possible to add logic to control pathPrefix.

I think the best option is to configure a new field inside our setup.xml file and have its value come from the server as we do with other fields.

We would have to consider the possibility that the value could be of two types:

- `/f/` -> default.
- `/deepLinkSuffix/f/` -> in case the value is filled in the server.

Are you in charge of creating the issue with the detailed solution and the name of the new field or do you prefer me to do it in this branch? @jesmrec

jesmrec:
> Are you in charge of creating the issue with the detailed solution and the name of the new field or do you prefer me to do it in this branch? @jesmrec

it's ok to do in this branch.

fesave:
> ### (4)
> 1. Enable passcode/pattern
> 2. Open deeplink from other app
> 
> Current: First time, passcode/pattern is asked. If app is killed (no lock delay), next times the deeplink is opened, pattern/passcode is not asked.
> 
> Expected: Every time a link is opened, pattern/passcode is asked
> 
> Nexus 6P Android 7 `6b4b0303`

First of all, I have only reproduced this error with image files and their preview. I have coded a solution to avoid it, but, we have lost the automatic preview of the images. I think it is not bad at all because we can unify all the previews with the three action buttons that we should add after the design discussion [here](https://github.com/owncloud/android/issues/3573#issuecomment-1088580308)

jesmrec:
That means, that deeplinks will point always to `Details` , no matter if the file is preview-able inside the app or not, right @fesave ? 

is that OK @michaelstingl? should we create issue or develop [this](https://github.com/owncloud/android/issues/3573#issuecomment-1088580308) to mitigate such effect?

fesave:
> That means, that deeplinks will point always to `Details` , no matter if the file is preview-able inside the app or not, right @fesave ?
> 
> is that OK @michaelstingl? should we create issue or develop [this](https://github.com/owncloud/android/issues/3573#issuecomment-1088580308) to mitigate such effect?

Yes, with the current state, all files (except videos) show their preview screen if the user accesses them through a deep link.

michaelstingl:
> > is that OK @michaelstingl? should we create issue or develop [this](https://github.com/owncloud/android/issues/3573#issuecomment-1088580308) to mitigate such effect?
> 
> Yes, with the current state, all files (except videos) show their preview screen if the user accesses them through a deep link.

What do you mean with "preview screen"? Images, Audio & Video should be shown or played if the app supports it. Only unsupported formats should open `Details` for this file. https://github.com/owncloud/android/issues/3573#issuecomment-1088580308 isn't a solution, this would open the image in a 3rd party app, not in the ownCloud app.

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
```diff
@@ -329,11 +332,33 @@ public void onClick(View v) {
...
+        } else if (PreviewAudioFragment.canBePreviewed(file)) {
+            // media preview
+            ((FileDisplayActivity) mContainerActivity).startAudioPreview(file, 0);
+            mContainerActivity.getFileOperationsHelper().syncFile(file);
+
+        } else {
+            // sync file content, then open with external apps
+            ((FileDisplayActivity) mContainerActivity).startSyncThenOpen(file);
+        }
+    }
```
abelgardep, line 360: I would say I have seen this logic before. I'll refactor this within the new arch 🏗️

fesave: Yeah, it is quite similar, but in this case, the video check is not necessary, be careful in the future 🤞🏻

jesmrec:
### (6)

1. Install the app, no accounts added
2. Click on any deeplink openable by the app

Current: crash happens. This is the stacktrace:

```2022-04-29 09:40:12.377 31539-31539/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.owncloud.android.debug, PID: 31539
    java.lang.RuntimeException: Unable to resume activity {com.owncloud.android.debug/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.datamodel.OCFile com.owncloud.android.datamodel.FileDataStorageManager.getFileByPrivateLink(java.lang.String)' on a null object reference
        at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3430)
        at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3470)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2733)
        at android.app.ActivityThread.-wrap12(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1478)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6121)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.datamodel.OCFile com.owncloud.android.datamodel.FileDataStorageManager.getFileByPrivateLink(java.lang.String)' on a null object reference
        at com.owncloud.android.ui.activity.FileDisplayActivity.getFileDiscovered(FileDisplayActivity.kt:1676)
        at com.owncloud.android.ui.activity.FileDisplayActivity.refreshListOfFilesFragment(FileDisplayActivity.kt:428)
        at com.owncloud.android.ui.activity.FileDisplayActivity.onResume(FileDisplayActivity.kt:668)
        at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1269)
        at android.app.Activity.performResume(Activity.java:6786)```

Expected: error message or similar

Pixel5 Android12, Nexus6P Android7
`08b091e6`

fesave:
> ### (6)
> 1. Install the app, no accounts added
> 2. Click on any deeplink openable by the app
> 
> Current: crash happens. This is the stacktrace:
> 
> ```
> 2022-04-29 09:40:12.377 31539-31539/? E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: com.owncloud.android.debug, PID: 31539
>     java.lang.RuntimeException: Unable to resume activity {com.owncloud.android.debug/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.datamodel.OCFile com.owncloud.android.datamodel.FileDataStorageManager.getFileByPrivateLink(java.lang.String)' on a null object reference
>         at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3430)
>         at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3470)
>         at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2733)
>         at android.app.ActivityThread.-wrap12(ActivityThread.java)
>         at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1478)
>         at android.os.Handler.dispatchMessage(Handler.java:102)
>         at android.os.Looper.loop(Looper.java:154)
>         at android.app.ActivityThread.main(ActivityThread.java:6121)
>         at java.lang.reflect.Method.invoke(Native Method)
>         at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
>         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
>      Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.datamodel.OCFile com.owncloud.android.datamodel.FileDataStorageManager.getFileByPrivateLink(java.lang.String)' on a null object reference
>         at com.owncloud.android.ui.activity.FileDisplayActivity.getFileDiscovered(FileDisplayActivity.kt:1676)
>         at com.owncloud.android.ui.activity.FileDisplayActivity.refreshListOfFilesFragment(FileDisplayActivity.kt:428)
>         at com.owncloud.android.ui.activity.FileDisplayActivity.onResume(FileDisplayActivity.kt:668)
>         at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1269)
>         at android.app.Activity.performResume(Activity.java:6786)
> ```
> 
> Expected: error message or similar
> 
> Pixel5 Android12, Nexus6P Android7 `08b091e6`

Fixed

jesmrec:
Approved. 

At this point:

- Deeplinks opened only over items that are already discovered
- Deeplinks opened only in single account mode
- Android12:  servers with name and URL must be branded and validated in device's `Settings` (not valid a  `*` in setup.xml)
- If server URL contains some extra chunk (subdomain or similar), needs to be branded as well
- Clicking on thumbnail inside `Details` will download and open the pointed file/folder

**Known restriction:**

After opening an item, navigation is not totally smooth. Browsing back from `Details` or the preview itself goes directly to root folder, not matter the location of the opened item (it could be placed very deep in the folder structure). This is a tricky one, that we expect to mitigate in the new architecture.

jesmrec:
Closing. Implemented in https://github.com/owncloud/android/pull/4191

PR #3628, opened 2022-04-13
rename verify_licnese_script back to check_script

theScrabi:
`verify_license_script.sh` is the wrong name for what the script does. This name implies that the license is checkt, where in fact more than this happens. Thus because of good coding we should go back to a more general description that could contain license verification as well as linting, or whatever is going to be added to in the future.
Therefore we should go back to `check_script`.

Library: https://github.com/owncloud/android-library/pull/479

### DIFF
--- a/check_code_script.sh
+++ b/check_code_script.sh
@@ -23,4 +23,4 @@ done
 
 echo "========= Check Library =========="
 cd owncloud-android-library
-./verify_licenses_script.sh
+./check_code_script.sh


jesmrec:
Maybe, we need something in the middle. 

If `verify_license_script.sh` is not correct, a new more accurate one should be set. IMHO `check_script` is too generical. What are you checking? For anyone that clones/forks the repo, this is not descriptive. 

As you described, not only license is checked. Also lint, and other stuff could be included later, right? would a name like `sanity_script.sh` or `repo_check_script.sh` fit better? what do you think?

CC @abelgardep @fesave @davcres

theScrabi:
How about `code_check_script.sh` then?

jesmrec:
> How about `code_check_script.sh` then?

works for me 👍

davcres (approved):
LGTM

PR #3631, opened 2022-04-20
[FIX] justified text library removed

davcres:


fesave (approved):
LGTM

jesmrec:
Checked in the release notes activity. That's correct... not much to test here

Issue #1414, opened 2016-01-09
Multiple files: improve copy/move dialog


elpraga:
They appear exactly the same and there is no way of knowing which is which as they appear only as an icon without any text

AndyScherzinger:
@elpraga can you check with the latest release (today) 20150109 since I have replaced the icons there and copy and move to be distinguishable :)

elpraga:
I can confirm. @AndyScherzinger in 20150109 the icons can be distinguished.

Do You think it would be nice to add "copy to" and "move to" respectively to the select folder dialog? Maybe just left from the three dots menu on the top right? This is just an idea though, if You feel like it is not good idea or it would be too much work to implement it, we can close the issue

tobiasKaminsky:
For me there are no icons? The list, which is opened after clicking the three dots, shows only text?

AndyScherzinger:
@tobiasKaminsky there should be ;) - what @elpraga means are the following icons you can see in the screenshots in  #1411

AndyScherzinger:
@elpraga can you provide some more details on the select folder comment? From what I understand you want to move/copy files from within a folder chooser. If so I would recommend against it since it mixed we the actions one would expect in the context of choosing. So imho when selecting a folder, selecting a folder should be the only action (except for creating a folder though)

elpraga:
I will as soon as I get home in a little bit.  The chance in icons you made
in 20160109 was great.
My last suggestion was to add an informative text in the choose folder
dialog that appears after clicking on respective icon.

If you click on the copy icon for example,  the dialog that appears would
say "copy to" in the top part,  just informing what are we selecting the
folder for.

But as I mentioned before,  I don't see it crucially important.  I just
think It would be nice if it doesn't require almost any extra work.

AndyScherzinger:
Ah, now I get what you mean, sorry! This shouldn't be to much work to be implemented imho and yes, this would totally make sense. But that is also just my opinion. @tobiasKaminsky @jancborchardt what do you think? I vote yes on implementing this. copy/move in the title to show this context information.

tobiasKaminsky:
This is a good idea.
cc @jancborchardt for the exact wording

jancborchardt:
Yeah, good call. »Copy file/foldername to« and »Move file/foldername to« in the header woulf be good wording.

Additionally if possible, it's always good to show the thumbnail of the file/folder to the left in that modal header. And have a relevant label for the button, like »Copy« / »Move«.

[...]

PR #3640, opened 2022-04-27
Improved copy/move dialog

fesave:
## Related Issues
App: #1414

_____

## Screenshots
Old Version / Select upload path  | New Copy Version | New Move Version
------------- | -------------  | -------------
  |  |  |

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
@@ -176,19 +176,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     private void setActionButtonText() {
         PickerMode actionButton = (PickerMode) getIntent().getSerializableExtra(EXTRA_PICKER_OPTION);
         Button chooseButton = findViewById(R.id.folder_picker_btn_choose);
-
-        if (actionButton != null) {
-            switch (actionButton) {
-                case MOVE:
-                    chooseButton.setText(getString(actionButton.getButtonString(PickerMode.MOVE)));
-                    break;
-                case COPY:
-                    chooseButton.setText(getString(actionButton.getButtonString(PickerMode.COPY)));
-                    break;
-                default:
-                    chooseButton.setText(getString(actionButton.getButtonString(PickerMode.CAMERA_FOLDER)));
-            }
-        }
+        chooseButton.setText(getString(actionButton.getButtonString()));
     }
 
     protected OCFileListFragment getListOfFilesFragment() {
@@ -536,8 +524,8 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     public enum PickerMode {
         MOVE, COPY, CAMERA_FOLDER;
 
-        public Integer getButtonString(PickerMode pickerMode) {
-            switch (pickerMode) {
+        public Integer getButtonString() {
+            switch (this) {
                 case MOVE:
                     return R.string.folder_picker_move_here_button_text;
                 case COPY:
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
@@ -69,7 +69,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
 
     private static final String TAG_LIST_OF_FOLDERS = "LIST_OF_FOLDERS";
 
-    private static final String ACTION_BUTTON = "ACTION_BUTTON";
+    public static final String EXTRA_PICKER_OPTION = "EXTRA_PICKER_OPTION";
 
     private LocalBroadcastManager mLocalBroadcastManager;
     private SyncBroadcastReceiver mSyncBroadcastReceiver;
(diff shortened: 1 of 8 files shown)


davcres (approved):
LGTM

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -509,4 +530,8 @@ private void refreshList(boolean ignoreETag) {
             }
         }
     }
+
+    public enum ACTION_BUTTON {
```
abelgardep, line 534: Adding an enum to do this, could make sense. But don't forget that enums are classes, so we should follow the convention and not use TRAILING_COMMA.

BTW, ActionButton is kind of generic, I would choose a more descriptive name, for example, PickerMode or something like that. Up to you 👍

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -115,6 +115,8 @@ public class OCFileListFragment extends ExtendedListFragment implements
 
     private static final String GRID_IS_PREFERED_PREFERENCE = "gridIsPrefered";
 
+    private static final String ACTION_BUTTON = "ACTION_BUTTON";
+
```
abelgardep, line 119: Since this is an EXTRA from the FolderPicker, and it already has the key there, we should not create the same key in any other file. Don't you think?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -69,6 +69,8 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
 
     private static final String TAG_LIST_OF_FOLDERS = "LIST_OF_FOLDERS";
 
+    private static final String ACTION_BUTTON = "ACTION_BUTTON";
```
abelgardep, line 72: Let's add the EXTRA prefix, make it public(so other classes can use it) and rename it according to the enum class name 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -509,4 +530,8 @@ private void refreshList(boolean ignoreETag) {
             }
         }
     }
+
+    public enum ACTION_BUTTON {
+        MOVE, COPY
```
abelgardep, line 535: Consider adding all the potential uses, not only move and copy. By the way, could be a good idea to create a toString() function to retrieve the actual string id directly

abelgardep (changes requested):
Dome suggenstins her e@fesave

owncloudApp/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
fesave: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
fesave: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
fesave: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -509,4 +532,19 @@ private void refreshList(boolean ignoreETag) {
             }
         }
     }
+
+    public enum PickerMode {
+        MOVE, COPY, CAMERA_FOLDER;
+
+        public Integer getButtonString(PickerMode pickerMode) {
+            switch (pickerMode) {
```
abelgardep, line 540: This is a class function. You don't need to receive the picker mode from parameters, because you already know which picker mode is when u call it. 

Try next code and let me know what happens 👍  (This will simplify setActionButtonText code too)
```suggestion
        public Integer getButtonString() {
            switch (this) {
```

abelgardep (changes requested):
Last change and it will be ready @fesave

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
fesave: OMG 😱  Marvelous!! Thanks.

abelgardep (approved):
LGTM

jesmrec:
Changes in the move/copy pickers are correct. Also, the button label was correctly adapted to the picture camera uploads' picker.

But, picture camera uploads' picker crashes. Check the steps out:

1. Open `Settings`
2. Open `Auto upload videos`
3. Enable `Video Uploads`
4. Open `Video upload path`

Current: app crashes
Expected: no crashh

Nexus6P Android7

Could you take a look @fesave 
Pixel5 Android12

fesave:
> 

Fixed :)

jesmrec:
Approved!!

Issue #3637, opened 2022-04-21
[FEATURE REQUEST] Prevent taking screenshots in Dialogs (Android 12)


davcres:
In #3615 the screenshots was prevented, but in Android 12 if you take a screenshot of a dialog, it will be taken.

The idea is to prevent the screenshots also in dialogs for Android 12. In Android 11 or below there is no problem, it doesn´t allow to take screenshots if the flag `allow_screenshots`is set to false.

||| 
|:-|:-:|
|![Screenshot_20220412-170534](https://user-images.githubusercontent.com/14894746/162994686-9699b158-bd6c-474e-89e0-150ae5321c75.png)|![Screenshot_20220412-170423](https://user-images.githubusercontent.com/14894746/162994713-62c9bae1-76fe-4ccc-a014-6d8450cfdc74.png)|

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

jesmrec:
Could be a solution to prevent this dialog specifically? as well as the passcode, pattern or login view? Only for Android 12

davcres:
`dialog.window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)` works to prevent screenshots in the dialogs.
There are too many, maybe we can do a dialog to inherit from it with the option.

PR #3642, opened 2022-04-27
[FEATURE] Prevent taking screenshots in Dialogs (Android 12)

davcres:
## Related Issues
App: #3637

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/DialogExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/DialogExt.kt
@@ -1,0 +1,31 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author David Crespo Ríos
+ * Copyright (C) 2022 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.extensions
+
+import android.app.Dialog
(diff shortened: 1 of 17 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/dialog/BaseDialogFragment.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.owncloud.android.R
+
+open class BaseDialogFragment : DialogFragment() {
+
+    fun avoidScreenshotsIfNeeded(dialog: Dialog) {
+        if (!BuildConfig.DEBUG && context?.resources?.getBoolean(R.bool.allow_screenshots) == false) {
+            dialog.window?.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
+        }
+    }
+}
```
abelgardep, line 35: `Base` classes are sometimes a good idea, not always. But in this case, is absolutely not needed. You could create a random class and that function would work since it does not perform anything related to itself nor overrides anything. It receives the dialog from the parameter and applies that flag.

You could instead create a Kotlin extension:
```
fun Dialog.avoidScreenshotsIfNeeded() {
    if (!BuildConfig.DEBUG && context.resources?.getBoolean(R.bool.allow_screenshots) == false) {
        window?.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
    }
}
```

And call it like this `dialog.avoidScreenshotsIfNeeded()` with no need of base classes.

What do you think?

abelgardep (changes requested):
Some changes here @davcres 👍

abelgardep (approved):
👍

jesmrec:
These are the dialogs/views not screenshot-prevented yet in Android12:

- Upload menu:

Screenshot 2022-05-05 at 09 02 50

- Auto uploads warning

Screenshot 2022-05-05 at 09 04 22

- Auto uploads behaviour

Screenshot 2022-05-05 at 09 05 01

- Auto uploads disabled

Screenshot 2022-05-05 at 09 05 35

- Prevent touches warning

Screenshot 2022-05-05 at 09 06 39

- Share sheet

Screenshot 2022-05-05 at 09 07 21

any of them could be also prevented or is this a dead way? 

Pixel5 Android12
`d35edb53`

davcres:
> any of them could be also prevented or is this a dead way?

@jesmrec I think they all can be prevented but the share sheet. I´m going to it

jesmrec:
Approved.

Issue #1419, opened 2016-01-10
User list for sharing contains signed in user


cwmke:
Using fdroid's owncloud beta20160109 on a Nexus 6p with PureNexus rom, I perform the following steps,
1. Long press on a folder I would like to share with another user.
2. Press the share button in the top bar.
3. Click 'Add User Or Group'.
4. Type in user's name.
5. List pops up with all users including my own.

This isn't very major, but it seems like (from what I can tell) it's unnecessary.

tobiasKaminsky:
Thank you for this report.
CC to responsible devs: @masensio @davivel

davivel:
Yes, the web UI filters it out, we could do the same.

PR #3643, opened 2022-04-27
Removed signed in user from the list of users for sharing

fesave:
## Related Issues
App: #1419 

_____

### DIFF
--- a/changelog/unreleased/3643
+++ b/changelog/unreleased/3643
@@ -1,4 +1,4 @@
-Bugfix: Fixed a bug in the list of users to be shared
+Bugfix: Prevented signed in user in the list of users to be shared
 
 Previously, user list for sharing contains signed in user,
 now this user is omitted to avoid errors.
--- a/changelog/unreleased/3643
+++ b/changelog/unreleased/3643
@@ -1,0 +1,7 @@
+Bugfix: Fixed a bug in the list of users to be shared
+
+Previously, user list for sharing contains signed in user,
+now this user is omitted to avoid errors.
+
+https://github.com/owncloud/android/issues/1419
+https://github.com/owncloud/android/pull/3643
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/providers/sharing/UsersAndGroupsSearchProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/providers/sharing/UsersAndGroupsSearchProvider.kt
@@ -24,6 +24,7 @@
 
 package com.owncloud.android.presentation.providers.sharing
 
+import android.accounts.AccountManager
 import android.app.SearchManager
 import android.content.ContentProvider
 import android.content.ContentValues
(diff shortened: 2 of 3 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/providers/sharing/UsersAndGroupsSearchProvider.kt
```diff
@@ -180,6 +179,9 @@ class UsersAndGroupsSearchProvider : ContentProvider() {
             try {
                 while (namesIt.hasNext()) {
                     item = namesIt.next()
+                    if (item.label == AccountUtils.getUsernameOfAccount(account.name) && item.shareType == ShareType.USER) {
+                        item = namesIt.next()
```
abelgardep, line 183: Please, don't do this.

If we use an iterator to iterate in a while loop, the condition that checks if we need to exit the loop is `hasNext()`. So, we use`next()` only when we are sure it `hasNext()`. If we want to proceed with the next step, [we use `continue`](https://kotlinlang.org/docs/returns.html)

abelgardep, line 183: Please, don't do this.

If we use an iterator to iterate in a while loop, the condition that checks if we need to exit the loop is `hasNext()`. So, we use`next()` only when we are sure it `hasNext()`. If we want to proceed with the next step, [we use `continue`](https://kotlinlang.org/docs/returns.html)

abelgardep (changes requested):
Some changes here @fesave

changelog/unreleased/3643
```diff
@@ -0,0 +1,7 @@
+Bugfix: Fixed a bug in the list of users to be shared
```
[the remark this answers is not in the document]
fesave: Changed

owncloudApp/src/main/java/com/owncloud/android/presentation/providers/sharing/UsersAndGroupsSearchProvider.kt
```diff
@@ -180,6 +179,9 @@ class UsersAndGroupsSearchProvider : ContentProvider() {
             try {
                 while (namesIt.hasNext()) {
                     item = namesIt.next()
+                    if (item.label == AccountUtils.getUsernameOfAccount(account.name) && item.shareType == ShareType.USER) {
```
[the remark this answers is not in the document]
fesave: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/providers/sharing/UsersAndGroupsSearchProvider.kt
fesave: Understood! Thank u a lot.

jesmrec:
- If username == displayname -> ✅   . Username is not in the list, not selectable

- If username != displayname  -> ❌   . It's selectable. For example, this user:

Screenshot 2022-05-04 at 13 32 55

searching:

Screenshot 2022-05-04 at 13 32 45

it should not appear either.

Could you take a look, @fesave ?

fesave:
> * If username == displayname -> ✅   . Username is not in the list, not selectable
> * If username != displayname  -> ❌   . It's selectable. For example, this user:
> 
> Screenshot 2022-05-04 at 13 32 55
> 
> searching:
> 
> Screenshot 2022-05-04 at 13 32 45
> 
> it should not appear either.
> 
> Could you take a look, @fesave ?

Good catch! Fixed :)

jesmrec:
Problem with the display name is fixed. So, everything done.

Approved

PR #3678, opened 2022-05-30
Add new items to release notes for 2.21

jesmrec:
To be translated in time.

### DIFF
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -673,9 +673,9 @@
     <string name="release_notes_proceed">Proceed</string>
     <string name="release_notes_icon">Release note icon</string>
 
-    <!-- Release Notes 2.21-beta.1 -->
+    <!-- Release Notes 2.21 -->
     <string name="release_notes_title_1">Release Notes</string>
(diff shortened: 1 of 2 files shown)


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -692,4 +692,10 @@
     <string name="release_notes_title_6">Hidden files</string>
     <string name="release_notes_subtitle_6">Added a new option in settings to decide to show or not hidden files.</string>
 
+    <string name="release_notes_title_7">Sharing from inside</string>
+    <string name="release_notes_subtitle_7">Added the sharing option to folders from the list of its contents</string>
+
+    <string name="release_notes_title_8">Bugfixing</string>
+    <string name="release_notes_subtitle_8">Some existing bugs were fixed</string>
```
abelgardep, line 699: Not 100% sure about this one, but ok

abelgardep (approved):

owncloudApp/src/main/res/values/strings.xml
jesmrec: Well, not an specific one like the others. But, fixes are also part of the release.

Issue #3677, opened 2022-05-27
[BUG] NPE in Google Play (Glide related)


jesmrec:
```
com.bumptech.glide.load.engine.CallbackException: 
  at com.bumptech.glide.load.engine.EngineJob.callCallbackOnResourceReady (EngineJob.java:161)
  at com.bumptech.glide.load.engine.EngineJob$CallResourceReady.run (EngineJob.java:428)
  at android.os.Handler.handleCallback (Handler.java:888)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at android.os.Looper.loop (Looper.java:213)
  at android.app.ActivityThread.main (ActivityThread.java:8178)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:513)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1101)
Caused by: java.lang.NullPointerException: 
  at com.owncloud.android.ui.preview.PreviewImageFragment.getBinding (PreviewImageFragment.kt:84)
  at com.owncloud.android.ui.preview.PreviewImageFragment.access$getBinding (PreviewImageFragment.kt:76)
  at com.owncloud.android.ui.preview.PreviewImageFragment$loadAndShowImage$1.onResourceReady (PreviewImageFragment.kt:337)
  at com.owncloud.android.ui.preview.PreviewImageFragment$loadAndShowImage$1.onResourceReady (PreviewImageFragment.kt:323)
  at com.bumptech.glide.request.SingleRequest.onResourceReady (SingleRequest.java:630)
  at com.bumptech.glide.request.SingleRequest.onResourceReady (SingleRequest.java:578)
  at com.bumptech.glide.load.engine.EngineJob.callCallbackOnResourceReady (EngineJob.java:159)
```

![Screenshot 2022-05-30 at 09 33 13](https://user-images.githubusercontent.com/14894746/170940919-580544df-44f0-49e9-987a-83427eba4b72.png)

abelgardep:
I was able to reproduce it locally:

1. Browse to a folder with lots of images
2. Preview an image
3. Swipe(to the next image) pretty fast till the app crashes

Pixel2 - Android 11

PR #3679, opened 2022-05-30
[Fix] Crash when loading files

abelgardep:
## Related Issues
App: Fix https://github.com/owncloud/android/issues/3677

Library PR (if needed):
_____


jesmrec:
Not able to reproduce this with the given steps. Let's move this forward and check if the crash is gone from Google Console.

jesmrec:
Not able to reproduce this with the given steps. Let's move this forward and check if the crash is gone from Google Console.

PR #3680, opened 2022-05-31
[UI IMPROVEMENT] Share screen

JuancaG05:
Some improvements have been applied to the Share screen. This is because I'm working personally on migrating the Share screen from XML to Jetpack Compose, which applies many accessibility guidelines by default, and I discovered that the original screen doesn't, so I think it's a good change (and it looks more attractive to me).

Here we can see a comparative:

| Old XML screen | Compose screen | New XML screen |
|:-:|:-:|:-:|
|  |  |  |

The main changes are the space around clickable icons, since accessibility guidelines state that a clickable icon should have a size of at least 48dpx48dp. Also, some tiny improvements in separators between the list items (previously they were duplicated unnecessarily) 
_____


jesmrec:
A new entry [here](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/presentation/viewmodels/releasenotes/ReleaseNotesViewModel.kt) would be appreciated as well.

Thanks a lot for the contribution!.

abelgardep (approved):
Thanks for the contribution @JuancaG05

PR #3690, opened 2022-06-07
New calens type for enterprise issues

abelgardep:
Let's check if we could add a new type for enterprise issues specifically so we can filter them easier.
_____


abelgardep:
Types are fixed, no way to customize them https://github.com/restic/calens/issues/4

PR #3711, opened 2022-07-06
Retrieve shares within the propfind

abelgardep:
Add a new property to the propfind. So that, we can get if the files in a folder are shared directly with just one request.
Previously, we needed a propfind and then a request to the shares api to retrieve the shares of the folder.

We save 1 network request and 2 requests to the database.

Library PR: https://github.com/owncloud/android-library/pull/496
_____


JuancaG05 (approved):
👍

jesmrec:
QA checks

Private link

- [ ] Private link created in app
- [ ] Private link created in other client
- [ ] Private link removed in app
- [ ] Private link removed in other client

Public link

- [ ] Public link created in app
- [ ] Public link created in other client
- [ ] Public link removed in app
- [ ] Public link removed in other client

Both 

- [ ] Public link & Private link created in app
- [ ] Public link & Private link created in other client
- [ ] Public link & Private link created + Public link removed -> only private link icon
- [ ] Public link & Private link created + Private link removed -> only public link icon
- [ ] Public link & Private link removed in app
- [ ] Public link & Private link removed in other client

jesmrec:
Approved on my side. Very good enhancement!

Issue #3707, opened 2022-07-04
[New arch]  Folder Synchronization


jesmrec:
Sync the folder's content recursively.

Base case: click on the circle arrow over a folder.
Other case: If file is downloaded/av,offline (also folder), a sync is triggered to assure local changes are pushed to the server.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App #3714 
- Library (if needed)

abelgardep:
Will be tested within https://github.com/owncloud/android/issues/2818

PR #3714, opened 2022-07-12
[New arch] Sync folder recursively

abelgardep:
## Related Issues
Implements #3707 
_____


abelgardep:
### Current state
Syncs recursively all content when clicking on the circle arrow over a folder.
Refresh folder syncs only already downloaded files within this folder, not recursively

### Known issues
1. Folder synchronization can't be cancelled at the moment. Uploads and Downloads are enqueued but can't be cancelled all at the same time
2. Refresh folder when picking a folder (move/copy/camera uploads path) triggers syncs for already downloaded files in the folder (and probably for av offline files in the future). Not sure if this is an issue, but it may takes longer than expected. Just a refresh could do the work when picking a folder

owncloudApp/build.gradle
```diff
@@ -93,7 +93,6 @@ dependencies {
 
     debugImplementation "androidx.fragment:fragment-testing:1.4.1"
     debugImplementation 'com.facebook.stetho:stetho:1.6.0'
-    debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.9.1'
```
JuancaG05, line 96: Was this intentional? 🤨 It also bothers me sometimes, but I don't know if you wanted to remove it definitely 😀

JuancaG05 (changes requested):
Nice improvement @abelgardep! 💯 Much better to send a String (`accountName`) than the whole account when it's not necessary 😀
Just a tiny comment about something you removed, well done!

JuancaG05 (approved):
LGTM 👍

Issue #3246, opened 2021-05-19
[New arch] Av. Offline


abelgardep:
Synchronization epic: https://github.com/owncloud/android/issues/2818

AC (To discuss):

- Av. offline means downloaded & synced
- Av. offline items will be marked to distinguish form the others
- Av. offline items will be filtered in the proper shortcut
- Items marked as Av. Offline will be synced automatically to keep them updated and available without network connection.
- All the content of a folder (files/subfolders) that is available offline will be available offline as well. In case of subfolders, all the content will be av. offline in a recursive way.
- The content of an av. offline folder will be av. offline till the folder loses the av. offline condition. That means, anything inside an av. offline folder will NOT have the option "unavailable offline"
- If an av. offline item is copied/moved, it keeps the av. offline condition in the new location unless it is av. offline because its parents are av. offline.
- If an item is copied/moved inside an av. offline folder, it becomes av. offline
- If an item inside an av. offline folder is copied/moved to any other location (not. av. offline), it loses the av. offline condition
- If an av. offline file is edited externally, in max 15 mins will be pushed to the server without opening the app  (Scoped Storage makes this one?, what happens with "open with")
- If any item is set as av.offline with no connection, it gets the av. offline condition and will be downloaded/synced when connection is recovered
- It is not posible to delete locally an av.offline item.

**REPLACE SERVICE BY WORKMANAGER**

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App #3715 
- Library (if needed)

abelgardep:
## Periodic worker.
⏲️ Interval: 15 Minutes
🛑 Constraints: NetworkType.CONNECTED and BatteryNotLow

Enqueued the first time the user skips settings. Probably should be done on installment or when setting a file as av. offline, not sure.

PR #3715, opened 2022-07-14
[New arch] Av. Offline

abelgardep:
## Related Issues
App: #3246 

_____


owncloudData/src/main/java/com/owncloud/android/data/availableoffline/AvailableOfflineRepositoryImpl.kt
```diff
@@ -0,0 +1,37 @@
...
+class AvailableOfflineRepositoryImpl(
+    val fileRepository: FileRepository,
+) : AvailableOfflineRepository {
+    override fun setFileAsAvailableOffline(fileToSetAsAvailableOffline: OCFile) {
+        fileRepository.updateFileWithNewAvailableOfflineStatus(fileToSetAsAvailableOffline, AvailableOfflineStatus.AVAILABLE_OFFLINE)
+    }
+
+    override fun unsetFileAsAvailableOffline(fileToSetAsAvailableOffline: OCFile) {
+        fileRepository.updateFileWithNewAvailableOfflineStatus(fileToSetAsAvailableOffline, AvailableOfflineStatus.NOT_AVAILABLE_OFFLINE)
+    }
```
JuancaG05, line 35: Is it correct that these methods belonging to a repository invoke methods from other repositories? 🤔. If that's the case for every method, maybe we can instance a `FileRepository` in the usecases that need this operations instead of adding boilerplate code like this. What do you think?

owncloudData/src/main/java/com/owncloud/android/data/files/datasources/LocalFileDataSource.kt
```diff
@@ -21,10 +21,11 @@
 package com.owncloud.android.data.files.datasources
 
 import androidx.lifecycle.LiveData
+import com.owncloud.android.domain.availableoffline.model.AvailableOfflineStatus
 import com.owncloud.android.domain.files.model.OCFile
 
 interface LocalFileDataSource {
-    fun copyFile(sourceFile: OCFile, targetFile: OCFile, finalRemotePath: String, remoteId: String)
+    fun copyFile(sourceFile: OCFile, targetFolder: OCFile, finalRemotePath: String, remoteId: String)
```
JuancaG05, line 28: In the second parameter of this method, do we specify the generated file that will result from the copy or the destination folder of the copy? I see you updated the name but the implementation continues being the same. If it is the second option of the ones previously mentioned, GREAT!! 🏅

owncloudData/src/main/java/com/owncloud/android/data/files/datasources/implementation/OCLocalFileDataSource.kt
```diff
@@ -134,12 +141,16 @@ class OCLocalFileDataSource(
     override fun renameFile(fileToRename: OCFile, finalRemotePath: String, finalStoragePath: String) {
         fileDao.moveFile(
             sourceFile = fileToRename.toEntity(),
-            targetFile = fileDao.getFileById(fileToRename.parentId!!)!!,
+            targetFolder = fileDao.getFileById(fileToRename.parentId!!)!!,
```
JuancaG05, line 144: *The shouting plugin when you write a `!!` didn't like this line 💔

owncloudDomain/src/main/java/com/owncloud/android/domain/availableoffline/AvailableOfflineRepository.kt
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.domain.availableoffline
+
+import com.owncloud.android.domain.files.model.OCFile
+
+interface AvailableOfflineRepository {
+    fun setFileAsAvailableOffline(fileToSetAsAvailableOffline: OCFile)
+    fun unsetFileAsAvailableOffline(fileToSetAsAvailableOffline: OCFile)
```
JuancaG05, line 25: I don't know if it's suitable that both methods have a parameter with the same name which indicates a same aim, when their objective is precisely the opposite. Maybe changing the name of the second one to `fileToUnsetAsAvailableOffline`?

owncloudDomain/src/main/java/com/owncloud/android/domain/availableoffline/usecases/GetFilesAvailableOfflineFromEveryAccountUseCase.kt
```diff
@@ -0,0 +1,29 @@
+/*
+ * ownCloud Android client application
+ *
```
JuancaG05, line 3: Who did this? 👻

JuancaG05 (changes requested):
Great work @abelgardep! 💯 
Some comments over here!

owncloudApp/src/main/java/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -360,32 +361,30 @@ private boolean anyFileDown() {
...
-//            if (file.getAvailableOfflineStatus() == OCFile.AvailableOfflineStatus.NOT_AVAILABLE_OFFLINE) {
-//                return true;
-//            }
-//        }
+        for (OCFile file : mFiles) {
+            if (file.getAvailableOfflineStatus() == AvailableOfflineStatus.NOT_AVAILABLE_OFFLINE) {
+                return true;
+            }
+        }
         return false;
```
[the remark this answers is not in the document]
abelgardep: The entire FileMenuFilter needs a refactor, so I didn't change anything there. 👀

owncloudData/src/main/java/com/owncloud/android/data/files/datasources/LocalFileDataSource.kt
abelgardep: `targetFolder` is the destination folder, probably `destinationFolder` would be a better name. I updated just the name because I think that it is easier to understand the logic, but it keeps the same behavior.

owncloudData/src/main/java/com/owncloud/android/data/files/datasources/implementation/OCLocalFileDataSource.kt
abelgardep: Not my best line of code, that's for sure 😢

owncloudDomain/src/main/java/com/owncloud/android/domain/availableoffline/usecases/GetFilesAvailableOfflineFromEveryAccountUseCase.kt
abelgardep: Not important, but I'll add my name there, so you can blame on me 😉

JuancaG05 (approved):
LGTM 👍

PR #3731, opened 2022-08-22
[New arch] Remove unused classes after uploads rearchitecture

JuancaG05:
Classes removed:
- `OCUpload.java`
- `UploadsStorageManager.java`
- `ExpandableUploadListAdapter.java`
- `ExpandableListFragment.java`
- `ExtendedListFragment.java`
- `OptionsInUploadListClickListener.kt`
- `UploadListFragment.java`

Also, Leak Canary dependency was removed from the app's level `build.gradle` because in this new arch branch it was not being used and hurts performance when developing.


abelgardep (approved):
Awesome! Removing lines is always a pleasure 🥇

Issue #3732, opened 2022-08-22
[FEATURE REQUEST] [oCIS] Respect Capability "privateLinks": false


michaelstingl:
**Is your feature request related to a problem? Please describe.**

oCIS doesn't have the oC10-style private links. It's also announced as false in the capabilities:

```
% cat ocis.capabilities | jq .ocs.data.capabilities.files
{
  "privateLinks": false,
  "bigfilechunking": false,
  "undelete": true,
  "versioning": true,
  "favorites": false,
  "blacklisted_files": [],
  "tus_support": {
    "version": "1.0.0",
    "resumable": "1.0.0",
    "extension": "creation,creation-with-upload",
    "max_chunk_size": 100000000,
    "http_method_override": ""
  },
  "archivers": [
    {
      "enabled": true,
      "version": "2.0.0",
      "formats": [
        "tar",
        "zip"
      ],
      "archiver_url": "/archiver",
      "max_num_files": "10000",
      "max_size": "1073741824"
    }
  ],
  "app_providers": [
    {
      "enabled": true,
      "version": "1.0.0",
      "apps_url": "/app/list",
      "open_url": "/app/open",
      "new_url": "/app/new"
    }
  ]
}
```

**Describe the solution you'd like**

Android app should respect the capability, and hide the private link entries on all screens.

**Additional context**

Desktop issues:

- https://github.com/owncloud/client/issues/9573
- https://github.com/owncloud/client/issues/8998

iOS issue:
- https://github.com/owncloud/ios-app/issues/1138

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/private_link_capability
- [x] Development tasks
  - [x] Add a new field in capabilities response
  - [x] Add a new field in capabilities domain model
  - [x] Add a new field in capabilities entity
  - [x] Show or hide the private link button when sharing depending on the capability
- [x] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/private_link_capability into master

### PR

- App https://github.com/owncloud/android/pull/3738
- Library (if needed) https://github.com/owncloud/android-library/pull/505

michaelstingl:
> Silly question came to my mind: private link means link like `http://<url>/f/<file-id>` , right?

Yes exactly. `http://<url>/f/<file-id>` or `http://<url>/index.php/f/<file-id>`

abelgardep:
With current behavior, we don't show the private link if it is not in the files propfind.

[...]

PR #3738, opened 2022-08-25
[Feature] Respect private link capability

abelgardep:
## Related Issues
App: https://github.com/owncloud/android/issues/3732

Library PR (if needed): https://github.com/owncloud/android-library/pull/505

_____

## QA

- [x] (1) oC10 + capability false https://github.com/owncloud/android/pull/3738#issuecomment-1229874826 [FIXED]
- [x] (2) Sharing API disabled https://github.com/owncloud/android/pull/3738#issuecomment-1229879710 [WONTFIX HERE]

### DIFF
--- a/owncloudApp/src/androidTest/java/com/owncloud/android/sharing/shares/ui/ShareFileFragmentTest.kt
+++ b/owncloudApp/src/androidTest/java/com/owncloud/android/sharing/shares/ui/ShareFileFragmentTest.kt
@@ -45,6 +45,7 @@ import com.owncloud.android.testutil.OC_ACCOUNT
 import com.owncloud.android.testutil.OC_CAPABILITY
 import com.owncloud.android.testutil.OC_SHARE
 import com.owncloud.android.utils.AppTestUtil.OC_FILE
+import com.owncloud.android.utils.matchers.isDisplayed
 import io.mockk.every
 import io.mockk.mockk
 import org.hamcrest.CoreMatchers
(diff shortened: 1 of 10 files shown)


JuancaG05 (approved):
Great work! LGTM 👍

jesmrec:
Three cases to check:

- `privateLink: false`(oCIS) -> private link not available 
- `privateLink: true` -> private link available 
- No `privateLink` in capabilities -> private link available

jesmrec:
### (1)

Something weird. Is the platform (oCIS/oC10) involved in the logic? 

in oCIS: `"privateLinks": false'  -> no private link
in oC10: `"privateLinks": false'  -> private link available. 

`a745542e4`

jesmrec:
### (2)

This case is also interesting to take in account:

1. In server admin dashboard, open Admin -> Sharing
2. Disable `Allow apps to use the Share API`

in that case, Share API is not available and users will not be allowed to shares with users or by link. But, the "Share" option is still available because the private link.

I mean, if: 

- Share API is disabled
- `privateLinks: false`

the sharing option in top bar could be disabled as well.

This is an improvement, not a bad behaviour

Samsung A51
`a745542e4`

abelgardep:
@jesmrec 
(1) Should be fixed now.
(2) It's not that simple to implement at the moment. I think we can create a new issue to do that

jesmrec:
@abelgardep after fixing (1), if `privateLink` capability is not retrieved, private links are not available.

jesmrec:
(1) fixed
(2) addressed to https://github.com/owncloud/android/issues/3740

Feature is approved

Issue #3708, opened 2022-07-05
Copy/Move to same folder


jesmrec:
In the following scenario, Android app behaves in a very different way comparing with other clients:

1. Select any item of the list
2. Select Copy or Move actions
3. Select as target folder the current folder

Current behaviour across the platform:

**Android:**

- Copy to same folder -> create a duplicate with (2)
- Move to same folder -> rename to (2)

**iOS:**

- Copy to same folder -> not allowed
- Move to same folder -> not allowed

**Web (classic):**

- Copy to same folder -> not allowed (by drag and drop)
- Move to same folder -> NA

**Web (oCIS):**

- Copy to same folder -> not allowed (error message)
- Move to same folder -> not allowed (error message)

**Desktop client:**

- Copy to same folder -> duplicate with `copy` (like any other file in the OS)
- Move to same folder -> NA

I'd go for the "not allowed" and prevent copy/move to the current folder

Question to solve/discuss: expected behaviour for copying/moving in document provider

jesmrec:
Checked also this related scenario:

1. Select item
2. Copy / Move
3. Select target folder that contains other item with same file/folder name

In iOS and web, it is prevented as well.

jesmrec:
Will follow the oCIS behaviour.

PR #3748, opened 2022-09-05
[New Arch] Copy/Move into the same folder

abelgardep:
Moving/Copying files into the same folder will fail. An error message will be shown in that case.

## Related Issues
App: https://github.com/owncloud/android/issues/3708
_____


JuancaG05 (approved):
Simple and accurate solution, nice work! 💯

jesmrec:
Approved

Issue #2818, opened 2020-02-18
[New arch] Synchronization


davigonz:
# Introduction
We have been working on updating the old Android App to a new architecture for a long time.

The new architecture is MVVM and we can find more details about it here: https://github.com/owncloud/android/issues/2351

To do this, we started with Shares functionality, we continued with Authentication, and finally, we are working on the most crytical part of the app: the Synchronization. This last epic is huge and we created a milestone to keep track of it: https://github.com/owncloud/android/milestone/43

# Synchronization
App PR: https://github.com/owncloud/android/pull/2934
Library PR: https://github.com/owncloud/android-library/pull/339

### Create folder [QA Passed] ✅ 
Issue: https://github.com/owncloud/android/issues/2861
PR: https://github.com/owncloud/android/pull/2923
QA: https://github.com/owncloud/android/issues/2818#issuecomment-683732010
QA reports fixed:
- [x] (1)

### Rename [QA Passed] ✅ 
Issue: https://github.com/owncloud/android/issues/2863
PR: https://github.com/owncloud/android/pull/3231
QA: https://github.com/owncloud/android/issues/2818#issuecomment-846984004
QA reports fixed:
- [x] (1)
- [x] (2)
- [x] (3)
- [x] (4)

### Remove [Bugfixing]  
Issue: https://github.com/owncloud/android/issues/2864
PR: https://github.com/owncloud/android/pull/3214
QA: https://github.com/owncloud/android/issues/2818#issuecomment-834301102
QA reports fixed:
- [x] (1)
- [x] (2)
- [ ] (3) [P2] Removing many files blocks the app
- [ ] (4) [P2] Txt file removal cleans up the file list

### Copy [Bugfixing] 
Issue: https://github.com/owncloud/android/issues/2866
PR: https://github.com/owncloud/android/pull/3253
QA: https://github.com/owncloud/android/issues/2818#issuecomment-849394483
QA reports fixed:
- [x] (1)
- [x] (2)
- [ ] (3) [P3] Error message too long

### Move [QA Passed] ✅ 
Issue: https://github.com/owncloud/android/issues/2865
PR: https://github.com/owncloud/android/pull/3232
QA: https://github.com/owncloud/android/issues/2818#issuecomment-844022500
QA reports fixed:
- [x] (1)

## Transfers
Transfers have changed completely from the previous version. We have moved from foreground services to WorkManager. They can be done parallely now.

[...]

PR #3757, opened 2022-09-15
[New Arch] Uploads open document

abelgardep:
Uses [ACTION_OPEN_DOCUMENT](https://developer.android.com/reference/android/content/Intent#ACTION_OPEN_DOCUMENT) instead of [ACTION_GET_CONTENT](https://developer.android.com/reference/android/content/Intent#ACTION_GET_CONTENT) to open the file picker from the floating action button.

It should retain the access for enough time so we have time to upload the file.

Should fix some reports we have in #2818 where we lose access to the file
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -38,6 +38,7 @@ import android.content.Intent
 import android.content.IntentFilter
 import android.net.Uri
 import android.os.Bundle
+import android.os.RemoteException
 import android.view.Menu
 import android.view.MenuItem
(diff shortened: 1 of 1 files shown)


JuancaG05 (approved):
👍

jesmrec:
Reports fixed from uploads: https://github.com/owncloud/android/issues/2818#issuecomment-1157608496

(2)
(4)
(15)
(17)


Report (8) with same behaviour (is this intended to be fixed here)? i guess not, just to be sure.

abelgardep:
> Report (8) with same behaviour (is this intended to be fixed here)? i guess not, just to be sure.

Nothing to do with chunking in this PR. 👍

jesmrec:
Go ahead with this. If new stuff happens, we can handle in a different PR

Issue #3005, opened 2020-10-26
[New Arch] Conflicts management


abelgardep:
Synchronization epic: #2818

Handle conflicts between local and remote files. Currently handled by SynchronizeFileOperation and SynchronizeFolderOperation. 

### TASKS

- [x] Research (if needed)
- [x] Create branch new_arch/conflicts_management
- [x] Development tasks
  - [x] Single item shows a red icon to show conflict
  - [x] Kotlinize `ConflictsResolveActivity.java`
  - [x] Kotlinize `ConflictsResolveDialog.java`
  - [x] Remove old `ConflictsResolveActivity.java` and `ConflictsResolveDialog.java`
  - [x] Show conflict dialog from the main file list
  - [x] Show conflict dialog from the file details view
  - [x] Folder shows a red icon if any descendants have conflicts [ADVANCED -> owncloud/android/pull/3766]
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch new_arch/conflicts_management into new_arch/synchronization

jesmrec:
One clarification. In `master` version, when a conflict happens a notification is raised up. By clicking on it, conflict solving dialog is open in the app. Conflict could also happen in the background (new version of av. offline that collides with new version in server side)

PR #3762, opened 2022-09-19
[New Arch] Basic conflicts management

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3005
_____

## QA

- [x] (1) local copy removed if solving with `Keep Both` https://github.com/owncloud/android/pull/3762#issuecomment-1265085401 [WONT FIX]


owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromFileSystemWorker.kt
```diff
@@ -119,24 +121,34 @@ class UploadFileFromFileSystemWorker(
...
+                        needsToUpdateThumbnail = true,
+                        etag = uploadFileOperation.etag,
+                        length = (File(ocTransfer.localPath).length()),
+                        lastSyncDateForData = currentTime,
+                        modifiedAtLastSyncForData = currentTime,
+                    )
+                } else {
+                    // Uploading a file should remove any conflicts on the file.
+                    ocFile.copy(
+                        etagInConflict = null,
```
abelgardep, line 141: It will be cleaned on the next cleanConflictUseCase, so I think this is not needed at all. 👍

abelgardep (changes requested):
Tiny change and ready to go @JuancaG05 👍

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromFileSystemWorker.kt
JuancaG05: My bad, I removed the `storagePath` instead of the `etagInConflict` field. Fixed!! 👍

abelgardep (approved):

jesmrec:
### (1)

1. Create a conflict 
2. Solve it with `Keep Both`

Current: Conflict is correctly solved, but local file is removed from device (only server copy)
Expected (`master`): local file is still stored in device. 

Samsung A51, Android 12
`2e2a0e2d`

jesmrec:
Report (1) will not be fixed here. This is a new behaviour, not intrusive, but could modify the user expectation (downloaded file is removed without any consent).

Open to discussion for beta 2

PR #3763, opened 2022-09-22
[New arch] Chunking bugfixing

JuancaG05:
When doing a manual content URI upload, now files are copied to local storage before uploading them.

Library PR: https://github.com/owncloud/android-library/pull/508
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFilesFromSystemUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFilesFromSystemUseCase.kt
@@ -58,13 +58,13 @@ class UploadFilesFromSystemUseCase(
 
             val uploadId = storeInUploadsDatabase(
                 localFile = localFile,
-                uploadPath = params.uploadFolderPath.plus(File.separator).plus(localFile.name),
+                uploadPath = params.uploadFolderPath.plus(localFile.name),
                 accountName = params.accountName,
             )
 
             enqueueSingleUpload(
                 localPath = localFile.absolutePath,
-                uploadPath = params.uploadFolderPath.plus(File.separator).plus(localFile.name),
+                uploadPath = params.uploadFolderPath.plus(localFile.name),
                 lastModifiedInSeconds = localFile.lastModified().div(1_000).toString(),
                 accountName = params.accountName,
                 uploadIdInStorageManager = uploadId,
--- a/owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
@@ -240,9 +240,10 @@ class UploadFileFromContentUriWorker(
     }
 
     private fun uploadChunkedFile(client: OwnCloudClient) {
+        val immutableHashForChunkedFile = SecurityUtils.stringToMD5Hash(uploadPath) + System.currentTimeMillis()
         // Step 1: Create folder where the chunks will be uploaded.
         val createChunksRemoteFolderOperation = CreateRemoteFolderOperation(
-            remotePath = SecurityUtils.stringToMD5Hash(uploadPath) + System.currentTimeMillis(),
+            remotePath = immutableHashForChunkedFile,
             createFullPath = false,
             isChunksFolder = true
         )
@@ -250,7 +251,7 @@ class UploadFileFromContentUriWorker(
 
         // Step 2: Upload file by chunks
         uploadFileOperation = ChunkedUploadFromFileSystemOperation(
-            transferId = SecurityUtils.stringToMD5Hash(uploadPath) + System.currentTimeMillis(),
+            transferId = immutableHashForChunkedFile,
             localPath = cachePath,
             remotePath = uploadPath,
             mimeType = mimeType,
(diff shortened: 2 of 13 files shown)


owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
```diff
@@ -134,6 +147,34 @@ class UploadFileFromContentUriWorker(
...
+
+        val outputStream = FileOutputStream(cachePath)
+        val buffer = ByteArray(4096)
+        if (inputStream != null) {
+            var count = inputStream.read(buffer)
+            while (count > 0) {
+                outputStream.write(buffer, 0, count)
+                count = inputStream.read(buffer)
+            }
+        }
```
abelgardep, line 172: I think we could try to open the input and output streams when we are using them and we need to close them just after using them to release their resources.

```suggestion
        val inputStream = appContext.contentResolver.openInputStream(contentUri)
        val outputStream = FileOutputStream(cachePath)
        outputStream.use { fileOut ->
            inputStream?.copyTo(fileOut)
        }
        inputStream?.close()
```

We can take advantage of `FileOutputStream.copyTo `extension 👍

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
```diff
@@ -134,6 +147,34 @@ class UploadFileFromContentUriWorker(
...
+
+        mimeType = cacheFile.extension
+        fileSize = cacheFile.length()
```
abelgardep, line 175: Not sure about this. If we store the new data into a variable but the worker fails, when it is retried, the mime-type and filesize, and local path would be still the same as the first time even when the file is already in cached storage.

I think that we should persist that data so we don't have issues when retrying failed ones.

An already copied to cache upload but failed, for example because server connection, should:
1. Option: retry as UploadFromFileSystem
2. Option: Detect if the file is already in the cache, and skip moving to caché and go straight to the upload

I would go with the second option. What do you think?

abelgardep (changes requested):
Some comments here

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
JuancaG05: Nice change!! Applied 👍

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
```diff
@@ -157,14 +199,80 @@ class UploadFileFromContentUriWorker(
...
+
+        val getStoredCapabilitiesUseCase: GetStoredCapabilitiesUseCase by inject()
+        val capabilitiesForAccount = getStoredCapabilitiesUseCase.execute(
+            GetStoredCapabilitiesUseCase.Params(
+                accountName = account.name
+            )
+        )
+        val isChunkingAllowed = capabilitiesForAccount != null && capabilitiesForAccount.isChunkingAllowed()
+        Timber.d("Chunking is allowed: %s", isChunkingAllowed)
```
abelgardep, line 214: This log could handle more information regarding not only if chunking is allowed but also if the file is big enough to be chunked

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromContentUriWorker.kt
```diff
@@ -157,14 +199,80 @@ class UploadFileFromContentUriWorker(
...
+        ).also {
+            it.addDataTransferProgressListener(this)
+        }
+
+        val result = executeRemoteOperation { uploadFileOperation.execute(client) }
+
+        if (result == Unit && behavior == UploadBehavior.MOVE) {
+            removeLocalFile()
         }
+    }
```
abelgardep, line 240: This is the code that should be placed just after copying the file to the cache 🍻

abelgardep (changes requested):
Some suggestions here and we are ready to go 👍

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFilesFromSystemUseCase.kt
```diff
@@ -58,13 +58,13 @@ class UploadFilesFromSystemUseCase(
...
                 localFile = localFile,
-                uploadPath = params.uploadFolderPath.plus(File.separator).plus(localFile.name),
+                uploadPath = params.uploadFolderPath.plus(localFile.name),
                 accountName = params.accountName,
             )
 
             enqueueSingleUpload(
                 localPath = localFile.absolutePath,
-                uploadPath = params.uploadFolderPath.plus(File.separator).plus(localFile.name),
+                uploadPath = params.uploadFolderPath.plus(localFile.name),
```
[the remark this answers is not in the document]
JuancaG05: Yes, I think it fixes the problems in the names of the paths due to the duplicated `/`.

abelgardep (approved):
LGTM, Let's QA this @jesmrec

jesmrec:
Some tests done here:

- [X] Manual uploads > 10MB
- [X] Manual uploads < 10 MB
- [X] Manual uploads -> same file uploaded twice
- [X] Manual uploads -> upload interrupted and retried
- [X] Upload from camera
- [X] Upload from external app
- [X] Camera uploads

jesmrec:
Ready to go, new round to be done in the global scope of the uploads after beta is released

Issue #3724, opened 2022-08-05
[RELEASE] 3.0


jesmrec:
Synchronization engine with [new architecture ](https://github.com/owncloud/android/milestone/43)

### TASKS:

 - [x] [GIT] Create branch release/3.0 in owncloud/android from new_arch/synchronization
 - [x] [DEV] Update version number and name in build.gradle in owncloudApp module
 - [x] [DIS] Create a folder for the new version like $majorVersion.$minorVersion.$patchVersion_YYYY-MM-DD
 - [x] [DIS] Move all changelog files from the unreleased folder to the new version folder
 - [x] [DIS] Update screenshots, if needed, in README.md
 - [x] [DIS] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
 - [x] [QA] Design Test plan
 - [x] [QA] Regression Test plan
 - [x] [GIT] Create and sign tag 'oc-android-3.0' in HEAD commit of release branch, in owncloud/android
 - [x] [GIT] Create and sign tag '2.0' in HEAD commit of release branch, in owncloud/android-library
 - [x] [DIS] Generate final bundle from signed commit in owncloud/android
 - [x] [GIT] Merge branch `release/3.0` in owncloud/android, into master
 - [x] [DIS] Upload & publish release bundle and changelog in Play Store
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [GIT] Publish a new release in owncloud/android
 - [x] [DIS] Create post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
 - [x] [COM] Inform `#updates` and `#marketing` in internal chat
 - [x] [DIS] Upload release APK and bundle to internal owncloud instance
 - [x] [GIT] Merge new_arch/synchronization branch into master, in owncloud/android-library
 - [x] [GIT] Merge new_arch/synchronization branch into master, in owncloud/android
 - [x] [GIT] Merge master branch into stable, in owncloud/android-library
 - [x] [GIT] Merge master branch into stable, in owncloud/android
 - [x] [DOC] Update documentation with new stuff (notify rocketchat #documentation-internal in advance!)

_____

### QA (pending reports)

S1:

- [x] (1) Refresh new content in previews https://github.com/owncloud/android/issues/3724#issuecomment-1302054391 [FIXED]
[...]

PR #3793, opened 2022-11-23
[FIX] Some fixes for 3.0

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3724

Includes fixes for the reports:
- (1) Refresh new content in previews https://github.com/owncloud/android/issues/3724#issuecomment-1302054391 [FIXED]
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -1191,17 +1191,17 @@ class FileDisplayActivity : FileActivity(),
                     }
                     is SynchronizeFileUseCase.SyncType.DownloadEnqueued -> {
                         if (fileWaitingToPreview != null) {
-                            showSnackMessage("A new version was found in server. Downloading...")
+                            showSnackMessage(getString(R.string.new_remote_version_found_msg))
                             startSyncThenOpen(fileWaitingToPreview!!)
                             fileWaitingToPreview = null
                         } else {
-                            showSnackMessage("Download enqueued")
+                            showSnackMessage(getString(R.string.download_enqueued_msg))
                         }
                     }
                     SynchronizeFileUseCase.SyncType.FileNotFound -> {
                         /** Nothing to do atm. If we are in details view, go back to file list */
                     }
-                    is SynchronizeFileUseCase.SyncType.UploadEnqueued -> showSnackMessage("Upload enqueued")
+                    is SynchronizeFileUseCase.SyncType.UploadEnqueued -> showSnackMessage(getString(R.string.upload_enqueued_msg))
                     null -> TODO()
                 }
             }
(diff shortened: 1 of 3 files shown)


jesmrec:
Checks done:

- [X] Text file with connection
- [X] Text file without connection
- [X] Audio file with connection
- [X] Audio file without connection 

Tested in both oC10 & oCIS

Samsung A51, Android 12
`8acd12d65`

(review pending)

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1176,13 +1178,30 @@ class FileDisplayActivity : FileActivity(),
...
+                    }
                     is SynchronizeFileUseCase.SyncType.ConflictDetected -> {
                         val showConflictActivityIntent = Intent(this, ConflictsResolveActivity::class.java)
                         showConflictActivityIntent.putExtra(ConflictsResolveActivity.EXTRA_FILE, file)
                         startActivity(showConflictActivityIntent)
                     }
-                    is SynchronizeFileUseCase.SyncType.DownloadEnqueued -> showSnackMessage("Download enqueued")
+                    is SynchronizeFileUseCase.SyncType.DownloadEnqueued -> {
+                        if (waitingToPreview) {
+                            showSnackMessage("A new version was found in server. Downloading...")
```
jesmrec, line 1197: Please, no hardcoded text. Make this translatable

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1176,13 +1178,30 @@ class FileDisplayActivity : FileActivity(),
...
+                            startSyncThenOpen(fileWaitingToPreview!!)
+                            waitingToPreview = false
+                            fileWaitingToPreview = null
+                        } else {
+                            showSnackMessage("Download enqueued")
```
jesmrec, line 1202: Same here

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
JuancaG05: Thanks, didn't realize!

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1176,55 +1176,67 @@ class FileDisplayActivity : FileActivity(),
...
-
-                            launch(Dispatchers.Main) {
-                                if (credentials is OwnCloudBearerCredentials) { // OAuth
-                                    showRequestRegainAccess()
-                                } else {
-                                    showRequestAccountChangeNotice(getString(R.string.auth_failure_snackbar), false)
+                if (fileWaitingToPreview != null) {
+                    startPreview(fileWaitingToPreview)
+                    fileWaitingToPreview = null
+                } else {
```
abelgardep, line 1212: If we get an error when syncing because credentials are not valid anymore, we probably want to ask to the user about login again. So I don't think we should add that into an else

abelgardep (changes requested):

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
JuancaG05: 👍

abelgardep (approved):

jesmrec:
Approved on my side. 

(1) fixed

Issue #3812, opened 2022-12-13
[TECHNICAL] Some refactoring after 3.0


JuancaG05:
During the long process of the rearchitecture of the app, there have been several points that we detected could be refactored to improve the code understandability and to unify it following some conventions. These points are:

- Prefix `OC-` in implementations: should we continue using it? Replacing with `-Impl` suffix or similar?
- Prefix `OC-` in domain classes: should we keep it? Maybe confusing with implementations?
- Make DAOs interfaces and not abstract classes where possible?
- Update SQL queries from "query" to """query"""?
- Align names of operations in database: `store-` or `save-` (in DataSources)? `insert` or `insertSomething` (in DAOs)? `delete-` or `remove-`? `deleteSomethingWithId` or `deleteSomethingById` (in Repositories, DataSources and DAOs)?
- Rearrange package structure (i.e. in the `presentation` package, arrange it by features, and each of the features contains standard packages `fragments`, `viewmodels`, etc.)
- Can we remove `DownloadReceiver` from `FileDisplayActivity.kt`?
- Can we remove `SquareImageView.java`?
- Can we remove `RemoteFileMapper.kt`?
- Anything else we can remove?
- Take a look at flaky tests, sometimes they prevent BitRise from passing at the first attempt and make work slower
- Update dependencies that we couldn't update all this time
- Check Sonar status of the code
- When Room 2.5 is stable, change the inserts with replace strategy and updates by `@Upsert`

This is all open to discussion, I will post in a comment the decisions we eventually take for each of the points above.

### TASKS

- [x] Create branch technical/some_refactoring
- [x] Development tasks
  - [x] Do what was decided with `OC-` prefixes
  - [x] Do what was decided with DAOs as interfaces and not abstract classes
  - [x] Do what was decided with format of SQL queries
  - [x] Do what was decided with names of DB operations
  - [x] Do what was decided with package structure
  - [x] Remove unnecessary stuff (@JuancaG05)
  - [x] Remove unnecessary stuff (@abelgardep)
  - [x] Update dependencies -> https://github.com/owncloud/android/pull/3810
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch technical/some_refactoring into master

JuancaG05:
**- Prefix `OC-` in implementations: should we continue using it? Replacing with `-Impl` suffix or similar?**
Let's keep implementations with `OC-` prefix.

[...]

PR #3815, opened 2022-12-14
[TECHNICAL] Some refactoring after 3.0

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3812
_____


owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackUpEntity.kt
```diff
@@ -31,12 +31,9 @@ data class FolderBackUpEntity(
     val uploadPath: String,
     val wifiOnly: Boolean,
     val chargingOnly: Boolean,
-    @ColumnInfo(name = folderBackUpEntityNameField) val name: String,
+    @ColumnInfo(name = "name")
+    val name: String,
```
abelgardep, line 35: If the name we want to apply is the same as the entity attribute, we don't need the column info. Don't know why I introduced it initially 👀 
```suggestion
    val name: String,
```

owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackupDao.kt
```diff
@@ -27,35 +28,40 @@ import com.owncloud.android.data.ProviderMeta
...
+        return insertOrReplace(folderBackUpEntity)
     }
 
-    @Query(
-        "DELETE from " + ProviderMeta.ProviderTableMeta.FOLDER_BACKUP_TABLE_NAME + " WHERE " +
-                FolderBackUpEntity.folderBackUpEntityNameField + " = :name"
-    )
-    abstract fun delete(name: String): Int
+    companion object {
+        private const val SELECT = """
```
abelgardep, line 55: If we want to add query constants on every dao, we should align query names too. For example: `SELECT_WITH_NAME` or `DELETE_WITH_NAME`.

What do u think?

owncloudData/src/test/java/com/owncloud/android/data/file/repository/OCFileRepositoryTest.kt
```diff
@@ -229,14 +229,14 @@ class OCFileRepositoryTest {
 
     @Test
     fun `remove file - ok`() {
-        every { remoteFileDataSource.removeFile(any(), any()) } returns Unit
+        every { remoteFileDataSource.deleteFile(any(), any()) } returns Unit
         every { localStorageProvider.deleteLocalFile(any()) } returns true
 
-        ocFileRepository.removeFile(listOfFilesToRemove = listOfFileToRemove, removeOnlyLocalCopy = false)
+        ocFileRepository.deleteFile(listOfFilesToDelete = listOfFileToRemove, removeOnlyLocalCopy = false)
```
abelgardep, line 235: I see that we use a singular name when deleting a list, so I would update it to `deleteFiles` by the way 👍

owncloudDomain/src/main/java/com/owncloud/android/domain/files/FileRepository.kt
```diff
@@ -34,20 +34,20 @@ interface FileRepository {
...
-    fun getFilesWithSyncInfoAvailableOfflineFromAccountAsStream(owner: String): Flow<List<OCFileWithSyncInfo>>
+    fun getSharedByLinkWithSyncInfoForAccountAsFlow(owner: String): Flow<List<OCFileWithSyncInfo>>
+    fun getFilesWithSyncInfoAvailableOfflineFromAccountAsFlow(owner: String): Flow<List<OCFileWithSyncInfo>>
     fun getFilesAvailableOfflineFromAccount(owner: String): List<OCFile>
     fun getFilesAvailableOfflineFromEveryAccount(): List<OCFile>
     fun moveFile(listOfFilesToMove: List<OCFile>, targetFile: OCFile)
     fun readFile(remotePath: String, accountName: String): OCFile
     fun refreshFolder(remotePath: String, accountName: String): List<OCFile>
-    fun removeFile(listOfFilesToRemove: List<OCFile>, removeOnlyLocalCopy: Boolean)
+    fun deleteFile(listOfFilesToDelete: List<OCFile>, removeOnlyLocalCopy: Boolean)
```
abelgardep, line 50: As the other comment, when the function applies to lists, I would rename them to match the parameters: `deleteFiles`

What do u think?

abelgardep (changes requested):
Good changes here to align naming across the whole app. Some suggestions here and we are ready 👍

owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackUpEntity.kt
JuancaG05: That's what I thought, but I didn't want to break it just in case 😬

owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackupDao.kt
JuancaG05: Yes, I totally agree, but there are maaany queries and each of them have a different style... I think this could be work for a second iteration of renaming that we can do in a future 🤔

owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackUpEntity.kt
abelgardep: If this change broke something, we would be notified by a new DB scheme change. If it is not generated, go ahead 👍

owncloudData/src/main/java/com/owncloud/android/data/folderbackup/db/FolderBackupDao.kt
abelgardep: Ok, let's do that on another round

owncloudData/src/test/java/com/owncloud/android/data/file/repository/OCFileRepositoryTest.kt
JuancaG05: True, I didn't realise! Change applied 👍

owncloudDomain/src/main/java/com/owncloud/android/domain/files/FileRepository.kt
JuancaG05: Yes, once refactored in the implementation the renaming is propagated here 🤣

abelgardep (approved):
Ready to go 🚀

JuancaG05:
@jesmrec fix for (1) pushed

jesmrec:
(1) fixed

jesmrec:
Approved on my side. Basic checks done after a bunch of renames.

Issue #3824, opened 2022-12-14
[BUG] oCIS - Remove share is correct but returns an error


jesmrec:
### Steps to reproduce
1.  In oCIS, share any item with another user
2.  In the app, remove the share

### Actual behaviour
Share is removed, but an error is shown: `Server could not be reached`

### Expected behaviour
Share removed without errors
 
This is the server response to the `DELETE` request:

```
{
    "ocs": {
        "data": {
            "additional_info_file_owner": "",
            "additional_info_owner": "",
            "displayname_file_owner": "",
            "displayname_owner": "",
            "expiration": "",
            "file_parent": "",
            "file_source": "",
            "file_target": "",
            "id": "",
            "item_source": "",
            "item_type": "",
            "mail_send": 0,
            "mimetype": "",
            "name": "",
            "parent": "",
            "path": "",
            "permissions": 17,
            "share_type": 0,
            "share_with": "fc20ad85-0600-4aa0-9de8-7411faf5114d",
            "share_with_additional_info": "",
            "share_with_user_type": 0,
            "state": 0,
            "stime": 1671024556,
            "storage": 0,
            "storage_id": "",
            "token": "",
            "uid_file_owner": "33e4fecc-b8c0-4aa2-b563-e0582ef9475c",
            "uid_owner": "33e4fecc-b8c0-4aa2-b563-e0582ef9475c"
        },
        "meta": {
            "message": "OK",
            "status": "ok",
            "statuscode": 200
        }
    }
}
```

`meta` object is slightly different to same request's response in oC10

```
{
    "ocs": {
        "data": [],
        "meta": {
            "itemsperpage": "",
            "message": null,
            "status": "ok",
            "statuscode": 200,
            "totalitems": ""
        }
    }
}
```

Samsung A51, Android 12

abelgardep:
As u post on the issue, there were a few differences in the responses.

We expected a list as we had on oC10 but an object was returned instead. And that's why the app failed to parse the response and showed that error.

PR #3831, opened 2022-12-15
Fix error when unsharing but network request was successfully done

abelgardep:
## Related Issues
App: https://github.com/owncloud/android/issues/3824

Library PR (if needed): https://github.com/owncloud/android-library/pull/523

There is no need to parse the response anymore. The app is not using that data, just checks if it was successful or not.

oC10 returned an empty list
oCIS returns an object

That's why it was not working properly on oCIS accounts

_____


JuancaG05 (approved):

jesmrec:
Checks done:

oCIS:

Remove share of...

- [X] File in root 
- [X] Folder in root
- [X] File in non-root
- [X] Folder in non-root

oC10:

Remove share of...

- [X] File in root 
- [X] Folder in root
- [X] File in non-root
- [X] Folder in non-root

Samsung A51, Android12

This is approved on my side

PR #3837, opened 2022-12-20
[FIX] Database upgrade to 3.0 app version

JuancaG05:
The app crashes if it's updated from a DB version previous to 34 to the latest one. This PR should fix that

_____

### DIFF
--- a/changelog/unreleased/3837
+++ b/changelog/unreleased/3837
@@ -1,0 +1,6 @@
+Bugfix: Fix crash when upgrading from 2.18
+
+Upgrading from 2.18 or older versions made the app crash due to camera uploads data migration.
+This problem has been solved and now the app upgrades correctly.
+
+https://github.com/owncloud/android/pull/3837
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -717,5 +717,7 @@
     <string name="release_notes_3.0_subtitle5">With the redesign of the sync engine, now several downloads and uploads can run at the same time</string>
     <string name="release_notes_3.0_title6">Browser login improved</string>
     <string name="release_notes_3.0_subtitle6">Only typing the URL, browser will be triggered</string>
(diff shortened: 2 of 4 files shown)


jesmrec:
- [X] 2.18 -> 3.0.1 
- [X] 2.20 -> 3.0.1 
- [X] 2.21.2 -> 3.0.1 
- [X] 3.0.0 -> 3.0.1

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -717,5 +717,7 @@
     <string name="release_notes_3.0_subtitle5">With the redesign of the sync engine, now several downloads and uploads can run at the same time</string>
     <string name="release_notes_3.0_title6">Browser login improved</string>
     <string name="release_notes_3.0_subtitle6">Only typing the URL, browser will be triggered</string>
+    <string name="release_notes_3.0.1_title1">Database migration fixed</string>
```
abelgardep, line 720: Users probably don't care about the technical issue, but the `Fix crash when upgrading from 2.18 ` or somehting like that, don't u think?

changelog/unreleased/3837
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix for database migration
```
abelgardep, line 1: Users probably don't care about the technical issue, but the `Fix crash when upgrading from 2.18 ` or somehting like that, don't u think?

changelog/unreleased/3837
```diff
@@ -0,0 +1,6 @@
...
+
+For those users who hadn't migrated the camera uploads data yet, the app crashed. This problem
+has been solved and now the app upgrades correctly.
```
abelgardep, line 4: This is something that users don't know: `For those users who hadn't migrated the camera uploads data yet` It's internal so I would start the summary with something like: `Upgrading from 2.18 or older versions made the app crash due to database migration...`

abelgardep (changes requested):
Just some wording @JuancaG05 👍

abelgardep (approved):

Issue #3805, opened 2022-12-12
[SPACES] Create new "Spaces" view


jesmrec:
It will be needed a new table in DB to support spaces. Much info comes from `drives` endpoint and should be persisted for Android app's use. Below, you'll find a sample JSON object with all the retrieved info.

At least, we should save:

- Space id
- Account id
- Description (subtitle)
- Name
- Last Modification date
- WebDavURL
- Etag
- Image (URL, name, etag...): not sure which fields will be compulsory here.
- Status (active/disabled)

(any other?)

Description is also interesting but, ftm, we are not handling markdown files. Future feature.
Permissions to be handled in a separate issue

Also, new activity to show list of available spaces will be created like 

![206656429-3944af86-c123-4d5a-9270-faf9f56409ed](https://user-images.githubusercontent.com/14894746/207038987-5530ebb9-e13e-4bc7-acea-d0f97d2a8dca.jpeg)

showing image, title and subtitle

If no spaces are available, a view with `No spaces` will be displayed instead.

Show only enabled spaces. If the space is disabled, it is not displayed.

Items in the list of spaces will not handle any operation. Just showing the list ftm.

This view could be refreshable. Good feature.

### TASKS

- [x] Research (if needed)
- [x] Create branch spaces/fetch_spaces
- [x] Development tasks
  - [x] Create new table
  - [x] Add logic to populate it when spaces account is added
  - [x] Create new usecase to fetch the spaces
  - [x] Call the new usecase when opening the spaces tab (we may refresh spaces from another place too, but keep it simple at the moment :))
  - [x] Create new usecase to retrieve spaces from database as flow
  - [x] Populate new fragment with the spaces retrieved from database(recycler grid, adapter etc)
  - [x] Display title and subtitle (name and description from space)
  - [x] Display image using the specials from space
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [X] Merge branch spaces/fetch_spaces into spaces/main

Example of JSON space object:

```
{
         "description" : "Subtitle",
         "driveAlias" : "project/space1",
         "driveType" : "project",
         "id" : "aeda45a3-9aae-408e-b050-4f1ea6258f5e$3ae8361b-51f0-4e3a-be77-17bf29fbbfe8",
         "lastModifiedDateTime" : "2022-12-09T13:57:08.380088722Z",
         "name" : "Space1",
         "owner" : {
            "user" : {
[...]

PR #3854, opened 2023-01-12
[SPACES] Fix empty view not showed when all spaces are disabled

abelgardep:
## Related Issues
https://github.com/owncloud/android/issues/3805#issuecomment-1378704787


jesmrec:
Fixed

PR #3855, opened 2023-01-12
[SPACES] Increase spaces thumbnails to support different screen sizes

abelgardep:
The thumbnails were a little bit too small for some resolutions. Now, they should escalate a little bit


jesmrec:
Checked in tablet emulator, with and without spaces, and using both orientations. This look matches better to the tablet users. It's OK from my side.

PR #3857, opened 2023-01-13
[SPACES] Remove attached spaces when removing account

JuancaG05:
Spaces are now deleted from database when the account they are attached to is removed from the app.
_____


abelgardep (approved):
LGTM

Issue #3809, opened 2022-12-12
[SPACES] Document Provider showing available Spaces


jesmrec:
Document provider feature will show the available spaces for an specific account.

By selecting an account in document provider, a list of spaces is displayed, every of them as a folder. Then, user will browse through it as usual. 

No changes for non-spaces accounts.

Article with info about DocumentsProvider: https://medium.com/androiddevelopers/building-a-documentsprovider-f7f2fb38e86a

### TASKS

- [x] Research
- [x] Create branch spaces/document_provider
- [x] Development tasks
  - [x] Create new `SpaceCursor` to show spaces
  - [x] Modify logic in `RootCursor` to set correct directory to `COLUMN_DOCUMENT_ID` depending on whether it's a spaces or non-spaces account
  - [x] Adapt logic in `DocumentsStorageProvider` to show the list of spaces
  - [x] Adapt logic to be able to browse through the hierarchy of spaces as if they were folders
  - [x] Check that create folder, rename and remove operations work properly
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch spaces/document_provider into spaces/main

PR #3875, opened 2023-01-26
[SPACES] Document Provider showing available Spaces

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3809

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_4.0/Spaces%20Document%20Provider.md

Reports:

- [X] (1) Spaces not switched between accounts https://github.com/owncloud/android/pull/3875#issuecomment-1410630204 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
@@ -1,0 +1,49 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Juan Carlos Garrote Gascón
+ *
+ * Copyright (C) 2023 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.presentation.documentsprovider.cursors
+
+import android.database.MatrixCursor
+import android.provider.DocumentsContract.Document
+import com.owncloud.android.R
+import com.owncloud.android.domain.spaces.model.OCSpace
+import com.owncloud.android.presentation.documentsprovider.cursors.FileCursor.Companion.DEFAULT_DOCUMENT_PROJECTION
+
+class SpaceCursor(projection: Array<String>?) : MatrixCursor(projection ?: DEFAULT_DOCUMENT_PROJECTION) {
+
+    fun addSpace(space: OCSpace) {
+        val iconRes = R.drawable.ic_spaces
+        val mimeType = Document.MIME_TYPE_DIR
+        var flags = Document.FLAG_DIR_SUPPORTS_CREATE
+
+        if (space.isPersonal) {
+            flags = flags or Document.FLAG_SUPPORTS_WRITE
+        }
+
+        newRow()
+            .add(Document.COLUMN_DOCUMENT_ID, space.id)
+            .add(Document.COLUMN_DISPLAY_NAME, space.name)
+            .add(Document.COLUMN_LAST_MODIFIED, space.lastModifiedDateTime)
+            .add(Document.COLUMN_SIZE, space.quota?.used)
+            .add(Document.COLUMN_FLAGS, flags)
+            .add(Document.COLUMN_ICON, iconRes)
+            .add(Document.COLUMN_MIME_TYPE, mimeType)
+    }
+}
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
@@ -24,14 +24,13 @@ import android.content.Context
 import android.database.MatrixCursor
 import android.provider.DocumentsContract.Document
 import com.owncloud.android.R
+import com.owncloud.android.domain.files.model.OCFile.Companion.ROOT_PARENT_ID
 import com.owncloud.android.domain.spaces.model.OCSpace
 import com.owncloud.android.presentation.documentsprovider.cursors.FileCursor.Companion.DEFAULT_DOCUMENT_PROJECTION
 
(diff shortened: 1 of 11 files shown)


owncloudData/src/main/java/com/owncloud/android/data/spaces/datasources/implementation/OCLocalSpacesDataSource.kt
```diff
@@ -104,7 +108,7 @@ class OCLocalSpacesDataSource(
                         used = spaceQuotaEntity.used,
                     )
                 },
-                root = space.root!!.let { spaceRootEntity ->
+                root = space.root.let { spaceRootEntity ->
```
abelgardep, line 111: If root is not nullable, `let` is not needed here, don't you think?

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
```diff
@@ -0,0 +1,64 @@
...
+            .add(Document.COLUMN_DISPLAY_NAME, name)
+            .add(Document.COLUMN_LAST_MODIFIED, space.lastModifiedDateTime)
+            .add(Document.COLUMN_SIZE, space.quota?.used)
+            .add(Document.COLUMN_FLAGS, flags)
+            .add(Document.COLUMN_ICON, iconRes)
+            .add(Document.COLUMN_MIME_TYPE, mimeType)
+    }
+
+    fun addRootForSpaces() {
+        val mimeType = Document.MIME_TYPE_DIR
```
abelgardep, line 54: This val is not needed, right? We could assign that value directly, right?

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
```diff
@@ -0,0 +1,64 @@
...
+
+        newRow()
+            .add(Document.COLUMN_DOCUMENT_ID, 0)
```
abelgardep, line 57: What does this 0 mean? Just curious 🤔

owncloudData/src/main/java/com/owncloud/android/data/spaces/datasources/implementation/OCLocalSpacesDataSource.kt
```diff
@@ -137,6 +141,44 @@ class OCLocalSpacesDataSource(
                 webDavUrl = webDavUrl
             )
 
+        @VisibleForTesting
+        fun SpacesEntity.toModel() =
+            OCSpace(
```
abelgardep, line 146: The only difference between this one and `SpacesWithSpecials.toModel()` is that we have specials on one of them. But the "main" parsing could be shared so if we need to apply any change to one class, we won't need to update 2 parsers. 

Is there a way to do this?

abelgardep (changes requested):
Some suggestions here @JuancaG05

owncloudData/src/main/java/com/owncloud/android/data/spaces/datasources/implementation/OCLocalSpacesDataSource.kt
JuancaG05: Not necessary but I think it improves readability, if `space.root` had a longer name it wouldn't be nice to repeat it in every line I think

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
JuancaG05: Yes, and in `addSpace` mimeType and iconRes as well

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
JuancaG05: Means nothing, just an ID to distinguish the root folder which contains all the spaces, which is not an existent ID in database (they start from 1). So, it is more like a flag that is checked afterwards in `DocumentsStorageProvider`. For more consistency, I replaced that raw 0 with `ROOT_PARENT_ID`, that is what is checked where I mentioned.

owncloudData/src/main/java/com/owncloud/android/data/spaces/datasources/implementation/OCLocalSpacesDataSource.kt
JuancaG05: I realised, and I thought I could reuse the `SpacesEntity.toModel()` in the other method, but in `OCSpace`, the `special` property is a val, and cannot be reassigned. I don't know if it is worth it to make it a var so that we can reuse the previous method when we are not going to reassign it anywhere else for the moment... My decision was not to do it 😬, what do you think?

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
abelgardep: I have just realized that it is only used as an identifier to show or not the spaces. It has no relation with the actual `ROOT_PARENT_ID`, right?

We probably should create a new constant to be used only on DcumentsProvider to show spaces or not. Similar to the one we use for nonexistent document id yet `NONEXISTENT_DOCUMENT_ID`

Something like: `const val DISPLAY_SPACES_DOCUMENT_ID = "-2"` instead of reusing the ROOT_PARENT_ID that we use for other topics

owncloudData/src/main/java/com/owncloud/android/data/spaces/datasources/implementation/OCLocalSpacesDataSource.kt
abelgardep: I've just pushed a commit updating this https://github.com/owncloud/android/pull/3875/commits/0f717e64ca9479ef2995907e329f3ca71786510c Check it out 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/cursors/SpaceCursor.kt
JuancaG05: Yes, that's a good idea, I'll apply it. I used `ROOT_PARENT_ID` because we were referring to the root parent folder, but it has other uses too as you mentioned 👍

abelgardep (approved):
LGTM, Ready to QA @jesmrec

jesmrec:
Let's QA this

jesmrec:
### (1)

1. Add two oCIS accounts with spaces to the app
2. Open any app with document provider access and open one of the oC accounts -> spaces listed ✅ 
3. In document provider app, switch to the other account

Current: listed spaces from the first account (step 2)
Expected: listed spaces from the second account (step 3)

Samsung A51, Android 12
`6b549881d`

jesmrec:
Approved on my side

Enhancement: https://github.com/owncloud/android/issues/3888

Issue #3747, opened 2022-09-05
[BUG] Adding a new account doesn't lead you to the correct screen


JuancaG05:
### Actual behaviour
After adding a new account, you can see that pressing the physical back button leads you to the files view of the user logged previously, creating an inconsistent state between the top app bar and the rest of the application. A more illustrative video here:

https://user-images.githubusercontent.com/57049315/188424182-feed7b9f-8060-46fa-8879-34cc5ba93fec.mp4

Here you can see that being logged in the "admin" account, and after adding the "demo" account, we keep accessing the files from the "admin" account, as well as noticing an inconsistent state between the top app bar (which shows that "demo" is logged) and the rest of the app (such as the drawer menu which shows that "admin" is logged, or the files view, which shows the files belonging to "admin").

NOTE: In `master`, this behaviour is only visual, it doesn't change anything from the "demo" account. In `new_arch/synchronization`, the shares created in "admin" are copied to "demo" once they are seen in the share screen.

### Expected behaviour
After adding a new account, the app should get you to a correct state in which the current logged user is the one you just added, showing it correctly in the top app bar, drawer menu and files view.
 
### Steps to reproduce
1. Login with an account in an ownCloud server
2. Create some shares in a file so that the inconsistent state is more visual afterwards
3. Add a new account
4. Press the physical back button
5. If you access the shares of the file you worked with before, you will see the shares created in the other account

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes

### Environment data
Android version: 9

Device model: Samsung Galaxy S9

Stock or customized system: No

ownCloud app version: 2.21.1

ownCloud server version: 10.10.0.3

#### ownCloud log (data/owncloud.log)
No logs

JuancaG05:
Fixed by https://github.com/owncloud/android/pull/3889

Issue #3216, opened 2021-04-29
[BUG] Not redirecting to login screen after detaching all accounts


JuancaG05:
### Actual behaviour
After logging out of the only account registered in the app via "Manage accounts" screen you are redirected to login screen. But, if you press the physical back button, you go back to the "Manage accounts" screen, and if you press back again, you reach the files view and you are allowed to navigate through the different folders. But once you try to download any file or press the avatar in the toolbar, the app crashes.

### Expected behaviour
After logging out of the only account registered in the app via "Manage accounts" screen you are redirected to login screen. If you press the physical back button, you go back for a second to the "Manage accounts" screen, but you are immediately redirected again to the login screen.
 
### Steps to reproduce
1. Login with any account
2. Go to the "Manage accounts" screen via the toolbar's avatar or the drawers' item with the same name
3. Press the rubbish bin to remove the account
4. When redirected to the login screen, press the physical back button
5. If you want to get to the files view, press the physical back button again

### Environment data
Android version: 9

Device model: Samsung Galaxy S9

Stock or customized system: No

ownCloud app version: 2.17

JuancaG05:
It doesn't happen with a Nexus 5 (Android 10) executed in an emulator. The behaviour is the expected. Exclusive bug for Samsung devices, such as some others I reported?

JuancaG05:
Fixed by https://github.com/owncloud/android/pull/3889

PR #3889, opened 2023-02-02
[Technical] Remove remote client retrieval from DI

abelgardep:
Previously, our source of clients was the SingleSessionManager and the client for the session was retrieved when opening the app/switching accounts. It caused some issues because sometimes to retrieve the client, a blocking function(to retrieve the credentials from the AccountManager) needs to be called. And it was done on the main thread when initializing the DI.

From now on, we will fetch the client on demand. So we don't need to reinitialize the DI each time we add/switch users. The most recent client will be cached on the [ClientManager](https://github.com/owncloud/android/blob/master/owncloudData/src/main/java/com/owncloud/android/data/ClientManager.kt) and we will retrieve that client on each RemoteDataSource (At this point, we are on a coroutine out of the main thread).

## Related Issues
Potential fix to https://github.com/owncloud/android/issues/3216
Potential fix to https://github.com/owncloud/android/issues/3747

Potential fix to crashes related to `com.owncloud.android.lib.common.accounts.AccountUtils.getCredentialsForAccount`

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RemoteDataSourceModule.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/dependecyinjection/RemoteDataSourceModule.kt
@@ -47,8 +47,6 @@ import com.owncloud.android.data.webfinger.datasources.implementation.OCWebfinge
 import com.owncloud.android.lib.common.ConnectionValidator
 import com.owncloud.android.lib.common.OwnCloudAccount
 import com.owncloud.android.lib.common.SingleSessionManager
-import com.owncloud.android.lib.resources.files.services.ChunkService
-import com.owncloud.android.lib.resources.files.services.implementation.OCChunkService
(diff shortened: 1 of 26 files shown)


JuancaG05 (approved):
This was really necessary! Great improvement! 🎉

jesmrec:
Stuff related with shares, sharees and capabilities is OK. But, some credentials lost happened in browser authentication accounts (OAuth2, OIDC). I got same logs, issue starts to happen about line 900:

[owncloud.2023-02-02.log](https://github.com/owncloud/android/files/10576625/owncloud.2023-02-02.log)

OTOH, there are many conflicts in this branch with `master`. Should they be fixed before testing?

jesmrec:
> Stuff related with shares, sharees and capabilities is OK. But, some credentials lost happened in browser authentication accounts (OAuth2, OIDC). I got same logs, issue starts to happen about line 900:

problem fixed with the latest commit. At this point, lost credentials problem is not reproducible anymore.

Issue #3808, opened 2022-12-12
[SPACES] Adapt Scoped Storage to Space's folder structure. Check Download action


jesmrec:
New spaces accounts will have another level of storage in Scoped Storage. From the account level will hang the space level. Inside it, the downloaded files and folders will be stored.

user1@myserver.com
|_ Personal
&emsp; |_ Folder1
&emsp; |_ Folder2
|_ Shares
|_ Space1
&emsp; |_ ...
|_ Space2
&emsp; |_ ...

Existing oCIS accounts should move to /Personal the existing stuff.

After migrating the content to the new location, we will need to update the database(storagePath field) accordingly.

Download action to add here, as check for the new locations.

Add space id to DB if space is personal ¿?

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Modify the path to save files
  - [ ] Migrate DB to include the new path.
  - [ ] Migrate existing ones to new location.
  - [ ] Unit tests
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App  https://github.com/owncloud/android/pull/3891
- Library (if needed)  https://github.com/owncloud/android-library/pull/537

abelgardep:
I've been checking how to introduce the hierarchy of the spaces on android storage, and I got some questions.

First, we thought about the "ideal" solution with space's name:

```
user1@myserver.com
|_ Personal
  |_ Folder1
  |_ Folder2
|_ Shares
|_ Space1
  |_ ...
|_ Space2
  |_ ...
```
The main issue here comes when we have several spaces with the same name.

Apparently, the only unique key to identify a space is the `id` so, the proper hierarchy would be like that:

```
user1@myserver.com
|_ space_id_151294u12
  |_ Folder1
  |_ Folder2
|_ space_id_2519284
|_ space_id_3184912
  |_ ...
|_ space_id_41u24124
  |_ ...
```
This is not human-readable. At the moment, we store the files on scoped storage, so users won't see the actual hierarchy, and we would be able to identify each space internally. So we could use the spaceId in the hierarchy (AFAIK).

The only issue I can think of at the moment is if we decide to open the storage in the future (somehow), users won't be able to identify the spaces by the space id in the hierarchy.

jesmrec:
a mixture would be posible? for example:

space name: `maths`
space id:  `345345baf54454....`

[...]

PR #3891, opened 2023-02-02
[Spaces] Local Storage

abelgardep:
## Related Issues
Implements https://github.com/owncloud/android/issues/3808

Library PR (if needed): https://github.com/owncloud/android-library/pull/537
_____

### DIFF
--- a/owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
@@ -62,12 +62,10 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
         remotePath: String,
         spaceId: String?,
     ): String {
-        val defaultPathWithoutSpace = getAccountDirectoryPath(accountName) + remotePath
-
         return if (spaceId != null) {
-            defaultPathWithoutSpace + getEncodedSpace(spaceId) + File.separator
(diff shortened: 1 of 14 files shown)


owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
```diff
@@ -58,9 +58,16 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
      * file.
      */
     fun getDefaultSavePathFor(
-        accountName: String?,
-        remotePath: String
-    ): String = getAccountDirectoryPath(accountName) + remotePath
+        accountName: String,
+        remotePath: String,
+        spaceId: String?,
```
JuancaG05, line 63: Maybe setting this parameter with default value `null`? Just like in `getTemporalPath`

JuancaG05 (approved):
Great! Just a tiny comment, but from my point of view, ready to go 🚀

owncloudData/src/main/java/com/owncloud/android/data/storage/LocalStorageProvider.kt
abelgardep: No default value is needed for this one. We want to make sure that we receive the spaceId, even if it's null.

I would change the one from `getTemporalPath` though

jesmrec:
Approved from my side

Issue #3881, opened 2023-01-31
[SPACES] Support file and folders operations with Spaces in DocumentsProvider


JuancaG05:
After adapting DocumentsProvider to show and browse available spaces with basic operations (create folder, remove folder/file and rename folder/file) in https://github.com/owncloud/android/issues/3809, we have to support the other operations with files and folders: upload, download, copy and move.

### TASKS

- [x] Research (if needed)
- [x] Create branch spaces/doc_provider_operations
- [x] Development tasks
  - [x] Support upload operation
  - [x] Support download operation
  - [x] Support copy operation
  - [x] Support move operation
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch spaces/doc_provider_operations into spaces/main

abelgardep:
Done, will be available on v4.0

Issue #3675, opened 2022-05-25
[BUG] "Copy to" fails sometimes in document provider


jesmrec:
### Steps to reproduce
1. Open any app with document provider support 
2. Select an external provider (f. ex Downloads)
3. Select and item and "Copy to"
4. As target, an oC account

### Actual behaviour
Operation does not always work, returning an error in notification

### Expected behaviour
Operation works fine and file is copied always to the correct location

Issue #2659, opened 2019-09-13
[BUG] Move files to a different oC account using the document provider doesn't work as expected


davigonz:
### Actual behaviour
- Notification Couldn't move 3 items appears.

### Expected behaviour
- Files are properly moved.
 
### Steps to reproduce
1. Attach two accounts to oC app
2. Open Files/Download app to work with document provider
3. Select several files (i selected three)
4. Move the selected files to the another oC account

### Environment data
Android version: 9

Device model: Xiaomi MI A2

Stock or customized system: Android One

#### ownCloud Android app log

`
[...]

Issue #2658, opened 2019-09-13
[BUG] Copy a folder with subfolders to different accounts doesn't work as expected


davigonz:
### Actual behaviour
- Folder, subfolders and containing files are not copied.

### Expected behaviour
- Folder, subfolders and containing files are properly copied.
 
### Steps to reproduce
1. Open Files app or other app supporting the document provider.
2. Select an ownCloud account from the side menu.
3. Long press a folder with subfolders and files inside and press copy to in the three-dots menu.
4. Select a different ownCloud account as destination to copy the files.

### Environment data
Android version: 10

Device model: Pixel 3 XL

Stock or customized system: stock

ownCloud app version: - master

#### ownCloud Android app log
```
2019-09-13 14:46:56.002 8069-9398/com.owncloud.android.debug E/CreateFolderOperation: /Test/ hasn't been created
2019-09-13 14:46:56.004 8069-9398/com.owncloud.android.debug E/DatabaseUtils: Writing exception to parcel
    android.os.ParcelableException: java.io.FileNotFoundException: Remote Operation failed
        at android.provider.DocumentsProvider.call(DocumentsProvider.java:1069)
        at android.content.ContentProvider.call(ContentProvider.java:2152)
        at android.content.ContentProvider$Transport.call(ContentProvider.java:477)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:277)
        at android.os.Binder.execTransactInternal(Binder.java:1021)
        at android.os.Binder.execTransact(Binder.java:994)
     Caused by: java.io.FileNotFoundException: Remote Operation failed
        at com.owncloud.android.providers.DocumentsStorageProvider.checkOperationResult(DocumentsStorageProvider.kt:377)
        at com.owncloud.android.providers.DocumentsStorageProvider.createFolder(DocumentsStorageProvider.kt:393)
        at com.owncloud.android.providers.DocumentsStorageProvider.createDocument(DocumentsStorageProvider.kt:299)
        at android.provider.DocumentsProvider.callUnchecked(DocumentsProvider.java:1121)
        at android.provider.DocumentsProvider.call(DocumentsProvider.java:1067)
        at android.content.ContentProvider.call(ContentProvider.java:2152) 
        at android.content.ContentProvider$Transport.call(ContentProvider.java:477) 
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:277) 
        at android.os.Binder.execTransactInternal(Binder.java:1021) 
[...]

PR #3941, opened 2023-02-28
[SPACES] Support file and folders operations with Spaces in DocumentsProvider 

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3881
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/DocumentsStorageProvider.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/documentsprovider/DocumentsStorageProvider.kt
@@ -466,6 +466,7 @@ class DocumentsStorageProvider : DocumentsProvider() {
         // File will be created at [openDocument] method.
         val tempDir = File(FileStorageUtils.getTemporalPath(parentDocument.owner, parentDocument.spaceId))
         val newFile = File(tempDir, displayName)
+        newFile.parentFile?.mkdirs()
         fileToUpload = OCFile(
             remotePath = parentDocument.remotePath + displayName, mimeType = mimeType, parentId = parentDocument.id, owner = parentDocument.owner, spaceId = parentDocument.spaceId
         ).apply {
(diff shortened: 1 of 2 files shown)


jesmrec:
One open question here:

"Move to other space in same account" operation is banned in the app, but available in doc provider. 

i guess that it's not straightforward to prevent this... but it is a shortcut to jump over the app restriction. Is this OK?

CC @michaelstingl

JuancaG05:
In Android, since we use the native documents provider, we cannot restrict this behaviour.

In fact, the move operation is not performed as it is, since we invoke the move usecase and an exception is thrown if we try to move to a different space. When this happens, the documents provider catches internally the exception and performs an "artificial" move, which means downloading the file from the source, upload it to the destination and deleting it from the source. This artificial move is performed as well when trying to move between accounts through the documents provider -> not allowed in the app but from the document provider.

In conclusion, a remote move is not performed in server, but we cannot restrict it since it is a native and internal behaviour from Android. We'll see if this is doable in future versions of documents provider.

/CC @michaelstingl

jesmrec:
A complete sanity test was done over the feature, including operations over oCIS accounts with spaces, oC10 accounts and other providers.

Some open issues are still reproducible: 

https://github.com/owncloud/android/issues/3675
https://github.com/owncloud/android/issues/2659
https://github.com/owncloud/android/issues/2658

These three, related with `Copy` and `Move` operations (read, upload and delete under the hood) are still there. The most unstable scenario is copying/moving tree structures with more than 2-deep levels. Sometimes works, sometimes fails, sometimes is only copied or moved a part of the files. If the operation is done over two different oC accounts/providers (no matter if oC10 or oCIS), it's unstable even with a few amount of items.  My guess: some internal memory restrictions (did not find anything in [official documentation](https://developer.android.com/guide/topics/providers/document-provider?hl=en) about this)

Keep in mind also this issue with improvements that point to the mentioned issues: https://github.com/owncloud/android/issues/2336

Since those are known issues and do not affect the oCIS' spaces implementation, i do not have objections to move this forward, keeping in mind that some scenarios are not working fine. If any other idea/suggestion/solution to improve the known bad scenarios, is totally welcome.

Issue #3947, opened 2023-03-02
[SPACES] Available offline


abelgardep:
In #3879 we introduced a new way to distinguish between uploads to different spaces.

We could extend that to available offline files to distinguish them easily. Otherwise, there is no way to distinguish them in the available offline section.

Options:
1. Add the space on the third row and show the entire path (including the name as it is atm)
2. Add the space on the third row and show the parent path (not including the name as it is in the uploads section)
3. Add a new row with the space.

| Uploads | Av Offline |
| --- | --- |
| ![uploads_spaces_dif](https://user-images.githubusercontent.com/47524927/222433115-11c2b758-342b-463f-91fb-d15bd9000a6d.png) | ![available_offline_diff_spaces](https://user-images.githubusercontent.com/47524927/222433137-9d6a1096-fed2-4ec3-8cfd-6e6e5df652b5.png) |

### TASKS

- [x] Research (if needed)
- [x] Create branch spaces/available_offline
- [x] Development tasks
  - [x] Show only parent folder path instead of whole path
  - [x] Add space name label and icon
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch spaces/available_offline into master

abelgardep:
I would go with option 2.

Option 1: space + full path is too long and we are already showing the name in the first row
Option 2: space + parent path will be kind of similar in length to the current row
Option 3: New row is too much from my point of view

jesmrec:
Number 2 is most congruent with the current stuff

PR #3968, opened 2023-03-14
[SPACES] Available offline

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3947
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
@@ -220,7 +220,7 @@ class FileListAdapter(
                             }
                             fileItem.space?.let { space ->
                                 it.fileSpaceIcon.isVisible = true
-                                it.fileSpaceName.isVisible= true
+                                it.fileSpaceName.isVisible = true
                                 if (space.isPersonal) {
                                     it.fileSpaceIcon.setImageResource(R.drawable.ic_folder)
                                     it.fileSpaceName.text = personalName
@@ -231,7 +231,7 @@ class FileListAdapter(
                         } else {
                             it.fileListPath.isVisible = false
                             it.fileSpaceIcon.isVisible = false
-                            it.fileSpaceName.isVisible= false
+                            it.fileSpaceName.isVisible = false
                         }
 
                     }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
@@ -2,7 +2,9 @@
  * ownCloud Android client application
  *
  * @author Fernando Sanz Velasco
- * Copyright (C) 2021 ownCloud GmbH.
+ * @author Juan Carlos Garrote Gascón
+ *
+ * Copyright (C) 2023 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -15,7 +17,6 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
  */
 
 package com.owncloud.android.presentation.files.filelist
@@ -41,6 +42,7 @@ import com.owncloud.android.datamodel.ThumbnailsCacheManager
 import com.owncloud.android.domain.files.model.FileListOption
 import com.owncloud.android.domain.files.model.OCFileWithSyncInfo
 import com.owncloud.android.domain.files.model.OCFooterFile
+import com.owncloud.android.domain.spaces.model.OCSpace
 import com.owncloud.android.presentation.authentication.AccountUtils
 import com.owncloud.android.utils.DisplayUtils
 import com.owncloud.android.utils.MimetypeIconUtil
@@ -52,15 +54,21 @@ class FileListAdapter(
     private val isPickerMode: Boolean,
     private val layoutManager: StaggeredGridLayoutManager,
     private val listener: FileListAdapterListener,
+    private val personalName: String,
 ) : SelectableAdapter<RecyclerView.ViewHolder>() {
 
     var files = mutableListOf<Any>()
     private var account: Account? = AccountUtils.getCurrentOwnCloudAccount(context)
     private var fileListOption: FileListOption = FileListOption.ALL_FILES
 
-    fun updateFileList(filesToAdd: List<OCFileWithSyncInfo>, fileListOption: FileListOption) {
+    fun updateFileList(filesToAdd: List<OCFileWithSyncInfo>, fileListOption: FileListOption, spaces: List<OCSpace>) {
+        val fileItems = filesToAdd.map { fileWithSyncInfo ->
+            val space = spaces.firstOrNull { it.id == fileWithSyncInfo.file.spaceId && it.accountName == fileWithSyncInfo.file.owner }
+            FileItem(fileWithSyncInfo, space)
+        }
+
         val listWithFooter = mutableListOf<Any>()
-        listWithFooter.addAll(filesToAdd)
+        listWithFooter.addAll(fileItems)
 
         if (listWithFooter.isNotEmpty()) {
             listWithFooter.add(OCFooterFile(manageListOfFilesAndGenerateText(filesToAdd)))
@@ -133,7 +141,7 @@ class FileListAdapter(
                 layoutManager.spanCount == 1 -> {
                     ViewType.LIST_ITEM.ordinal
                 }
-                (files[position] as OCFileWithSyncInfo).file.isImage -> {
+                (files[position] as FileItem).fileWithSyncInfo.file.isImage -> {
                     ViewType.GRID_IMAGE.ordinal
                 }
                 else -> {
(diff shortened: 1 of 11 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -203,10 +214,27 @@ class FileListAdapter(
                         it.Filename.text = file.fileName
                         it.fileListSize.text = DisplayUtils.bytesToHumanReadable(file.length, context)
                         it.fileListLastMod.text = DisplayUtils.getRelativeTimestamp(context, file.modificationTimestamp)
-                        it.fileListPath.apply {
-                            text = file.remotePath
-                            isVisible = !fileListOption.isAllFiles()
+                        if (fileListOption.isAvailableOffline()) {
+                            it.fileListPath.apply {
+                                text = File(file.remotePath).parent
```
abelgardep, line 219: There is a function at file to retrieve the parent path.

```suggestion
                                text = file.getParentRemotePath()
```

owncloudApp/src/main/res/layout/item_file_list.xml
```diff
@@ -95,6 +95,37 @@
             app:layout_constraintTop_toBottomOf="@id/Filename"
             tools:text="Mod Date" />
 
+        <ImageView
```
abelgardep, line 98: This one could be extracted to a new component and used wherever we need it. At the moment, we have the same item at two different places and we will need to update both to keep consistency.

A new component could keep the icon, space name, and parent path. Just an idea..

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListViewModel.kt
```diff
@@ -82,6 +83,10 @@ class MainFileListViewModel(
     private val sortTypeAndOrder = MutableStateFlow(Pair(SortType.SORT_TYPE_BY_NAME, SortOrder.SORT_ORDER_ASCENDING))
     val space: MutableStateFlow<OCSpace?> = MutableStateFlow(null)
 
+    private val _spaces: MutableStateFlow<List<OCSpace>> = MutableStateFlow(emptyList())
+    val spaces: StateFlow<List<OCSpace>>
+        get() = _spaces
+
```
abelgardep, line 89: At the moment, we have an observable for the `currentFolderDisplayed`, another one for its `space`, and now we have an observable for every single space in the app, not only the ones from this account, to match each file with the space it belongs.

I think that we could introduce the space associated with each file within a wrapper as we already did with `OCFileWithFileSync`. It could contain also the space associated. There are already foreign keys, so the database is already prepared for this. If we introduce this, we could get rid of the space, and spaces observables since they would not be needed as we already had the space info for each file 🤔

abelgardep (changes requested):
Some suggestions here 👍 @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -203,10 +214,27 @@ class FileListAdapter(
...
+                                isVisible = true
+                            }
+                            fileItem.space?.let { space ->
+                                it.fileSpaceIcon.isVisible = true
+                                it.fileSpaceName.isVisible = true
+                                if (space.isPersonal) {
+                                    it.fileSpaceIcon.setImageResource(R.drawable.ic_folder)
+                                    it.fileSpaceName.text = personalName
```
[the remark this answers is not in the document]
JuancaG05: Aaaah this is nice! Didn't think about that, let's change it on `TransfersAdapter` as well!

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -203,10 +214,27 @@ class FileListAdapter(
                         it.Filename.text = file.fileName
                         it.fileListSize.text = DisplayUtils.bytesToHumanReadable(file.length, context)
                         it.fileListLastMod.text = DisplayUtils.getRelativeTimestamp(context, file.modificationTimestamp)
-                        it.fileListPath.apply {
-                            text = file.remotePath
-                            isVisible = !fileListOption.isAllFiles()
+                        if (fileListOption.isAvailableOffline()) {
```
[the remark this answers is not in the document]
JuancaG05: Ok, I'll change it, but instead of `!fileListOption.isAllFiles()`, I will add `fileListOption.isSharedByLink()` because there's no need to show the path in the Shares section for oCIS users (files will be always located in the root folder of the Shares space).

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
JuancaG05: Aaah nice! Didn't know about this 👍

owncloudApp/src/main/res/layout/item_file_list.xml
JuancaG05: I extracted this to a new component, `space_path_line.xml` 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListViewModel.kt
JuancaG05: This is a bit complex and implies some changes throughout the code. I will create a new issue to address this 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -203,9 +212,27 @@ class FileListAdapter(
...
                         it.fileListLastMod.text = DisplayUtils.getRelativeTimestamp(context, file.modificationTimestamp)
-                        it.fileListPath.apply {
-                            text = file.remotePath
-                            isVisible = !fileListOption.isAllFiles()
+                        if (fileListOption.isAvailableOffline() || (fileListOption.isSharedByLink() && fileItem.space == null)) {
+                            it.spacePathLine.path.apply {
+                                text = file.getParentRemotePath()
+                                isVisible = true
+                            }
+                            if (fileListOption.isAvailableOffline()) {
```
abelgardep, line 220: This `if` is redundant, right? 

if it's `availableOffline` this is true.
If it's `isSharedByLink`, space will be null, so the s`pace?.let` clause won't be executed anyway

🤔

abelgardep (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListViewModel.kt
```diff
@@ -67,9 +67,10 @@ class MainFileListViewModel(
     private val getFileByRemotePathUseCase: GetFileByRemotePathUseCase,
     private val getSpaceWithSpecialsByIdForAccountUseCase: GetSpaceWithSpecialsByIdForAccountUseCase,
     private val sortFilesWithSyncInfoUseCase: SortFilesWithSyncInfoUseCase,
+    private val synchronizeFolderUseCase: SynchronizeFolderUseCase,
+    private val getSpacesFromEveryAccountUseCase: GetSpacesFromEveryAccountUseCase,
```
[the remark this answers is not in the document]
JuancaG05: vThere's no specific use case for that. There's a query in `SpacesDao`, but no use case associated to it. IMO, at this point the best is to keep this as technical debt because in the short term this new issue will be addressed: https://github.com/owncloud/android/issues/3971, making this being deleted.

owncloudData/src/main/java/com/owncloud/android/data/spaces/db/SpacesDao.kt
```diff
@@ -129,7 +129,7 @@ interface SpacesDao {
         private const val SELECT_ALL_SPACES = """
             SELECT *
             FROM ${ProviderMeta.ProviderTableMeta.SPACES_TABLE_NAME}
-            WHERE $SPACES_DRIVE_TYPE LIKE '$DRIVE_TYPE_PROJECT' OR $SPACES_DRIVE_TYPE LIKE '$DRIVE_TYPE_PERSONAL'
+            WHERE $SPACES_DRIVE_TYPE LIKE '$DRIVE_TYPE_PROJECT' OR $SPACES_DRIVE_TYPE LIKE '$DRIVE_TYPE_PERSONAL' OR $SPACES_ID LIKE '$SPACE_ID_SHARES'
```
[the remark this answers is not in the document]
JuancaG05: Same as in the previous comment. We'll keep it like this for the moment. Anyway, it's only one additional space (the Shares one).

I added some TODO's to keep it in mind.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
JuancaG05: True 💯

abelgardep (approved):

jesmrec:
QA checks

oCIS

- [X] Av. offline file in root of personal
- [X] Av. offline folder in root of personal
- [X] Av. offline file in non-root of personal
- [X] Av. offline folder in non-root of personal
- [X] Av. offline file in root of custom space
- [X] Av. offline folder in root of custom space
- [X] Av. offline file in non-root of custom space
- [X] Av. offline folder in non-root of custom space

oC10 (regressive behaviour)

- [X] Av. offline file in root of personal
- [X] Av. offline folder in root of personal
- [X] Av. offline file in non-root of personal
- [X] Av. offline folder in non-root of personal

UI 

- [X] No av. offline stuff
- [X] Long space name (oCIS)
- [X] Long path 
 
Issue detected while testing (regression): https://github.com/owncloud/android/issues/3976

JuancaG05:
> Please, let me know if the conflict fix affects the current behaviour and new round should be passed.

It doesn't, the conflict came from merging https://github.com/owncloud/android/pull/3974, but it has nothing to do with this, so this is ready to be moved on 👍

Issue #3617, opened 2022-04-11
[TECHNICAL] Bump target SDK to Android 13


abelgardep:
**Deadlines**
New apps: August 1, 2023
Updates: November 1, 2023

Source: https://support.google.com/googleplay/android-developer/answer/11926878

Android 13 overview: https://developer.android.com/about/versions/13/overview

All apps: https://developer.android.com/about/versions/13/behavior-changes-all
Target 33: https://developer.android.com/about/versions/13/behavior-changes-13

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App https://github.com/owncloud/android/pull/3972
- Library (if needed) https://github.com/owncloud/android-library/pull/550

abelgardep:
In Android 13, the notifications are off by default. And once we target Android 13 #3617 the app will have complete control over when the permission dialog is displayed. 

AFAIK, this is the only runtime permission we request that involves core functionality. But in the future could be more. So, it could be an option to ask for it within wizard #3955  explaining to the user why we need to request this permission. 
https://developer.android.com/develop/ui/views/notifications/notification-permission#new-apps

Also, there are some recommendations about permission requests:
https://developer.android.com/develop/ui/views/notifications/notification-permission#best-practices

abelgardep:
[Behavior changes: Apps targeting Android 13 or higher](https://developer.android.com/about/versions/13/behavior-changes-13)
## Privacy
- [X] Notification permission affects foreground service appearance (Notification dialog is shown and manage notification row added in Settings screen)
- [ ] ~~New runtime permission for nearby Wi-Fi devices~~ oC does not need that permission
- [ ] ~~Granular media permissions~~ We get files via SAF
- [ ] ~~Use of body sensors in the background requires new permission~~ oC does not need that permission
## Performance and battery
- [X] Battery Resource Utilization should have no impact in the current app
## User experience
- [ ] Media controls derived from PlaybackState No real impact atm
- [ ] ~~App color theme applied automatically to WebView content~~ No dark theme (yet), so should have no impact
## Google Play services
- [ ] ~~Permission required for advertising ID~~ no ads
## Updated non-SDK restrictions
No impact

PR #3972, opened 2023-03-16
[TECHNICAL] Bump target sdk to 33

abelgardep:
## Related Issues
App: #3617 

Library PR: https://github.com/owncloud/android-library/pull/550

_____

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         // SDK
         sdkCompileVersion = 33
         sdkMinVersion = 21
-        sdkTargetVersion = 31
+        sdkTargetVersion = 33
 
         // Android jetpack
         androidxArchCore = "2.2.0"
--- a/changelog/unreleased/3973
+++ b/changelog/unreleased/3973
@@ -1,0 +1,8 @@
+Change: Bump target SDK to 33
+
+Target SDK was upgraded to 33 to keep the app updated with the latest android changes.
+A new setting was introduced to manage notifications in an easier way.
+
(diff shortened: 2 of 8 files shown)


JuancaG05 (approved):

jesmrec:
QA

Notifications in...

- [ ] API < 26 ->  No notifications menu in `Settings`
- [ ] API >= 26 && API < 33 -> Notifications menu in `Settings` but no popup on first start 
- [ ] API >= 33 -> Popup on first start

jesmrec:
About the notification dialog in API >= 33

If notifications are disabled in `Settings`, after moving to the file list, the notification is asked again. In steps:

1. Install app from scratch and add account
2. Notifications' dialog displayed -> allow them
3. Go to `Settings` and open `Manage notifications` 
4. Disable notifications
5. Go the file list

Current: Asked again to allow notifications. Are we managing this? if happens every time the notification status switches from enabled to disabled.

Expected: not sure. Only asking the very first time? 

Open discussion.

abelgardep:
We trust the [shouldShowRequestPermissionRationale](https://developer.android.com/reference/androidx/core/app/ActivityCompat#shouldShowRequestPermissionRationale(android.app.Activity,%20java.lang.String)) method. If it returns true, it means that the user may be confused about the permission. Showing a specific screen showing permission is needed would be great, but in this case, we just ask the user about it.

It the user denies it, we won't ask for it again. Only if the user enables them, and then disables them. Not sure how common would be this case.

Issue #3971, opened 2023-03-15
[TECHNICAL] [SPACES] Create entity for OCFile + OCFile sync info + OCSpace


JuancaG05:
There are several cases in which we retrieve all the spaces from database just to be able to associate each file to its space. For example, `FileListAdapter` receives from `MainFileListFragment` a list with all the spaces, provided by `MainFileListViewModel`. The same happens with `TransfersAdapter`, `TransfersListFragment` and `TransfersViewModel`, respectively.

A better solution would consist in retrieve the space all together with the file, so that we don't have to call again the database and retrieve all the spaces we have saved for all accounts. This could be done by adding the space entity to `OCFileWithSyncInfo`, which would help removing the observable with the list of spaces in the previous mentioned ViewModels.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch spaces/entity_file_space
- [ ] Development tasks
  - [ ] Add space to `OCFileWithSyncInfo` (and all the layers behind)
  - [ ] Adapt code to the new entity that already contains the space
  - [ ] Remove observables with the list of spaces
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch spaces/entity_file_space into master

PR #3973, opened 2023-03-17
Add space to OCFileWithSyncInfo

abelgardep:
## Related Issues
App: https://github.com/owncloud/android/issues/3971
_____

### DIFF
--- a/owncloudDomain/src/main/java/com/owncloud/android/domain/files/model/OCFileWithSyncInfo.kt
+++ b/owncloudDomain/src/main/java/com/owncloud/android/domain/files/model/OCFileWithSyncInfo.kt
@@ -20,14 +20,13 @@
 
 package com.owncloud.android.domain.files.model
 
-import android.os.Parcelable
-import kotlinx.parcelize.Parcelize
+import com.owncloud.android.domain.spaces.model.OCSpace
 import java.util.UUID
 
-@Parcelize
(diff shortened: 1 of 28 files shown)


owncloudDomain/src/main/java/com/owncloud/android/domain/files/model/OCFileWithSyncInfo.kt
JuancaG05, line 1: I would change the name of this... Maybe `OCFileWithCompleteInfo`? IMHO, `OCFilePopulated` doesn't convince me since it seems that `OCFile` is the same version but with empty properties, so I don't know. Maybe you can suggest a better name of course 👍

JuancaG05 (changes requested):
A tiny suggestion here @abelgardep. Also, as mentioned in the issue, now that we have this we can maybe clean the transfers VM, fragment and adapter as well 👍

owncloudDomain/src/main/java/com/owncloud/android/domain/files/model/OCFileWithSyncInfo.kt
JuancaG05: Ok, seems this implies too many changes, we can let it stay for the moment

jesmrec:
Technical one, CI green and basic tests passed. Ready to go.

Issue #3890, opened 2023-02-02
[SPACES] WebDAV permissions


jesmrec:
Every pair space-member is led by a permission. Three different permissions are available for a space:

- Viewer: Download and preview
- Editor: Upload, edit, delete, download and preview
- Manager:  Upload, edit, delete, download, preview and share

Depending on the member permission, some actions will be allowed, other ones will be banned. Info fetched from the `drives` endpoint, `permissions` array:

```
"permissions" : [
               {
                  "grantedTo" : [
                     {
                        "user" : {
                           "id" : "52a81e9f-1c64-4384-8371-b1ebafd14d53"
                        }
                     }
                  ],
                  "roles" : [
                     "manager"
                  ]
               },
               {
                  "grantedTo" : [
                     {
                        "user" : {
                           "id" : "b7ea9393-fd39-4627-a3c7-11ad55cdcf11"
                        }
                     },
                     {
                        "user" : {
                           "id" : "f4cc5b14-7bf2-45c1-b470-e3d94969083e"
                        }
                     }
                  ],
                  "roles" : [
                     "editor"
                  ]
               },
               {
                  "grantedTo" : [
                     {
                        "user" : {
                           "id" : "2c487709-18db-4e48-8227-99af69274db6"
                        }
                     }
                  ],
                  "roles" : [
                     "viewer"
                  ]
               }
            ]
```

Depending on the permission level

| | Viewer | Editor | Manager |
| :-: | :-: | :-: | :-: |
| Create folder | ❌  | ✅  | ✅ |
| Upload | ❌  | ✅  | ✅ |
| Download | ✅  | ✅  | ✅ |
| Preview | ✅  | ✅  | ✅ |
| Set as av. offline | ✅  | ✅  | ✅ |
| Open with | ✅  | ✅  | ✅ |
| Send | ✅   | ✅  | ✅ |
| Move | ❌  | ✅  | ✅ |
| Copy |   **  | ✅  | ✅ |
| Rename | ❌  | ✅  | ✅ |
| Remove | ❌  | ✅  | ✅ |
| Share | ❌  | ❌   | ✅ |

** Only to other space with enough permission. Maybe it's OK as ❌ 

That means: inside a space there should be available only those actions that are allowed. It should be refreshed somehow, because the permission level is not static.

Extra ball, just an idea: in the sticky header, show with a small icon in the corner the permission level of the user in the space. F. ex:

- If it is viewer, show an eye icon
- If it is editor, a pencil icon 
[...]

PR #3978, opened 2023-03-20
[SPACES] WebDAV permissions in folders

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3890
In this PR we handle permissions related to folders (for oC10 as well). Specifically, we handle these cases:
- In the main file list, if we don't have `C` permission (add new files) in the current folder we hide the "Upload" mini FAB, if we don't have `K` permission (add new subdirectories) we hide the "New folder" mini FAB, and if we have none of them, we hide the main FAB as well
- In the documents provider, if we don't have the `C` permission in the current folder we hide the choose folder buttons (for copy and move), and if we don't have the `K` permission we disable the "New folder" option
- In the folder picker, if we don't have the `C` permission in the current folder we hide the choose folder button (for copy, move and sharing with oC) and we show a message, and if we don't have the `K` permission we hide the "New folder" icon button
_____

## QA

TP: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_4.0/Permissions.md

Reports:

- [X] (1) Share from other apps with oC10 allows to select restricted folders https://github.com/owncloud/android/pull/3978#issuecomment-1484918964 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -201,7 +201,7 @@ class MainFileListFragment : Fragment(),
             }
         }
 
-        showOrHideFab(requireArguments().getParcelable(ARG_FILE_LIST_OPTION)!!)
+        showOrHideFab(requireArguments().getParcelable(ARG_FILE_LIST_OPTION)!!, requireArguments().getParcelable(ARG_INITIAL_FOLDER_TO_DISPLAY)!!)
     }
 
     private fun toggleSelection(position: Int) {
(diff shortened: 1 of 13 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -348,30 +348,37 @@ class MainFileListFragment : Fragment(),
...
         } else {
             toggleFabVisibility(true)
-
+            if (currentFolder.permissions != null) {
+                if (!currentFolder.hasAddFilePermission) {
+                    binding.fabUpload.isVisible = false
+                } else if (!currentFolder.hasAddSubdirectoriesPermission) {
+                    binding.fabMkdir.isVisible = false
+                }
+            }
```
abelgardep, line 374: This logic and the 
```
|| (currentFolder.permissions != null && !currentFolder.hasAddFilePermission && !currentFolder.hasAddSubdirectoriesPermission
```
looks like they are redundant and could be unified. At this point, we show the fab and then we hide it depending on permissions. And the logic I think it would be permissions and then check filelistoption only in case there are permsisions

Also, we do not need to register the fab listeners in case they are hidden.

abelgardep (changes requested):
Some suggestions here @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -310,6 +316,20 @@ open class FolderPickerActivity : FileActivity(),
...
+            findViewById<TextView>(R.id.folder_picker_no_permissions_message).isVisible = !it
+        }
+        findViewById<AppCompatImageView>(R.id.view_type_selector).let {
+            if (currentFolder.hasAddSubdirectoriesPermission) {
+                it.visibility = VISIBLE
+            } else {
+                it.visibility = INVISIBLE
+            }
+        }
+    }
```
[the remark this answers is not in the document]
JuancaG05: Ok, I'll manage it externally from `MainFileListFragment` 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
JuancaG05: The alternative would be repeating the `toggleFabVisibility(false)` logic instead of the conditions logic, and also the main FAB would be made visible and then maybe invisible again if we go into the "else", so I think we have the same logic structure problem in both cases :S

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -367,16 +368,14 @@ class MainFileListFragment : Fragment(),
...
-            if (currentFolder.permissions != null) {
-                if (!currentFolder.hasAddFilePermission) {
-                    binding.fabUpload.isVisible = false
-                } else if (!currentFolder.hasAddSubdirectoriesPermission) {
-                    binding.fabMkdir.isVisible = false
-                }
+            if (!currentFolder.hasAddFilePermission) {
+                binding.fabUpload.isVisible = false
+            } else if (!currentFolder.hasAddSubdirectoriesPermission) {
+                binding.fabMkdir.isVisible = false
```
abelgardep, line 378:
```suggestion
            binding.fabUpload.isVisible = currentFolder.hasAddFilePermission
            binding.fabMkdir.isVisible = currentFolder.hasAddSubdirectoriesPermission
            binding.fabMain.isVisible = current.Folder.hasAddFilePermission || currentFolder.hasAddSubdirectoriesPermission
```

JuancaG05: That's not the same logic! Now only one of the two mini FABs is hidden here, since if the folder has none of the two permissions, it will enter in the previous `if`, which is also in charge for hiding the main FAB with `toggleFabVisibility(false)`

jesmrec:
### (1)

1. in oC10, user1 shares folder with user2 removing edit permission
2. user2 opens Gallery, Photos, or any app that contains files
3. user2 shares some photos with oC, selecting the shared folder without permissions

Current: Upload option is available and selectable in folder picker.
Expected: Upload option not available since the folder does not have "W" option.

Note:  Uploads view shows the right error, so the issue is not critical. Just allignment with other cases

Samsung A51, Android12
`7d4002442`

jesmrec:
(1) fixed

Approved on my side. Great job!

PR #3979, opened 2023-03-21
Add specific forbidden exception to transfer errors pool

abelgardep:
## Related Issues
Add a new message to the errors pool when uploading or doing any operation to a space where the user does not have permission.

Also, the `when` clause was sorted alphabetically
_____


JuancaG05 (approved):

abelgardep:
> i mean, a message like No permissions or similar is more accurate than Upload not allowed

I agree 100%. But it is kind of tricky at the moment because there are a lot of strings and translated strings over there. Sometimes we just want to show the reason, sometimes we need to show the action and the reason:

For example:
- When creating a folder: `Folder could not be created because permission error`
- When uploading a file: `Permission error` in the notification message and the list

We can proceed with `permission error` or `no permissions` string, but the problem here is a little bit deeper

jesmrec:
i understand... if we can get `permission error`, it's OK. 

So, messaging will need a look too.

PR #3997, opened 2023-03-28
[SPACES] WebDAV permissions in files

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3890

In this PR we handle permissions related to files (for oC10 as well). Specifically, we handle these cases:
- When a file doesn't have the `W` permission (write), we show "Open with (read only)" instead of "Open with", in the main file list with only the file selected, in the details view and in its preview. Also, this file cannot be modified by external apps since we send it without the write permission flag. NOTE: oC10 always sends `W` in the permissions although the file doesn't actually have it, so the error is handled in backend in this case (error message shown in the app)
- When a file or folder doesn't have the `R` permission (reshare), we hide the "Share" option in the main file list with only the file selected, in the details view and in its preview. NOTE: in oCIS accounts, at the moment it is not possible to share files from a space that is not the personal, so this option won't appear anyway

The next points do not apply to oC10 since there, a user can have lack of permissions over a file or folder only when it is shared with them, and when this occurs, a new synchronized copy of the file is created for them, so that they will always have these permissions because the operations apply to their local copy:
- When a file (or folder) doesn't have the `D` permission (delete), we hide the "Remove" option in the main file list with the file selected (also if it is selected among some others), in the details view and in its preview
- When a file (or folder) doesn't have the `N` permission (rename), we hide the "Rename" option in the main file list with only the file selected, in the details view and in its preview (in the preview wasn't shown before either, it is work to do)
- When a file (or folder) doesn't have the `V` permission (move), we hide the "Move" option in the main file list with only the file selected (also if it is selected among some others)
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/files/FileMenuFilter.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/files/FileMenuFilter.java
@@ -118,7 +118,7 @@ public class FileMenuFilter {
      */
     public void filter(Menu menu, boolean displaySelectAll, boolean displaySelectInverse,
                        boolean onlyAvailableOffline, boolean sharedByLinkFiles, boolean hasWritePermission,
-                       boolean hasDeletePermission) {
+                       boolean hasDeletePermission, boolean hasRenamePermission) {
         if (mFiles == null || mFiles.size() <= 0) {
             hideAll(menu);
 
@@ -127,7 +127,7 @@ public class FileMenuFilter {
             List<Integer> toHide = new ArrayList<>();
 
             filter(toShow, toHide, displaySelectAll, displaySelectInverse, onlyAvailableOffline, sharedByLinkFiles,
-                    hasDeletePermission);
+                    hasDeletePermission, hasRenamePermission);
 
             MenuItem item;
             for (int i : toShow) {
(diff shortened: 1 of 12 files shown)


owncloudApp/src/main/java/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -117,22 +117,32 @@ public FileMenuFilter(List<OCFile> targetFiles, Account account, ComponentsGette
      * @param menu Options or context menu to filter.
      */
     public void filter(Menu menu, boolean displaySelectAll, boolean displaySelectInverse,
-                       boolean onlyAvailableOffline, boolean sharedByLinkFiles) {
+                       boolean onlyAvailableOffline, boolean sharedByLinkFiles, boolean hasWritePermission,
+                       boolean hasDeletePermission, boolean hasRenamePermission, boolean hasMovePermission,
+                       boolean hasResharePermission) {
```
abelgardep, line 122: The `FileMenuFilter` already maintains the list of files we need to filter. We could get the permissions from there instead of adding 5 new parameters, right?

abelgardep (changes requested):
A suggestion here @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/files/FileMenuFilter.java
JuancaG05: Yes, it's true. I removed them and accessed their value directly from `FileMenuFilter` 👍

abelgardep (approved):

jesmrec:
Some comments about QA, related with the exposed in the top message

- `Can Edit` only tested in oCIS, because a potential bug in oC10 (`W` permission listed even though the item does not have such permission). Issue created https://github.com/owncloud/core/issues/40716
- `Can Share` only tested in oC10. Resharing not implemented in oCIS yet.
- Move, Delete and Rename operations, only tested in oCIS, for spaces with no edit permission (viewers)

jesmrec:
Approved on my side

PR #4005, opened 2023-04-05
Settings: Add divider between all entries

mueller-ma:
IMO it looks weird that there's no divider between picture upload and video upload:

grafik

This PR adds a divider there.

## Related Issues
App:

Library PR (if needed):

_____

### DIFF
--- a/owncloudApp/src/main/res/xml/settings.xml
+++ b/owncloudApp/src/main/res/xml/settings.xml
@@ -48,6 +48,7 @@
         app:title="@string/prefs_subsection_picture_uploads" />
 
     <Preference
+        app:allowDividerAbove="true"
         app:fragment="com.owncloud.android.presentation.settings.autouploads.SettingsVideoUploadsFragment"
         app:icon="@drawable/ic_video_uploads"
         app:key="video_uploads_subsection"


jesmrec:
We did not send a divider considering that both are variants of the same Setting. What do you think @JuancaG05 @abelgardep @michaelstingl? Should be divided?

JuancaG05:
Yes, as @jesmrec comments, it was like that originally because of grouping reasons. It may look weird because currently, those are the only 2 items which are stretch-related, but if in a future we add more items here and they are related to some of the ones already there, we'll group them in the same way

JuancaG05:
> IMHO grouping makes sense. No divider please. Otherwise single menu item „media upload”, and move image and video to submenu. (low priority)

Adding a "Media upload" item would add a new level to the settings hierarchy, so that for example, if the user wants to enable the picture auto-uploads, they will have to go to Settings > Media upload > Auto upload pictures >  Picture uploads. IMO this is not necessary at the moment because there's not much complexity, better keep it simple and save a click to the user 👍

mueller-ma:
Ok, now I see why there's no divider between photo and video.

Issue #3995, opened 2023-03-27
[FEATURE REQUEST] Create new file via Open in Web


abelgardep:
As a user, I want to create a new file and open it directly on the web.

On the floating action button, we need a new item for each available file we can create on the web.

https://owncloud.dev/services/app-registry/apps/#creating-a-file-with-the-app-provider

Create mini-fab if providers available for open
Set all options in bottom sheet
Ask for file name
Open in web
Refreshing list of files just after closing file in web

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/create_file_web
- [x] Development tasks
  - [x] Create network call to create file via app provider
  - [x] Create use case to create file via app provider
  - [x] Add new mini FAB and bottom sheet with the different available file options
  - [x] Hide mini FAB if there are no available file options to create
  - [x] When picking the document type in the bottom sheet, ask for file name and call use case
  - [x] Open file in web just after creating it
  - [x] Refresh list of files when closing file in web
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/create_file_web into master

JuancaG05:
Not sure about which title show in the new mini floating button that will allow to create new files and then automatically open them in web.

The iOS app says "New document" as shown in the first screenshot. The other 2 are possible options for the floating button title:

| iOS title | Option 1 | Option 2 |
|---|---|---|
| ![new_file_ios](https://user-images.githubusercontent.com/57049315/232769158-9b736ab8-2039-485b-9b9c-ea90d27ca4ce.png) | ![new_file1](https://user-images.githubusercontent.com/57049315/232768919-fe2726b7-1e72-4eec-9256-96c0fcbf24ea.png) | ![new_file2](https://user-images.githubusercontent.com/57049315/232768975-295b0055-a66d-46bf-a1fc-762836ada91c.png) |

The difference here is that in iOS, the file is only created and not automatically opened in web, and in Android we do open it just after creating it, so it could be a bit weird for the user if the browser is opened with no previous warn. For this, the Option 2 would be more suitable, but we can also set Option 1 and let the user discover it the first time they create a new document.

Also open to other suggestions!

What do you think? @michaelstingl @tbsbdr @jesmrec

michaelstingl:
Let's try option 1, and we'll iterate on it…

JuancaG05:
This is how the document type selector looks like at the moment (just visually, no action linked yet):

[...]

PR #4006, opened 2023-04-05
Replace upload bottom sheet with additional FABs

mueller-ma:
Previously you had to press the FAB => 'Upload' and then select 'Camera' or 'File' in the bottom sheet. Now these two options are included directly in the FAB.

Now:

grafik

Before:
1. 
grafik

2. 
grafik

## Related Issues
App:

Library PR (if needed):

_____


abelgardep:
Hi @mueller-ma

Thanks for the contribution, I understand that the current FAB may not be that useful now that we only have 2 items that we have plans to add new items there.

But with the current way we can introduce new items in an easier way. For example the ones for https://github.com/owncloud/android/issues/3995. My first idea was to introduce a new FAB for file creation and display the list of potential formats in the bottom sheet.

Or in case we introduce the [new photo/video picker](https://developer.android.com/training/data-storage/shared/photopicker) it would be introduced in the bottom sheet.

So, the FAB is something that could receive more importance in the upcoming months.

Thoughts @michaelstingl @JuancaG05 @JuancaG05 ?

jesmrec:
Agreed, we could run into an overpopulated FAB in case of adding there all the options we have/we plan to have. 

- Create folder
- Upload file (2 options inside)
- Create new file (at least 3 options inside)

some hierarchy is useful, imo.

mueller-ma:
I haven't found any guidelines on https://m3.material.io/components/floating-action-button/overview about how many small FABs are ok. IMO 4 should be fine, but if you plan to add more options, feel free to close this PR.

JuancaG05:
Hi @mueller-ma! Since the "Create file via open in web" feature was developed, we have several dynamic options in a new bottom sheet (that we arrive to from a new mini FAB), as you can see here: https://github.com/owncloud/android/issues/3995
Now, since we have an undefined number of options and they could be a lot, I think the best is to keep the FAB hierarchy as it is.
Anyway, thanks a lot for your suggestion! Your contributions will be always welcome to be discussed 😁
I'm closing this, if you have any other idea just open another new issue 🍻

PR #4010, opened 2023-04-10
Bump dependencies

abelgardep:
Bump several dependencies at the same time:
Kotlin 1.8.20, lifecycle, room, etc

Library PR (if needed): https://github.com/owncloud/android-library/pull/557

_____


abelgardep:
Dependencies are already outdated. So, let's trust dependabot 👍

Issue #3862, opened 2023-01-19
[FEATURE REQUEST] OIDC parameters as brandable: scope and prompt


jesmrec:
https://github.com/owncloud/enterprise/issues/5551

Scope and prompt parameters to be handled in app. Now, scope is hardcoded to a default value and prompt is not supported:

**Scope**

mandatory parameter. It is hardcoded:

https://github.com/owncloud/android/blob/5a7df234f924320fc61f24e4698d00abec99c9c9/owncloudData/src/main/java/com/owncloud/android/data/authentication/AuthenticationConstants.kt#L38

it should be set as a brandable string with default value. We could add it also via MDM.

**Prompt**

Optional parameter. But, it could be required somehow. It is like an enum, with 4 different values. As the RFC:

```
prompt
   OPTIONAL. Space delimited, case sensitive list of ASCII string values that specifies whether the Authorization Server prompts the End-User for reauthentication and consent. The defined values are:

   none
The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be login_required, interaction_required, or another code defined in [Section 3.1.2.6](https://openid.net/specs/openid-connect-core-1_0.html#AuthError). This can be used as a method to check for existing authentication and/or consent.
   login
The Authorization Server SHOULD prompt the End-User for reauthentication. If it cannot reauthenticate the End-User, it MUST return an error, typically login_required.
  consent
The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client. If it cannot obtain consent, it MUST return an error, typically consent_required.
   select_account
The Authorization Server SHOULD prompt the End-User to select a user account. This enables an End-User who has multiple accounts at the Authorization Server to select amongst the multiple accounts that they might have current sessions for. If it cannot obtain an account selection choice made by the End-User, it MUST return an error, typically account_selection_required.
```

Source: https://openid.net/specs/openid-connect-core-1_0.html

For vanilla app not included, just for branding.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] ...
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

abelgardep:
[...]

PR #4011, opened 2023-04-10
Add prompt to the authentication flow

abelgardep:
## Related Issues
App: #3862 

- [X] Added changelog files for the fixed issues in folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
_____


JuancaG05 (approved):

jesmrec:
QA

Connection:

- [X] demo.owncloud.com 
- [X] ocis.owncloud.works
- [X] OAuth2 server

Branding

- [ ] No values in prompt
- [X] Values in prompt

jesmrec:
In case the prompt field is empty (by branding), prompt is sent in the request anyway. Should we prevent this?

michaelstingl:
> In case the prompt field is empty (by branding), prompt is sent in the request anyway. Should we prevent this?

I’d think, empty should fallback to default. I’m not aware of use cases with empty prompts.

jesmrec:
Ok, in the vanilla app, the default case (`select account`) is the only available case. Fallback to default will be managed in oB issue.

Issue #3948, opened 2023-03-02
[BUG] Error handling in uploads


jesmrec:
Comes from https://github.com/owncloud/android/pull/3913#issuecomment-1450544756

When a upload fails for a known issue (check list in link above), the feedback to the user is not descriptive. Almost all errors are labelled with `Unknown error` instead a friendly or helpful message. 

Upload errors are set in two different places: uploads view (persisted) and notification (not persisted). Both should match somehow.

jesmrec:
After checking with servers deployed out of docker, everything works but the case in which an upload is cut off because lost connection

PR #4013, opened 2023-04-11
[FIX] Introduce new error message for ProtocolException

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3948

Library PR (if needed): https://github.com/owncloud/android-library/pull/558

_____


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -475,6 +475,7 @@
     <string name="network_error_socket_timeout_exception">The server took too long to respond</string>
     <string name="network_error_connect_timeout_exception">The server took too long to connect</string>
     <string name="network_host_not_available">Server could not be reached</string>
+    <string name="network_error_message">An error occurred in the network</string>
```
abelgardep, line 478: We already have a similar message some lines above `network_error_socket_exception`. Would it be enough for this error?

abelgardep (commented):
Little suggestion here @JuancaG05

owncloudApp/src/main/res/values/strings.xml
JuancaG05: IMO, `network_error_socket_exception` ("An error occurred while connecting with the server.") suggests the server is or could be involved. In this new error message, we only blame the network since we are not able to reach the server, it's like a previous step

abelgardep (approved):

jesmrec:
Approved

Issue #3796, opened 2022-11-26
Start using Binding instead of using findViewByID 


nangarita-coder:
The findViewById method is often used in the application to retrieve a view from a view hierarchy based on an identifier but the problem is that this seemingly harmless method can, in fact, cause performance issues in your Android application if used carelessly, as it is a very demanding method in terms of resource usage.

To solve this problem it is possible to make use of view binding, this is a feature that allows you to more easily write code that interacts with views. Once view linking is enabled in a module, a link class is generated for each XML layout file present in that module. An instance of a link class contains direct references to all views that have an ID in the corresponding layout.

An example of a method in which many are made is the getView method located in the following path owncloudApp/src/main/java/com/owncloud/android/ui/adapter/AccountListAdapter.java
in this method calls are made to the findViewById method 7 times which is evidenced in the code snippet below.

![image_2022-11-25_222958101](https://user-images.githubusercontent.com/60239309/204070839-41d67ce0-ddcc-4237-a31f-ef7300bd8d44.png)

### TASKS

- [ ] Investigate the different view binding libraries to find the implementation that best suits the project architecture. ex https://developer.android.com/topic/libraries/view-binding
- [ ] Create branch name binding Transition
- [ ] Development tasks
  - [ ] Identify classes where there is excessive use of the findViewByID method. 
  - [ ] Implement view binding on the files identified in the previous step
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

abelgardep:
Hi @nangarita-coder!

First of all, thanks for opening an issue!

Currently, we are working hard to refactor a lot of classes in the app and release a 3.0 version. During this process, we have already moved some classes from `findViewById` to `ViewBinding`. For example, the [LoginActivity](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/presentation/ui/authentication/LoginActivity.kt) is already using ViewBinding

We will take a deeper look after releasing 3.0 to keep replacing the `findViewById` with `ViewBinding` 👍

dev-divyansh:
Hey , i would like to work on this ;;

abelgardep:
Hey, @dev-divyansh Go ahead! 

We already use ViewBinding for some screens but there are still several uses of `findViewById`, so take it easy!

dev-divyansh:
Hey @abelgardep  I am beginner in OpenSource would need your guidance ; 
[...]

PR #4021, opened 2023-04-21
updated the findViewById to use view binding

dev-divyansh:
## Related Issues
App: #3796 

Refracted findViewById to use ViewBinding in [this Activity](https://github.com/owncloud/android/blob/000a0d97ac8b8f72f29f80818d64e17d0929e2fd/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java)

@abelgardep please review my PR and let me know what I can improve

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
@@ -194,8 +194,7 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
         public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
                                  @Nullable Bundle savedInstanceState) {
 
-            WhatsNewElementBinding binding_element;
-            binding_element = WhatsNewElementBinding.inflate(getLayoutInflater());
+            WhatsNewElementBinding binding_element = WhatsNewElementBinding.inflate(getLayoutInflater());
 
             ImageView iv = binding_element.whatsNewImage;
             if (mItem.shouldShowImage()) {
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
@@ -188,11 +188,12 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
             mItem = getArguments() != null ? (FeatureItem) getArguments().getParcelable("feature") : null;
         }
 
+
         @Nullable
         @Override
         public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
                                  @Nullable Bundle savedInstanceState) {
-            View v = inflater.inflate(R.layout.whats_new_element, container, false);
+
 
             ImageView iv = binding_element.whatsNewImage;
             if (mItem.shouldShowImage()) {
@@ -209,7 +210,7 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
                 tv2.setText(mItem.getContentText());
             }
 
-            return v;
+            return null;
         }
     }
 }
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
@@ -44,6 +44,8 @@ import androidx.viewpager.widget.ViewPager;
 import com.owncloud.android.BuildConfig;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.databinding.WhatsNewActivityBinding;
+import com.owncloud.android.databinding.WhatsNewElementBinding;
 import com.owncloud.android.wizard.FeatureList;
 import com.owncloud.android.wizard.FeatureList.FeatureItem;
 import com.owncloud.android.presentation.authentication.LoginActivity;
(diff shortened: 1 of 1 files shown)


JuancaG05:
Hi @dev-divyansh! Thanks a lot for your contribution! 🍻
Unfortunately, @abelgardep is not part of the ownCloud Android team anymore 😞, so if you don't mind I'll review your PR and I'll suggest you some changes if needed, I'll try to do it as good as @abelgardep did 😀

dev-divyansh:
Ya sure @JuancaG05  !

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -58,13 +60,16 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
...
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.whats_new_activity);
 
-        mProgress = findViewById(R.id.progressIndicator);
-        mPager = findViewById(R.id.contentPanel);
+        mProgress = binding_activity.progressIndicator;
+        mPager = binding_activity.contentPanel;
```
JuancaG05, line 72: Neither `binding_activity` nor `binding_element` have been initialized, they were only declared 😉. Before the `setContentView` line, you should initialize them:
```
binding_activity = WhatsNewActivityBinding.inflate(getLayoutInflater());
```
Same with the other variable, I won't write the answer for you to practise 😁

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -58,13 +60,16 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
     private ProgressIndicator mProgress;
     private ViewPager mPager;
 
+    private WhatsNewActivityBinding binding_activity;
+    private static WhatsNewElementBinding binding_element;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.whats_new_activity);
```
JuancaG05, line 69: Now that we are using ViewBinding, we must tell the `setContentView` method:
```suggestion
        setContentView(binding_activity.getRoot());
```

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -183,17 +189,17 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
                                  @Nullable Bundle savedInstanceState) {
             View v = inflater.inflate(R.layout.whats_new_element, container, false);
```
JuancaG05, line 190: As this variable is only used for finding the UI elements and we won't need it anymore after using `binding_element`, it can be removed, do you agree?
BTW, this method returns this variable currently. Let's see if after removing it you can make out how to solve it 😉

JuancaG05 (changes requested):
Some comments here @dev-divyansh! 🚀 
BTW, this class is still written in Java, and it has to be moved to Kotlin. Do you dare to take the challenge (not necessarily in this PR)? I think it would be a great practice with a lot of learning! 😀

dev-divyansh:
> Some comments here @dev-divyansh! rocket BTW, this class is still written in Java, and it has to be moved to Kotlin. Do you dare to take the challenge (not necessarily in this PR)? I think it would be a great practice with a lot of learning! grinning

yes , I would like to take up the challenge ( converting Java file to Kotlin ) in the next PR

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -177,28 +188,29 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
             }
 
-            tv2 = v.findViewById(R.id.whatsNewText);
+            tv2 = binding_element.whatsNewText;
             if (mItem.shouldShowContentText()) {
                 tv2.setText(mItem.getContentText());
             }
 
-            return v;
+            return null;
```
JuancaG05, line 213: Nope 😁. What we expect from the `onCreateView` method is to return a view to show in screen, as we did before when returning the variable `v`, so we cannot return null. You have to continue returning a view, but the difference is that now you have to do it with the binding variables instead of `v`. This can be done like:
```suggestion
            return binding_element.getRoot();
```

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -58,13 +60,21 @@ public class WhatsNewActivity extends FragmentActivity implements ViewPager.OnPa
     private ProgressIndicator mProgress;
     private ViewPager mPager;
 
+    private WhatsNewActivityBinding binding_activity;
+    private static WhatsNewElementBinding binding_element;
```
JuancaG05, line 64: BTW, this variable is only used in the inner `FeatureFragment`, so maybe it's better if you move it there (declaration and initialization) since it belongs there. That way, you can save that unnecessary `static` as well 👍

JuancaG05 (changes requested):
Some more suggestions here and we're ready to go @dev-divyansh! 🚀

dev-divyansh:
Hey @JuancaG05  I have updated the changes , please review it

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
```diff
@@ -177,28 +188,30 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
 
+
         @Nullable
         @Override
         public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
                                  @Nullable Bundle savedInstanceState) {
-            View v = inflater.inflate(R.layout.whats_new_element, container, false);
 
-            ImageView iv = v.findViewById(R.id.whatsNewImage);
+            WhatsNewElementBinding binding_element = WhatsNewElementBinding.inflate(getLayoutInflater());
```
JuancaG05, line 197: Just to polish it definitely, I would initialize this here but put the declaration around line 175, so that we have the binding variable available for all the methods of the fragment and not only for `onCreateView`:
```suggestion
            binding_element = WhatsNewElementBinding.inflate(getLayoutInflater());
```

And just before or after to the declaration of `mItem` (line 175):
```
private WhatsNewElementBinding binding_element;
```

JuancaG05 (changes requested):
Last changes to polish a bit @dev-divyansh! Also, I recommend you to press in Android Studio before committing: Code > Reformat Code, so that it removes extra blank lines and other tiny code smells that ktLint could complain of 😁

owncloudApp/src/main/java/com/owncloud/android/ui/activity/WhatsNewActivity.java
JuancaG05: Hey @dev-divyansh, any news here? Just this change and I think we're ready to go 🚀

JuancaG05 (approved):
I will apply some last polishments, but this is ready to go! Thanks for your contribution @dev-divyansh  💯

PR #4023, opened 2023-04-25
[FEATURE REQUEST] Create new file via Open in Web

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3995

Library PR (if needed): https://github.com/owncloud/android-library/pull/562

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_4.0/New%20Document%20-%20FAB.md

Reports:

- [x] (1) Create file with existing name https://github.com/owncloud/android/pull/4023#issuecomment-1532672814
- [x] (2) Question about thumbnails https://github.com/owncloud/android/pull/4023#issuecomment-1532719238
- [x] (3) Improvement in creation dialog https://github.com/owncloud/android/pull/4023#issuecomment-1532852015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Summary
 * Enhancement - Link in drawer menu: [#3949](https://github.com/owncloud/android/pull/3949)
 * Enhancement - Send language header in all requests: [#3980](https://github.com/owncloud/android/issues/3980)
 * Enhancement - Open in specific web provider: [#3994](https://github.com/owncloud/android/issues/3994)
+* Enhancement - Create file via web: [#3995](https://github.com/owncloud/android/issues/3995)
 * Enhancement - Updated WebFinger flow: [#3998](https://github.com/owncloud/android/issues/3998)
 * Enhancement - Monochrome icon for the app: [#4001](https://github.com/owncloud/android/pull/4001)
 * Enhancement - Add prompt parameter to OIDC flow: [#4011](https://github.com/owncloud/android/pull/4011)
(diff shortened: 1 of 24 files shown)


jabarros (approved):
LGTM. Good work!

jesmrec:
### (1)

1. Using FAB, create a new document using a file name that already exists in the current folder

Current: `File could not be created because permission error`

Expected: i guess that such message happens because lack of permissions somehow. Maybe, it could be prevented by detecting that the name already exists. Web client does not allow to create duplicated names:

Screenshot 2023-05-03 at 10 56 04

could we do something about this?

Pixel2, Android 12
`17fb17499`

jesmrec:
### (2)

question about the thumbnails in `Pick document type to create` bottom sheet.

I could only create presentation, sharesheet and document files, in which, afaik, the icon is got from the app. Will this work for any type? for example, if an app provider supports txt files, will the txt file thumbnail be there? 

if a server supports a type of file for which we don't have a thumbnail, will we show the placeholder like in list of files?

jesmrec:
### (3) (improvement - not blocker)

1. Create new document including forbidden characters like "/" or "\"

Current: Dialog closes and error message in snackbar: `File name contains at least one forbidden character`

This is a valid approach, but, i compared this with same error in dialog when sharing text from outside:

Screenshot 2023-05-03 at 12 44 38

same approach in web. 

Advantage: user has the clue about what was wrong. With current approach, user must type again the whole name, guessing which one is the forbidden error. Likely, user could make another mistake.

Since this is also reproducible in the "Create folder" dialog, is not totally related to this issue and could be addressed to another issue.

Pixel2, Android 12
`17fb17499`

JuancaG05:
About (1): it comes from this bug in the server https://github.com/owncloud/ocis/issues/6140
About (2): exactly, the icon comes from the app. In the case of any other type supported by the app, the thumbnail will be the same as in the file list since the method to obtain it is the same. If we don't have a thumbnail for some kind of file type, the default one will be shown, just like in the file list.
About (3): it would be nice to have, I would move it to another issue 👍

jesmrec:
(1) will wait for server fix
(2) thanks for clarifying
(3) new issue -> https://github.com/owncloud/android/issues/4031

jesmrec:
Approved on my side

PR #4035, opened 2023-05-07
[Technical] Gradle Version Catalog

abelgardep:
Migrates current dependencies declaration to Gradle Version Catalog. More info about it can be found [here](https://developer.android.com/build/migrate-to-catalogs)

`Dependabot ` already supports this feature, so it should send PR to update the versions defined in the TOML file 🥇 

I created 2 bundles to illustrate its use, but feel free to create as many as u consider 👍 
```
espresso = ["androidx-test-espresso-contrib", "androidx-test-espresso-core", "androidx-test-espresso-intents", "androidx-test-espresso-web"]
markwon = ["markwon-core", "markwon-ext-tables", "markwon-ext-strikethrough", "markwon-ext-tasklist", "markwon-html"]
```
So we can add several dependencies that are related and are introduced all together like this:
`implementation libs.bundles.espresso`
_____

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,37 +4,6 @@ buildscript {
         sdkCompileVersion = 33
         sdkMinVersion = 23
         sdkTargetVersion = 33
-
-        // Android jetpack
-        androidxArchCore = "2.2.0"
-        androidxLifecycle = "2.5.1"
-        androidxRoom = "2.5.1"
-        androidxCore = "1.9.0"
-        androidxFragment = "1.5.5"
-        androidxAppcompat = "1.5.1"
-        androidxLegacy = "1.0.0"
-
-        // Kotlin
-        orgJetbrainsKotlin = "1.8.10"
-        orgJetbrainsKotlinx = "1.6.4"
-
-        // Koin
-        ioInsertKoin = "3.3.3"
-
-        // Markwon
-        markwon = "4.6.2"
-
-        // Moshi
-        comSquareupMoshi = '1.14.0'
-
-        // Testing
-        ioMockk = "1.13.3"
-        junitVersion = "4.13.2"
-        androidxTestExt = "1.1.5"
-        androidxTest = "1.4.0"
-        androidxTestEspresso = "3.5.1"
-        androidxTestUiautomator = "2.2.0"
-        androidxAnnotation = "1.6.0"
     }
 
     repositories {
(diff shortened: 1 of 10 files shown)


jesmrec:
Thanks a lot!! will check your contribution.

build.gradle
```diff
@@ -43,15 +12,15 @@ buildscript {
...
-        classpath "com.android.tools.build:gradle:7.4.2"
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$orgJetbrainsKotlin"
-        classpath "org.jlleitschuh.gradle:ktlint-gradle:11.1.0"
+        classpath libs.android.gradlePlugin
+        classpath libs.kotlin.gradlePlugin
+        classpath libs.ktlint.gradlePlugin
     }
 }
 
 plugins {
```
JuancaG05, line 21: Since we are using version 7.4.2 of Gradle, I think we need to use an annotation here as the Android documentation says:

> Note: If you are using a version of Gradle below 8.1, you need to annotate the plugins{} block with @Suppress("DSL_SCOPE_VIOLATION") when using version catalogs. Refer to [issue #22797](https://github.com/gradle/gradle/issues/22797) for more info.

JuancaG05 (changes requested):
Thanks for your contribution @abelgardep! Take a look at the comments I added and, if possible, rebase it with the current status of `master` branch 😉

gradle/libs.versions.toml
```diff
@@ -0,0 +1,123 @@
+[versions]
+androidGradlePlugin = "7.4.1"
```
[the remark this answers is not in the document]
abelgardep: I see, my bad, I added the wrong number when migrating the dependencies..

build.gradle
abelgardep: We are not using kotlin dsl yet. We would need to do that in case we had `build.gradle.kts` files. It's not the case and there is no error shown in the IDE either (or at least I cannot see it), so I think there is nothing to do here.

abelgardep:
Ok, changes applied. You can check how dependabot updates the versions in the [fork](https://github.com/abelgardep/android/pulls)

Let me know if there is something else I could do @JuancaG05 

And, of course, this is just the first iteration, we can update it in the way we consider. New bundles, other way to declare the versions, or whatever..

For example, we could consider introducing directly the version for one-time uses. Instead of:
```
[versions]
stetho = "1.6.0"

[libraries]
stetho = { group = "com.facebook.stetho", name = "stetho", version.ref = "stetho" }
```
Use it directly:
```
[libraries]
stetho = { group = "com.facebook.stetho", name = "stetho", version = "1.6.0" }
```
Feel free to customize it 👍

JuancaG05 (approved):
Thanks a lot for your contribution @abelgardep! This will be very useful to keep ownCloud technically updated. BitRise will provide the final approve 😁 
BTW I think this needs a rebase again since we merged several stuff in master, could you do it?

JuancaG05:
Done! Thanks again @abelgardep! 🚀

Issue #4012, opened 2023-04-11
[RELEASE] 4.0


jesmrec:
### TASKS:

 - [x] [DOC] Ping in #documentation-internal about the new release
 - [x] [GIT] Create branch `release/4.0` in owncloud/android from master
 - [x] [DEV] Warn users with oCIS accounts that accounts must be removed and reattached
 - [x] [DEV] Update version number and name in build.gradle in owncloudApp module
 - [x] [DIS] Create a folder for the new version like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
 - [x] [DIS] Move all changelog files from the unreleased folder to the new version folder
 - [x] [DIS] Update screenshots, if needed, in README.md
 - [x] [DIS] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
 - [x] [QA] Design Test plan
 - [x] [QA] Regression Test plan
 - [x] [DOC] Ping in #documentation-internal to let them merge their PR before creating the tag
 - [x] [GIT] Create and sign tag `oc-android-4.0` in HEAD commit of release branch, in owncloud/android
 - [x] [GIT] Create and sign tag `2.1` in HEAD commit of release branch, in owncloud/android-library
 - [x] [DIS] Generate final bundle from signed commit in owncloud/android
 - [x] [GIT] Merge branch `release/4.0` in owncloud/android, into `stable`
 - [x] [GIT] Rebase with `master` in owncloud/android and merge into it
 - [x] [GIT] Merge branch `release/4.0` in owncloud/android-library, into `stable`
 - [x] [GIT] Rebase with `master` in owncloud/library and merge into it
 - [x] [DIS] Upload & publish release bundle and changelog in Play Store
 - [x] [DIS] Update screenshots and store listing, if needed, in Play Store
 - [x] [GIT] Publish a new release in owncloud/android
 - [x] [DIS] Create post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
 - [x] [COM] Inform `#updates` and `#marketing` in internal chat
 - [x] [DIS] Upload release APK and bundle to internal owncloud instance
 - [ ] [DOC] Update documentation with new stuff by creating [issue](https://github.com/owncloud/docs-client-android/issues)

### QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Release_4.0/Regression.md

Reports:

[...]

PR #4037, opened 2023-05-10
[Release] 4.0

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4012

Library PR (if needed): https://github.com/owncloud/android-library/pull/567
_____


owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/usecases/GetPersonalSpaceForAccountUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.domain.spaces.usecases
+
+import com.owncloud.android.domain.BaseUseCase
+import com.owncloud.android.domain.spaces.SpacesRepository
+import com.owncloud.android.domain.spaces.model.OCSpace
+
+class GetPersonalSpaceForAccountUseCase(
```
manuelplazaspalacio, line 27: For only accessing to a method of a repository, I am not sure that we need an UseCase.

manuelplazaspalacio (changes requested):

owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
```diff
@@ -118,13 +131,64 @@ class MainApp : Application() {
 
                     } else {
                         ReleaseNotesActivity.runIfNeeded(activity)
+
+                        val pref = PreferenceManager.getDefaultSharedPreferences(appContext)
```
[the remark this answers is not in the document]
JuancaG05: We don't have currently a ViewModel for the MainApp, and I don't know if it would be suitable for this case... Anyway, this code is going to be predictably for 2 versions, then it won't be necessary anymore 👍

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/usecases/GetPersonalSpaceForAccountUseCase.kt
JuancaG05: You are right, but in ownCloud we follow the clean architecture structure always and I decided to do it like that to keep consistency, although it is not necessary

Issue #4009, opened 2023-04-10
[TECHNICAL] Move available operations in file list to UseCase


jesmrec:
Let's move the legacy `FileMenuFilter.java` class to Kotlin and transform it to use case.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/filter_menu_usecase
- [x] Development tasks
  - [x] Create new enum with the different menu options
  - [x] Move `FileMenuFilter` to use case
  - [x] Create Fragment extension to show or hide menu options
  - [x] Replace `FileMenuFilter` uses by the new use case
  - [x] Remove old `FileMenuFilter`
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/filter_menu_usecase into master

### QA

Test plan:

Reports & improvements

- [x] (1) Sharing missing in oC10 https://github.com/owncloud/android/issues/4009#issuecomment-1570497306  [FIXED]
- [x] (2) Links view does not allow to share in oC10 https://github.com/owncloud/android/issues/4009#issuecomment-1570526571  [FIXED]
- [x] (3) Sharing not available in Details view (oC10) https://github.com/owncloud/android/issues/4009#issuecomment-1570540502  [FIXED]

michaelstingl:
Needed for:
- https://github.com/owncloud/android/issues/2885
- https://github.com/owncloud/android/issues/3860

JuancaG05:
The old filter, and thus the new use case, is used in the next screens:
- Main file list (when selecting at least 1 file)
- Details screen
- Audio preview screen
- Image preview screen
- Text preview screen
- Video preview screen

The filter consists of various menu options and parameters to select which of them must be shown in each case:
- **SELECT ALL:** shown always in the main file list except when all the files in the screen are already selected.
- **SELECT INVERSE:** shown always in the main file list.
- **DOWNLOAD:** shown when none of the selected files is synchronizing, no videos are being previewed, we're not in the av. offline screen, we're not in the shared by link screen, there are no folders among the selected files and there are no downloaded files among the selected ones.
- **RENAME:** shown when none of the selected files is synchronizing, no videos are being previewed, we're not in the av. offline screen, we're not in the shared by link screen, it is a single file or folder and we do have permission to rename it.
- **MOVE:** shown when none of the selected files is synchronizing, no videos are being previewed, we're not in the av. offline screen, we're not in the shared by link screen and we do have permissions to move the selection.
[...]

PR #4039, opened 2023-05-16
[TECHNICAL] Move available operations in file list to UseCase

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4009

Library PR (if needed):

_____


owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -329,55 +335,20 @@ public void onPrepareOptionsMenu(Menu menu) {
...
-                    mContainerActivity,
-                    getActivity()
-            );
-            mf.filter(
-                    menu,
-                    false,
-                    false,
-                    false,
-                    false
+            OCFile safeFile = getFile();
```
manuelplazaspalacio, line 338: This whole code seems to need refactoring. I assume that this is beyond the scope of this issue.

owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewVideoFragment.java
```diff
@@ -307,18 +312,16 @@ public void onCreateOptionsMenu(@NonNull Menu menu, @NonNull MenuInflater inflat
...
-                mContainerActivity,
-                getActivity()
-        );
-        mf.filter(
-                menu,
-                false,
-                false,
-                false,
-                false
+        OCFile safeFile = getFile();
```
manuelplazaspalacio, line 315: The same that I comment before.

owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
JuancaG05: Exactly, 2 out of the 4 preview fragments are still in Java and need to be refactored, and even the other 2 need to be moved to the new architecture as well, so we'll create a issue for that out of the scope of this one 👍

Issue #3946, opened 2023-03-01
[FEATURE REQUEST] File name conflict starting by (1)


jesmrec:
Related with https://github.com/owncloud/android/issues/3935

Copying a file or folder into a location that contains another item with the same name. 

Conflict should be fixed by adding (1) to the new item, if new conflict then (2), (3) etc... Now is (2), then (2) (2), then (2) (2) (2) etc...

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/file_conflict_naming
- [x] Development tasks
  - [x] Start by `MainFileListFragment` and look for the copy action until we arrive at the suitable use case
  - [x] From the use case, navigate through the different layers of the architecture until we find the conflict naming algorithm
  - [x] Change the algorithm
  - [x] Add unit tests
- [ ] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/file_conflict_naming into master

PR #4040, opened 2023-05-17
[FEATURE REQUEST] File name conflict starting by (1)

manuelplazaspalacio:
## Related Issues
App: https://github.com/owncloud/android/issues/3946

_____


owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCFile.kt
```diff
@@ -83,3 +83,4 @@ val OC_FILES_WITH_SYNC_INFO = listOf(OC_FILE_WITH_SYNC_INFO, OC_FILE_WITH_SYNC_I
 val OC_AVAILABLE_OFFLINE_FILES = listOf(OC_AVAILABLE_OFFLINE_FILE, OC_AVAILABLE_OFFLINE_FILE, OC_AVAILABLE_OFFLINE_FILE)
 val OC_FILES_EMPTY = emptyList<OCFile>()
 val OC_FILES_WITH_SYNC_INFO_EMPTY = emptyList<OCFileWithSyncInfo>()
+val OC_FILE_SPACE_WEB_DAV_URL = "url"
```
JuancaG05, line 86: I think you can use here the `webUrl` property of one of the space objects we already have in the `OCSpace.kt` file in the owncloudTestUtil module, so that it is more realistic and we don't need to create new variables. For example, you can use `OC_SPACE_PROJECT_WITH_IMAGE.webUrl` 👍

JuancaG05 (changes requested):
Some comments here and we're ready to go! 🚀 @manuelplazaspalacio

JuancaG05 (approved):
LGTM. Great job @manuelplazaspalacio! Here goes your first PR in ownCloud! 🚀

jesmrec:
Technical solution is correct. 

Mi open question to all: would it make sense to allow creating folders with same name as other existing ones using the same renaming algorithm? i mean, copying or moving files with same name is allowed (adding (1), (2)...), but folders is not. 

what do you think?

JuancaG05:
> Mi open question to all: would it make sense to allow creating folders with same name as other existing ones using the same renaming algorithm? i mean, copying or moving files with same name is allowed (adding (1), (2)...), but folders is not.
> 
> what do you think?

It could make sense, but if we want to keep consistency with the web client, the behaviour is like that there, so I think it's fine as it is. What do you think @manuelplazaspalacio?

manuelplazaspalacio:
I think is a good idea, but it is beyond the scope of this issue. As @JuancaG05 said I think we have to maintain the consistency with the web client. Can we propose this feature to the web team?

jesmrec:
Let's move this forward and wait for valuable feedback.

Issue #3860, opened 2023-01-16
[FEATURE REQUEST] Create "Open in web" option in main file list


jesmrec:
[First iteration](https://github.com/owncloud/android/pull/3737). `Open in web` available from Details view. 

- Create option in main list of files
- Keep option in Details 
- List available app providers for every file (https://owncloud.dev/services/app-registry/apps/#app-registry). 
- Out of multiselection
- Refactor of class

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/open_in_web_main_list
- [x] Development tasks
  - [x] Move "Open in web" to a new extension
  - [x] Use this new extension in the main file list
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/open_in_web_main_list into master

abelgardep:
What should be the label here for each provider?

```
Open in "Provider"
Open in web ("Provider")
Open in "Provider" (web)
```

Other considerations are also welcome @michaelstingl

michaelstingl:
`Open in "Provider"` works for me 👍

Issue #2885, opened 2020-05-29
[FEATURE REQUEST] Show "More" button for every file list item


hosy:
At the moment if I want to run an action on a file item like Delete, Copy, Move, ... I have to select the item before.

**Describe the solution you'd like**
It would be cool to have the "…" button on every file list item, like in the iOS app, which no longer needs to select the item, before running an action on it.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

abelgardep:
Thanks for creating the issue!! This is an improvement we want to develop as part of https://github.com/owncloud/android/issues/2742. There are still some issues required before, but this is in our backlog for sure.

jesmrec:
To do:

- Add three-dot-button to every cell
- Add sheet with actions available for every item (same as displayed in current way)
- Re-allocate share icons (subtitle line, f. ex.)
- Re-allocate multiselection selector? multiselection actions? (bottom bar?) (2nd iteration)
- Add `Open in web` to new actions menu.

JuancaG05:
This is the order I propose:

| File is not downloaded | File is downloaded | Folder |
|---|---|---|
| ![](https://github.com/owncloud/android/assets/57049315/6e5970d0-d77f-4ac3-843d-2019313614a2) | ![](https://github.com/owncloud/android/assets/57049315/009632b1-2ea1-46fb-929f-2ea146f921ff) | ![](https://github.com/owncloud/android/assets/57049315/9cd3fe2b-9a77-4cee-a8ca-8a311d630ca0) | 

First we'll find the `Details` option, that is where we're redirected when we press on the row. Then we'll find `Download`, which is substituted by `Synchronize` in the same position when the file is downloaded. After this, we'll find some file operations like `Rename`, `Move` and `Copy` and then, operations related with other apps such as `Open with`, `Share` and `Send`. Finally, we can find the `Set as available offline` or `Unset as available offline` options, which are related with the sync status of the file, and last, we can find `Remove`. In the case of folders, the order is the same but skipping the options that do not proceed.

JuancaG05:
BTW, the shared icons have been improved here as well, taking advantage that we had to move them to include the 3-dot icon:

[...]

PR #4058, opened 2023-06-06
[FEATURE REQUEST] Create "Open in web" option in main file list

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/3860

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -65,6 +65,7 @@ import com.owncloud.android.domain.files.model.OCFileSyncInfo
 import com.owncloud.android.domain.files.model.OCFileWithSyncInfo
 import com.owncloud.android.domain.spaces.model.OCSpace
 import com.owncloud.android.domain.utils.Event
+import com.owncloud.android.extensions.addOpenInWebMenuOptions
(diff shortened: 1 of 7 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -919,7 +930,23 @@ class MainFileListFragment : Fragment(),
                     false
                 }
                 menu?.filterMenuOptions(menuOptions, hasWritePermission)
+            }
 
+            if (checkedFiles.size == 1) {
```
manuelplazaspalacio, line 935: Esta lógica de aquí, ¿podría estar en el viewModel?

manuelplazaspalacio (approved):

jesmrec:
Tested over two servers that support app providers. So, tests are restricted to

- [X] Files with MIME supported by two providers
- [X] Files with MIME supported by one provider (images, in this test case)
- [X] Files without MIME supported
- [X] Folders (not supported) 
- [X] Multiselection (not supported)
- [X] Details view (regression behaviour)

The only minor detail is that the string overflows the menu width in list of files (does not happen in `Details`):

Screenshot 2023-06-08 at 13 35 41

Since we plan to [move all the options to bottom sheet ](https://github.com/owncloud/android/issues/2885), it's not a problem to keep that behaviour.

No problem from my side to move this forward.

PR #4064, opened 2023-06-12
[TECHNICAL] Remove ignore from AndroidManifest in debug flavour

JuancaG05:
I think a `tools:ignore` property in the Android manifest file specific for the debug flavour is not necessary anymore since it only prevents a warning to appear, and it doesn't happen anymore. Let's check if BitRise passes in this PR to confirm that.

_____


manuelplazaspalacio (approved):
LGTM 🎖️

Issue #4025, opened 2023-04-25
[FEATURE REQUEST] Open markdown files in ASCII mode


jesmrec:
Some days ago, we merged a [contribution](https://github.com/owncloud/android/pull/4017) that renders markdown files, very nice and handy.

New option to open `.md` files in ascii mode (not rendered) could be included, just in case the user wants to check md's source code.

Suggested behaviour:

- By clicking on the file row: file is rendered (contribution)
- New option in menu: "Open ascii" (more ideas for the name) that will open source code 

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

parneet-guraya:
Hi, I would like to work on it!

parneet-guraya:
#4083

PR #4083, opened 2023-06-21
Add show ascii format functionality

parneet-guraya:
## Related Issues
App: #4025 

Library PR (if needed):

_____

## QA
_____

## TASKS

- [x] Switching back and forth from/to ASCII/Markdown format.
- [x] Showing the related MenuItem options only for .md files
- [ ] Retaining state due to configuration change.
- [ ] Retaining state when user comes back to the fragment from home screen.

https://github.com/owncloud/android/assets/111801812/64e39f30-428b-444b-921e-f31a6c26d26f


parneet-guraya:
I have added the functionality to change view state(in this case menu items & rendering text) according to a property, which is changed to change the UI state.

onStart() callback contains the method which will load and show the preview text. It means whenever there's any configuration change or user comes back to the fragment from homescreen, It will be triggered and so the load and preview method. This is causing the properties and rendered text to be reset to the state where it was on the initial startup of the fragment. This is causing UI state to reset. Any suggestions about how should I approach this?

Thanks!

JuancaG05:
Hi @parneet-guraya! Nice to see you again here! 😀

I'm going to suggest you something to address this issue in a different way than the one you did here (and the one that was indeed mentioned in the issue). After seeing this approach technically, I think this may not be the best way. This proposal would solve the problem you posted in the last comment here as well.

I think this would be way better if, instead of adding two new options in the three-dot-menu in the text preview screen, we add two different tabs (https://m2.material.io/components/tabs), one of them (let's say, the left one and the one that is opened by default) the `Markdown mode` and the other (the right one) the `Text mode` (I think it's easier to understand for users to say "Text" instead of "ASCII"). This is because of several reasons:

Firstly, we would keep more or less the same appearance as in the web version, that is usually our reference:
Captura de pantalla 2023-06-22 a las 8 46 54
In the case of the web both views are shown at the same time, but we need to take into account that this is thought to be seen from a wider screen. Adapting it to the mobile world, it would make sense to have each of these views in one tab, just like I suggested.

Secondly, and more technically, I don't think adding these 2 new options to `FileMenuOption` is the best. This recent class I added in a PR some weeks ago is intended to contain common options that we can find in every file at some point. In this case, these options are reserved for a very specific case: text files, and more specifically, Markdown files. That's why I'm not sure this fits here.

In any case, this was way nicely implemented and I will do a quick CR with some details I could see here for you to learn, if that's OK with you. So, thanks a looooot for your contribution! 🚀 

Let me know your opinion about this and, if you agree with my proposal, you can close this PR (no need to solve the CR comments) and open a new one with the new solution 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/previews/PreviewTextViewModel.kt
```diff
@@ -66,6 +69,15 @@ class PreviewTextViewModel(
                 remove(FileMenuOption.MOVE)
                 remove(FileMenuOption.COPY)
                 remove(FileMenuOption.SYNC)
+
+               if(isTypeMarkdown){
+                   if(isAscii){
```
JuancaG05, line 74: KtLint would complain as well about not separating the `if` from the condition and the `{`
```suggestion
               if (isTypeMarkdown) {
                   if (isAscii) {
```
Android Studio can make this tiny changes automatically for you in Code > Reformat code 👍

parneet-guraya:
> Hi @parneet-guraya! Nice to see you again here! grinning
> 
> I'm going to suggest you something to address this issue in a different way than the one you did here (and the one that was indeed mentioned in the issue). After seeing this approach technically, I think this may not be the best way. This proposal would solve the problem you posted in the last comment here as well.
> 
> I think this would be way better if, instead of adding two new options in the three-dot-menu in the text preview screen, we add two different tabs (https://m2.material.io/components/tabs), one of them (let's say, the left one and the one that is opened by default) the `Markdown mode` and the other (the right one) the `Text mode` (I think it's easier to understand for users to say "Text" instead of "ASCII"). This is because of several reasons:
> 
> Firstly, we would keep more or less the same appearance as in the web version, that is usually our reference: Captura de pantalla 2023-06-22 a las 8 46 54 In the case of the web both views are shown at the same time, but we need to take into account that this is thought to be seen from a wider screen. Adapting it to the mobile world, it would make sense to have each of these views in one tab, just like I suggested.
> 
> Secondly, and more technically, I don't think adding these 2 new options to `FileMenuOption` is the best. This recent class I added in a PR some weeks ago is intended to contain common options that we can find in every file at some point. In this case, these options are reserved for a very specific case: text files, and more specifically, Markdown files. That's why I'm not sure this fits here.
> 
> In any case, this was way nicely implemented and I will do a quick CR with some details I could see here for you to learn, if that's OK with you. So, thanks a looooot for your contribution! rocket
> 
> Let me know your opinion about this and, if you agree with my proposal, you can close this PR (no need to solve the CR comments) and open a new one with the new solution :+1: 

Hi, it's nice to see you too! and yes I took the menu item approach (because it was mentioned in the issue).But I can think about the other solution you proposed and I can work on that and see how it goes. 

Also I have taken into account about suitability of  menu items(Show ASCII and Show Markdown) in the `FileMenuOption`.

PS: I've also seen CR and thanks I learnt some things :smile:

JuancaG05:
> Hi, it's nice to see you too! and yes I took the menu item approach (because it was mentioned in the issue).But I can think about the other solution you proposed and I can work on that and see how it goes.
> 
> Also I have taken into account about suitability of menu items(Show ASCII and Show Markdown) in the `FileMenuOption`.
> 
> PS: I've also seen CR and thanks I learnt some things 😄

Cool! Then, let me close this PR and if you want to work on the other approach, open a new one so that PR are not overpopulated 😁. As always, if you need help or you have any kind of question just ping me and I'll try to help you out ASAP 😀.

Issue #4031, opened 2023-05-03
[FEATURE REQUEST] Improve UX of creation dialog when forbidden characters are typed


jesmrec:
Comes from https://github.com/owncloud/android/pull/4023#issuecomment-1532672814

Using FAB, create a new document using a file name that already exists in the current folder

Currently, dialog closes and error message in snackbar: `File name contains at least one forbidden character`

This is a valid approach, but, i compared this with same error in dialog when sharing text from outside:

![Screenshot 2023-05-03 at 12 44 38](https://user-images.githubusercontent.com/14894746/235900627-e0c1964b-6115-4fd4-877b-783c653c9a49.png)

same approach in web. 

Advantage: user has the clue about what was wrong. With current approach, user must type again the whole name, guessing which one is the forbidden error. Likely, user could make another mistake.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/feature_name
- [x] Development tasks
  - [x] Implement whatever
  - [x] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

PR #4097, opened 2023-07-11
[FEATURE REQUEST] Improve UX of creation dialog when forbidden characters are typed

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4031

_____

## QA

https://github.com/owncloud/android/pull/4097#issuecomment-1655113827

Reports 

- [x] (1) Different error messages https://github.com/owncloud/android/pull/4097#issuecomment-1655182565 [FIXED]
- [x] (2) No error message in rename dialog https://github.com/owncloud/android/pull/4097#issuecomment-1655190529 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -909,7 +909,7 @@ public class ReceiveExternalFilesActivity extends FileActivity
                     error = String.format(getString(R.string.uploader_upload_text_dialog_filename_error_length_max),
                             MAX_FILENAME_LENGTH);
                 } else if (matcher.find()) {
-                    error = getString(R.string.filename_forbidden_characters_from_server);
+                    error = getString(R.string.filename_forbidden_characters);
                 } else {
                     okButton.setEnabled(true);
                     error = null;


owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -43,6 +46,8 @@ class CreateFolderDialogFragment : DialogFragment() {
     private lateinit var parentFolder: OCFile
     private lateinit var createFolderListener: CreateFolderListener
     private var isButtonEnabled: Boolean = false
+    private val MAX_FILENAME_LENGTH = 223
+    private val pattern = "[/\\\\]".toRegex()
```
JuancaG05, line 50: I think here should be enough with the `[/\\]` expression for the regular expression. The `[]` indicate that the String matches the regex if it contains one of the characters inside them, and in this case you are repeating the `\` character (with a total of 4 `\` because you need to escape this character)

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,6 +97,29 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+                var error: String? = null
+                if (fileName.length > MAX_FILENAME_LENGTH) {
+                    error = String.format(
+                        getString(R.string.uploader_upload_text_dialog_filename_error_length_max),
+                        MAX_FILENAME_LENGTH
+                    )
+                } else if (fileName.isEmpty()) {
+                    error = getString(R.string.uploader_upload_text_dialog_filename_error_empty)
+                } else if (pattern.containsMatchIn(fileName)) {
+                    error = getString(R.string.filename_forbidden_characters)
```
JuancaG05, line 112: Maybe this part and the check of `MAX_FILENAME_LENGTH` could be moved to the `TextChangedListener` for the `inputText` so that we can see the error appearing or disappearing in real time just after adding the forbidden character or exceeding the maximum length, without the need to press the OK button and thus saving one click and some time, what do you think? 🤔

owncloudApp/src/main/res/layout/edit_box_dialog.xml
```diff
@@ -17,8 +17,9 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
-<LinearLayout
+<com.google.android.material.textfield.TextInputLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/user_input_text_layout"
```
JuancaG05, line 22: Maybe it's better if we name the ID differently, since this is a very general name and just by seeing it we cannot infer that it belongs to this dialog, or even worse, confuse it in a future if we add some more TextInputLayout and we retrieve it with `findViewById` 😓. A proposal:
```suggestion
    android:id="@+id/edit_box_input_text_layout"
```

JuancaG05 (changes requested):
Nice job @Aitorbp! 💯
Take a look at the comments I left, and in addition, I have 2 questions:
- You correctly stated that the only forbidden characters following the backend rules are `/` and `\`, but, did you change the logic to apply this also in the dialog shown in the screenshot (sharing a text with ownCloud from an external app)?
- We have another dialog in which we have to apply this. This dialog is only available in servers with app providers which allow the creation of new files via web, for example ocis.ocis-wopi.latest.owncloud.works, and we can see it by pressing the new mini FAB that appears when pressing the main FAB. As a hint, in the code this dialog is located in the class `MainFileListFragment`, in the method `showFilenameTextDialog`

owncloudApp/src/main/res/layout/edit_box_dialog.xml
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: We need 4 \ for creating the regular expression in a good way. There are more details in this link: https://stackoverflow.com/a/57477268

changelog/unreleased/4031
```diff
@@ -0,0 +1,6 @@
+Enhancement: Improve UX of creation dialog
+
+Improve UX of creation dialog when forbidden characters are typed
```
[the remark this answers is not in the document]
Aitorbp: Done

changelog/unreleased/4031
```diff
@@ -0,0 +1,6 @@
...
+
+https://github.com/owncloud/android/pull/4097
+https://github.com/owncloud/android/issues/4031
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: Done in the three dialogs: create a folder, create a file, and create a file when is receive from externa

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
             val okButton = alertDialog.getButton(AlertDialog.BUTTON_POSITIVE)
             okButton.isEnabled = isButtonEnabled
+
+            okButton.setOnClickListener {
+                var fileName: String = inputText.text.toString()
+                if (fileName.isEmpty()) {
+                    error = getString(R.string.uploader_upload_text_dialog_filename_error_empty)
+                } else {
+                    null
+                }
```
JuancaG05, line 97: I think this case is not reachable, since the OK button is always disabled by default when the filename is empty (empty string or string made up by spaces), so we'll never be able to click on it, we can remove this code.
But anyway, given that code, the `else` branch is doing nothing right now. Maybe you wanted to write:
```suggestion
                error = if (fileName.isEmpty()) {
                    getString(R.string.uploader_upload_text_dialog_filename_error_empty)
                } else {
                    null
                }
```
In this case, the `else` block does makes sense 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+                if (error != null) {
+                    inputLayout.isErrorEnabled = true
+                    inputLayout.error = error
+                } else {
```
JuancaG05, line 101: In theory, this `if` won't be executed ever as well. As we changed the concept of the solution, now we'll show the error in real time (just when the forbidden character is written or the filename exceeds the maximum length) and in that moment, the OK button will be disabled. This means that the OK button will only be clickable when the filename is correct, and in consequence, it will always execute what we have inside the `else` now

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+            }
         }
 
         inputText.doOnTextChanged { text, _, _, _ ->
             val okButton = alertDialog.getButton(AlertDialog.BUTTON_POSITIVE)
-            isButtonEnabled = !text.isNullOrBlank()
-            okButton.isEnabled = isButtonEnabled
+
+            if (text.isNullOrBlank()) {
+                okButton.isEnabled = false
```
JuancaG05, line 112: I think this behaviour is provided by the InputText by default (or at least it was), so check it and if it's not needed, we can remove this code 😀

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+            } else if (text.length > MAX_FILENAME_LENGTH) {
+                error = String.format(
+                    getString(R.string.uploader_upload_text_dialog_filename_error_length_max),
+                    MAX_FILENAME_LENGTH
+                )
```
JuancaG05, line 117: In this case we should also disable the OK button

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+            } else if (forbiddenChars.any { text.contains(it) }) {
+                error = getString(R.string.filename_forbidden_characters_from_server)
+            } else {
```
JuancaG05, line 120: Here we should disable the OK button as well

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -82,12 +87,48 @@ class CreateFolderDialogFragment : DialogFragment() {
...
+                okButton.isEnabled = true
+                error = null
+                inputLayout.error = null
+                inputLayout.isErrorEnabled = false
+            }
+
```
JuancaG05, line 126: You can add here `inputLayout.error = error` and you can save 2 lines: assigning that value in the previous `else` and in the next `if`

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
```diff
@@ -107,6 +148,7 @@ class CreateFolderDialogFragment : DialogFragment() {
     companion object {
         const val CREATE_FOLDER_FRAGMENT = "CREATE_FOLDER_FRAGMENT"
         private const val IS_BUTTON_ENABLED_FLAG_KEY = "IS_BUTTON_ENABLED_FLAG_KEY"
+        private val forbiddenChars = listOf('/', '\\')
```
JuancaG05, line 151: I prefer this approach rather than the regex one! In this case it's a simple case and I think it's better understood this way 🥇

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
JuancaG05, line 1: Follow the same indications I provided for the other dialog 😃

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
JuancaG05, line 1: Follow the previous indications here as well 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -161,6 +163,8 @@ public class ReceiveExternalFilesActivity extends FileActivity
 
     private static final String DIALOG_WAIT_COPY_FILE = "DIALOG_WAIT_COPY_FILE";
 
+    Pattern pattern = Pattern.compile("[/\\\\]");
```
JuancaG05, line 166: This is Java but, can we follow a similar approach as before so that we can avoid regex?

JuancaG05 (dismissed):
Some changes requested here @Aitorbp! 🚀

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: The behavior is ok because after the` if` we validate if the `error `is not null. When it has a value okButton is disable.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: Yes, this behavior was before the issue in that class. I just readapt the code. It wasn't provided by the InputText.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
Aitorbp: The option I see in Java would be to create an array with those two values ​​and then loop through the characters. I think the current option is cleaner using a pattern. In Kotlin it is easy to do it like this because it is cleaner, but in Java, I see it as more cumbersome. So I would leave it as it is.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createfolder/CreateFolderDialogFragment.kt
Aitorbp: Done

manuelplazaspalacio (approved):
LGTM! 🎖

jesmrec:
## QA

Checking if the following dialogs detects the forbidden character when typing:

- [ ] Create folder
- [ ] Rename item
- [ ] Share text with oC
- [ ] Create new document (app providers)

jesmrec:
### (1)

Error messages should be congruent across all cases:

- Create folder with forbidden characters: `File name contains at least one invalid character`
- Share text:  `File name contains at least one invalid character`
- Create new document: `Forbidden characters: / \ `

All of them should show the same message.

jesmrec:
### (2)

1. Open file list
2. Select any item and option `Rename`
3. In the dialog, add forbidden characters to the new name

Current: No error message like other cases
Expected: Same error like in other cases

Pixel2 Android 11
`c1ba55e5b`

jesmrec:
(2) is fixed

(1) Share text scenario shows `File name contains at least one invalid character` as error instead of `Forbidden characters: / \ ` that is the one in the other scenarios

jesmrec:
(1) fixed and PR approved

Ready to go!

PR #4124, opened 2023-08-10
Fix typo in lenght overflow error message

jesmrec:
Before: `Filename must not be more than %d characters`

Now: `Filename must not be longer than %d characters`

more is not the correct comparative for the length

No Calens added since this is just a one-line fix out of the logic.


michaelstingl (approved):
LGTM 👍

PR #4132, opened 2023-08-18
[BUG] Crash when the token is expired v2

Aitorbp:
## Related Issues
App:
 

_____


manuelplazaspalacio (approved):
LGTM!!

jesmrec:
With this approach

- There is no crashes when switching from grid to list (and viceversa)
- There is no crazy behaviours when killing the app after a token revokation

Works for me 👍 

Will be checked again in regression testing stage.

PR #4136, opened 2023-08-20
Use invoke operator to execute usecases

abelgardep:
Minimal refactor. It removes all the `execute `verbosity for use cases.

Previous approach:
`fun execute(params: Params): Type = run(params)`

Operator approach:
`operator fun invoke(params: Params): Type = run(params)`

Instead of `useCase.execute(useCaseParams)`, `operator` allows us to `useCase(useCaseParams)`

If you are OK with the current execute approach, just close the PR 👍 
_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/MainApp.kt
@@ -166,7 +166,7 @@ class MainApp : Application() {
 
                     val getStoredCapabilitiesUseCase: GetStoredCapabilitiesUseCase by inject()
                     val capabilities = withContext(CoroutineScope(CoroutinesDispatcherProvider().io).coroutineContext) {
-                        getStoredCapabilitiesUseCase.execute(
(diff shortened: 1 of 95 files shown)


jesmrec:
Will review. Thanks a lot!!

JuancaG05 (approved):
LGTM 👍 Thanks a lot @abelgardep! I think this will be very handy and more understandable 😃

JuancaG05:
There are some conflicts after we merged other PR, could you solve them please? 🙏 @abelgardep

abelgardep:
Done @JuancaG05 🍻

JuancaG05:
I think there's something related with the library that is not correctly rebased @abelgardep, could you check please? You can see that builds are not correctly created in the latest BitRise triggers. Let's move this on as soon as it is correctly rebased 🚀

abelgardep:
Overseeded by https://github.com/owncloud/android/pull/4179

PR updated and pushed within oC. It should be easier to fix any issue this way.

@JuancaG05

PR #4144, opened 2023-08-30
More improvements in release checklist

jesmrec:
## Changes:

- Add step to rebase the release branch with `stable` just after creation, in order to fast forward when it's approved.
- Add step to prepare post central (separate from publishing step)
- Add step to warn documentation team that tag is closed to be signed
- Add code review step
- Add step to wait for the release published in GP (steps below this one must wait till such event)
- Change `Bugs and improvements` section for `QA` to include more stuff beyond reports

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -18,23 +18,31 @@ For OEM releases, keep the OEM Release template and remove the Open Release one
 
  - [ ] [DOC] Ping in #documentation-internal about the new release
  - [ ] [GIT] Create branch `release/M.m.p` in owncloud/android from master
+ - [ ] [GIT] Rebase `release/M.m.p` against `stable` in owncloud/android
+ - [ ] [GIT] Create branch `release/x.y.z` in owncloud/android-library from master
+ - [ ] [GIT] Rebase `release/x.y.z` against `stable` in owncloud/android-library
  - [ ] [DEV] Update version number and name in build.gradle in owncloudApp module
  - [ ] [DEV] Update [SBOM](https://cloud.owncloud.com/f/6072870)
  - [ ] [DIS] Create a folder for the new version like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
  - [ ] [DIS] Move all changelog files from the unreleased folder to the new version folder
  - [ ] [DIS] Update screenshots, if needed, in README.md
- - [ ] [DIS] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
- - [ ] [QA] Design Test plan
- - [ ] [QA] Regression Test execution
- - [ ] [GIT] Create and sign tag `oc-android-M.m.p` in HEAD commit of release branch, in owncloud/android
- - [ ] [GIT] Create and sign tag `x.y.z` in HEAD commit of release branch, in owncloud/android-library
+ - [ ] [DEV] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
+ - [ ] [DIS] Prepare post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
  - [ ] [DIS] Generate final bundle from signed commit in owncloud/android
+ - [ ] [QA] Design Test plan
+ - [ ] [DEV] Code Review
+ - [ ] [QA] Regression Test execution
+ - [ ] [QA] QA Approval
+ - [ ] [DIS] Upload release APK and bundle to internal owncloud instance
+ - [ ] [DOC] Ping in #documentation-internal that we are close to sign the new tags
+ - [ ] [GIT] Create and sign tag `vM.m.p` in HEAD commit of release branch, in owncloud/android
+ - [ ] [GIT] Create and sign tag `vx.y.z` in HEAD commit of release branch, in owncloud/android-library
  - [ ] [DIS] Upload & publish release bundle and changelog in Play Store
  - [ ] [DIS] Update screenshots and store listing, if needed, in Play Store
  - [ ] [GIT] Publish a new release in [owncloud/android](https://github.com/owncloud/android/releases)
- - [ ] [DIS] Create post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
- - [ ] [COM] Inform `#updates` and `#marketing` in internal chat
- - [ ] [DIS] Upload release APK and bundle to internal owncloud instance
+ - [ ] [DIS] Release published in Play Store
+ - [ ] [DIS] Publish post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
+ - [ ] [COM] Inform `#updates` and `#marketing` in internal chat that release is out
  - [ ] [GIT] Merge `release/M.m.p` branch into `stable`, in owncloud/android-library
  - [ ] [GIT] Merge `release/M.m.p` branch into `stable`, in owncloud/android
  - [ ] [GIT] Merge `release/M.m.p` branch into `master`, in owncloud/android-library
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,31 +18,44 @@ For OEM releases, keep the OEM Release template and remove the Open Release one
...
  - [ ] [DIS] Create a folder for the new version like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
  - [ ] [DIS] Move all changelog files from the unreleased folder to the new version folder
  - [ ] [DIS] Update screenshots, if needed, in README.md
  - [ ] [DIS] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
+ - [ ] [DIS] Prepare post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
  - [ ] [QA] Design Test plan
  - [ ] [QA] Regression Test execution
- - [ ] [GIT] Create and sign tag `oc-android-M.m.p` in HEAD commit of release branch, in owncloud/android
- - [ ] [GIT] Create and sign tag `x.y.z` in HEAD commit of release branch, in owncloud/android-library
  - [ ] [DIS] Generate final bundle from signed commit in owncloud/android
```
JuancaG05, line 33: I think this step should go before `[QA] Regression Test execution` since this bundle is used in QA and if everything is OK, we use this same bundle to upload to Google Play. If there's something to change, I think the generation of a new bundle is implicit in the `[QA] Regression Test execution` step, otherwise we should add another additional step specifying this or add it in that step like `[QA] Regression Test execution and bugfixing`.

.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,31 +18,44 @@ For OEM releases, keep the OEM Release template and remove the Open Release one
...
  - [ ] [GIT] Create branch `release/M.m.p` in owncloud/android from master
+ - [ ] [GIT] Rebase `release/M.m.p` against `stable` in owncloud/android
+ - [ ] [GIT] Create branch `release/x.y.z` in owncloud/android-library from master
+ - [ ] [GIT] Rebase `release/x.y.z` against `stable` in owncloud/android-library
  - [ ] [DEV] Update version number and name in build.gradle in owncloudApp module
  - [ ] [DEV] Update [SBOM](https://cloud.owncloud.com/f/6072870)
  - [ ] [DIS] Create a folder for the new version like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
  - [ ] [DIS] Move all changelog files from the unreleased folder to the new version folder
  - [ ] [DIS] Update screenshots, if needed, in README.md
  - [ ] [DIS] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
```
JuancaG05, line 29: Not sure if this step should have `[DIS]` or `[DEV]` prefix 🤔

.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,31 +18,44 @@ For OEM releases, keep the OEM Release template and remove the Open Release one
...
+ - [ ] [DIS] Prepare post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
  - [ ] [QA] Design Test plan
  - [ ] [QA] Regression Test execution
- - [ ] [GIT] Create and sign tag `oc-android-M.m.p` in HEAD commit of release branch, in owncloud/android
- - [ ] [GIT] Create and sign tag `x.y.z` in HEAD commit of release branch, in owncloud/android-library
  - [ ] [DIS] Generate final bundle from signed commit in owncloud/android
+ - [ ] [DIS] Upload release APK and bundle to internal owncloud instance
+ - [ ] [DEV] Code Review
```
JuancaG05, line 35: Maybe this fits better just after the adding ReleaseNotes step, since if there is anything to change there, it would be more consistent that the bundle used for QA (and even uploading to Google Play) references the same commit as the final bundle.

.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,31 +18,44 @@ For OEM releases, keep the OEM Release template and remove the Open Release one
...
+ - [ ] [DOC] Ping in #documentation-internal that we are close to sign the new tags
+ - [ ] [GIT] Create and sign tag `vM.m.p` in HEAD commit of release branch, in owncloud/android
+ - [ ] [GIT] Create and sign tag `vx.y.z` in HEAD commit of release branch, in owncloud/android-library
  - [ ] [DIS] Upload & publish release bundle and changelog in Play Store
  - [ ] [DIS] Update screenshots and store listing, if needed, in Play Store
  - [ ] [GIT] Publish a new release in [owncloud/android](https://github.com/owncloud/android/releases)
- - [ ] [DIS] Create post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))
- - [ ] [COM] Inform `#updates` and `#marketing` in internal chat
- - [ ] [DIS] Upload release APK and bundle to internal owncloud instance
+ - [ ] [DIS] Release published in Google Play
```
JuancaG05, line 42:
```suggestion
 - [ ] [DIS] Release published in Play Store
```
Let's call it always the same way to avoid confusions :D

JuancaG05 (changes requested):
Some suggestions here @jesmrec :)

.github/ISSUE_TEMPLATE/release_template.md
jesmrec: Let's use [DEV], since the stuff is inside the code

.github/ISSUE_TEMPLATE/release_template.md
jesmrec: Sure 👍

.github/ISSUE_TEMPLATE/release_template.md
jesmrec: The final approval is implicit in the QA stage, so, it can go before. During QA, bugfixing will trigger another bundle creation and then, more QA. Will change.

.github/ISSUE_TEMPLATE/release_template.md
jesmrec: i had also that idea, but, after QA (and also, release notes) bugs fixes could need another code review. It will trigger another loop in case code review changes important code and more QA is required, then new bugs could raise up, more fixes and another new review etc...

So, finally, i left the code review after the first loop (release notes, fixed and everything) because no matter where you place the "code review" step, always you may find another code review.

Is this ok?

JuancaG05: Well, if the final approval is implicit in the QA process as you mentioned, then I think it should go before anyways. Although you need to start another loop, I think CR should be always made before QA.
Just thinking on going step by step reading this template, I think it would make sense finding the CR step before QA, both when it is only one loop or several ones. But this is just my point of view 👀

JuancaG05 (commented):
Not sure about something, let me know what do you think about the comment I added in this review 👍

.github/ISSUE_TEMPLATE/release_template.md
JuancaG05: Well, if the final approval is implicit in the QA process as you mentioned, then I think it should go before anyways. Although you need to start another loop, I think CR should be always made before QA.
Just thinking on going step by step reading this template, I think it would make sense finding the CR step before QA, both when it is only one loop or several ones. But this is just my point of view 👀

jesmrec: Ok, i will move to before, and create another step `QA approval`, separating the aproval itself from the test execution. So, these two steps:

```
 - [ ] [DEV] Code Review
 - [ ] [QA] Regression Test execution
 ```

are in a loop till QA clicks the new one `QA approval`

JuancaG05: Nice! I like it! 👍

JuancaG05 (approved):
LGTM 👍

PR #4182, opened 2023-10-07
language review

EParzefall:
Fixed various formal and some language issues I noticed while translating.

### DIFF
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -448,8 +448,8 @@
     <string name="confirmation_disable_camera_uploads_title">Confirm</string>
     <string name="confirmation_disable_pictures_upload_message">Are you sure you want to disable this feature? The pending pictures will not be uploaded.</string>
     <string name="confirmation_disable_videos_upload_message">Are you sure you want to disable this feature? The pending videos will not be uploaded.</string>
-    <string name="proper_pics_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the pictures. Otherwise, the feature will not be able to detect your pictures. Keep in mind that pictures will be uploaded in at least 15 minutes after taking them."</string>
-    <string name="proper_videos_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the videos. Otherwise, the feature will not be able to detect your videos. Keep in mind that videos will be uploaded in at least 15 minutes after recording them."</string>
+    <string name="proper_pics_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the pictures taken. Otherwise, the feature will not be able to detect your pictures. Keep in mind that pictures will be uploaded in at least 15 minutes after taking them."</string>
+    <string name="proper_videos_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the videos recorded. Otherwise, the feature will not be able to detect your videos. Keep in mind that videos will be uploaded in at least 15 minutes after recording them."</string>
     <string name="conflict_title">File in conflict</string>
     <string name="conflict_description">There is a conflict in file %1$s, tap to solve it.</string>
     <string name="conflict_message">Which files do you want to keep? If you select both versions, the local file will have a number added to its name.</string>
@@ -518,7 +518,7 @@
     <string name="actionbar_select_inverse">Select inverse</string>
 
     <string name="actionbar_move">Move</string>
-    <string name="file_list_empty_moving">Nothing in here. You can add a folder.</string>
+    <string name="file_list_empty_moving">Nothing in here. You can add a folder!</string>
     <string name="folder_picker_choose_button_text">Choose</string>
     <string name="folder_picker_move_here_button_text">Move here</string>
     <string name="folder_picker_copy_here_button_text">Copy here</string>
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -129,7 +129,7 @@
 
     <string name="error_reason">because</string>
     <string name="auth_failure_snackbar">Authentication failed</string>
-    <string name="server_not_supported">Your server version is lower than 10 so is not supported, please upgrade your server to a newer version.</string>
+    <string name="server_not_supported">Your server version is lower than 10 so it is not supported, please upgrade your server to a newer version.</string>
     <string name="auth_failure_snackbar_action">Change account</string>
     <string name="auth_oauth_failure">The access token has expired or become invalid. Sign in again to regain access.</string>
     <string name="auth_oauth_failure_snackbar_action">Sign in</string>
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -56,11 +56,11 @@
     <string name="prefs_touches_with_other_visible_windows">Touches with other visible windows</string>
     <string name="prefs_touches_with_other_visible_windows_summary">Allow touches when the view is obscured by another visible window. Enable it to use light filtering apps.</string>
     <string name="confirmation_touches_with_other_windows_title">Are you sure you want to enable this feature?</string>
-    <string name="confirmation_touches_with_other_windows_message">Use this feature under your own responsibility, a malicious application could try to spoof you into performing some actions, unaware, using other views.</string>
-    <string name="prefs_subsection_picture_uploads">Auto upload pictures</string>
-    <string name="prefs_subsection_picture_uploads_summary">Manage location and behaviour of the pictures uploaded automatically</string>
-    <string name="prefs_subsection_video_uploads">Auto upload videos</string>
-    <string name="prefs_subsection_video_uploads_summary">Manage location and behaviour of the videos uploaded automatically</string>
+    <string name="confirmation_touches_with_other_windows_message">Use this feature at your own risk. A malicious application could try to spoof you into unknowingly performing some actions, using other views.</string>
+    <string name="prefs_subsection_picture_uploads">Auto-upload pictures</string>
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
LGTM

jesmrec:
Thanks a lot for reviewing!

We'll check it.

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -56,11 +56,11 @@
...
-    <string name="confirmation_touches_with_other_windows_message">Use this feature under your own responsibility, a malicious application could try to spoof you into performing some actions, unaware, using other views.</string>
-    <string name="prefs_subsection_picture_uploads">Auto upload pictures</string>
-    <string name="prefs_subsection_picture_uploads_summary">Manage location and behaviour of the pictures uploaded automatically</string>
-    <string name="prefs_subsection_video_uploads">Auto upload videos</string>
-    <string name="prefs_subsection_video_uploads_summary">Manage location and behaviour of the videos uploaded automatically</string>
+    <string name="confirmation_touches_with_other_windows_message">Use this feature at your own risk. A malicious application could try to spoof you into unknowingly performing some actions, using other views.</string>
+    <string name="prefs_subsection_picture_uploads">Auto-upload pictures</string>
+    <string name="prefs_subsection_picture_uploads_summary">Manage location and behavior of the pictures uploaded automatically</string>
```
JuancaG05, line 61: Well, I was taking a look at the code and for variable names and comments sometimes we use the UK variant ("behaviour") and sometimes the USA variant ("behavior"), so I guess both are valid

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -129,9 +129,9 @@
 
     <string name="error_reason">because</string>
     <string name="auth_failure_snackbar">Authentication failed</string>
-    <string name="server_not_supported">Your server version is lower than 10 so is not supported, please upgrade your server to newer version</string>
+    <string name="server_not_supported">Your server version is lower than 10 so is not supported, please upgrade your server to a newer version.</string>
```
JuancaG05, line 132:
```suggestion
    <string name="server_not_supported">Your server version is lower than 10 so it is not supported, please upgrade your server to a newer version.</string>
```

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -105,16 +105,16 @@
...
+    <string name="pattern_not_same_pattern">The patterns are not the same.</string>
+    <string name="pattern_error_set">An error occurred while setting the pattern lock.</string>
+    <string name="pattern_error_remove">An error occurred while removing the pattern lock.</string>
+    <string name="pattern_already_set">Pattern and passcode locks cannot be enabled at the same time, please disable pattern first.</string>
+    <string name="passcode_already_set">Pattern and passcode locks cannot be enabled at the same time, please disable passcode first.</string>
+    <string name="pattern_no_longer_required">The pattern will not be asked for anymore.</string>
+    <string name="pattern_configure_your_pattern_explanation">The pattern will be requested every time the app is started.</string>
     <string name="document_provider_locked">Locked</string>
-    <string name="illegal_argument_exception_message">A valid ACTION is needed in the Intent passed to</string>
+    <string name="illegal_argument_exception_message">A valid ACTION is needed in the intent passed to</string>
```
JuancaG05, line 117: In this case with "Intent" we refer to the Kotlin object, which starts with capital letter, I would let it as it was before 😄

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -219,22 +219,22 @@
...
-    <string name="uploader_upload_text_dialog_filename_error_empty">Filename must not be empty</string>
-    <string name="uploader_upload_text_dialog_filename_error_length_max">Filename must not be longer than %d characters</string>
-    <string name="uploader_upload_text_filename_hint">Filename</string>
+    <string name="uploader_upload_failed_content_single">Upload of %1$s could not be completed.</string>
+    <string name="uploader_upload_failed_credentials_error">Upload failed, you need to log in again.</string>
+    <string name="uploader_upload_failed_ssl_certificate_not_trusted">Server certificate is not trusted.</string>
+    <string name="uploader_upload_text_dialog_title">Insert a name for the new file.</string>
+    <string name="uploader_upload_text_dialog_filename_error_empty">File name must not be empty.</string>
+    <string name="uploader_upload_text_dialog_filename_error_length_max">File name must not be longer than %d characters.</string>
+    <string name="uploader_upload_text_filename_hint">File name</string>
```
JuancaG05, line 228: "Filename" and "file name" are both accepted, so as you prefer: https://en.wikipedia.org/wiki/Filename (we have some variables called `filename` and some others are `fileName`, so both cases in the code again)

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -518,22 +518,22 @@
     <string name="actionbar_select_inverse">Select inverse</string>
 
     <string name="actionbar_move">Move</string>
-    <string name="file_list_empty_moving">Nothing in here. You can add a folder!</string>
```
JuancaG05, line 521: This was intended to sound more user-friendly, so I would let it stay as it is 😄

JuancaG05 (changes requested):
Hey @EParzefall thanks a lot for this review! I was taking a look and I left some comments 😃.
BTW, do you know if strings that are updated here are uploaded to transifex as new strings so that community can translate them again?

EParzefall:
> Hey @EParzefall thanks a lot for this review! I was taking a look and I left some comments 😃. BTW, do you know if strings that are updated here are uploaded to transifex as new strings so that community can translate them again?

I only heard/read that there is a daily sync with Transifex. I expected you know how it works.

owncloudApp/src/main/res/values/strings.xml
EParzefall: Sure, both are valid, but this PR is part of an effort to improve the user experience by making the language more consistent/unified across all ownCloud products. Since we use US English in the documentation, it make sense to do so in the UI as well.

owncloudApp/src/main/res/values/strings.xml
EParzefall: Again, this PR is about wording and spelling consistency. Webster doesn't accept filename as a variant yet (although it likely will soon). Variable names aren't a problem, but in the UI and docs we should be consistent and correct. https://www.merriam-webster.com/dictionary/filename

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -438,42 +438,42 @@
...
     <string name="placeholder_timestamp">2012/05/18 12:23 PM</string>
     <string name="placeholder_media_time">12:23:45</string>
 
     <string name="confirmation_disable_camera_uploads_title">Confirm</string>
-    <string name="confirmation_disable_pictures_upload_message">Are you sure you want to disable this feature? The pending pictures will not be uploaded</string>
-    <string name="confirmation_disable_videos_upload_message">Are you sure you want to disable this feature? The pending videos will not be uploaded</string>
-    <string name="proper_pics_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the pictures taken. Otherwise, the feature will not be able to detect your pictures. Keep in mind that pictures will be uploaded in at least 15 minutes after taking them."</string>
-    <string name="proper_videos_folder_warning_camera_upload">"Please make sure the folder selected is where the camera you are using saves the videos recorded. Otherwise, the feature will not be able to detect your videos. Keep in mind that videos will be uploaded in at least 15 minutes after recording them."</string>
```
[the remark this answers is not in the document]
EParzefall: Well, since it's the camera saving the pictures and videos that should be implied... But if you really want to keep "taken" and "recorded", feel free to add them again.

owncloudApp/src/main/res/values/strings.xml
EParzefall: A friendly command? I don't think so.

JuancaG05: Well, just as we do with `You do not have permission to add content here!`, also reviewed by you in this PR

JuancaG05:
> I only heard/read that there is a daily sync with Transifex. I expected you know how it works.

I think these modified strings are uploaded as well, it was just to make sure. And no, Transifex is not our responsibility, it's just a service that our product uses, so our team doesn't need to know deeply how it works.

JuancaG05 (approved):

phil-davis:
The flow is that this PR first needs to be merged. Then a nightly job will add/adjust the source strings that need to be translated in Transifex.
Then as people add translations of the strings, a nightly Transifex job will push them back to this repo.

JuancaG05:
Hi @phil-davis, thanks for the explanation! 😃 I was aware of the basic flow you describe, I know that happens for newly added strings. It's just that I wasn't sure if that flow is triggered as well for already existent strings which are modified in the default language resource file, as it happens in this PR.

phil-davis:
> In case an existing string is re-translated (because an error, a bad translation or whatever...), is the previous translation cleaned up in tfx?

yes, AFAIK. And it is a bit annoying, IMO, for translators when a string only changes a tiny bit. It would be nice if Transifex would show the previous string and previous translation, and give the translator the option to use the previous translation (adjusting it a tiny bit)

So this sort of language review of strings that get translated makes a lot of "noise" for translators.

EParzefall:
> > In case an existing string is re-translated (because an error, a bad translation or whatever...), is the previous translation cleaned up in tfx?
> 
> yes, AFAIK. And it is a bit annoying, IMO, for translators when a string only changes a tiny bit. It would be nice if Transifex would show the previous string and previous translation, and give the translator the option to use the previous translation (adjusting it a tiny bit)
> 
> So this sort of language review of strings that get translated makes a lot of "noise" for translators.

Actually at least the translation memory in Transifex should show the previous translation with an percentage how well it matches. In many cases 99%- That's why I dared to add all the missing periods after complete sentences. One click. But yes, this cleaning up will be a nuisance for translators.

Issue #4165, opened 2023-09-20
[RELEASE] 4.1.1


jesmrec:
### TASKS:

 - ~~[DOC] Ping in #documentation-internal about the new release~~
 - [x] [GIT] Create branch `release/4.1.1` in owncloud/android from master
 - ~~[GIT] Rebase `release/4.1.1` against `stable` in owncloud/android~~
 - [x] [DEV] Update version number and name in build.gradle in owncloudApp module
 - [x] [DEV] Update [SBOM](https://cloud.owncloud.com/f/6072870)
 - [x] [DIS] Create a folder for the new version like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
 - [x] [DIS] Move all changelog files from the unreleased folder to the new version folder
 - ~~[DIS] Update screenshots, if needed, in README.md~~
 - [x] [DEV] Add ReleaseNotes replacing `emptyList` with `listOf` and adding inside `ReleaseNote()` with String resources
 -  ~~[DIS] Prepare post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))~~
 - [x] [DIS] Generate final bundle from last commit in owncloud/android
 - [x] [QA] Design Test plan
 - [x] [DEV] Code Review
 - [x] [QA] Regression Test execution
 - [x] [QA] QA Approval
 - [x] [DIS] Upload release APK and bundle to internal owncloud instance
 - ~~[DOC] Ping in #documentation-internal that we are close to sign the new tags~~
 - [x] [GIT] Create and sign tag `v4.1.1` in HEAD commit of release branch, in owncloud/android
 - [x] [DIS] Upload & publish release bundle and changelog in Play Store
 - ~~[DIS] Update screenshots and store listing, if needed, in Play Store~~
 - [x] [GIT] Publish a new release in [owncloud/android](https://github.com/owncloud/android/releases)
 - [x] [DIS] Release published in Play Store
 - ~~[DIS] Publish post in central.owncloud.org ([`Category:News + Tag:android`](https://central.owncloud.org/tags/c/news/5/android))~~
 - [x] [COM] Inform `#updates` and `#marketing` in internal chat that release is out
 - [x] [GIT] Merge `release/4.1.1` branch into `stable`, in owncloud/android
 - [ ] [GIT] Merge `release/4.1.1` branch into `master`, in owncloud/android
[...]

PR #4189, opened 2023-10-16
[RELEASE] 4.1.1

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4165

_____


owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -168,10 +168,14 @@ class OCFileRepository(
         val personalSpace = localSpacesDataSource.getPersonalSpaceForAccount(owner)
         if (personalSpace == null) {
             val legacyRootFolder = localFileDataSource.getFileByRemotePath(remotePath = ROOT_PATH, owner = owner, spaceId = null)
-            return legacyRootFolder!!
+            try {
+                return legacyRootFolder ?: throw IllegalStateException("LegacyRootFolder not found")
+            } catch (e: IllegalStateException) {
+                Timber.i("There was an error: $e")
```
manuelplazaspalacio, line 174: Placing this catch here could mute the error to the user.

manuelplazaspalacio (changes requested):
One comment here

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
JuancaG05: This comes from this CR: https://github.com/owncloud/android/pull/4170#discussion_r1338841950
It's true that the user won't receive any feedback, but legacyRootFolder should never be `null`, this allows us to catch the exception and prevent the app from crashing

manuelplazaspalacio (approved):

jesmrec:
`4.1.1` OK from my side

**Approved commits:** 

app: `cde9bb841`
library: `c82d75b1`

**Android Studio version:** 

`Giraffe | 2022.3.1 Patch 1`

**Reports:**

vs oC 10.3.2: https://reports.cucumber.io/reports/c5cecd13-d2d5-476c-85a6-99ca70a2b4c4
vs oCIS 4.0.0: https://reports.cucumber.io/reports/3e3420a5-b9ac-48bb-ad73-41039968b83e

**Smoke test checklist:** 

https://github.com/owncloud/android/issues/4165#issuecomment-1768877683

**Comments:**

Thumbnails fix is a first approach by updating the cache version that triggers a clean up. Ultimate fix is removing the thumbnails cache by replacing for another library/solution. Should be addressed to a different issue.

jesmrec (approved):

Issue #4103, opened 2023-07-14
[FEATURE REQUEST] Auto-refresh when a file is uploaded


Aitorbp:
**Is your feature request related to a problem? Please describe.**
When I create a file on the home screen after I've created it, this item doesn't appear in the feed. When a user uploads a file to the app, this item doesn't appear in the home screen feed in the first instance. You have to refresh the screen to see the new item. I think this functionality is not intuitive because when I come back from the upload file I don't see anything, so I think something is wrong or the file didn't upload.

**Describe the solution you'd like**
From my point of view, when the client returns to the main activity, the screen should refresh immediately. This way, users will see your download in the feed without the need to refresh the screen. It will improve the user experience.

**Additional context**
Next, I show you a video where you can see the current behavior. 

https://github.com/owncloud/android/assets/34248951/2ce8bb08-86c4-4bce-9750-51d214b9b7bd

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/refresh_after_upload
- [x] Development tasks
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/refresh_after_upload into master

JuancaG05:
Some ideas here:
- Detect when the upload worker has finished and, if we continue in the same folder, refresh it so that the view is updated with the new file
- Add a badge in the uploads section of the bottom bar when there is something uploading (also when an upload has finished or failed)

More ideas welcome!

abelgardep:
Maybe you can trigger a `SynchronizeFileUseCase` in the final step of the upload, so you save the new file info in the database within the upload and it should appear immediately when u are in the same folder.

Or adding a file to the database when the operation finishes. At that point you know some info of the upload: `path`, `owner`, `spaceId`, I think that u also know the `eTag`, maybe the `remoteId`. The file should appear and will be synchronized with all the info the next time the user refreshes the folder.

michaelstingl:
For files being created and edited in web office now also trigger an auto refresh of the parent dir.

PR #4199, opened 2023-10-23
[FEATURE REQUEST] Auto-refresh when a file is uploaded

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4103

_____

## QA

Checklist:

https://github.com/owncloud/android/pull/4199#issuecomment-1784616876

Reports:

- [x] (1) `NEW CONTENT` staying in screen when browsing https://github.com/owncloud/android/pull/4199#issuecomment-1794294182
- [x] (2) `NEW CONTENT` in wrong folder https://github.com/owncloud/android/pull/4199#issuecomment-1798956977

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -159,6 +159,7 @@ class MainFileListFragment : Fragment(),
     private var fileSingleFile: OCFile? = null
     private var fileOptionsBottomSheetSingleFileLayout: LinearLayout? = null
     private var succeededTransfers: List<OCTransfer>? = null
+    private var numberOfUploadsRefreshed: Int = 0
 
     override fun onCreateView(
         inflater: LayoutInflater, container: ViewGroup?,
@@ -247,6 +248,17 @@ class MainFileListFragment : Fragment(),
             )
         }
 
+        // Set Refresh FAB and its listener
+        binding.fabRefresh.setOnClickListener {
+            fileOperationsViewModel.performOperation(
+                FileOperation.RefreshFolderOperation(
+                    folderToRefresh = mainFileListViewModel.getFile(),
+                    shouldSyncContents = false,
+                )
+            )
+            it.animate().translationY(-it.height.toFloat()*2).withEndAction { it.isVisible = false }
+        }
+
         // Set SortOptions and its listeners
         binding.optionsLayout.onSortOptionsListener = this
         setViewTypeSelector(SortOptionsView.AdditionalView.CREATE_FOLDER)
@@ -288,6 +300,7 @@ class MainFileListFragment : Fragment(),
             } else {
                 setViewTypeSelector(SortOptionsView.AdditionalView.HIDDEN)
             }
+            numberOfUploadsRefreshed = 0
         }
 
         // Observe the current space to update the toolbar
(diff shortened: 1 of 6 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -593,6 +598,24 @@ class MainFileListFragment : Fragment(),
                 }
             }
         }
+
```
manuelplazaspalacio, line 601: This should have it's own method.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -127,6 +130,7 @@ class MainFileListFragment : Fragment(),
         )
     }
     private val fileOperationsViewModel by sharedViewModel<FileOperationsViewModel>()
+    private val transfersViewModel by viewModel<TransfersViewModel>()
```
manuelplazaspalacio, line 133: This could not be managed with the MainFileListViewModel or by fileOperationsViewModel? I think it's a better solution to not add not view logic to the fragment.

manuelplazaspalacio (changes requested):
Some change here @JuancaG05

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
JuancaG05: Technically it could, but we already have this logic in the TransfersViewModel and:

1) We would need to duplicate code (not a good practice)
2) Semantically, this is something related with transfers so we shouldn't mix it up with other ViewModels

So, taking into account we already have several ViewModels in this fragment, adding another one to prevent duplication of code, I think it's the best option

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -593,6 +598,24 @@ class MainFileListFragment : Fragment(),
                 }
             }
         }
+
+        /* TransfersViewModel observables */
+        // Observe the transfers
+        collectLatestLifecycleFlow(transfersViewModel.transfersWithSpaceStateFlow) { transfers ->
```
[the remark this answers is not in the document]
JuancaG05: If you take a look at the container method, it is `subscribeToViewModels()`, and it is where we add the observers of the different ViewModels in fragments usually. Moving each of the observers out of this method as new methods could be an option, but since they are used just once and `subscribeToViewModels()` has a simple structure, I think we can keep it like that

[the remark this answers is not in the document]
manuelplazaspalacio: A method should do one thing, this method do a lot of thing. Actually, you had to add some comments to clarify what is done in that part of the code. Don't you think it would be better to isolate that action in a method? You would save on comments, and the code would be more readable, as you wouldn't have to read such a long method.

manuelplazaspalacio (approved):
LGTM!!

jesmrec:
Manual uploads

- [X] 1 file
- [X] 10 files
- [x] 100 files
- [x] 500 files: performance reduced
- [ ] 1000 files
- [ ] 1000 files (outside the target folder)

Other (regression):

- [ ] Auto uploads
- [ ] Uploads from camera
- [ ] Uploads from 3rd party

manuelplazaspalacio (approved):
LGTM!!

jesmrec:
### (1)

1. Upload a number of files >  `max_uploads_to_refresh`
2. When `NEW CONTENT` is displayed, browse to another folder without refreshing. 

Current:  `NEW CONTENT` stays in the screen
Expected: not totally sure... i'd think for the `NEW CONTENT` to update the content of the current folder in case new items where uploaded to that location. If i move to another location, it loses the meaning because this place is updated since it was just browsed. So, i'd go to remove the `NEW CONTENT` if browsing to another place

Pixel 2 Android 11
`e2404875`

JuancaG05:
@jesmrec (1) should be fixed now 👍

jesmrec:
i'd raise the default number of uploads to refresh to 200. In terms of performance is affordable and reduce the number of interactions.

jesmrec:
### (2)

1. Upload a number of files > max_uploads_to_refresh
2. Just after submitting the upload, browse to another folder, no matter which one and stay there

Current: After some time (when the number of finished uploads overtakes the limit), the fab to refresh is displayed
Expected: Never shown in that folder since this is not the folder to refresh

Pixel 2 Android 11
`d7c56dd61`

JuancaG05:
@jesmrec (2) should be fixed now.
I also updated the maximum number of uploads to refresh automatically to 200, and I introduced a change that I hope helps with performance: manual refresh button will trigger a new refresh only when there is not another one in progress, otherwise it will show a snackbar informing the user there is already a synchronization in progress

jesmrec:
This is approved on my side. 

Just as clarification: for a very big amount of files, performance could be affected since device affords too many operations at the time: uploads management including statuses changes, UI updates and long PROPFINFDs parsing that triggers also DB updates. For a reasonable amount of files (1000 - 1500) at the time, this is affordable for a budget device. But, i can not assure that a much bigger amount is going to be performant, depending on the device.

Open to iterate here somehow.

Issue #4194, opened 2023-10-18
[FEATURE REQUEST]  Select correct user and navigate to the correct file when opening via deep link


manuelplazaspalacio:
**Describe the solution you'd like**
Things to do:
- Select the user owner of the file between the logged users if the user is not logged show the login view.
- When the user is selected, navigate to the file with id contained in the deep link.
- Show error if no user for the file.
- Show error if file couldn't be finde in the user account.

**Additional context**
Spike related: https://github.com/owncloud/android/issues/4068

## oC10 link / fileid resolve

```
% curl -sI -u "test:test" "https://demo.owncloud.com/f/9" | grep "location"
location: /apps/files/?dir=/Photos
webdav-location: /remote.php/dav/files/test/Photos
```

- https://github.com/owncloud/core/pull/30383

## oCIS link / fileid resolve

`/remote.php/dav/meta/` could be used to get more information.

```
% curl -s 'https://ocis.ocis-traefik.latest.owncloud.works/remote.php/dav/meta/427a5e74-6921-4768-8d53-cd5e8b30f5b0%24534bb038-6f9d-4093-946f-133be61fa4e7!a130bbe3-5681-4193-9ab0-2a714934722a' \
-X 'PROPFIND' \
-H 'Authorization: Bearer REDACTED' \
-H 'Depth: 0' \
--data-binary $'<?xml version="1.0"?>\n<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n  <d:prop>\n    <oc:meta-path-for-user />\n  </d:prop>\n</d:propfind>' | xmllint --format -

<?xml version="1.0"?>
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
  <d:response>
    <d:href>/remote.php/dav/meta/427a5e74-6921-4768-8d53-cd5e8b30f5b0$534bb038-6f9d-4093-946f-133be61fa4e7%21a130bbe3-5681-4193-9ab0-2a714934722a/</d:href>
    <d:propstat>
      <d:prop>
        <oc:meta-path-for-user>/Hacker-Songs</oc:meta-path-for-user>
        <oc:id>427a5e74-6921-4768-8d53-cd5e8b30f5b0$534bb038-6f9d-4093-946f-133be61fa4e7!a130bbe3-5681-4193-9ab0-2a714934722a</oc:id>
[...]

PR #4212, opened 2023-11-14
[FEATURE REQUEST] Select correct user and navigate to the correct file when opening via deep link 

manuelplazaspalacio:
## Related Issues
App: https://github.com/owncloud/android/issues/4194

_____

## QA

- [x] (1) Downloaded files in root folder of oC10 not displayed https://github.com/owncloud/android/pull/4212#issuecomment-1827677419 [FIXED]
- [X] (2) Fille id collation https://github.com/owncloud/android/pull/4212#issuecomment-1827753077 [WONT FIX]
- [ ] (3) Lack of connection not correctly handled https://github.com/owncloud/android/pull/4212#issuecomment-1840703242
- [ ] (4) mountpoint not supported https://github.com/owncloud/android/pull/4212#issuecomment-1840841042


jabarros (approved):
LGTM

jesmrec:
Starting QA...

jesmrec:
### (1)

Open files in root of oC10 does not work for me. Tested with two different servers, getting `Unknown error` or `The user doesn't have access to the file`

Files pointed are stored in the root folder of oC10 account (that means, already discovered) and they are already downloaded.


https://github.com/owncloud/android/assets/14894746/e5df9929-d05e-4e2f-b139-01a018107fbc

Pixel 2 Android 11
`c3bb396b5`

jesmrec:
### (2)

When there are some accounts in the device, the way to check which account is the pointed file in, is checking whether the file id exists in an account.

But, is posible that two accounts on the device have a file with the same id (likely in oC10 that is numerical) . In that case, link is not correctly solved.

Pixel 2 Android 11
`c3bb396b5`

michaelstingl:
> ### (2)
> But, is posible that two accounts on the device have a file with the same id (likely in oC10 that is numerical) . In that case, link is not correctly solved.

Coordinate with iOS team, then maybe document behaviour.

Aitorbp (approved):

Issue #4156, opened 2023-09-07
Sharing Log Files seamlessly within the ownCloud Android App


dpapac:
### Actual behaviour

When attempting to share log files with our own application, it does not appear as a selectable option in the sharing menu.

### Expected behaviour

Our own application should be listed among the available apps for sharing when trying to share log files.

michaelstingl:
@dpapac what exact version? Post a screenshot?

JuancaG05:
This is a explicit restriction specified in the current code:

```
// Show dialog, without the own app
val packagesToExclude = arrayOf<String>(this.packageName)
```

So the best thing to do is create a new brandable parameter (`allow_sharing_with_own_app`?) which states if we can share with the own app or not, keeping `false` as default value to keep current behaviour. Just in case there is a historical reason for that restriction to be there.

michaelstingl:
No problem to show own app from my POV

jesmrec:
Done. To be released in 4.2.

PR #4215, opened 2023-11-16
Sharing Log Files seamlessly within the ownCloud Android App

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4156

_____

### DIFF
--- a/changelog/unreleased/4215
+++ b/changelog/unreleased/4215
@@ -1,0 +1,6 @@
+Enhancement: Log file sharing allowed within ownCloud Android app
+
+Sharing log files to the ownCloud app itself is now possible from the logs screen.
+
+https://github.com/owncloud/android/issues/4156
+https://github.com/owncloud/android/pull/4215
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -243,17 +243,16 @@ fun AppCompatActivity.sendFile(file: File?) {
     if (file != null) {
         val sendIntent: Intent = makeIntent(file, this)
 
-        val packagesToExclude = arrayOf<String>()
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             val shareSheetIntent = ShareSheetHelper().getShareSheetIntent(
-                sendIntent,
-                this,
-                R.string.activity_chooser_send_file_title,
-                packagesToExclude
+                intent = sendIntent,
+                context = this,
+                title = R.string.activity_chooser_send_file_title,
+                packagesToExclude = arrayOf()
             )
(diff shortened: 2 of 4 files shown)


changelog/unreleased/4215
```diff
@@ -0,0 +1,6 @@
+Enhancement: Log file sharing allowed within ownCloud Android app
+
+With this improvement, sharing log files to the Owncloud application itself is possible from the logs screen.
```
JuancaG05, line 3:
```suggestion
Sharing log files to the ownCloud app itself is now possible from the logs screen.
```
We shouldn't start all the calens files with "With this improvement,...". That's information we can save for te changelog
Also, always respect `ownCloud` name as it is, it's a proper brand name

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -242,8 +242,8 @@ fun Activity.openFileWithIntent(intentForSavedMimeType: Intent, intentForGuessed
 fun AppCompatActivity.sendFile(file: File?) {
     if (file != null) {
         val sendIntent: Intent = makeIntent(file, this)
-        // Show dialog, without the own app
-        val packagesToExclude = arrayOf<String>(this.packageName)
+
+        val packagesToExclude = arrayOf<String>()
```
JuancaG05, line 246: What is the sense of creating a variable for an empty array? Send `arrayOf<String>()` instead of creating a variable for that empty array

JuancaG05 (changes requested):

changelog/unreleased/4215
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
Aitorbp: done

JuancaG05 (approved):
LGTM

jesmrec:
Feature works fine, ownCloud is now an option to share logs.

But i noticed that the shared log file uses `.bin` extension. Maybe not related with this but it deserves a look.

Aitorbp:
There are internal android studio method that do not recognize .log extensions, specifically this one:  `MimeTypeMap.getSingleton().getMimeTypeFromExtension(displayName.substring(index + 1))`. Since it does not recognize it, it tries to create a new extension, therefore it puts the `.bin` in front of the `.log`.

The `.log` extension is recognized as a `mimeType` `application/octet-stream`, which is a generic `MIME` type indicating that the contents of the file are binary and not specified more specifically.
 
The solution that has been carried out in this issue is to put a condition when it is a `.log` file so that it does not add the `.bin `extension.

jesmrec:
Fixed and approved

Issue #4219, opened 2023-11-17
[FEATURE REQUEST] Improvements in "New content" refresh button


JuancaG05:
Recently, we added a new clickable floating action button that we use to trigger a manual refresh of the folder:

![Screenshot_20231117_120331](https://github.com/owncloud/android/assets/57049315/5bfda8f9-b852-4e02-b1d8-387b7ae88bfc)

There was some feedback and we can improve it in several ways:

- [ ] Wording: change `New content` to `Refresh`, so that the user knows exactly this button triggers that action
- [ ] Not showing everything in uppercase (this is a Material default, but could investigate how to change it)

More ideas welcome: color, icon...

CC/ @tbsbdr 

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/improvements_refresh_fab
- [x] Development tasks
  - [x] Implement whatever
  - [x] ...
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/improvements_refresh_fab into master

jesmrec:
@rishabh241 before developing, a proposal will help. This issue was created after getting some feedback after first iteration. Just to hit the correct thing, we could discuss about how will be the final look

JuancaG05:
Hi @rishabh241! Cool, go for it! 🚀 
But as @jesmrec says, let's first post a proposal so that everyone can tell their thoughts, and also maybe some feedback from @tbsbdr which is an expert in UI/UX 😃

rishabh241:
Instead of writing refresh or new content, we can just put a symbol for refresh which is more effective, and for color we should choose a color from the color palette of owncloud app

tbsbdr:
thanks @rishabh241 for your idea!
yes, the color should be selected from the color palette 👍
> Wording: change New content to Refresh, so that the user knows exactly this button triggers that action
I'd stick to this approach (just rename to "refresh") since its more polite to add some text to labels - good read: https://www.nngroup.com/articles/icon-usability/

JuancaG05:
Thanks a lot @tbsbdr! I also prefer some label in the button, so that the click area is bigger as well.

So now we can go for it, these are the changes required @rishabh241:

- [ ] Rename "New content" to "Refresh"
- [ ] Not showing everything in uppercase
- [X] The color being used is already the primary color from the palette

[...]

PR #4226, opened 2023-11-23
Feature/improvements refresh fab 

rishabh241:
feature #4219


JuancaG05:
Hey @rishabh241! We still have 7 commits in this PR, so it's not what we wanted 😅. To achieve what we wanted you have to:

1. Create a new branch (if you create a new PR with the same branch, it will keep the same commits as you can see here)
2. Add only the necessary changes to the new branch in a single commit
3. Push it to GitHub and create a new PR from that branch

That way we'll have the clean PR we want 😄

rishabh241:
Sorry sir, I didn't get how to delete the commits and generate a new branch with a new name but it also shows these 7 commits.

Issue #4207, opened 2023-11-03
[BUG] Some NPEs from Google Play Console


jesmrec:
Some crashes coming from Google Play Console

1.

```
Exception java.lang.NullPointerException:
  at com.owncloud.android.ui.activity.FileDisplayActivity.navigateTo (FileDisplayActivity.kt:1555)
  at com.owncloud.android.ui.activity.FileDisplayActivity.navigateTo$default (FileDisplayActivity.kt:1531)
  at com.owncloud.android.ui.activity.FileDisplayActivity.navigateToOption (FileDisplayActivity.kt:1582)
  at com.owncloud.android.ui.activity.DrawerActivity.bottomBarNavigationTo (DrawerActivity.kt:278)
  at com.owncloud.android.ui.activity.DrawerActivity.setupNavigationBottomBar$lambda$7 (DrawerActivity.kt:248)
  at com.owncloud.android.ui.activity.DrawerActivity.$r8$lambda$D_YMO0EvQvgnxMxwdaUwNHSdh-w
  at com.owncloud.android.ui.activity.DrawerActivity$$ExternalSyntheticLambda5.onNavigationItemSelected
  at com.google.android.material.navigation.NavigationBarView$1.onMenuItemSelected (NavigationBarView.java:291)
  at androidx.appcompat.view.menu.MenuBuilder.dispatchMenuItemSelected (MenuBuilder.java:836)
  at androidx.appcompat.view.menu.MenuItemImpl.invoke (MenuItemImpl.java:159)
  at androidx.appcompat.view.menu.MenuBuilder.performItemAction (MenuBuilder.java:987)
  at com.google.android.material.navigation.NavigationBarMenuView$1.onClick (NavigationBarMenuView.java:139)
  at android.view.View.performClick (View.java:7892)
  at android.view.View.performClickInternal (View.java:7869)
  at android.view.View.-$$Nest$mperformClickInternal
  at android.view.View$PerformClick.run (View.java:30891)
  at android.os.Handler.handleCallback (Handler.java:942)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8762)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:604)
[...]

PR #4238, opened 2023-11-29
[BUG] Some NPEs from Google Play Console

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4207

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -1626,8 +1626,8 @@ class FileDisplayActivity : FileActivity(),
 
             FileListOption.SHARED_BY_LINK -> {
                 if (previousFileListOption != newFileListOption || initialState) {
-                    val rootFolderForShares = storageManager?.getRootSharesFolder()
-                    val personalFolder = storageManager?.getRootPersonalFolder()
+                    val rootFolderForShares = storageManager.getRootSharesFolder()
+                    val personalFolder = storageManager.getRootPersonalFolder()
                     if (rootFolderForShares == null && personalFolder?.spaceId != null) {
                         fileListOption = newFileListOption
                         initAndShowListOfShares()
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -1626,8 +1626,8 @@ class FileDisplayActivity : FileActivity(),
 
             FileListOption.SHARED_BY_LINK -> {
                 if (previousFileListOption != newFileListOption || initialState) {
-                    val rootFolderForShares = storageManager.getRootSharesFolder()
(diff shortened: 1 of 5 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -613,7 +613,9 @@ private void startSyncFolderOperation(OCFile folder) {
 
         mSyncInProgress = true;
 
-        mReceiveExternalFilesViewModel.refreshFolderUseCase(folder);
+        if (mReceiveExternalFilesViewModel != null) {
+            mReceiveExternalFilesViewModel.refreshFolderUseCase(folder);
+        }
```
JuancaG05, line 618: Not sure if the NPE refers to this line or something inside the `refreshFolderUseCase` method... `mReceiveExternalFilesViewModel` shouldn't be `null` because it's obtained by Koin dependency injection in a previous executed method 🤔

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1622,8 +1622,8 @@ class FileDisplayActivity : FileActivity(),
 
             FileListOption.SHARED_BY_LINK -> {
                 if (previousFileListOption != newFileListOption || initialState) {
-                    val rootFolderForShares = storageManager.getRootSharesFolder()
-                    val personalFolder = storageManager.getRootPersonalFolder()
+                    val rootFolderForShares = storageManager?.getRootSharesFolder()
+                    val personalFolder = storageManager?.getRootPersonalFolder()
```
JuancaG05, line 1626: Looking at the report, it seems that the same is happening in the line 1642, where `storageManager` is used as well

JuancaG05 (changes requested):
Some things to check here @Aitorbp, maybe a little research on what would happen if those variables are `null` and, now that we handle nullability, we won't get their results? 🤔

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
Aitorbp: Yes, true, but with the information we have in the trace we do not know exactly what is causing the null. Since we were also unable to reproduce the error, I would opt for this approach.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
Aitorbp: In relation to the first case of error. We are going to do the following approach: In the `getStorageManager` getter of the `BaseActivity`, we are going to put a condition whether `mStorageManager` is `null`. In this case we will create a new instance of `FileDataStorageManager` using the current account.

JuancaG05 (approved):
LGTM, let's see if this works now!

jesmrec:
I did some checks regarding the FileDataStorageManager and the ReceiveExternalFileActivity to check if nullability checks affect somehow. For me, we can move forward this one, waiting for the next release to check whether crash reports are not there anymore.

Moving this forward

Issue #4236, opened 2023-11-28
[BUG] Two logs are displayed when you open the app


Aitorbp:
### Actual behaviour
Two log files are displayed when you delete all the logs and close and open the app.

### Expected behaviour
One log should be displayed
 
### Steps to reproduce
1. Browse the application with logs enabled.
2. Go to the logs screen and delete this log. If you have several, delete all of them until you have none.
3. From that screen, close the app and open it again.
4. Two log files are displayed on the screen, when it should be 1

This problem is related to this PR:
- https://github.com/owncloud/android/pull/4204

PR #4239, opened 2023-11-29
[BUG] Two logs are displayed when you open the app

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4236

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/LogsListActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/LogsListActivity.kt
@@ -124,6 +124,10 @@ class LogsListActivity : AppCompatActivity() {
     private fun setData() {
         val items = viewModel.getLogsFiles()
 
+        if (items.isEmpty()) {
+            LogsProvider(applicationContext).stopLogging()
+        }
+
         binding.recyclerViewActivityLogsList.isVisible = items.isNotEmpty()
         binding.logsListEmpty.apply {
             root.isVisible = items.isEmpty()
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/LogsListActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/LogsListActivity.kt
@@ -125,7 +125,8 @@ class LogsListActivity : AppCompatActivity() {
         val items = viewModel.getLogsFiles()
 
         if (items.isEmpty()) {
-            LogsProvider(applicationContext).stopLogging()
+            val mdmProvider = MdmProvider(applicationContext)
+            LogsProvider(applicationContext, mdmProvider).stopLogging()
         }
 
         binding.recyclerViewActivityLogsList.isVisible = items.isNotEmpty()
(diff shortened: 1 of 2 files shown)


JuancaG05 (changes requested):
Some thing to check here @Aitorbp!

owncloudApp/src/main/java/com/owncloud/android/presentation/logging/LogsListActivity.kt
```diff
@@ -122,6 +122,10 @@ class LogsListActivity : AppCompatActivity() {
     private fun setData() {
         val items = viewModel.getLogsFiles()
 
+        if (items.isEmpty()) {
+            LogsProvider(applicationContext).stopLogging()
+        }
```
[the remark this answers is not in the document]
Aitorbp: No, it will start logging again when you close the app and re-open: it runs in `MainApp` and `enabledLogging` is `true`.

changelog/unreleased/4239
```diff
@@ -0,0 +1,6 @@
+Bugfix: Two log files displayed when re-opening the app
+
+When the application is closed from the log files screen (with no files) and the app is reopened, a single log file appears and not two.
```
jesmrec, line 3: "instead of two"

jesmrec:
> Two log files are displayed when you delete all the logs and close and open the app.

Unfortunately, it happens in a easier way. Just closing the app and opening it again, we have a new log file. No matter the date or deletion of other log files. Looks like this:


https://github.com/owncloud/android/assets/14894746/ac8e15af-4453-4312-adca-0759aa46c802


Reproduced with `292e9910`.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/OCFileLoggingTree.kt
```diff
@@ -52,7 +52,17 @@ class OCFileLoggingTree(
...
+
+            var fileNameTimeStamp = SimpleDateFormat(LOG_FILE_TIME_FORMAT, Locale.getDefault()).format(Date())
+
+            if (it.list().isNotEmpty()) {
+                val lastDayDateLogFile = getDayFromLogFileName(it.list().last())
+                val newDayDateLogFile = getDayFromLogFileName(fileNameTimeStamp)
+                if (lastDayDateLogFile == newDayDateLogFile) {
+                    fileNameTimeStamp = getDateFromLogFileName(it.list().last())
+                }
+            }
```
JuancaG05, line 64: We should write quality code as long as we can, free of warnings. In this case, Android Studio is already adding several warnings in this snippet. The method `list()` returns a nullable array, so let's cover that:
```suggestion
            it.list()?.let { logFiles ->
                if (logFiles.isNotEmpty()) {
                    val lastDayDateLogFile = getDayFromLogFileName(logFiles.last())
                    val newDayDateLogFile = getDayFromLogFileName(fileNameTimeStamp)
                    if (lastDayDateLogFile == newDayDateLogFile) {
                        fileNameTimeStamp = getDateFromLogFileName(logFiles.last())
                    }
                }
            }
```

JuancaG05 (changes requested):

JuancaG05 (approved):
Much better now!! Approved on my side

jesmrec:
Please, remove Calens stuff. This is a regression and we have already entry for Logging Changes

Issue #4065, opened 2023-06-12
[BUG] NPE related to MainFileListViewModel


jesmrec:
Reported in Google Play Console

```
Exception java.lang.NullPointerException:
  at com.owncloud.android.presentation.files.filelist.MainFileListViewModel$manageBrowseUp$1.invokeSuspend (MainFileListViewModel.kt:232)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:106)
  at kotlinx.coroutines.internal.LimitedDispatcher.run (LimitedDispatcher.kt:42)
  at kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:95)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:570)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:750)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker (CoroutineScheduler.kt:677)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:664)
```

marekchen:
Do you have any idea about this issue?
I have the same issue in my project.

jesmrec:
@marekchen no for the moment, we got the stacktrace from the console, but we lack of steps to reproduce. Do you have extra info about the problem?

marekchen:
I think the crash reason is use !! operation.
You can check it.

jesmrec:
will check. Also, if you want to contribute with a potential fix, team will review it.

PR #4241, opened 2023-12-01
[BUG] NPE related to MainFileListViewModel

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4065

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ ownCloud admins and users.
 
 ## Summary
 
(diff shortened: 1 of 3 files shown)


JuancaG05 (approved):
Good! Now we'll have to see if this has consequences in case `fileById` is null and thus `parentDir` is null 👀

jesmrec:
Nothing to test here, just waiting for next release. If the crash does not happen anymore, please close the issue then

Issue #4118, opened 2023-08-04
Show/Hide "Personal" space notation in different places in the app


jesmrec:
As @JanAckermann suggestion, when any item is placed in the "Personal" space, should not be notated in the app. Space notations just for project spaces.

Android app is showing "Personal" in different scenarios:

1) New details view

![Screenshot 2023-08-04 at 10 16 11](https://github.com/owncloud/android/assets/14894746/4ce483f4-6624-417f-bdac-a8bcff6f8458)

2) Uploads view

![Screenshot 2023-08-04 at 10 17 52](https://github.com/owncloud/android/assets/14894746/43aeb4be-933e-4486-8de2-ec2290270804)

3) Av. offline shortcut

![Screenshot 2023-08-04 at 10 18 45](https://github.com/owncloud/android/assets/14894746/c2fe0dc9-4223-4b1d-acba-a09fda01b328)

Open to discuss.

PR #4242, opened 2023-12-05
tried fixing the personal space notation

yvarun03:
## Related Issues
App:

Tried fixing the #4118 issue. In the FileListAdapter, I added logic on line 205 to check if a file is in the "Personal" space (isInPersonalSpace). If it is, I customized the UI to hide the space path, space icon, and space name, providing a distinct representation for "Personal" space files.


jesmrec:
Will take a look to your contribution, i think we should discuss which is the best approach since the issue is open to discussion. 

Thanks a lot for contributing!!

JuancaG05:
Hi @yvarun03! Thanks a lot for your contribution! 🍻

This is being discussed and we'll let you know when we have a conclusion for it. In any case, I have several points here:
- We don't check if a file is in the personal space by checking if it contains `/Personal` in its remote path. Indeed, the space will never appear here, nor the personal neither other spaces, so this check is wrong.
- The issue proposes some other places in the app where this would need to be changed. It would be nice to have all of these changes in a single PR.

Issue #4217, opened 2023-11-17
[BUG] Some bugs in download m


JuancaG05:
### Actual behaviour
-Tell us what happens

### Expected behaviour
-Tell us what should happen
 
### Steps to reproduce
1. 
2. 
3. 

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Android version:

Device model: 

Stock or customized system:

ownCloud app version:

ownCloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### ownCloud log (data/owncloud.log)
```
Insert your ownCloud log here
```

parneet-guraya:
Want to work on this.

JuancaG05:
Hi @parneet-guraya! Sorry, this is already planned for a developer of the team 😢, just assigned the issue to him to avoid confusions

PR #4244, opened 2023-12-05
added logfiles

ishlaga:
we  added a suffix (1), (2), etc. The name right now is owncloud.YYYY-MM-DD.log (1).log and should be owncloud.YYYY-MM-DD (1).log as mentioned in the issue #4217 

we made sure that log files are downloaded into the correct "download" files, as a result the log file upload process should be simple.


jesmrec:
Thanks a lot for your contribution. We'll review it

JuancaG05:
This is not a valid solution. Too many comments and code with no sense in classes where it doesn't fit.
Closing this one, but feel free to send any other PR with contributions and we'll review it, they are always welcome! 👍

Issue #4248, opened 2023-12-07
[BUG] Thumbnails glitches if there are many files in the folder


Aitorbp:
### Actual behaviour
- When we have many files without thumbnails in the feed, there are some of them that appear without text. It happens especially when scrolling. The first ones you see on the screen are usually good.

### Expected behaviour
- All the files that do not have a thumbnail must have text below.
 
### Steps to reproduce
1. Upload hundreds of file without thumbnail.
2. See the result in the feed. 

This issue is related to this PR:
- https://github.com/owncloud/android/pull/4237

Check the following video out with ~100 heic files that does not have server-thumbnail:

https://github.com/owncloud/android/assets/14894746/615bdd97-19be-4eb5-be34-ac9d09317caa

Some logs:
[owncloud.2023-12-05_16.57.05.log](https://github.com/owncloud/android/files/13562429/owncloud.2023-12-05_16.57.05.log)

Pixel2, Android11

PR #4256, opened 2023-12-12
[BUG] Thumbnails glitches if there are many files in the folder

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4248

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
@@ -33,7 +33,6 @@ import android.widget.ImageView
 import android.widget.LinearLayout
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
-import androidx.core.view.setMargins
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
 import androidx.recyclerview.widget.StaggeredGridLayoutManager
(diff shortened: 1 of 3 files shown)


jesmrec:
IMO, Calens entries not needed here because there is no change, just a fix over a new feature that already exists in the changelog.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -250,16 +249,25 @@ class FileListAdapter(
 
                 ViewType.GRID_IMAGE.ordinal -> {
                     val view = holder as GridImageViewHolder
+                    val layoutParams = fileIcon.layoutParams as ViewGroup.MarginLayoutParams
```
JuancaG05, line 252: Why `ViewGroup.MarginLayoutParams`?

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -250,16 +249,25 @@ class FileListAdapter(
...
+
                     if (thumbnail == null) {
                         view.binding.Filename.text = file.fileName
+                        //Reset layout params values default
```
JuancaG05, line 256:
```suggestion
                        // Reset layout params values default
```
Correct format of comments is separating `//` from the message of the commit. But real correct format for comments is not writing comments 😉, because that means the code is not clear enough. In this case I think it's ok, but just FYI for the future 👍

JuancaG05 (commented):
Some comments here @Aitorbp, no changes required FTM, just some doubts 👍

owncloudApp/src/main/res/values/dims.xml
```diff
@@ -62,6 +62,12 @@
     <!-- File List -->
     <dimen name="file_list_progress_bar_height">4dp</dimen>
 
+    <!-- item file grid dimens-->
+    <dimen name="item_file_grid_margin">10dp</dimen>
+    <dimen name="item_file_image_grid_margin">1.5dp</dimen>
```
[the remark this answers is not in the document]
Aitorbp: If it is checked, they give almost identical results, not perceptible to the eyes. The problem was that 4 I don't know to what extent it was. Of course it wasn't in dp. So if we put 4dp it would not be correct. Debugging I found that the margin was dp1, but in the print I made it did not come with decimals. I found that when I set 1dp it was a little smaller than the original, almost not noticeable to the eye. Setting 1.5dp I can't find the differences, at least with the naked eye.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
Aitorbp: Because if we do not set `MarginLayoutParams` we cannot access the `setMargins()` method that we need to adjust the margins.

JuancaG05 (approved):
LGTM

jesmrec:
Fixed

Issue #4249, opened 2023-12-08
[FEATURE REQUEST] Don't redact auth type HTTP request logging


michaelstingl:
**Describe the solution you'd like**

I'd like to log the auth method, and redact only the actual credentials.

**Additional context**

Desktop example:
```
        "header":
        {
            "Authorization": "Bearer [redacted]",
```

iOS example:
```
Authorization: Bearer [redacted:1]
```

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/reveal_redacted_info_logs
- [x] Development tasks
  - [x] Process the string to show just the authentication method
  - [x] Create a brandable parameter to show or hide redacted info
  - [x] Create a MDM parameter for the same
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/reveal_redacted_info_logs into master

jesmrec:
New requirement:

Add new option to reveal `[redacted]` info with debugging purposes. Options:

- New branding/MDM option
- Reveal always in `debug` mode (not sure if this is the best option)

more options, welcome.

PR #4257, opened 2023-12-13
[FEATURE REQUEST] Don't redact auth type HTTP request logging

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4249

_____

### DIFF
--- a/owncloudApp/src/main/res/values/setup.xml
+++ b/owncloudApp/src/main/res/values/setup.xml
@@ -140,4 +140,7 @@
     <!-- Number of max uploads to refresh automatically -->
     <integer name="max_uploads_to_refresh">200</integer>
 
+    <!-- Redact authorization header info in logs -->
+    <bool name="redact_auth_header_logs">true</bool>
+
 </resources>
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
@@ -85,7 +85,10 @@ class LogInterceptor : Interceptor {
     private fun logHeaders(headers: Headers): Map<String, String> {
         val auxHeaders = headers.toMap().toMutableMap()
         if (auxHeaders.contains(AUTHORIZATION_HEADER)) {
-            auxHeaders[AUTHORIZATION_HEADER] = "[redacted]"
+            val authHeaderList = auxHeaders[AUTHORIZATION_HEADER]!!.split(" ")
+            val authType = authHeaderList[0]
+            val authInfo = if (redactAuthHeader) "[redacted]" else authHeaderList[1]
+            auxHeaders[AUTHORIZATION_HEADER] = "$authType $authInfo"
         }
         return auxHeaders
     }
@@ -181,6 +184,7 @@ class LogInterceptor : Interceptor {
 
     companion object {
         var httpLogsEnabled: Boolean = false
+        var redactAuthHeader: Boolean = true
         private const val LIMIT_BODY_LOG: Long = 1000000
         private const val BINARY_OMITTED = "<-- Body end for response -- Binary -- Omitted:"
         private const val BYTES = "bytes -->"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ ownCloud admins and users.
 * Enhancement - Use invoke operator to execute usecases: [#4179](https://github.com/owncloud/android/pull/4179)
 * Enhancement - Deep link open app correctly: [#4181](https://github.com/owncloud/android/issues/4181)
 * Enhancement - Select user and navigate to file when opening via deep link: [#4194](https://github.com/owncloud/android/issues/4194)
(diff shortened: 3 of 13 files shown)


JuancaG05:
When using the MDM parameter, if logs are already enabled, we need to disable and re-enable them again for this option to work properly. This is because we cannot perform an active listening to changes in MDM parameters, we just read them when necessary, and in this case, we decide if we show sensitive information or not (by reading the MDM parameter) every time we enable logs.

michaelstingl:
> This is because we cannot perform an active listening to changes in MDM parameters, we just read them when necessary, and in this case, we decide if we show sensitive information or not (by reading the MDM parameter) every time we enable logs.

Okay for me, that MDM config changes require a restart of the app. iOS app for example triggers a notification to tell the user to close and re-open the app.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
```diff
@@ -85,7 +85,10 @@ class LogInterceptor : Interceptor {
     private fun logHeaders(headers: Headers): Map<String, String> {
         val auxHeaders = headers.toMap().toMutableMap()
         if (auxHeaders.contains(AUTHORIZATION_HEADER)) {
-            auxHeaders[AUTHORIZATION_HEADER] = "[redacted]"
+            val authHeaderList = auxHeaders[AUTHORIZATION_HEADER]!!.split(" ")
```
Aitorbp, line 88: Could this value be null? I don´t know if in this case is necessary put the exclamations !!. If it is not I would write something like this:
```
            val authHeaderList = auxHeaders[AUTHORIZATION_HEADER]?.split(" ")
            val authType = authHeaderList?.get(0)
            val authInfo = if (redactAuthHeader) "[redacted]" else authHeaderList?.get(1)
```

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
```diff
@@ -85,7 +85,10 @@ class LogInterceptor : Interceptor {
     private fun logHeaders(headers: Headers): Map<String, String> {
         val auxHeaders = headers.toMap().toMutableMap()
         if (auxHeaders.contains(AUTHORIZATION_HEADER)) {
-            auxHeaders[AUTHORIZATION_HEADER] = "[redacted]"
```
Aitorbp, line 88: I would move this string `"[redacted]" `to the companion object below.

owncloudApp/src/main/res/values/setup.xml
```diff
@@ -140,4 +140,7 @@
     <!-- Number of max uploads to refresh automatically -->
     <integer name="max_uploads_to_refresh">200</integer>
 
+    <!-- Redact authorization header info in logs -->
+    <bool name="redact_auth_header_logs">true</bool>
```
Aitorbp, line 144: Does this value have to default to true?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
```diff
@@ -85,7 +85,10 @@ class LogInterceptor : Interceptor {
...
+            val authHeaderList = auxHeaders[AUTHORIZATION_HEADER]!!.split(" ")
+            val authType = authHeaderList[0]
+            val authInfo = if (redactAuthHeader) "[redacted]" else authHeaderList[1]
```
Aitorbp, line 90: The same that before. I would write the string in the companion object.

Aitorbp (changes requested):
Hello, I leave you some doubts here:

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
JuancaG05: This is not going to be used more than once in the class, neither out of this class, so I see no point moving it to a variable. If you check the rest of the class (and in general, the places where we log messages), you can see that we use raw strings in-situ, because they're used just once and they are not translatable

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
JuancaG05: This is the same string as before, so the answer is the same

owncloudApp/src/main/res/values/setup.xml
JuancaG05: Of course, by default we want to make it as safer as we can!!

Aitorbp (approved):
LGTM

jesmrec:
## QA checks

- [X] Check that authentication method is separated from `[redacted]` in Authorization header
- [X] Check `redact_auth_header_logs = true` does not reveal secrets in logs
- [X] Check `redact_auth_header_logs = false` reveals secrets in logs
- [X] Check both values using MDM build (need disable and enable feature)

OK from my side

Issue #4188, opened 2023-10-16
[BUG] Some bugs related to Details view


JuancaG05:
There are 2 glitches related to the details screen:

### Actual behaviour
When coming from a preview which is hosted in an activity (currently images and videos previews), the top bar shown is not the one that should be shown.

### Expected behaviour
The top bar should show the back arrow and the file title, not the top bar shown in the root folder.
 
### Steps to reproduce
1. Go to an image/video preview
2. Press the 3-dot menu and go to Details
3. See the incorrect top bar

https://github.com/owncloud/android/assets/111801812/6fa5b922-d213-482d-a47e-ef945b9aecbc

(Thanks @parneet-guraya)

### Actual behaviour
When coming from a preview which is hosted in an activity (currently images and videos previews), if we press back until we are in the root folder and we press back again, we go back to the preview we were seeing once more.

### Expected behaviour
When we are in the root folder and press back, we should exit the app.
 
### Steps to reproduce
1. Go to an image/video preview
2. Press the 3-dot menu and go to Details
3. Press back until we are in the root folder
4. Press back again
5. See we are in the preview screen again

This is not a loop, it only happens twice due to the back stack we have (when we are in a preview hosted in an activity and we go to the details screen, FileDisplayActivity is opened again, more details here: https://github.com/owncloud/android/pull/4177#issuecomment-1763360444).

https://github.com/owncloud/android/assets/111801812/99fb54c6-b236-42b3-8a04-7fe08e83349c

(Thanks @parneet-guraya)

**Can these problems be reproduced with the official owncloud server?**
**(url: https://demo.owncloud.org, user: test, password: test)**
Yes they can, they are not related to the server but to the app itself.

jesmrec:
Let's QA this one

PR #4261, opened 2023-12-15
[BUG]  Incorrect top bar when coming from preview & Repetitive navigation when coming from preview

jrjithin:
## Related Issues
App: [#4188](https://github.com/owncloud/android/issues/4188)

_____


JuancaG05:
Hi @jrjithin! 👋  Thanks a lot for your valuable contribution!

Unfortunately, I have to close it 😭. The issue was already assigned to a developer (in this case myself) and it's already under development, so the great work done here may overlap with the one we are doing.
For future cases I recommend first taking a look to whether the issue is already assigned (which likely means that it's under development), and if it is, at least post a comment in the same issue asking about its state. From our side, we'll try to remove the label `Contributions are welcome` when it is already being worked to avoid these confusions.

Sorry again, and we encourage you to keep contributing! 🚀

jrjithin:
@JuancaG05 No problem. I recently started using my own cloud setup and am fixing a few issues I met. I hope you will update the "Contributions are welcome" labels.

PR #4265, opened 2023-12-18
[FIX] Some bugs related to Details view

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4188

_____

## QA

Checks done: https://github.com/owncloud/android/pull/4265#issuecomment-1880888664

- [X] (1) Details not opened in oCIS non-root images https://github.com/owncloud/android/pull/4265#issuecomment-1880897104 [FIXED]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ ownCloud admins and users.
 
 * Security - Improve biometric authentication security: [#4180](https://github.com/owncloud/android/issues/4180)
 * Bugfix - Some Null Pointer Exceptions in MainFileListViewModel: [#4065](https://github.com/owncloud/android/issues/4065)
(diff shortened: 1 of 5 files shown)


Aitorbp (approved):
Everything good for my side

jesmrec:
Let's QA this one

jesmrec:
(1) is reproducible in 4.1.1 so it could be moved to another issue (not a regression). But, that problem prevents to reproduce correctly the 2nd report in the issue. 

Checking with an image in root folder, the 2nd item is fixed.

So, here we have to choose between two options:

- Fixing here and getting rid of the problem, if feasible.
- Move to another issue assuming that the behaviour could not no be correct for items in non-root folder (not testable with the problem)

open to discussion

JuancaG05:
Since it was an ugly bug, I decided to fix it here. (1) should be solved and fixes in this PR already testable 👍

Issue #4231, opened 2023-11-24
[FEATURE REQUEST] Block UI when deleting files


JuancaG05:
**Is your feature request related to a problem? Please describe.**
Currently, when performing operations like copy or move, the UI blocks until the operation is finished by showing a dialog on screen. This doesn't happen with the delete operation.

**Describe the solution you'd like**
Block the UI with the delete operation as well.

**Describe alternatives you've considered**
Showing the same dialog as in the copy and delete operations to block the screen.

**Additional context**
The dialog is the following:

![block_ui](https://github.com/owncloud/android/assets/57049315/2e361333-b51c-4455-bda2-ce96dbdbe36b)

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/block_ui_delete
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/block_ui_delete into master

parneet-guraya:
I want to work on it.

PR #4270, opened 2023-12-20
[FEATURE REQUEST] Block UI when deleting files

parneet-guraya:
Fix: #4231

https://github.com/owncloud/android/assets/111801812/6fb3ed46-de0e-4b1f-a628-0a25dff91137

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
@@ -43,7 +43,9 @@ import com.owncloud.android.domain.files.model.OCFile
 import com.owncloud.android.domain.files.model.OCFile.Companion.ROOT_PARENT_ID
 import com.owncloud.android.domain.files.usecases.SortFilesUseCase
 import com.owncloud.android.domain.utils.Event
+import com.owncloud.android.extensions.showErrorInSnackbar
 import com.owncloud.android.presentation.authentication.AccountUtils
+import com.owncloud.android.presentation.common.UIResult
 import com.owncloud.android.presentation.files.SortOrder
 import com.owncloud.android.presentation.files.SortType
 import com.owncloud.android.presentation.files.operations.FileOperation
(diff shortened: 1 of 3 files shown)


parneet-guraya:
@JuancaG05 Have a look :-)

owncloudApp/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
```diff
@@ -124,9 +126,18 @@ class PreviewImageActivity : FileActivity(),
...
+                is UIResult.Error -> {
+                    dismissLoadingDialog()
+                    showErrorInSnackbar(R.string.remove_fail_msg, uiResult.getThrowableOrNull())
+                }
+
+                is UIResult.Loading -> showLoadingDialog(R.string.wait_a_moment)
+                is UIResult.Success -> {
+                    dismissLoadingDialog()
+                    finish()
+                }
```
JuancaG05, line 140: Ok!! Thanks for this improvement in image and video previews, so that we can keep a consistent behaviour with other previews 👍

JuancaG05 (approved):
LGTM!! Thanks again @parneet-guraya for your valuable contributions!! 😄

JuancaG05:
(For QA purposes) Things changed here:
- Blocking UI when removing a file in the file list
- Blocking UI when removing a file in its preview when it's a fragment (audio, text)
- Blocking UI when removing a file in its preview when it's an activity (image, video)

jesmrec:
hi @parneet-guraya! thanks again for your valuable contribution. After passing code review, QA stage is ongoing and i detected a behaviour that deserves a look:

It's related with videos preview, that also implements the blocking behaviour as well as pictures preview and list of files.

The point is that videos can be streamed, that means, no local copy in device while the preview is running. If you try to remove a non-downloaded video from the preview (streamed), an error is displayed:

`Remove failed because unknown error`

Behaviour should be the same as downloaded and streamed videos.

Could you take a look?

parneet-guraya:
@jesmrec Hi :wave: ,

I believe you're talking about deleting file while the video preview is playing. But, strangely I couldn't reproduce the same behaviour you're talking about.


https://github.com/owncloud/android/assets/111801812/7b442ed0-78b0-42a5-b537-fb64e84a06a7

Issue #3192, opened 2021-04-07
[BUG] White title on white background in AlertDialogs in Samsung devices


JuancaG05:
### Actual behaviour
In Samsung devices, when you open any AlertDialog in the app, the title is shown white-colored, and since the background is also white, it becomes illegible. A screenshot where we can see the problem:

![device-2021-04-06-155007](https://user-images.githubusercontent.com/57049315/113828239-76d3c780-9784-11eb-8b0a-324a24675463.png)

### Expected behaviour
The title should appear in black so that it is perfectly legible, as it does in any other not-Samsung device.
 
### Steps to reproduce
1. Select with a long tap any folder or file
2. Tap the share button in the Action Bar
3. Tap the "+" button next to "Public links"

This is an example of an AlertDialog with title in the app, we can find more through the different screens.

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes it can, it's a bug in the front-end part of the app, so the server is not important here.

### Environment data
It's been checked with 2 different Samsung devices:

Android version: 9
Device model: Samsung Galaxy S9
ownCloud app version: 2.17

Android version: 11
Device model: Samsung Galaxy S10+
ownCloud app version: 2.17

It has also been tested with another devices such as Nexus 5 (Android 10), Nexus 6P (Android 7), Huawei P20 Lite (Android 9) and Pixel 2 (Android 11), and working fine in all of them.

jesmrec:
Some black magic in Samsung devices. It will be checked.

PR #4277, opened 2023-12-30
Alert Dialog title color issue in Samsung devices

jrjithin:
## Related Issues
App: #3192 

- [x] Added changelog files for the fixed issues in the folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
_____

## QA
The issue was reproduced on Samsung F41 and verified after the fix.


changelog/unreleased/3192
JuancaG05, line 1: The title of this file should be `4277`, we name them with the # of the associated PR 😃

owncloudApp/src/main/res/layout/share_public_dialog.xml
```diff
@@ -11,7 +11,7 @@
         android:layout_height="wrap_content"
         android:padding="@dimen/standard_half_padding"
         android:text="@string/share_via_link_create_title"
-        android:textAppearance="@android:style/TextAppearance.DeviceDefault.DialogWindowTitle" />
+        android:textAppearance="@style/Theme.ownCloud.AlertDialog.WindowTitle" />
```
JuancaG05, line 14: I just checked and an easier option is applying:

`android:textAppearance="@android:style/TextAppearance.Material.DialogWindowTitle"`

It implies just 1 change and the looking is similar to the current one (but correct color). The new style you created makes the title a bit smaller 😄

owncloudApp/src/main/res/values/styles.xml
```diff
@@ -96,6 +96,10 @@
         <item name="windowNoTitle">true</item>
     </style>
 
+    <!-- Alert Dialog title -->
+    <style name="Theme.ownCloud.AlertDialog.WindowTitle" parent="TextAppearance.Material3.TitleMedium">
+    </style>
```
JuancaG05, line 101: This is not needed anymore in case you apply what I suggest in the previous comment 👍

JuancaG05 (changes requested):
Hi @jrjithin! Thanks a lot for your contribution! Here I left some comments to help improve your PR 🍻

owncloudApp/src/main/res/values/styles.xml
jrjithin: Thanks for the suggestions. Removed as suggested. However, I would like to clarify the following points.

 1) Isn't it better to move eventually to a material3 day-night theme rather than using a device default theme?
2) I noticed the app doesn't have any effect on Day/night mode changes. In that case, using components from the device's default theme may change based on the device theme right?

JuancaG05:
Hey @jrjithin! You still have 2 pending comments that you didn't applied (maybe you did it locally and didn't push the commits?). I leave it here opened so that you can commit the solution and we can merge your contribution 👍

owncloudApp/src/main/res/values/styles.xml
JuancaG05: Sure, Material3 will be a future aim in our app 😃. We're thinking of including big UI changes soon, so all of these will be taken into account, for the moment the best is keeping it as simple as possible

jrjithin:
@JuancaG05 Thank you for pointing this out. It's resolved now.

JuancaG05 (approved):
Approved from my side! Let's QA this @jesmrec.
Thanks again for your valuable contribution @jrjithin! 🍻

jesmrec:
Nothing to say :D

PR #4280, opened 2024-01-04
[BUG] Some bugs in log files downloads

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4217

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
@@ -21,9 +21,11 @@
 package com.owncloud.android.presentation.logging
 
 import android.content.Context
+import android.os.Build
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.view.isVisible
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
 import com.owncloud.android.R
@@ -49,6 +51,7 @@ class RecyclerViewLogsAdapter(
         holder.binding.apply {
             textViewTitleActivityLogsList.text = log.name
             textViewSubtitleActivityLogsList.text = log.toLegibleStringSize(context)
+            hideDownloadButton(holder)
             imageViewShareActivityLogsList.setOnClickListener {
                 listener.share(log)
             }
@@ -83,5 +86,12 @@ class RecyclerViewLogsAdapter(
         fun delete(file: File, isLastLogFileDeleted: Boolean)
         fun open(file: File)
         fun download(file: File)
+    }
+
+
+    private fun hideDownloadButton(holder: ViewHolder) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+            holder.binding.imageViewDownloadActivityLogsList.isVisible = false
+        }
     }
 }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
@@ -51,7 +51,9 @@ class RecyclerViewLogsAdapter(
         holder.binding.apply {
             textViewTitleActivityLogsList.text = log.name
             textViewSubtitleActivityLogsList.text = log.toLegibleStringSize(context)
-            hideDownloadButton(holder)
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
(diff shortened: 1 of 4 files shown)


Aitorbp:
First of all, to access the device's downloads folder we need read permissions. But we have said that we do not want to ask the user for these permissions for security.

Another option would be to download the log to the Application Downloads folder in the internal storage (`/storage/emulated/0/Android/data/com.owncloud.android.debug/files/Download`). To download here we do not need permissions, but the problem is that I have not found a way to navigate to this route. Whatever behavior you choose, it always takes me either to the generic Downloads folder or to the Recent folder.

Android 9 (Pie) introduced additional restrictions on URI security, and this may affect how certain types of URIs are opened, especially those related to file access.

In any case, I would rule out this last option, the ideal would be to download in the external Downloads folder, not in an internal one, which is difficult for the user to locate.

In short, I would ask the user for read permissions or restrict the download of logs for users with devices lower than Android 10.

Aitorbp:
In versions previous to API 29 we need permissions to access the public Downloads folder. When we call `checkSelfPermission` to check if `Manifest.permission.WRITE_EXTERNAL_STORAGE` is allowed by the user, we receive a denied permission, even if we have it set in the `Manifest`. For this reason it is necessary to ask permissions to the user at runtime to be able to make the download to the DOWNLOADS public folder.

I leave an extract of where I have extracted the [information](https://medium.com/@heysahilpatel/android-scoped-storage-with-example-e4dc8f2a234d): 
`Request permission to read and write to external storage if necessary. If you need to access files in the public Downloads directory, you will need to request permission to read and write to external storage. You can do this by calling the checkSelfPermission() method and then requesting permission if the app does not already have it.
`
I also leave this other documentation where it clarifies for which version the permissions are requested:
https://medium.com/@wanxiao1994/android-storage-permission-adaptation-and-reading-writing-media-files-6b85121c9461

Aitorbp:
To add a little more information, we are going to use `MediaStore` for downloading logs on Android 10 onwards. 
[Mediastore](https://developer.android.com/training/data-storage/shared/media) allows users to retrieve and update media files more easily. Even after uninstalling the app, these files remain on the user's device. 

It is a more optimal tool than the ones used before. In fact, thanks to this we no longer need to manipulate the name of the log if it is repeated, it will be done internally.

Unfortunately, the APIs used in Android 10 are not available in earlier versions of Android, which means we must request permissions at runtime for earlier versions. More information about this [here](https://alessandromautone.medium.com/public-file-storage-on-android-10-622c49a874a9).

So in this PR, we limit this new functionality to Android versions equal to or greater than 10.

michaelstingl:
> Unfortunately, the APIs used in Android 10 are not available in earlier versions, which means we must request permissions at runtime for earlier versions. More information about this [here](https://alessandromautone.medium.com/public-file-storage-on-android-10-622c49a874a9).
> 
> So in this PR, we limit this new functionality to Android versions equal to or greater than 10.

Works for me 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
```diff
@@ -84,4 +87,11 @@ class RecyclerViewLogsAdapter(
         fun open(file: File)
         fun download(file: File)
     }
+
+
+    private fun hideDownloadButton(holder: ViewHolder) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+            holder.binding.imageViewDownloadActivityLogsList.isVisible = false
+        }
```
JuancaG05, line 95: I would add these 3 lines directly in the `onBindViewHolder` method, having a method called `hide...` that is always called can lead to confusion, and we can save calling to `holder.binding` again

JuancaG05 (changes requested):
Good job @Aitorbp! Just one tiny change to improve readability 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/logging/RecyclerViewLogsAdapter.kt
Aitorbp: Done

JuancaG05 (approved):
LGTM!

jesmrec:
Checks: 

- [ ] File name (.log)
- [ ] File name in repetition 
- [ ] Download button in Android 9
- [ ] Download button in Android 11
- [ ] Download button in Android 13
- [ ] Download button in Android 14
- [ ] Download file functionality

jesmrec:
About the PR, i'd not add a new Calens entry, since it's fixing something that is not released. We already have a Calens entry for the "download log file". I'd modify the existing one, clarifying the strict availability (Android10+) instead of adding two entries about same feature.

Aitorbp:
I have done some changes and removed calens. Could you check this @jesmrec ?

jesmrec:
Ready to go on my side!

Issue #4282, opened 2024-01-09
Handle error 423 (locked file)


jesmrec:
Comes from: https://github.com/owncloud/android/pull/4270#issuecomment-1882626155

The app has the [423](https://http.dev/423) error defined 

https://github.com/owncloud/android/blob/c4c2ac624b5ec4209848b6b157b88f779881c0c7/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/HttpConstants.java#L187

but not handled. In case it happens, `Unknown error`  is displayed.

Expected: if 423 is retrieved from the server in any situation, `Locked file` or similar should be displayed.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

parneet-guraya:
I want to work on this

jesmrec:
go ahead!!

PR #4285, opened 2024-01-16
Handle error 423 (locked file)

parneet-guraya:
Fixes: #4282

What this PR includes?
- throwing exception when get HTTP 423 (resource locked)
- handling the exception by showing snackbar

https://github.com/owncloud/android/assets/111801812/5a395242-3125-4c7d-90b1-8fc38aaa17fa

@jesmrec The reason I believe this exception occurs is when file is being accessed, like in this case when the video is being fetched and we try to remove the file then it throws the exception.

### DIFF
--- a/owncloudDomain/src/main/java/com/owncloud/android/domain/exceptions/ResourceLockedException.kt
+++ b/owncloudDomain/src/main/java/com/owncloud/android/domain/exceptions/ResourceLockedException.kt
@@ -1,0 +1,24 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Parneet Singh
+ * Copyright (C) 2024 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.domain.exceptions
+
+import java.lang.Exception
+
+class ResourceLockedException : Exception()
--- a/changelog/unreleased/4285
+++ b/changelog/unreleased/4285
@@ -1,0 +1,6 @@
+Bugfix: Handle Http 423 (resource locked)
+
+App can gracefully show if the file is locked when done certain operations on it.
+
+https://github.com/owncloud/android/issues/4282
+https://github.com/owncloud/android/pull/4285
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ThrowableExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ThrowableExt.kt
@@ -43,6 +43,7 @@ import com.owncloud.android.domain.exceptions.OAuth2ErrorAccessDeniedException
 import com.owncloud.android.domain.exceptions.OAuth2ErrorException
 import com.owncloud.android.domain.exceptions.QuotaExceededException
 import com.owncloud.android.domain.exceptions.RedirectToNonSecureException
+import com.owncloud.android.domain.exceptions.ResourceLockedException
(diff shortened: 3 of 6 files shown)


parneet-guraya:
@JuancaG05 hey could you give it a quick look :-)

JuancaG05:
I'll take a look ASAP. In this case @Aitorbp will take a look as well so that he gets familiar with how errors flow from network responses to UI 🙂

owncloudDomain/src/main/java/com/owncloud/android/domain/exceptions/ResourceLockedException.kt
```diff
@@ -0,0 +1,5 @@
+package com.owncloud.android.domain.exceptions
+
+import java.lang.Exception
+
+class ResourceLockedException : Exception()
```
Aitorbp, line 5: Here should be a line break in the end of the class. And also add the license text with your name.

Aitorbp (changes requested):
The development seems correct to me. Follow the logic of the other cases. I only saw a small formatting correction. 👍

owncloudDomain/src/main/java/com/owncloud/android/domain/exceptions/ResourceLockedException.kt
JuancaG05: Indeed there is a line break, otherwise GitHub would indicate it

parneet-guraya: I did have the line break not sure why it isn't showing here. I added the license header.

parneet-guraya:
I updated the requested changes 👍

JuancaG05 (approved):
Approved from my side! Good job @parneet-guraya! 🚀

jesmrec:
Testing the PR i notice a side-effect (not regressive):

1. Lock a picture file
2. Open the picture in the app to preview it
3. Remove the picture from the preview

snackbar is not displayed

Locking a file can be easily done in oC10 servers. Log in an admin account > Settings (top right menu) > General (left menu, admin side) > tick `Enable manual file locking in the web interface`

Once enabled, every file (not folder) in the list will have an option `Lock file` available. Operations over the locked file will return `423`

this could be fixed here or moved to another PR.

All the places in which i tested and the message appears are correct

tested with different kind of files and previews (pictures, txt, video), performing operations: Move, Remove, Move, Rename (operations not allowed for a locked file). All of them fine.

parneet-guraya:
@jesmrec Will check and push the hotfix here if it's simple enough otherwise we can solve it by raising another issue

parneet-guraya:
Hi @jesmrec, I have been using `demo.owncloud.com` for accessing limited demo server. So, I think this one does not have the privilege to enable/disable file locking and you mentioned `admin` account and I'm not sure how to get access to that. Can you help out here? :-)

jesmrec:
`demo.owncloud.com` can be accessed with `admin/admin` credentials

parneet-guraya:
@jesmrec Hi, the case you said, I was able to reproduce at first attempt. After investigating, the error was getting thrown correctly but at UI level the throwable turns out to be null (couldn't get the chance to see further) that's why it was showing the generic unknown error message.

But, now I can't reproduce it anymore. Let me know if you still see this otherwise we may push this one and later on create separate issue and pickup from there if it ever starts coming again.

https://github.com/owncloud/android/assets/111801812/32446f43-d9f6-42f6-8d83-336106e7e346

jesmrec:
i don't see the `unknown` message anymore, when i try to remove a picture from its preview, there is no snackbar. I still reproduce this consistently, but, as i commented, it's not a regression.

Let's move this forward and i will create another issue since it seems that reproducing is not so straightforward.

JuancaG05:
Need a rebase here if you don't mind @parneet-guraya 😄.

BTW, to automatize changelog, we are using a tool called Calens since some time ago. What we need to do is adding in each PR a plain text file in a certain folder with certain features for it to be recognised. Do you feel like adding them from now on? You can find more info [here](https://github.com/owncloud/android/blob/master/changelog/README.md), and one example [here](https://github.com/owncloud/android/pull/4257/files) (but you can check all of our PR to see more examples 😉). Just if you want, otherwise we can take care of it, thanks! 👍

parneet-guraya:
Sure, I can do it :-)

JuancaG05 (approved):
Thanks a lot @parneet-guraya! It's perfect like that! `CHANGELOG.md` will be modified automatically when merged this into master or when creating a new branch, that's the magic of Calens! 😄

Issue #4279, opened 2024-01-03
[BUG] 4.1.1 composes /app/list path wrong when instance is installed in subfolder (/owncloud)


michaelstingl:
### Actual behaviour
- 4.1.1 Android app queries `/owncloud/owncloud/index.php/app/list` when  instance is installed in subfolder `/owncloud`

### Expected behaviour
- Leading `/` should add `"apps_url": "/owncloud/index.php/app/list",` to host, not to server URL
 
### Steps to reproduce
1. Connect Android app to ownCloud instance installed in subfolder `/owncloud`

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

no

### Environment data

ownCloud app version:
4.1.1 cde9bb841

ownCloud server version:

```
            "version":
            {
                "major": 10,
                "minor": 13,
                "micro": 4,
                "string": "10.13.4",
                "edition": "Community",
                "product": "ownCloud"
            },
```

### Logs

Capabilities:

```
                    "app_providers":
                    [
                        {
                            "enabled": true,
                            "version": "1.1.0",
                            "apps_url": "/owncloud/index.php/app/list",
                            "open_web_url": "/owncloud/index.php/app/open-with-web",
                            "new_url": "/owncloud/index.php/app/new"
                        }
                    ],

```

Request to composed URL:

```
D: 2024-01-03 08:07:44:103(LogBuilder.kt:38) .logHttp()[Network, request] [info] [05f707f0-4c10-4d1e-8d46-27f04e6d2671] Method: GET URL: https://example.com/owncloud/owncloud/index.php/app/list
D: 2024-01-03 08:07:44:104(LogBuilder.kt:38) .logHttp()[Network, request] [header] [05f707f0-4c10-4d1e-8d46-27f04e6d2671] X-Request-ID: 05f707f0-4c10-4d1e-8d46-27f04e6d2671
D: 2024-01-03 08:07:44:105(LogBuilder.kt:38) .logHttp()[Network, request] [header] [05f707f0-4c10-4d1e-8d46-27f04e6d2671] User-Agent: Mozilla/5.0 (Android) ownCloud-android/4.1.1
D: 2024-01-03 08:07:44:106(LogBuilder.kt:38) .logHttp()[Network, request] [header] [05f707f0-4c10-4d1e-8d46-27f04e6d2671] Accept-Language: en
[...]

PR #4287, opened 2024-01-18
[BUG] 4.1.1 composes /app/list path wrong when instance is installed in subfolder (/owncloud)

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4279

_____

### DIFF
--- a/changelog/unreleased/4287
+++ b/changelog/unreleased/4287
@@ -1,0 +1,7 @@
+Bugfix: Fix in the handling of the base URL
+
+Base URL has been formatted in GetRemoteAppRegistryOperation
+when server instance is installed in subfolder, so that the endpoint is formed correctly.
+
+https://github.com/owncloud/android/issues/4279
+https://github.com/owncloud/android/pull/4287
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetRemoteAppRegistryOperation.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetRemoteAppRegistryOperation.kt
@@ -77,14 +77,10 @@ class GetRemoteAppRegistryOperation(private val appUrl: String?) : RemoteOperati
 
     private fun removeSubfolder(url: String): String {
         val doubleSlashIndex = url.indexOf("//")
-        return if (doubleSlashIndex >= 0) {
-            val nextSlashIndex = url.indexOf('/', doubleSlashIndex + 2)
-            if (nextSlashIndex >= 0) {
-                val result = url.substring(0, nextSlashIndex)
-                return if (result.endsWith("/")) result else "$result/"
-            } else {
-                if (url.endsWith("/")) url else "$url/"
-            }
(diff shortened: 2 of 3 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetRemoteAppRegistryOperation.kt
```diff
@@ -75,4 +74,19 @@ class GetRemoteAppRegistryOperation(private val appUrl: String?) : RemoteOperati
 
         return result
     }
+
+    private fun removeSubfolder(url: String): String {
+        val doubleSlashIndex = url.indexOf("//")
```
JuancaG05, line 79: Why `//`? What is useful for checking if the URL contains `//`? I don't understand the rest of the algorithm though

JuancaG05 (changes requested):
Here I leave some changes requested and doubts @Aitorbp!

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetRemoteAppRegistryOperation.kt
Aitorbp: The algorithm is formed so that if it finds a slash `/` it deletes everything in front of it, thus eliminating the subfolders. But the algorithm can detect these slashes // and will think that it has to remove the content in front since it has found a slash. Therefore, it must be validated that if there are two `//` slashes in a row, it does not delete the content afterwards.
In a more practical way this would be the result in each case:
Sent URL:
https://test.issue.com
Returned URL:
https://test.issue.com

URL with subfolder:
https://test.issue.com/folder1/folder2
Returned URL:
https://test.issue.com

By the way, I changed a little bit the logic because Base URL is always supposed to have `https://` or `htpp://`.

changelog/unreleased/4287
```diff
@@ -0,0 +1,7 @@
+Bugfix: Fix in the handling of the base url
+
+Url base has been formatted in GetRemoteAppRegistryOperation
+when server instance is installed in subfolder.
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/GetRemoteAppRegistryOperation.kt
JuancaG05: The returned URLs in your comment are not correct, they must finish in `/`. But the code is correct, so let's resolve this

JuancaG05 (approved):
LGTM

jesmrec:
QA checks

- [X] URL without subfolder without trailing slash
- [X] URL without subfolder with trailing slash
- [X] URL with subfolder without trailing slash
- [X] URL with subfolder with trailing slash

jesmrec:
approved on my side

Issue #4288, opened 2024-01-19
[FEATURE REQUEST] [OIDC] Branding option to remove "login_hint" and "user" parameter from re-login URL


michaelstingl:
**Is your feature request related to a problem? Please describe.**

The ownCloud-internal username not always works fine on the OpenID Connect IdP. When ownCloud clients append `&login_hint=test&user=test` to the re-login URL, in some cases the IdP has no way to let the user log in again.

**Describe the solution you'd like**

I'd need a branding option to remove "login_hint" and "user" parameter from re-login URL.

**Describe alternatives you've considered**

We discussed requesting other user attributes from the `/ocs/v1.php/cloud/user` endpoint, and using those for the "login_hint" and "user", but it sounds like a nightmare to maintain and test.

**Additional context**

Re-login URL:

```
https://demo.owncloud.com/index.php/apps/oauth2/authorize?response_type=code
&client_id=xdXOt13JKxym1B1QcEncf2XDkLAexMBFwiT9j6EfhhHFJhs2KM9jbjTmf8JBXE69
&redirect_uri=http://localhost:61901
&code_challenge=DSdzUn4on0L_NTAXYQ-_Iwk8y2KGsn9bzatz0roYy5I
&code_challenge_method=S256
&scope=openid offline_access email profile
&prompt=select_account consent
&state=ifvvlCpuaSiQxC7UhvjyHoXWFFDgvaltQmkr7wLiQPM%3D
&login_hint=test
&user=test
```

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

@JuancaG05 @Aitorbp @jesmrec any chance we can get this in the upcoming 4.2 release?

michaelstingl:
Related:

- Desktop: https://github.com/owncloud/client/issues/11489
- Android: https://github.com/owncloud/android/issues/4288

Aitorbp:
We will take a look, I think it could be in the upcoming 4.2 release. 👍 @michaelstingl

jesmrec:
1. Add `login_hint`
2. Add branding / MDM parameter

Default: true ¿?

PR #4291, opened 2024-01-22
[FEATURE REQUEST] [OIDC] Branding option to remove `login_hint` and `user` parameter from re-login URL

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4288

_____

## QA

https://github.com/owncloud/android/pull/4291#issuecomment-1907946457

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/oauth/OAuthUtils.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/oauth/OAuthUtils.kt
@@ -2,7 +2,9 @@
  * ownCloud Android client application
  *
  * @author David González Verdugo
- * Copyright (C) 2020 ownCloud GmbH
+ * @author Juan Carlos Garrote Gascón
+ *
+ * Copyright (C) 2024 ownCloud GmbH
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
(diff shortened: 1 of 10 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/oauth/OAuthUtils.kt
```diff
@@ -113,7 +117,10 @@ class OAuthUtils {
                 appendQueryParameter(QUERY_PARAMETER_CODE_CHALLENGE, codeChallenge)
                 appendQueryParameter(QUERY_PARAMETER_CODE_CHALLENGE_METHOD, CODE_CHALLENGE_METHOD)
                 appendQueryParameter(QUERY_PARAMETER_STATE, state)
-                if (!username.isNullOrEmpty()) appendQueryParameter(QUERY_PARAMETER_USERNAME, username)
+                if (sendLoginHintAndUser && !username.isNullOrEmpty()) {
+                    appendQueryParameter(QUERY_PARAMETER_USER, username)
```
Aitorbp, line 121: why have you changed the name from `QUERY_PARAMETER_USERNAME` to `QUERY_PARAMETER_USER`?. Why haven't you also changed it in the `username` variable?

Aitorbp (changes requested):
Look good for me 👍  just one doubt in comments and Bitrise is returning an error. I think you have to review the unit tests.

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/oauth/OAuthUtils.kt
JuancaG05: Because the name of the parameter is `user` and not `username` as you can see in the definition of that constant. I kept `username` like that because otherwise it can be confused with an user (`OCAccount`) object, in this case it's just the name. So, this way we keep consistency for both network request (keeping the same name as the parameter) and business logic (not confusing it with the user object)

Aitorbp (approved):
LGTM 🥇

jesmrec:
QA Checks

## Branding parameter 

### send_login_hint_and_user = true

- [X] OAuth2 login -> no user, no login_hint
- [X] OAuth2 re-login -> user and login hint
- [X] OIDC login -> no user, no login_hint
- [X] OIDC re-login -> user and login hint

### send_login_hint_and_user = false

- [X] OAuth2 login -> no user, no login_hint
- [X] OAuth2 re-login -> no user, no login_hint
- [X] OIDC login -> no user, no login_hint
- [X] OIDC re-login -> no user, no login_hint

## MDM parameter 

### send_login_hint_and_user = true

- [X] OAuth2 login -> no user, no login_hint
- [X] OAuth2 re-login -> user and login hint
- [X] OIDC login -> no user, no login_hint
- [X] OIDC re-login -> user and login hint

### send_login_hint_and_user = false

- [X] OAuth2 login -> no user, no login_hint
- [X] OAuth2 re-login -> no user, no login_hint
- [X] OIDC login -> no user, no login_hint
- [X] OIDC re-login -> no user, no login_hint

jesmrec:
One consideration:

I noticed this is affecting the legacy webfinger flow somehow: if `send_login_hint_and_user` is true, the username is propagated from the login view to the idP. If if `send_login_hint_and_user` is false, it is not. Since the feature and the new parameter should only affect the re-login process as defined, this might need a look.

jesmrec:
I also noticed that `scope` and `prompt` are sent empty to OAuth2, because they are OIDC parameters. This is not in the scope of this PR, i will check and open issue in order to discuss whether that behaviour could be improved.

jesmrec:
> I noticed this is affecting the legacy webfinger flow somehow: if send_login_hint_and_user is true, the username is propagated from the login view to the idP. If if send_login_hint_and_user is false, it is not. Since the feature and the new parameter should only affect the re-login process as defined, this might need a look.

this is the expected behaviour since the login and relogin calls are the same. We can not prevent the side-effect is attached to the branding decision regarding the  `send_login_hint_and_user` parameter.

PR #4296, opened 2024-01-25
[FIX] Hardcoded name in logs leads to crash

Aitorbp:

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/OCFileLoggingTree.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/OCFileLoggingTree.kt
@@ -37,7 +37,7 @@ import java.util.UUID
 class OCFileLoggingTree(
     externalCacheDir: File,
     context: Context? = null,
-    var filename: String = UUID.randomUUID().toString(),
+    filename: String = UUID.randomUUID().toString(),
     private val newLogcat: Boolean = true,
 ) : Timber.DebugTree() {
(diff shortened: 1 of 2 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/utils/OCFileLoggingTree.kt
```diff
@@ -58,7 +58,11 @@ class OCFileLoggingTree(
             it.list()?.let { logFiles ->
                 if (logFiles.isNotEmpty()) {
 
-                    val lastDateLogFileString = logFiles.last().substringAfterLast("owncloud.").substringBeforeLast(".log")
+                    val lastDateLogFileString = if (context != null) {
+                        logFiles.last().substringAfterLast("${context.packageName}.").substringBeforeLast(".log")
+                    } else {
+                        logFiles.last().substringBeforeLast(".log")
```
JuancaG05, line 64: The name we give a log file in case that `context` is null (shouldn't happen normally) is like `$filename.$fileNameTimestamp.log`. If we consider this, this condition will take `$filename.$fileNameTimestamp`, which is not correct and will lead to a parse exception again

JuancaG05 (changes requested):
Check the comment here @Aitorbp

JuancaG05 (approved):

jesmrec:
Tested with branded build from reporting customer, no crash. Will report back to assure that does not happen in customer side.

PR #4310, opened 2024-02-12
Modify GH templates

jesmrec:
Changes:

- In features template, new entry to add feature to release notes. So, it could be translated with more time in advance.
- In release template, new task to check if new screenshots are required to be generated before/during QA phase to be ready in deployment time.
- Add a new task to the release template: move the `latest` tag to the new release commit
- Replace GitFlow for OneFlow in `CONTRIBUTING.md` file
- Remove references to `stable` in release template

## Related Issues

App:

_____

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -27,6 +27,8 @@ Add any other context or screenshots about the feature request here.
 - [ ] Development tasks
   - [ ] Implement whatever
   - [ ] ...
+  - [ ] Implement unit tests (if needed)
+  - [ ] Add feature to Release Notes (if required)
 - [ ] Code review and apply changes requested
 - [ ] Design test plan
 - [ ] QA
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,15 @@ Before we're able to merge your code into the ownCloud app for Android, please, 
 	Please, use the mentioned prefixes because CI system is ready to match with them. Be sure your feature, fix, improvement or technical branches are updated with latest changes in official `android/master`, it will give us a better chance to test your code before merging it with stable code.
 * Once you are done with your code, start a pull request to merge your contribution into official `android/master`.
 * Keep on using pull requests for your next contributions although you own write permissions.
-* Important to mention that ownCloud Android team uses [GitFlow](https://datasift.github.io/gitflow/IntroducingGitFlow.html) as branching model. Please take a look to the link to understand better what it is and how it works. It's something as useful as easy.
+* Important to mention that ownCloud Android team uses OneFlow as branching model. It's something as useful as easy:
+
+  * `master` will stay as main branch. Everything will work around it.
+  * Feature branch: new branch created from `master`. Once it is finished and DoD accomplished, rebased and merged into `master`.
+  * Release branch: will work as any feature branch. Before rebasing and merging into `master`, release tag must be signed.
+  * Hotfix branch: created from latest tag. Once it is finished, tag must be signed. Then, rebased and merged into `master`.
+  * The way to get an specific version is browsing through the tags.
+
+	Interesting [link](https://www.endoflineblog.com/oneflow-a-git-branching-model-and-workflow) about this.
 
 [contribution]: https://owncloud.com/contribute/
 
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -30,4 +30,5 @@ Add any other context or screenshots about the feature request here.
 - [ ] Code review and apply changes requested
 - [ ] Design test plan
 - [ ] QA
+- [ ] Add feature to Release Notes (if required)
```
JuancaG05, line 33: I would add this inside the "Development tasks" sublevel checks, the last one, or even like an additional check at the same level after "Development tasks". What do you think?

JuancaG05: Also, taking advantage of this PR, we could add a new check for "Implement unit tests (if needed)", so that we don't forget about them (we'll let UI tests stay for the moment).
Same, inside the "Development tasks" as a new check after `- [ ] ...` or as a new item at the same level. Personally, I would introduce both inside the "Development tasks" section, so that we englobe everything related to the development phase in one same item 👍

JuancaG05 (changes requested):
Some suggestions here, let me know what do you think @jesmrec 👍

.github/ISSUE_TEMPLATE/feature_request.md
jesmrec: Will add both to the Development tasks. So, it will look like:

```
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
  - [ ] Add feature to Release Notes (if required)
```

JuancaG05 (approved):

CONTRIBUTING.md
```diff
@@ -36,7 +36,15 @@ Before we're able to merge your code into the ownCloud app for Android, please,
 	Please, use the mentioned prefixes because CI system is ready to match with them. Be sure your feature, fix, improvement or technical branches are updated with latest changes in official `android/master`, it will give us a better chance to test your code before merging it with stable code.
 * Once you are done with your code, start a pull request to merge your contribution into official `android/master`.
 * Keep on using pull requests for your next contributions although you own write permissions.
-* Important to mention that ownCloud Android team uses [GitFlow](https://datasift.github.io/gitflow/IntroducingGitFlow.html) as branching model. Please take a look to the link to understand better what it is and how it works. It's something as useful as easy.
+* Important to mention that ownCloud Android team uses OneFlow. as branching model. It's something as useful as easy:
```
JuancaG05, line 39:
```suggestion
* Important to mention that ownCloud Android team uses OneFlow as branching model. It's something as useful as easy:
```

CONTRIBUTING.md
```diff
@@ -36,7 +36,15 @@ Before we're able to merge your code into the ownCloud app for Android, please,
...
+
+  * `master` will stay as main branch. Everything will work around it.
+  * Feature branch: new branch created from `master`. Once it is finished and DoD accomplished, rebased and merged into `master`.
+  * Release branch: will work as any feature branch. Before rebasing and merging into `master`, release tag must be signed.
+  * Hotfix branch: created from latest tag. Once it is finished, tag must be signed. Then, rebased and merged into `master`.
+  * The way to get an specific version is browsing through the tags.
```
JuancaG05, line 45: Idk if we talk about tags in some other point of the documentation, but would be interesting mentioning that it requires a separate `fetch` to update them, and including the command used to do it. What do you think?

JuancaG05 (commented):
A tiny correction and a comment, give me your insight about it @jesmrec 😁

CONTRIBUTING.md
jesmrec: you're right, i will fix 👍

CONTRIBUTING.md
jesmrec: Not sure about this, because it's an specific detail of the git use. We should guide users through the process, not teaching how to use an external (involved) tool beyond the steps to handle the repo. I get your point, but also, i don't know in which point i'd add that.

JuancaG05: In this same document, in section `2. Create pull request` for example, we include several git commands to help the users. It's true that this is not totally part of the process but would be good to know in case someone wants to work among different versions. Not sure where  and if including it either 🤷🏼‍♂️

jesmrec: Added a new section in `CONTRIBUTING.md` file about "Versioning" . What do you think?

JuancaG05 (approved):
Nice!! I like it 💯

Issue #4301, opened 2024-01-29
[FEATURE REQUEST] Make automatic discovery of the account


jesmrec:
EPIC:
- https://github.com/owncloud/android/issues/4314

The `Manage accounts` view looks like that:

![Screenshot 2024-01-29 at 13 04 21](https://github.com/owncloud/android/assets/14894746/ff08922b-de89-4d01-80f3-e27381bcc593)

The rounded arrow icon discovers the whole app by clicking on it. This is a process that can be automated to get rid of that icon and use another one instead.

The process of discovering the app should be done at the first login and should affect:

- Personal space (the only existing one in case of oC10)
- Spaces (in case of oCIS account)
- Shared jail (in case of oCIS account) -> BLOCKED due to https://github.com/owncloud/android/issues/4250

Discovery process to be done in background mode.

Once discovered, if new items are uploaded, will be discovered in browsing mode.

After finishing that migration process, the rounded arrow icon will be removed from the `Manage accounts` view.

PR #4325, opened 2024-02-21
[FEATURE REQUEST] Make automatic discovery of the account 

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4301

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.3/Discovery.md

Some regards: https://github.com/owncloud/android/pull/4325#issuecomment-1982771824

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/accounts/AccountsManagementActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/accounts/AccountsManagementActivity.kt
@@ -2,8 +2,9 @@
  * ownCloud Android client application
  *
  * @author Javier Rodríguez Pérez
+ * @author Aitor Ballesteros Pavón
  *
- * Copyright (C) 2022 ownCloud GmbH.
+ * Copyright (C) 2024 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -25,15 +26,12 @@ import android.accounts.AccountManager
 import android.accounts.AccountManagerCallback
 import android.accounts.AccountManagerFuture
 import android.accounts.OperationCanceledException
-import android.content.ContentResolver
 import android.content.Intent
-import android.content.SyncRequest
 import android.os.Bundle
 import android.view.MenuItem
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.owncloud.android.MainApp.Companion.accountType
-import com.owncloud.android.MainApp.Companion.authority
 import com.owncloud.android.MainApp.Companion.initDependencyInjection
 import com.owncloud.android.R
 import com.owncloud.android.presentation.accounts.RemoveAccountDialogFragment.Companion.newInstance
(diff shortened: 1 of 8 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/accounts/AccountsManagementActivity.kt
```diff
@@ -2,8 +2,9 @@
  * ownCloud Android client application
  *
  * @author Javier Rodríguez Pérez
+ * @author Aitor Ballesteros Pavón
  *
- * Copyright (C) 2022 ownCloud GmbH.
+ * Copyright (C) 2024 ownCloud GmbH.
```
JuancaG05, line 7: When we're just removing content, there's no need to update the header, you didn't add any new content and so you are not author of anything in this class

owncloudApp/src/main/java/com/owncloud/android/presentation/releasenotes/ReleaseNotesViewModel.kt
```diff
@@ -46,60 +46,10 @@ class ReleaseNotesViewModel(
     companion object {
         val releaseNotesList = listOf(
             ReleaseNote(
-                title = R.string.release_notes_4_2_0_title_1,
-                subtitle = R.string.release_notes_4_2_0_subtitle_1,
+                title = R.string.release_notes_4_3_0_title_2,
+                subtitle = R.string.release_notes_4_3_0_subtitle_2,
                 type = ReleaseNoteType.CHANGE,
```
JuancaG05, line 51: More than change, I would say it is an enhancement because this is done automatically now

JuancaG05 (changes requested):
Some tiny changes here and ready to go @Aitorbp

owncloudApp/src/main/java/com/owncloud/android/presentation/accounts/AccountsManagementActivity.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/releasenotes/ReleaseNotesViewModel.kt
Aitorbp: Done

JuancaG05 (approved):
LGTM

jesmrec:
It seems that the branch is conflicting with the current `master`, could you take a look?

jesmrec:
Feature works file but is not good-performant. The process to discover is the same as before (so, correct) but the complete discovery in accounts with great amount of folders is pretty slow. Just an example

Account with:

~13.000 folders
+20.000 files

took longer than 3 hours to be completely discovered. As first approach is OK, because the given example is just an extreme one to test the app. But, probably, there are other ways to do this task in a more optimal way like [depth infinity](https://github.com/owncloud/ios-app/issues/950) (needs server logic as well).

Other feature (not a bug): If server connection is lost somehow during the discovery process, this is never completed and there is no way to retry/trigger it because of the removal of the circled arrow icon. Know restriction not to forget.

jesmrec:
Let's move this forward, without forgetting the known restrictions

Issue #4311, opened 2024-02-13
[BUG] Lens icon in folder picker


jesmrec:
### Steps to reproduce

1.  In a oCIS account, select any item and select `Copy`

1. Open a 3rd party app and share with oCIS account with spaces

1. Open Settings > Auto Uploads > Picture/Video uploads path

### Actual behaviour

All of those options shows the list of spaces including the lens icon

![Screenshot 2024-02-13 at 13 52 21](https://github.com/owncloud/android/assets/14894746/c2dbbefa-9710-42ca-8c29-ac4f631f5ac5)

### Expected behaviour

Lens icon hidden till https://github.com/owncloud/android/issues/3865 is implemented.
 
Pixel2 Android11
`v4.2.0`

Aditya13s:
can i start working on it? @jesmrec

JuancaG05:
Hi @Aditya13s! Sure! Go for it 🚀

jesmrec:
> can i start working on it? @jesmrec

yes, go ahead

Aditya13s:
@jesmrec do i need to create a new server or is there any other way to test the app?

jesmrec:
server [ocis.owncloud.works](https://ocis.owncloud.works) is always available. You can use it

JuancaG05:
Hi @Aditya13s! It's been a while since you don't reply, so we've assigned this issue to other person in the team as we want to include it in the next release. In any case, feel free to take any other issue that is not assigned! 😃🍻

PR #4339, opened 2024-03-14
[FIX] Lens icon in folder picker

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4311

_____

## QA

Checks done: https://github.com/owncloud/android/pull/4339#issuecomment-2003730072

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
@@ -159,14 +159,18 @@ open class FolderPickerActivity : FileActivity(),
 
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
         menuInflater.inflate(R.menu.main_menu, menu)
-        val menuItem = menu!!.findItem(R.id.action_search)
-        val searchView = menuItem.getActionView() as SearchView
-        searchView.maxWidth = Int.MAX_VALUE
-        searchView.queryHint = resources.getString(R.string.actionbar_search)
-        if (isInSpacesScreen) {
-            menuItem.setVisible(false)
+        menu?.let {
+            it.findItem(R.id.action_search)?.let { menuItem ->
+                (menuItem.actionView as? SearchView)?.apply {
+                    maxWidth = Int.MAX_VALUE
+                    queryHint = resources.getString(R.string.actionbar_search)
+                }
+                if (isInSpacesScreen) {
+                    menuItem.isVisible = false
+                }
+            }
+            menu.removeItem(menu.findItem(R.id.action_share_current_folder)?.itemId ?: 0)
         }
-        menu!!.removeItem(menu!!.findItem(R.id.action_share_current_folder).itemId)
         return true
     }
 
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
@@ -57,9 +57,6 @@ open class FolderPickerActivity : FileActivity(),
 
     private var isInSpacesScreen: Boolean = false
 
-    private lateinit var menuItem: MenuItem
-
-    private lateinit var searchView: SearchView
 
     override fun onCreate(savedInstanceState: Bundle?) {
         Timber.d("onCreate() start")
@@ -162,8 +159,8 @@ open class FolderPickerActivity : FileActivity(),
 
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
         menuInflater.inflate(R.menu.main_menu, menu)
-        menuItem = menu!!.findItem(R.id.action_search)
-        searchView = menuItem.getActionView() as SearchView
+        val menuItem = menu!!.findItem(R.id.action_search)
+        val searchView = menuItem.getActionView() as SearchView
         searchView.maxWidth = Int.MAX_VALUE
         searchView.queryHint = resources.getString(R.string.actionbar_search)
         if (isInSpacesScreen) {
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
@@ -160,16 +160,15 @@ open class FolderPickerActivity : FileActivity(),
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
         menuInflater.inflate(R.menu.main_menu, menu)
         menu?.let {
-            it.findItem(R.id.action_search)?.let { menuItem ->
-                (menuItem.actionView as? SearchView)?.apply {
-                    maxWidth = Int.MAX_VALUE
-                    queryHint = resources.getString(R.string.actionbar_search)
-                }
-                if (isInSpacesScreen) {
-                    menuItem.isVisible = false
-                }
+            val menuItem = it.findItem(R.id.action_search)
+            (menuItem.actionView as SearchView).apply {
+                maxWidth = Int.MAX_VALUE
+                queryHint = resources.getString(R.string.actionbar_search)
             }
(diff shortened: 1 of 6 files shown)


owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -179,6 +179,10 @@ public class ReceiveExternalFilesActivity extends FileActivity
     private OCSharedPreferencesProvider sharedPreferencesProvider;
     private OCSpace personalSpace;
 
+    private boolean isInSpacesList;
+
+    private MenuItem menuItem;
```
JuancaG05, line 184: Do we really need this as an attribute of the class? We're just using it in one place 🤔

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -54,6 +55,12 @@ open class FolderPickerActivity : FileActivity(),
 
     private lateinit var binding: FilesFolderPickerBinding
 
+    private var isInSpacesScreen: Boolean = false
+
+    private lateinit var menuItem: MenuItem
+
+    private lateinit var searchView: SearchView
```
JuancaG05, line 62: Same as in the other: these 2 properties are just used in 1 place, they don't need to be properties. You can remove them. Also, in `ReceiveExternalFilesActivity` you can remove `mSearchView` as an attribute and transform it into a local variable 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -155,6 +162,14 @@ open class FolderPickerActivity : FileActivity(),
...
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
         menuInflater.inflate(R.menu.main_menu, menu)
+        menuItem = menu!!.findItem(R.id.action_search)
+        searchView = menuItem.getActionView() as SearchView
+        searchView.maxWidth = Int.MAX_VALUE
+        searchView.queryHint = resources.getString(R.string.actionbar_search)
+        if (isInSpacesScreen) {
+            menuItem.setVisible(false)
+        }
+        menu!!.removeItem(menu!!.findItem(R.id.action_share_current_folder).itemId)
```
JuancaG05, line 172: We should avoid `!!` as much as we can, since it could lead to `NullPointerException`s. We can just surround all these lines by a `menu?.{ ... }`, which will only run if menu is not `null`

JuancaG05 (changes requested):
Good job @joragua! Here I left some comments for you to review 😃

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
joragua: Yeah. I have changed those two propertis and now are local values in `onCreateOptionsMenu` function.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
joragua: In this case, `isInSpacesList` must to be an attribute because it's used in `onCreateOptionsMenu` which is part of lifecycle activity. On the other hand, I have changed `menuItem` because it's use only once in `onCreateOptionsMenu` function and now it's a local variable.

joragua: Ok. I will keep that in my mind for next time. 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
joragua: Ok. It's done ✅

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -735,7 +735,8 @@
     <string name="release_notes_4_3_0_subtitle_3">Removed the key icon, avoid overlap account name with icons and redirect to login when the snackbar appears in authentication failure</string>
     <string name="release_notes_4_3_0_title_4">Name conflicts dialog more Android-alike</string>
     <string name="release_notes_4_3_0_subtitle_4">Name conflicts dialog appearance was changed to look Android-alike and more similar to other dialogs in the app</string>
-
+    <string name="release_notes_4_3_0_title_5">Lens icon in folder picker</string>
+    <string name="release_notes_4_3_0_subtitle_5">The lens icon in the toolbar was removed in folder picker</string>
```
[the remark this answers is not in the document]
joragua: Ok. I will keep that in my mind for next time. 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -753,12 +756,14 @@ private void initTargetFolder() {
     public boolean onCreateOptionsMenu(Menu menu) {
         MenuInflater inflater = getMenuInflater();
         inflater.inflate(R.menu.main_menu, menu);
-
-        mSearchView = (SearchView) menu.findItem(R.id.action_search).getActionView();
+        MenuItem menuItem = menu.findItem(R.id.action_search);
```
JuancaG05, line 759: Cool! You can do the same with `mSearchView` 😀

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -155,6 +159,18 @@ open class FolderPickerActivity : FileActivity(),
 
     override fun onCreateOptionsMenu(menu: Menu?): Boolean {
         menuInflater.inflate(R.menu.main_menu, menu)
+        menu?.let {
+            it.findItem(R.id.action_search)?.let { menuItem ->
```
JuancaG05, line 163: In this case, we don't need a nullable check for `findItem` since it returns `MenuItem!`, or which is the same, a non-nullable `MenuItem`

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -155,6 +159,18 @@ open class FolderPickerActivity : FileActivity(),
...
+                (menuItem.actionView as? SearchView)?.apply {
```
JuancaG05, line 164: Why using `as?`? There's no need to in this case 😀

JuancaG05, line 164: `apply` is correctly used, nice! 🥇

JuancaG05 (changes requested):
Some more changes suggested here @joragua! 😁

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
joragua: Perfect. It's done ✅

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
joragua: Ok, I have deleted that nullable check and it's working.  👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
joragua: I changed that and now it's `as` instead `as?` ✅

JuancaG05 (approved):
LGTM! Good job @joragua! 😄

Issue #4309, opened 2024-02-12
[BUG] "Clear data" button malfunctioning


JuancaG05:
EPIC:
- https://github.com/owncloud/android/issues/4314

### Actual behaviour
When going to the device Settings > Apps > ownCloud > Storage > Manage storage/Clear data, a screen of our own is opened with a button that says "Clear data". Pressing it lets the app and its database in a weird and inconsistent state (a new folder /Shares is created, spaces ID are removed from the files...), which leads to unexpected behaviours.

### Expected behaviour
The usual behaviour here is removing all the data from the app, letting the OS take care of it. The key here is, if we currently remove `ManageSpaceActivity.java` and the `android:manageSpaceActivity="com.owncloud.android.ui.activity.ManageSpaceActivity"` property from the Android Manifest, the  OS takes care of clearing the data but the strange status continues happening. So, there is something in the logic of our app that makes it put itself into this inconsistent status, that's what we have to research here.
 
### Steps to reproduce
1. Go to device Settings > ownCloud > Storage > Manage storage/Clear data
2. Press the button "Clear data" from our screen
3. Notice the strange status when the app is opened again (a /Shares folder appears in root in Personal space, deep links are not working...)

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Yes, this seems to be a problem related to the app itself so the server shouldn't affect.

### Environment data
Android version: 9

Device model: Samsung Galaxy S9

Stock or customized system: Stock

ownCloud app version: 4.1.1

ownCloud server version: doesn't mind here

Aitorbp:
In the following text I will explain in detail the research I have done on this issue.
This problem happens only in `OCIS`.
First of all, let's clarify what the Clear data button is for. The Android message that appears when we click on this option is the following: `All of this application's data will be deleted permanently. This includes all files, settings, accounts, databases etc`. I want to point out that according to this text the accounts associated to the application should be deleted, but as we have the application this is not happening and it could be one of the focuses to solve the problem. We manage the accounts in the `ManageAccount`, so the `Clear data` does not affect this object.
The data that is being deleted by pressing the Clear data button are the following: **Sharedpreferences, temporary files and the local database being cleared**.

[...]

PR #4343, opened 2024-03-19
[BUG] "Clear data" button malfunctioning

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4309

_____


JuancaG05 (changes requested):
Cannot approve this. This is deleting the data of the app (including shared preferences) but not the logged in accounts. It results in removing the security lock methods when you press the "Clear data" button but keep the access to the accounts. We'll need to discuss how to address this issue in an alternative way.

Issue #4284, opened 2024-01-16
[FEATURE REQUEST] Add warning in http connections


jesmrec:
When `http://` url is set on the login view it's straight approved with no warning to the user. http connections are not secure, and it's relevant that users are warned about that issue.

This is the iOS warning:

<img width="327" alt="Screenshot 2024-01-16 at 08 45 59" src="https://github.com/owncloud/android/assets/14894746/491a0e47-8f10-4fc7-af31-18c86f35a195">

This is de Desktop client warning:

![Screenshot 2024-01-16 at 08 47 01](https://github.com/owncloud/android/assets/14894746/3e6ec35e-d38f-4940-9e81-859da6d1c2fd)

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

michaelstingl:
@tbsbdr @EParzefall anything you'd recommend? We can also change the existing iOS + Desktop warnings.

EParzefall:
My suggestion for all clients:

Insecure HTTP URL
The URL you provided uses HTTP rather than the encrypted HTTPS protocol. If you continue, your communication will not be encrypted.

For consistency reasons all affirmative buttons should be labeled "Continue".

rohnsha0:
@EParzefall Shall I implement? the warnings?

jesmrec:
go ahead @rohnsha0

PR #4345, opened 2024-03-19
[FEATURE REQUEST] Add warning in http connections

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4284

_____


owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -345,53 +348,67 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
                     text = getString(R.string.auth_connection_established)
                     setCompoundDrawablesWithIntrinsicBounds(R.drawable.ic_lock_open, 0, 0, 0)
+                    val builder = AlertDialog.Builder(context)
+                    builder.apply {
+                        setTitle(context.getString(R.string.title_dialog_insecure_http_url))
+                        setMessage(context.getString(R.string.message_dialog_insecure_http_url))
+                        setPositiveButton("Continue") { dialog, which ->
+                            checkServerType(serverInfo)
+                        }
+                        setNegativeButton("Cancel", null)
```
JuancaG05, line 362: "Continue" and "Cancel" strings should be also part of the string resources of the project. Otherwise, for example, a Spanish user using the app in Spanish would see these English strings, which is not what we want 👍. Check `strings.xml` because there can be already some values with those texts, or even some Android default ones (especially for "Cancel")

JuancaG05 (changes requested):
Nice! Just one comment here @joragua 🙂

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: Ok. I changed it! Thanks for the feedback 😃

JuancaG05 (approved):
LGTM! Good job @joragua 🤠

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -737,6 +737,8 @@
     <string name="release_notes_4_3_0_subtitle_4">Name conflicts dialog appearance was changed to look Android-alike and more similar to other dialogs in the app</string>
     <string name="release_notes_4_3_0_title_5">Lens icon in folder picker</string>
     <string name="release_notes_4_3_0_subtitle_5">The lens icon in the toolbar was removed when listing spaces in folder picker</string>
+    <string name="release_notes_4_3_0_title_6">Add warning in http connections</string>
+    <string name="release_notes_4_3_0_subtitle_6">Added a warning dialog in the login screen when it is a http connection</string>
```
jesmrec, line 741: I'd change `screen` (hardware concept) for view here

jesmrec:
QA checks

- [X] http server: warning displayed
- [X] http server: `continue`
- [X] http server: `cancel`
- [X] URL check after 40x
- [ ] https non-secured: certificate warning (existing behaviour)
- [ ] https secured: no warning

PR #4346, opened 2024-03-19
Modify commit messages in calens and dependabot to fit "Conventional commits"

jesmrec:
More info: 

https://www.conventionalcommits.org/en/v1.0.0/#summary

Calens commits to add `docs: ` as prefix
Dependabot commits to add `feat: ` as prefix

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/calens.yml
+++ b/.github/workflows/calens.yml
@@ -29,5 +29,5 @@ jobs:
       - name: Commit files
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
-          commit_message: Calens changelog updated
+          commit_message: "docs: calens changelog updated"
           file_pattern: CHANGELOG.md
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,13 @@ updates:
       interval: "daily"
     labels:
       - "Dependencies"
+    commit-message:
+      prefix: "feat"
+      include: "scope"
   - package-ecosystem: "github-actions"
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: "feat"
+      include: "scope"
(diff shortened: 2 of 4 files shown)


Aitorbp (approved):
LGTM 🥇

.github/workflows/calens.yml
```diff
@@ -29,5 +29,5 @@ jobs:
       - name: Commit files
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
-          commit_message: Calens changelog updated
+          commit_message: "docs: Calens changelog updated"
```
JuancaG05, line 32: I guess you wrote `Calens` with capital letter because it is a proper noun, right?

JuancaG05 (commented):
I added a comment for you to review @jesmrec 👍

.github/workflows/calens.yml
jesmrec: I just let the same message we had before, adding the prefix. Did not think on switching to non-capital, but i can change it

JuancaG05: Since we talked about writing everything in lowercase letters (except for proper nouns I guess), I let the decision in your hands. If you think it is just a normal message, then I'd write it in lowercase, but if you consider it as a proper noun, then it's ok that it has a capital letter

jesmrec: as i understood, lowercase was intended for the pretty first letter in the message, in the prefix (the `d` of `docs`). I will move to lowercase anyway, but we should keep the same criteria in every message. Not a problem i guess.

JuancaG05: Yes, let's keep also the first letter of the description per se in lowercase as well, as agreed and as they show in the official [Examples](https://www.conventionalcommits.org/en/v1.0.0/#examples) 👍

JuancaG05 (approved):
LGTM 👍

Issue #4342, opened 2024-03-19
[FEATURE REQUEST] Improvements in remove confirmation dialog


JuancaG05:
**Is your feature request related to a problem? Please describe.**
The remove confirmation dialog is currently pretty simple, and could be improved by showing some valuable information (being aware of not overloading the dialog, that's not the aim).

**Describe the solution you'd like**
I think it would be nice to apply 2 different improvements:

1. When removing a file with thumbnail, it would be cool to show the thumbnail in the dialog to know exactly which file we are going to delete (also to avoid misclicks). Currently, we just show the title, which can be non-descriptive sometimes:

<img width="300" src="https://github.com/owncloud/android/assets/57049315/f76cc773-7fa2-4d30-9b54-4fcf47f89472">

2. When removing several files, it would be nice to show the number of elements selected. Currently, we show a common message for every case. It's true that there's a conflictive case, in which we select a folder: should we consider it as just 1 element or count recursively all the elements contained there? Right now it looks like this:

<img width="300" src="https://github.com/owncloud/android/assets/57049315/b9444c34-0e41-48bb-b28c-39e595bd7643">

**Describe alternatives you've considered**
Proposals are welcome, and more ideas or suggestions as well 🙂.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/improvements_remove_dialog
- [x] Development tasks
  - [x] Add thumbnail in remove dialog, just for files with thumbnails
  - [x] Add number of elements in remove dialog when selecting several files to remove
  - [x] Implement unit tests (if needed)
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/improvements_remove_dialog into master

jesmrec:
> When removing an image, it would be cool to show the thumbnail in the dialog to know exactly which image we are going to delete (also to avoid misclicks). Currently, we just show the title, which can be non-descriptive sometimes:

Should we show everything that contains a thumbnail? for example a txt file, not only an image. If we are ready for this, app will be ready to show any supported thumbnail in the future.

JuancaG05:
> Should we show everything that contains a thumbnail? for example a txt file, not only an image. If we are ready for this, app will be ready to show any supported thumbnail in the future.

Sure! I didn't realise about them. I'll update the main comment, thanks 😄

PR #4348, opened 2024-03-21
[FEATURE REQUEST] Improvements in remove confirmation dialog

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4342

_____

### DIFF
--- a/changelog/unreleased/4348
+++ b/changelog/unreleased/4348
@@ -1,7 +1,7 @@
 Enhancement: Improvements in remove dialog alert
 
 A custom dialog alert has been added when the file that is going to be deleted has thumbnail.
-There have also been some changes in multiple selection.
+Also, when removing files in multiple selection, the number of elements that are going to be removed is displayed in the dialog.
 
 https://github.com/owncloud/android/issues/4342
 https://github.com/owncloud/android/pull/4348
--- a/changelog/unreleased/4348
+++ b/changelog/unreleased/4348
@@ -1,0 +1,7 @@
+Enhancement: Improvements in remove dialog alert
+
+A custom dialog alert has been added when the file that is going to be deleted has thumbnail.
+There have also been some changes in multiple selection.
+
+https://github.com/owncloud/android/issues/4342
+https://github.com/owncloud/android/pull/4348
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -1133,7 +1133,7 @@ class MainFileListFragment : Fragment(),
 
             R.id.action_remove_file -> {
                 if (checkedFiles.size == 1 && !checkedFiles[0].isFolder) {
-                    context?.let { showRemoveCustomDialog(checkedFiles[0], it) }
+                    showRemoveCustomDialog(checkedFiles[0], requireContext())
                 } else {
                     filesToRemove = checkedFiles
                     fileOperationsViewModel.showRemoveDialog(filesToRemove)
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -5,8 +5,9 @@
  * @author Jose Antonio Barros Ramos
  * @author Juan Carlos Garrote Gascón
  * @author Manuel Plazas Palacio
+ * @author Jorge Aguado Recio
  *
- * Copyright (C) 2023 ownCloud GmbH.
+ * Copyright (C) 2024 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -24,6 +25,8 @@
 package com.owncloud.android.presentation.files.filelist
 
 import android.annotation.SuppressLint
+import android.app.Dialog
+import android.content.Context
 import android.content.Intent
 import android.content.res.ColorStateList
 import android.net.Uri
@@ -35,6 +38,7 @@ import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
+import android.widget.Button
 import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.TextView
@@ -475,8 +479,13 @@ class MainFileListFragment : Fragment(),
                                 }
 
                                 FileMenuOption.REMOVE -> {
-                                    filesToRemove = listOf(file)
-                                    fileOperationsViewModel.showRemoveDialog(filesToRemove)
+                                    if (file.isFolder) {
+                                        filesToRemove = listOf(file)
+                                        fileOperationsViewModel.showRemoveDialog(filesToRemove)
+                                    } else {
+                                        showRemoveCustomDialog(file, context)
+                                    }
+
                                 }
 
                                 FileMenuOption.OPEN_WITH -> {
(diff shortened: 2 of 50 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -1123,8 +1132,12 @@ class MainFileListFragment : Fragment(),
             }
 
             R.id.action_remove_file -> {
-                filesToRemove = checkedFiles
-                fileOperationsViewModel.showRemoveDialog(filesToRemove)
+                if (checkedFiles.size == 1 && !checkedFiles[0].isFolder) {
+                    context?.let { showRemoveCustomDialog(checkedFiles[0], it) }
```
JuancaG05, line 1136: Instead of checking nullability of `context`, let's use better `requireContext()`, which is a function that always returns a non-nullable context or throws a `NullPointerException` (in this case there should always be a context, so NPE shouldn't be thrown ever):
```suggestion
                    showRemoveCustomDialog(checkedFiles[0], requireContext())
```

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -1357,6 +1370,52 @@ class MainFileListFragment : Fragment(),
...
+    private fun showRemoveCustomDialog(file: OCFile, context: Context) {
+        val removeDialog = Dialog(context)
+        removeDialog.apply {
+            setContentView(R.layout.remove_files_dialog)
+            setCancelable(false)
+            show()
+        }
+
+        val imageView = removeDialog.findViewById<ImageView>(R.id.dialog_remove_thumbnail)
+        val content = removeDialog.findViewById<TextView>(R.id.dialog_remove_information)
```
JuancaG05, line 1382: Let's set more accurate variable names, like `thumbnailImageView` and `dialogMessage` (examples), so that when we find the variable in the code we know what it refers to 👍. `content` could lead to a misunderstanding thinking it refers to a container layout, for example

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -1357,6 +1370,52 @@ class MainFileListFragment : Fragment(),
...
+        val localRemoveButton = removeDialog.findViewById<Button>(R.id.dialog_remove_local_only)
+        val yesRemoveButton = removeDialog.findViewById<Button>(R.id.dialog_remove_yes)
+        val noRemoveButton = removeDialog.findViewById<Button>(R.id.dialog_remove_no)
+
+        content.text = String.format(getString(R.string.confirmation_remove_file_alert), file.fileName)
+
+        // Show the thumbnail when the file has it
```
JuancaG05, line 1389:
```suggestion
        // Show the thumbnail when the file has one
```

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -1357,6 +1370,52 @@ class MainFileListFragment : Fragment(),
...
+        val thumbnail = ThumbnailsCacheManager.getBitmapFromDiskCache(file.remoteId)
+
+        if (thumbnail != null) {
+            imageView.setImageBitmap(thumbnail)
+        } else {
+            imageView.visibility = View.GONE
+        }
+
+        // Hide local remove only button when the file it is no available locally
```
JuancaG05, line 1398:
```suggestion
        // Hide "Local only" remove button when the file is not available locally
```

owncloudApp/src/main/res/layout/remove_files_dialog.xml
```diff
@@ -0,0 +1,92 @@
...
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:layout_gravity="center">
+
+    <androidx.cardview.widget.CardView
+        android:id="@+id/cardView"
```
JuancaG05, line 11: An ID like this is not suitable, it is very general and could collide easily with other `CardView`s in the app, in addition that we don't know which `CardView` in the app we're referring to 😄. I think we don't even use this ID, we can safely delete it 👍

owncloudApp/src/main/res/layout/remove_files_dialog.xml
```diff
@@ -0,0 +1,92 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:elevation="10dp"
```
JuancaG05, line 14: All these measures that are written raw here (except for `0dp`) should be part of the `dims.xml` resource file, or use one of the values we already have there.
In this case, is it necessary that we set the elevation? The one by default is not suitable enough for a dialog?

owncloudApp/src/main/res/values-ar/strings.xml
```diff
@@ -487,8 +487,7 @@
   <string name="manage_space_clear_data">مسح البيانات</string>
   <string name="manage_space_error">تعذّر حذف بعض الملفات.</string>
   <string name="local_file_not_found_toast">لم يتم العثور على الملف في نظام الملفات المحلية</string>
-  <string name="confirmation_remove_files_alert">هل ترغب حقًا في إزالة العناصر المحددة؟</string>
-  <string name="confirmation_remove_folders_alert">هل ترغب حقًا في إزالة العناصر المحددة ومحتوياتها؟</string>
+    <string name="confirmation_remove_folders_alert">هل ترغب حقًا في إزالة العناصر المحددة ومحتوياتها؟</string>
```
JuancaG05, line 490: When you remove strings with the translator editor, some spaces are added in the line next to the last string that was removed. It's a bit boring, but we need to remove these spaces by hand so that Transifex doesn't complain in the future 😖

JuancaG05 (changes requested):
Good job here @joragua! I left some comments for you to review 😄

changelog/unreleased/4348
```diff
@@ -0,0 +1,7 @@
+Enhancement: Improvements in remove dialog alert
+
+A custom dialog alert has been added when the file that is going to be deleted has thumbnail.
+There have also been some changes in multiple selection.
```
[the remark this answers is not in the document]
joragua: Ok, it's done.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
joragua: ✅

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
joragua: Changed. ✅

owncloudApp/src/main/res/layout/remove_files_dialog.xml
joragua: In this case, I've deleted that attribute because it could be confusing with others card views.

owncloudApp/src/main/res/layout/remove_files_dialog.xml
joragua: I've changed it with a dimensions that was in dimensions.xml

owncloudApp/src/main/res/values-ar/strings.xml
joragua: I've removed all spaces before string tag.

JuancaG05 (approved):
LGTM! Nice work here @joragua 🥇

jesmrec:
Some QA checks

- [x] Dialog in portrait
- [x] Dialog in landscape
- [x] Dialog with local removal
- [x] Dialog without local removal
- [x] Thumbnail for image
- [x] Thumbnail for text file
- [x] File with no thumbnail
- [x] Folder (no thumbnail)

joragua:
> Feature is functionally correct.
> 
> Just one detail: our convention is that every PR has to contain only two calens' commits: one for our input and another one from the calens job. In this PR, there are three commits, one of them should be removed (good chance to play with interactive rebase)

Ok, I removed those calens' commits with interactive rebase and now it should be ready. 😃

jesmrec:
great! then is totally correct and approved!

Issue #4308, opened 2024-02-06
[FEATURE REQUEST] Password generator for public links


TheOneRing:
**Is your feature request related to a problem? Please describe.**
Along with the password policy, the iOS client implements a password generator.

**Describe the solution you'd like**
The generator creates passwords that follow the policy and in general should have a higher entropy than a user generated password.
The implementation should be aligned with the iOS implementation.

**Additional context**
iOS code: https://github.com/owncloud/ios-sdk/blob/develop/ownCloudSDK/Password%20Policy/OCPasswordPolicy%2BGenerator.m#L24

Possible password policies to fulfill are the following (values are server-customizable):
 ```
 "password_policy": {
                    "max_characters": 72,
                    "min_characters": 8,
                    "min_digits": 1,
                    "min_lowercase_characters": 1,
                    "min_special_characters": 1,
                    "min_uppercase_characters": 1
}
```

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/password_generator_public_links
- [x] Development tasks
  - [x] Create password generator algorithm
  - [x] Add a button to generate password (only for oCIS)
  - [x] Add a button to copy password (only for oCIS)
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/password_generator_public_links into master

jesmrec:
Needs research to check for available solutions in the Android world

JuancaG05:
> Needs research to check for available solutions in the Android world

None of the available solutions was suitable (in addition to the fact that none of them is native).
Developing our own solution then 👨🏼‍💻

PR #4349, opened 2024-03-21
[FEATURE REQUEST] Password generator for public links

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4308

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.3/Password%20Policy%20-%20Generator.md

Reports: 

- [x] (1) Length of label `generate password` https://github.com/owncloud/android/pull/4349#issuecomment-2022325144 [FIXED]
- [x] (2) oC10 support for passwords https://github.com/owncloud/android/pull/4349#issuecomment-2022340640 [FIXED]
- [X] (3) Password initially visible https://github.com/owncloud/android/pull/4349#issuecomment-2022705807 [WONT FIX]


owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
```diff
@@ -325,6 +338,7 @@ class PublicShareDialogFragment : DialogFragment() {
         binding.shareViaLinkPasswordValue.doOnTextChanged { text, _, _, _ ->
             capabilities?.passwordPolicy?.let { passwordPolicy ->
                 updateRequirementsPasswordPolicy(text.toString(), passwordPolicy)
+                enableCopyPasswordButton(text.toString().isNotEmpty())
```
Aitorbp, line 341: Why is the `copy` button activated when a character is typed in the message entry? Wouldn't it be more correct to allow copying the password when it has fulfilled the requirements? Following the same logic as the `save` button. I think it is also better in terms of UX. If I generate a password and delete some characters until the password is invalid, the copy option will still be enabled.

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
```diff
@@ -871,13 +901,30 @@ class PublicShareDialogFragment : DialogFragment() {
         if (!hasPasswordEnforcedFor && capabilities?.filesSharingPublicPasswordEnforced == CapabilityBooleanType.TRUE) {
             setPasswordEnforced()
         }
+
+        capabilities?.passwordPolicy?.let { passwordPolicy ->
```
Aitorbp, line 905: The `setOnClickListener` could be reformatted.

owncloudApp/src/main/res/layout/share_public_dialog.xml
```diff
@@ -112,6 +112,34 @@
                 android:inputType="textPassword"
                 android:visibility="gone" />
 
+            <LinearLayout
+                android:id="@+id/layoutPasswordGeneratorButtons"
```
Aitorbp, line 116: I noticed that in this `xml` we don't put the word `layout` in the `id` when we are referring to a layout. Anyway, looking a bit at other layouts in most cases we are putting it at the end and not at the beginning. Just an observation.

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -590,6 +590,8 @@
     <string name="share_via_link_expiration_date_label">Expiration</string>
     <string name="share_via_link_password_label">Password</string>
     <string name="share_via_link_password_title">Secured</string>
+    <string name="share_via_link_generate_password_button">Generate password</string>
+    <string name="share_via_link_copy_password_button">Copy</string>
```
Aitorbp, line 594: This could be sorted alphabetically. Maybe in another issue...

owncloudApp/src/test/java/com/owncloud/android/presentation/sharing/PublicLinkPasswordUtilsTest.kt
```diff
@@ -0,0 +1,102 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.presentation.sharing
+
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class PublicLinkPasswordUtilsTest {
```
Aitorbp, line 26: To be consistent we should call this test class in the same way as the parent class. There is an `"s"` left over or to be added: `PublicLinkPasswordUtilsTest`

owncloudApp/src/test/java/com/owncloud/android/presentation/sharing/PublicLinkPasswordUtilsTest.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val charsetUppercase = ('A'..'Z').toList()
+    private val charsetDigits = ('0'..'9').toList()
+    private val charsetSpecial = listOf('!', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',', '-', '.', '/', ':', ';', '<', '=', '>', '?', '@', '[', '\\', ']', '^', '_', '`', '{', '|', '}', '~')
+
+    private val minCharacters = 10
+    private val maxCharacters = 100
+    private val minDigits = 2
+    private val minLowercaseCharacters = 2
+    private val minUppercaseCharacters = 2
+    private val minSpecialCharacters = 2
```
Aitorbp, line 38: why don't we put the values that are coming right now from the server?

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
```diff
@@ -213,6 +218,14 @@ class PublicShareDialogFragment : DialogFragment() {
         binding.saveButton.setOnClickListener { onSaveShareSetting() }
         binding.cancelButton.setOnClickListener { dismiss() }
 
+        binding.copyPasswordButton.setOnClickListener {
+            val passwordText = binding.shareViaLinkPasswordValue.text.toString()
+            val clipboard = requireActivity().getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+            val clip = ClipData.newPlainText("Public link", passwordText)
```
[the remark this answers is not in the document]
JuancaG05: Not necessary, it's just a label for the clip data, it doesn't need to be translatable and it's not even visible in some devices

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
```diff
@@ -213,6 +218,14 @@ class PublicShareDialogFragment : DialogFragment() {
...
+            clipboard.setPrimaryClip(clip)
+            showMessageInSnackbar(getString(R.string.clipboard_text_copied))
```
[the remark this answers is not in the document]
JuancaG05: I see this message appears in a Toast by default but just from a specific One UI version, the UI layer for Samsung devices. We can find the same behaviour by going to Settings > App version. Knowing this, I'd say it's better to keep both messages, better show the message twice to the user than not giving them feedback at all

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
JuancaG05: I think it's a separate feature, on the one hand you can generate a password and on the other hand you can copy it. I would say copying it doesn't need to be related with the fulfilment of the policies, and in terms of UX, could be confusing for the user because "Copy" buttons attached to a text field are usually enabled when there's some non-empty text in it. What do you think @jesmrec?

owncloudApp/src/main/res/values/strings.xml
JuancaG05: No sense to order strings alphabetically, they should be grouped by meaning. In any case, `strings.xml` is a chaos right now, this PR is obviously not the scope to work on it

owncloudApp/src/test/java/com/owncloud/android/presentation/sharing/PublicLinkPasswordUtilsTest.kt
JuancaG05: Renamed

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
JuancaG05: Not that important (KtLint won't complain about it), but ok

owncloudApp/src/test/java/com/owncloud/android/presentation/sharing/PublicLinkPasswordUtilsTest.kt
JuancaG05: We're doing a isolation unit test, how are we going to know which values are coming from the server? That doesn't make sense

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/shares/PublicShareDialogFragment.kt
jesmrec: you are right that usually is enough with input characters, but this is because the most usual condition is "not empty" to avoid corner cases with non-valid strings. In this case, the condition is not only "not empty" but also "fulfilling conditions". As @Aitorbp stated, i'd only enable it when password is really useful, that means, when password passes policies. What's the aim of copying a non-valid password? we could lead users to confusion. that's what i mean.

JuancaG05: Ok, done 👍

Aitorbp (approved):
LGTM 💯

jesmrec:
### (1)

We could have a problem if the translation of `Generate password` is too long in some languages. I guess, there will not be enough room for the `copy` option or it will be misplaced somehow.

Since the focus is in the `Password` field, should it be enough with the label `Generate`? (iOS approach) wdyt? @JuancaG05 @Aitorbp

jesmrec:
### (2)

oC10 servers are showing the options `generate password` and `copy` even though they don't support the feature. It's not a bad idea to let users generate secure passwords, also if they are not mandatory or must not fulfill any policy. 

In that case, options should be hidden and only displayed if the user enables the password switch.

Pixel 2, Android 11
`9acf1fe75`

JuancaG05:
> Since the focus is in the `Password` field, should it be enough with the label `Generate`? (iOS approach) wdyt? @JuancaG05 @Aitorbp

Ok! Yes, I think it's enough. Will change it 👍

JuancaG05:
About (2), in fact, the behaviour of "Generate password" button would also differ, since the password is generated in function of the policies coming in the capabilities, and not sure if that would be working if the `passwordPolicy` (not only the different policies) itself is null. I would say it's better to hide it for oC10 servers to avoid complexity about this, what do you think? @jesmrec

jesmrec:
> About (2), in fact, the behaviour of "Generate password" button would also differ, since the password is generated in function of the policies coming in the capabilities, and not sure if that would be working if the passwordPolicy (not only the different policies) itself is null. I would say it's better to hide it for oC10 servers to avoid complexity about this, what do you think? @jesmrec

IMO, "password policies" and "generate passwords" should be (theoretically) isolated features. Not sure if we can set any "default" policies in case there is no server ones, but, this is out of the scope here. For me in this PR are valid any of these two:

- Remove options from oC10 servers, so that, the feature is only available in oCIS
- Let users generate passwords with any default policy (good to have but not mandatory here)

Pick the easiest one for you, then, we will decide whether there is place for improvement in another moment.

JuancaG05:
> IMO, "password policies" and "generate passwords" should be (theoretically) isolated features. Not sure if we can set any "default" policies in case there is no server ones, but, this is out of the scope here. For me in this PR are valid any of these two:
> 
> * Remove options from oC10 servers, so that, the feature is only available in oCIS
> * Let users generate passwords with any default policy (good to have but not mandatory here)
> 
> Pick the easiest one for you, then, we will decide whether there is place for improvement in another moment.

We can set default policies, and indeed they are already set, but as I commented, it depends on the structure of the capabilities, which in terms of password policy is not the same in oC10 and oCIS. To avoid more technical complexity here, I'll just remove the options from oC10 servers and we can think about this in a future 👍

jesmrec:
(1) and (2) fixed

jesmrec:
### (3)

I noticed a tiny detail: when clicking on `Generate`, the default is that the password is initially visible. But, if user types it, it is initially hidden. I think that both should behave in the same way, i'd say with a default hidden password.

Pixel 2, Android 11
`d17c425b`

tbsbdr:
I'd follow the following:

> Whenever users interact with a system, they need to know whether the interaction was successful.


[Source: Visibility of System Status (Usability Heuristic #1)](https://www.nngroup.com/articles/visibility-system-status/)

- **User typed password**: hide password, because user knows what he/she types and can see via growing ****** field that the input was very likely succesful
- **Generate password**: show the password, because the user does not know what had been typed, so it should be shown 

I hope that makes sense. thanks for paying attention to these details!

jesmrec:
So, the brush-broom icon will be the first approach. We'll work in future improvements

jesmrec:
Then, current approach will stay as is

jesmrec:
This is approved on my side

Issue #4197, opened 2023-10-20
Improve av. offline performance


jesmrec:
Checking issue https://github.com/owncloud/android/issues/4192, i notice av. offline behaves in a non-performant way:

1. Set a folder with files inside as av. offline -> everything marked as av. offline
2. Kill the app
3. Open the app

**Current**: one `PROPFIND` request sent for every item inside the folder recursively. Even if no changes happened in the tree structure. In case a deep structure or populated folders, many requests are being wasted making the app non-performant.

**Expected**: Checking folder etag is enough to know whether the av. offline folder has changed or not. Only if etag changed, sync operation is triggered.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/improve_available_offline_performance
- [x] Development tasks
  - [x] Etag condition in refreshFolder()
  - [x] Release notes
  - [x] Calens
- [ ] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

Aitorbp:
For this issue I have decided to make a change in the `refreshFolder` method inside the `FileRepository`. Currently within the refresh method all files are being returned, both those that have been modified from the server and those that have not. 
Theoretically, we are only interested in refreshing those files that have been modified. Therefore, the `etag` we have on the server will be compared with the etag of the local database to extract only those files that have been modified, thus avoiding unnecessary requests.
Also, this change implies a change in the behaviour of the method, which is used by a number of classes in the application. At the generic level, it is used in the `SynchronizeFolderUseCase` and in the `getFileFromRemoteId` (which is only used in the `ManageDeepLinkUseCase`).
To facilitate the QA work @jesmrec, here are the uses of `SynchronizeFolderUseCase` in the application:

- When we try to choose the destination path where to make an upload
- When we copy a file, we choose the space, it makes a sync of the content.
- When we do a move operation
- When we open the application
- When we click on set available offline on a folder/file in the operations menu
- When we click on unset available offline on a folder/file in the operations menu
- When switching accounts 
- When clicking on synchronize in the operations list
- When entering a folder
- When exiting the folder, by clicking on back
- when cancelling a copy/move file operation
- When triggering the AccountDiscoveryWorker worker
- When the AvailableOfflinePeriodicWorker is activated, the files that are available offline are synchronised.
- When we receive an external file and we choose the destination place.

Places where it is not clear to me what it does:
[...]

PR #4354, opened 2024-04-04
Improve av. offline performance 

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4197

_____

## QA

Checks:

Reports:

- [x] (1) Av. offline does not sync https://github.com/owncloud/android/pull/4354#issuecomment-2045170517 [FIXED]
- [x] (2) Adding external files to av. offline does not trigger a sync https://github.com/owncloud/android/pull/4354#issuecomment-2046659982 [FIXED]
- [x] (3) Local changes not pushed https://github.com/owncloud/android/pull/4354#issuecomment-2047280691 [FIXED]

### DIFF
--- a/changelog/unreleased/4354
+++ b/changelog/unreleased/4354
@@ -1,0 +1,7 @@
+Enhancement: Avoid unnecessary requests when an av. offline folder is refreshed
+
+The available offline folders will only be refreshed when their eTag from the server and the corresponding one of the local database are different,
+avoiding sending unnecessary request.
+
+https://github.com/owncloud/android/issues/4197
+https://github.com/owncloud/android/pull/4354
--- a/owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
@@ -424,23 +424,21 @@ class OCFileRepository(
                                 if (remoteFolder.isAvailableOffline) AVAILABLE_OFFLINE_PARENT else NOT_AVAILABLE_OFFLINE
 
                         })
-                } else {
-                    if (localChildToSync.etag != remoteChild.etag) {
-                        // File exists in the database, we need to check several stuff.
-                        folderContentUpdated.add(
-                            remoteChild.apply {
-                                copyLocalPropertiesFrom(localChildToSync)
-                                // DO NOT update etag till contents are synced.
-                                etag = localChildToSync.etag
-                                needsToUpdateThumbnail =
-                                    (!remoteChild.isFolder && remoteChild.modificationTimestamp != localChildToSync.modificationTimestamp) || localChildToSync.needsToUpdateThumbnail
-                                // Probably not needed, if the child was already in the database, the av offline status should be also there
-                                if (remoteFolder.isAvailableOffline) {
-                                    availableOfflineStatus = AVAILABLE_OFFLINE_PARENT
-                                }
-                                // FIXME: What about renames? Need to fix storage path
-                            })
-                    }
+                } else if (localChildToSync.etag != remoteChild.etag) {
+                    // File exists in the database, we need to check several stuff.
+                    folderContentUpdated.add(
+                        remoteChild.apply {
+                            copyLocalPropertiesFrom(localChildToSync)
+                            // DO NOT update etag till contents are synced.
+                            etag = localChildToSync.etag
+                            needsToUpdateThumbnail =
+                                (!remoteChild.isFolder && remoteChild.modificationTimestamp != localChildToSync.modificationTimestamp) || localChildToSync.needsToUpdateThumbnail
+                            // Probably not needed, if the child was already in the database, the av offline status should be also there
+                            if (remoteFolder.isAvailableOffline) {
+                                availableOfflineStatus = AVAILABLE_OFFLINE_PARENT
+                            }
+                            // FIXME: What about renames? Need to fix storage path
+                        })
                 }
             }
 
--- a/owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
+++ b/owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
@@ -462,7 +462,7 @@ class OCFileRepository(
         if (!anyConflictInThisFolder) {
             remoteFolder.etagInConflict = null
         }
-        localFileDataSource.saveFilesInFolder(
+        localFileDataSource.saveFilesInFolderAndReturnThem(
             folder = remoteFolder,
             listOfFiles = folderContentUpdated
         )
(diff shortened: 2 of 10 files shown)


owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -458,11 +460,11 @@ class OCFileRepository(
         if (!anyConflictInThisFolder) {
             remoteFolder.etagInConflict = null
         }
-
-        return localFileDataSource.saveFilesInFolderAndReturnThem(
+        localFileDataSource.saveFilesInFolder(
```
JuancaG05, line 463: There's no need to use another method and repeat code in every layer underneath (datasource, DAO...). Just use the same method but don't use what it returns

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -423,20 +423,22 @@ class OCFileRepository(
...
-                            etag = localChildToSync.etag
-                            needsToUpdateThumbnail =
-                                (!remoteChild.isFolder && remoteChild.modificationTimestamp != localChildToSync.modificationTimestamp) || localChildToSync.needsToUpdateThumbnail
-                            // Probably not needed, if the child was already in the database, the av offline status should be also there
-                            if (remoteFolder.isAvailableOffline) {
-                                availableOfflineStatus = AVAILABLE_OFFLINE_PARENT
-                            }
-                            // FIXME: What about renames? Need to fix storage path
-                        })
+                    if (localChildToSync.etag != remoteChild.etag) {
```
JuancaG05, line 426: This is an `else if`, KtLint will probably complain about this

JuancaG05 (changes requested):
Some changes here to avoid code repetition @Aitorbp

changelog/unreleased/4354
```diff
@@ -0,0 +1,7 @@
+Enhancement: Avoid unnecessary requests when an av. offline folder is refreshed
+
+The available offline folders will only be refreshed when the eTag parameter of the server and that of the local database are different,
+avoiding to send unnecessary request to the server.
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
Aitorbp: Done

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
Aitorbp: Changed, although Ktlint wasn´t complaining

JuancaG05 (approved):
LGTM now

jesmrec:
### (1)

Check the following steps out. We are losing the sync condition in the basic av. offline case.

1. Create a folder and add some files inside (better images)
2. Set the folder as av. offline (without browsing inside)
3. After some seconds, remove device connection
4. Browse into the folder and click on any file

**Current**: not posible to preview

**Expected**: file displayed. It is available offline, so the preview must be always available. When the folder was set as av. offline in step 2. all content inside had to be synced. It doesn't

Pixel 2 Android 11
`d86367b9c`

Aitorbp:
> ### (1)
> Check the following steps out. We are losing the sync condition in the basic av. offline case.
> 
> 1. Create a folder and add some files inside (better images)
> 2. Set the folder as av. offline (without browsing inside)
> 3. Wait till everything download
> 4. Unset the folder as available offline
> 5. Remove locally the content of the folder
> 6. Repeat step 2.
> 7. After some seconds, remove device connection
> 8. Browse into the folder and click on any file
> 
> **Current**: not posible to preview
> 
> **Expected**: file displayed. It is available offline, so the preview must be always available. When the folder was set as av. offline in step 2. all content inside had to be synced. It doesn't
> 
> Pixel 2 Android 11 `d86367b9c`

With the change of refresh functionality, from now on only files that have been modified will be returned. This means that the local etag and the server etag are different. With this known, we are presented with the case discovered by @jesmrec.

This means if we delete the local copy and we want to make a set available offline, it will compare between two equal etags, as the file on the server has not changed and the local etag has remained as it was before deleting it. 

So the condition has not detected that there was a change if we make the files available offline after deleting the content of the folder locally. This means that the download is not done even if it is available online. If we take the connection offline and enter a file it will not be available.

So to solve this we must set the `etag` parameter to null if we want to delete files locally. This way, they will be downloaded again.

jesmrec:
### (2)

may be related with (1)

1. Set a folder as av. offline
2. Using other client, web f. ex., add more files to the folder
3. In app, pull to refresh (not browsing inside)
4. Try to open the file added in 2.

Current: file not opened

Expected: file should be opened. Pull to refresh action should trigger a sync in the av. offline folder because the etag changed after new item was added inside.

Pixel 2 Android 11
`b1cad6fb`

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -460,11 +460,11 @@ class OCFileRepository(
         if (!anyConflictInThisFolder) {
             remoteFolder.etagInConflict = null
         }
-
-        return localFileDataSource.saveFilesInFolderAndReturnThem(
+        localFileDataSource.saveFilesInFolderAndReturnThem(
```
JuancaG05, line 463: Just 1 thing before merging this: the `saveFilesInFolderAndReturnThem` result is never used after changing this, we could transform it to `saveFilesInFolder`. Adapt the existing tests as well

JuancaG05 (changes requested):

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
Aitorbp: I have realized that we need the return of that function. Because if not in the current code if we upload a file to a folder available offline, the file id will be null and it will not download the file. So I am going to change this function `insertFilesInFolderAndReturnThem` to return only the changed files.

jesmrec:
### (3)

1. Set a folder as av. offline
2. Modify any file inside the folder using the `open with` option (i used a txt file with the quickedit free editor) 
3. Save modifications in external editor
4. Go again to the app

Current: changes are not synced back to the server. Even pulling to refresh, changes not pushed

Expected: just entering the to the containing folder, changes are synced back to the server (checked current `master`)

Pixel 2 Android 11
`d75ebeae0`

JuancaG05 (approved):
LGTM

jesmrec:
Go ahead with this! approved.

Issue #4213, opened 2023-11-14
[BUG] Request review in fresh install


jesmrec:
Installing the app from scratch, and adding an oC10 account, i realised that many requests are repeated somehow. This is the request flow after clicking the `Login` button:

![Screenshot 2023-11-14 at 18 28 00](https://github.com/owncloud/android/assets/14894746/ed967988-fd74-443a-8528-9ebd9dcb0747)

I see there:

- 3 calls to `capabilities`
- 7 calls to avatar 😨 
- 3 calls to app/list
- 4 `PROPFIND` to the root (2 required: one for the list, other for the quota)

etc...

in oCIS account is also visible, not so many request but also repeated:

![Screenshot 2023-11-14 at 18 41 32](https://github.com/owncloud/android/assets/14894746/d2914c15-9ce1-4aab-b0b0-d85ed995064d)

this might be reviewed to get throttled in needed requests and no repeated stuff.

PR #4385, opened 2024-04-16
[BUG] Request review in fresh install

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4213

_____

### DIFF
--- a/changelog/unreleased/4385
+++ b/changelog/unreleased/4385
@@ -1,0 +1,6 @@
+Bugfix: Removed unnecessary requests when the app is installed from scratch
+
+Some requests to the server that were not necessary when installing the app from scratch have been removed.
+
+https://github.com/owncloud/android/issues/4213
+https://github.com/owncloud/android/pull/4385
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -84,7 +84,6 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCo
 import com.owncloud.android.lib.resources.status.OwnCloudVersion
 import com.owncloud.android.operations.SyncProfileOperation
 import com.owncloud.android.presentation.authentication.AccountUtils.getCurrentOwnCloudAccount
-import com.owncloud.android.presentation.capabilities.CapabilityViewModel
 import com.owncloud.android.presentation.common.UIResult
 import com.owncloud.android.presentation.conflicts.ConflictsResolveActivity
 import com.owncloud.android.presentation.files.details.FileDetailsFragment
@@ -119,7 +118,6 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import org.koin.android.ext.android.inject
 import org.koin.androidx.viewmodel.ext.android.viewModel
-import org.koin.core.parameter.parametersOf
 import timber.log.Timber
 import java.io.File
 import kotlin.coroutines.CoroutineContext
@@ -235,6 +233,7 @@ class FileDisplayActivity : FileActivity(),
         setupRootToolbar(
             isSearchEnabled = true,
             title = getString(R.string.default_display_name_for_root_folder),
+            isAvatarRequested = true,
         )
 
         // setup drawer
@@ -297,15 +296,6 @@ class FileDisplayActivity : FileActivity(),
         super.onPostCreate(savedInstanceState)
 
         if (savedInstanceState == null && mAccountWasSet) {
-            val capabilitiesViewModel: CapabilityViewModel by viewModel {
-                parametersOf(
-                    account?.name
-                )
-            }
-            capabilitiesViewModel.refreshCapabilitiesFromNetwork()
-            capabilitiesViewModel.capabilities.observe(this, Event.EventObserver {
-                onCapabilitiesOperationFinish(it)
-            })
             navigateTo(fileListOption, initialState = true)
         }
 
@@ -940,7 +930,7 @@ class FileDisplayActivity : FileActivity(),
                     FileListOption.ALL_FILES -> getString(R.string.default_display_name_for_root_folder)
                     FileListOption.SPACES_LIST -> getString(R.string.bottom_nav_spaces)
                 }
-            setupRootToolbar(title, isSearchEnabled = fileListOption != FileListOption.SPACES_LIST)
+            setupRootToolbar(title, isSearchEnabled = fileListOption != FileListOption.SPACES_LIST, false)
             mainFileListFragment?.setSearchListener(findViewById(R.id.root_toolbar_search_view))
         } else if (space?.isProject == true && chosenFile.remotePath == OCFile.ROOT_PATH) {
             updateStandardToolbar(title = space.name, displayHomeAsUpEnabled = true, homeButtonEnabled = true)
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -84,6 +84,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCo
 import com.owncloud.android.lib.resources.status.OwnCloudVersion
 import com.owncloud.android.operations.SyncProfileOperation
 import com.owncloud.android.presentation.authentication.AccountUtils.getCurrentOwnCloudAccount
+import com.owncloud.android.presentation.capabilities.CapabilityViewModel
 import com.owncloud.android.presentation.common.UIResult
 import com.owncloud.android.presentation.conflicts.ConflictsResolveActivity
 import com.owncloud.android.presentation.files.details.FileDetailsFragment
@@ -118,6 +119,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import org.koin.android.ext.android.inject
 import org.koin.androidx.viewmodel.ext.android.viewModel
+import org.koin.core.parameter.parametersOf
 import timber.log.Timber
 import java.io.File
 import kotlin.coroutines.CoroutineContext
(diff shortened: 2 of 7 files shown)


Aitorbp:
Below is the detail of the research that has been done to eliminate some unnecessary requests when installing the app from scratch:

### Capabilities interactions when installing the application from scratch:

* `AuthenticationViewModel`: call to `refreshCapabilitiesFromServerAsyncUseCase`
* `AccountDiscoveryWorker`: call to `refreshCapabilitiesFromServerAsyncUseCase` -> Removed 
* `DrawerActivity`: Capabilities observer
* `CapabilityViewModel` call to r`efreshCapabilitiesFromNetwork()`
* `FileDisplayActivity`: call to `capabilitiesViewModel.refreshCapabilitiesFromNetwork()` and Observer -> Removed 

The capabilities are first called in the `AuthenticationViewModel` and then called again in the `AccountDiscoveryWorker`. This is redundant.

When you enter the `setupNavigationBottomBar` function in the `DrawerActivity` there is a capabilities observer that makes a capabilities refresh (`CapabilitiesViewModel`). Then this request goes through the `FileDisplaActivity` in the `onPostCreate` and makes a call to the `refreshCapabilitiesFromNetwork`, I think this last call is redundant and you could leave only the observer of the `DrawerActivity`, because it makes a double refresh.

The result of the capabilities would look like this: When the application is started from scratch, the capabilities will be called twice: in the `authenticatorViewModel` and through the observer of the `DrawerActivity`. If we open the app again it will only be called from the `DrawerActivity` (this case is not tested because it doesn't give the App inspector time to see the requests that are sent). But searching the whole app there are only two observers to the capabillities: `DrawerActivity` and `FileDisplayActivity`.

### Avatar interactions when installing the application from scratch:

The `setupRootToolbar` function contained in `ToolbarActivity` has a request to the `loadAvatarForAccount()`. This function is called from 3 places in the app:
* `FileDisplayActivity`: In onCreated
* `FileDisplayActivity`: In the `updateToolbar` function.  It is used in 16 places in this class, but when the app is started from scratch in oc10 it is called in:
    * `cleanSecondFragment()`
    * `navigateTo(newFileListOption: FileListOption, initialState: Boolean = false)`
    * `onCurrentFolderUpdated(newCurrentFolder: OCFile, currentSpace: OCSpace?)`

* `UploadListActivity`: Not run on when installing from scratch

In `AccountManagerAdapter` there is a call to the function `loadAvatarForAccount()` but it doesn't affect when you install the application from scratch. Not applicable.

The result of the avatar request will be as follows: One will be executed from `FileDisplayActivity` in the onCreated and one in the `SyncProfileOperation` in its `syncUserProfile()` function. We remove from the `updateToolbar` in the `FileDisplayActivity` that the avatar is called. Remember that this function is called 16 times in the class, but only in three of the calls it enters here when we install the app from scratch. 
In short, with these changes we reduce the avatar requests to two when the app is installed from scratch.

Sometimes when you run the application from scratch the request that is executed in `SyncProfileOperation` is made twice.... So we will have three calls from scratch.

### App/list interactions when installing the application from scratch:

With all the above changes on an oc10 server, two calls to app/list are executed:
- `AuthenticationViewModel`
- From the `ViewModelExt` `runUseCaseWithResultAndUseCachedData`: used in the `ShareViewModel` and in the `CapabilityViewModel` (called through this viewmodel).
By removing the capabilities call in the `AccountDiscoveryWorker` the call to the app/list is no longer made at this point.

### Interactions of files/admin when installing the application from scratch:

Runs in four places in the application when installing from scratch:
- `AccountDiscoveryWorker`
- `SyncProfileOperation`
- `MainFileListViewModel`
- Observer of a `SynchronizeFolderUseCase`
In this call I wouldn't touch anything because it can be critical to remove one of these calls. The first two are necessary and the next two are used on more occasions throughout the app's lifecycle.

In short, the following image shows the requests that are made with the changes discussed above in an oc10 server:
oc10 after install from scratch
The requests that are made with the changes discussed above in an ocis server:
ocis after install from scratch

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -940,7 +930,7 @@ class FileDisplayActivity : FileActivity(),
                     FileListOption.ALL_FILES -> getString(R.string.default_display_name_for_root_folder)
                     FileListOption.SPACES_LIST -> getString(R.string.bottom_nav_spaces)
                 }
-            setupRootToolbar(title, isSearchEnabled = fileListOption != FileListOption.SPACES_LIST)
+            setupRootToolbar(title, isSearchEnabled = fileListOption != FileListOption.SPACES_LIST, false)
```
JuancaG05, line 933: Since we have 2 different boolean parameters now (and who knows if more in the future), let's add the parameter name to avoid confusions

owncloudApp/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -74,7 +75,7 @@ protected void onCreate(Bundle savedInstanceState) {
         setFile(null);
 
         // setup toolbar
-        setupRootToolbar(getString(R.string.uploads_view_title), false);
+        setupRootToolbar(getString(R.string.uploads_view_title), false, true);
```
JuancaG05, line 78: Do we really need to fetch the avatar in this activity as well? Don't we have it in cache or something like that? Just asking

JuancaG05 (changes requested):

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -744,6 +744,8 @@
     <string name="release_notes_4_3_0_subtitle_clear_data_button_hard_reset">The "Clear data" button has been enabled to delete the application data from the app settings in the device settings. The application will perform a hard reset if this button is pressed</string>
     <string name="release_notes_4_3_0_title_improvements_remove_dialog">Improvements in remove dialog</string>
     <string name="release_notes_4_3_0_subtitle_improvements_remove_dialog">Added a custom dialog when the file that is going to be deleted has thumbnail and added the number of files that are going to be deleted in a multiple selection</string>
+    <string name="release_notes_4_3_0_title_remove_unnecessary_requests">Improved installation of the application from scratch</string>
+    <string name="release_notes_4_3_0_subtitle_remove_unnecessary_requests">By installing the application from scratch, the communication between the server and the app is more optimal</string>
```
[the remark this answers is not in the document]
Aitorbp: Removed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
Aitorbp: Added

changelog/unreleased/4385
```diff
@@ -0,0 +1,6 @@
+Bugfix: Removed unnecessary requests when the application is installed from scratch
+
+By installing the application from scratch, some additional requests to the server that are not necessary are eliminated.
```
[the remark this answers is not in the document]
Aitorbp: done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
Aitorbp: If we navigate to Uploads view the avatar is not going to be requested. We get the capabilities and there is a condition `shouldFetchAvatar` to fetch the avatar from the server.

JuancaG05: In any case, this is not needed here

JuancaG05 (approved):

jesmrec:
First launch oCIS

- capabilities: 2 calls
- avatar: no calls (does this endpoint exist in oCIS?)
- app/list: 2 calls
- PROPFIND to root: 1 call (recursive to discover)

First launch oC10

- capabilities: 2 calls
- avatar: 2 calls
- app/list: 2 calls
- PROPFIND to root: 1 call (recursive to discover)


Before:

Opening uploads tab, a request to app/list and capabilities sent
Opening shares tab, request to to shares
Opening spaces tab, request to drives
Opening drawer, no requests
Opening settings, no requests
Opening manage accounts, no requests


|| Capabilities | Avatar | /app/list | 
|:-|:-:|:-:|:-:|
| **Browsing** | No sent | No sent | No sent |
| **Reopening** | Sent | Sent only in oC10 | Sent |
| **Share** | Sent | No sent | **Sent** |
| **Switch to uploads tab** | **Sent** | No sent | **Sent** |
| **Switch from uploads tab** | Sent | **Sent only in oC10** | Sent |

Aitorbp:
> Let's check the new status:
> 
> ### 1. First launch
> First launch oCIS
> 
> * capabilities: 2 calls
> * avatar: no calls (does this endpoint exist in oCIS?)
> * app/list: 2 calls
> * PROPFIND to root: 1 call (recursive to discover)
> 
> First launch oC10
> 
> * capabilities: 2 calls
> * avatar: 2 calls
> * app/list: 2 calls
> * PROPFIND to root: 1 call (recursive to discover)
> 
> the new status after fresh install is much better and approved on my side
> 
> ### 2. Regressions
> Before (4.2.1):
> 
> Opening uploads tab, a request to app/list and capabilities sent Opening shares tab, request to to shares Opening spaces tab, request to drives Opening drawer, no requests Opening settings, no requests Opening manage accounts, no requests
> 
> With such info, i will take the three relevant endpoints that were handled (capabilities, avatar and app/list) against relevant scenarios to send requests to them. This is the result:
> 
> Capabilities	Avatar	/app/list
> **Browsing**	No sent	No sent	No sent
> **Reopening**	Sent	Sent only in oC10	Sent
> **Share**	Sent	No sent	**Sent**
> **Switch to uploads tab**	**Sent**	No sent	**Sent**
> **Switch from uploads tab**	Sent	**Sent only in oC10**	Sent
> Regards (no regressions):
> 
> * When opening the `Share` option over any item, a call to `app/list` is triggered. I guess it could be something related with capabilities, please confirm. Sharing and app/list have nothing to do, so, it could be avoided.
> * When switching from any tab to `Uploads` , a call to `capabilities` and `app/list` is triggered. Same as before.
> * When switching from uploads tab to any other, a call to `avatar` is triggered. I guess it is because loading `FileListActivity`, right?
> 
> In general, the status of this PR is stable. Just, the mentioned improvements are something we can live with but may deserve a quick look, in order to know if it is OK addressing to another issue or let them be... @Aitorbp


In relation to what you say @jesmrec: The `app/list` is always called in the `refreshCapabilitiesAccount()` when the `filesAppProviders` is `not null`. Therefore, these requests that you indicate are being made. You should see if it is necessary to make that call even if `filesAppProviders` is different from null. I would leave it for another issue.

Finally, in reference to the `avatar`, when we are on the uploads screen, two calls are made to the avatar:

- `SyncProfileOperation`: We cannot remove it because it is also executed when the application is opened from scratch. We would have to make a flag variable, which is not the most correct thing to do.

- `ToolbarActivity`: We cannot eliminate this call, the avatar in the toolbar would disappear.

jesmrec:
Thanks for the input. Behaviour was improved in comparison with the initial status and that was the target of the issue. 

It can be merged into `master`

Issue #4233, opened 2023-11-27
[TECHNICAL] Unit tests for repository classes - Part 2


JuancaG05:
Epic:
- https://github.com/owncloud/android/issues/4063

In this issue we'll address the unit tests for the class:
- OCFileRepository

Remember we want to create unit tests that **check that every method returns the expected result** and **verify that every method calls the corresponding methods of other classes**.

### TASKS

- [x] Research (if needed)
- [x] Create branch technical/testing_repositories_2
- [x] Development tasks
  - [x] Create unit tests for `OCFileRepository`
- [ ] Code review and apply changes requested
- [ ] Merge branch technical/testing_repositories_2 into master

PR #4389, opened 2024-04-18
[TECHNICAL] Unit tests for repository classes - Part 2

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4233

_____

### DIFF
--- a/owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
+++ b/owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
@@ -26,7 +26,9 @@ import com.owncloud.android.data.files.datasources.LocalFileDataSource
 import com.owncloud.android.data.files.datasources.RemoteFileDataSource
 import com.owncloud.android.data.providers.LocalStorageProvider
 import com.owncloud.android.data.spaces.datasources.LocalSpacesDataSource
+import com.owncloud.android.domain.availableoffline.model.AvailableOfflineStatus
 import com.owncloud.android.domain.exceptions.ConflictException
+import com.owncloud.android.domain.exceptions.FileAlreadyExistsException
 import com.owncloud.android.domain.exceptions.FileNotFoundException
 import com.owncloud.android.domain.files.model.FileListOption
 import com.owncloud.android.domain.files.model.MIME_DIR
@@ -59,6 +61,7 @@ import org.junit.Assert.assertNull
 import org.junit.Assert.assertThrows
 import org.junit.Before
 import org.junit.Test
+import java.util.UUID
 
 @ExperimentalCoroutinesApi
 class OCFileRepositoryTest {
@@ -79,32 +82,17 @@ class OCFileRepositoryTest {
     private val storagePath = "/local/storage/path/username@demo.owncloud.com/Folder/Photos/image2.jpt"
     private val remoteId = "remoteId"
     private val searchText = "image"
-
-    /*private val listOfFilesRetrieved = listOf(
-        OC_FOLDER,
-        OC_FOLDER.copy(remoteId = "one"),
-        OC_FOLDER.copy(remoteId = "two")
-    )
-    private val listOfFileToRemove = listOf(
-        OC_FOLDER.copy(id = 1),
-        OC_FILE.copy(id = 2),
-        OC_FILE.copy(id = 3)
-    )
-
-    private val timeInMilliseconds = 3600000L*/
+    private val etagInConflict = "5efb0c13c688i"
+    private val fileWithConflict = OC_FILE_WITH_SPACE_ID.copy(etagInConflict = etagInConflict)
 
     @Before
     fun setUp() {
+        val commonSpaceId = OC_FILE_WITH_SPACE_ID.spaceId
+        val commonAccountName = OC_FILE_WITH_SPACE_ID.owner
         every {
             localSpacesDataSource.getWebDavUrlForSpace(
-                spaceId = OC_FILE_WITH_SPACE_ID.spaceId,
-                accountName = OC_FILE_WITH_SPACE_ID.owner
-            )
-        } returns OC_SPACE_PERSONAL.root.webDavUrl
-        every {
-            localSpacesDataSource.getWebDavUrlForSpace(
-                spaceId = OC_FOLDER_WITH_SPACE_ID.spaceId,
-                accountName = OC_FOLDER_WITH_SPACE_ID.owner
+                spaceId = commonSpaceId,
+                accountName = commonAccountName
             )
         } returns OC_SPACE_PERSONAL.root.webDavUrl
     }
@@ -1113,9 +1101,6 @@ class OCFileRepositoryTest {
 
     @Test
     fun `moveFile returns an empty list with no OCFiles in conflict when replace parameter is empty, expected path doesn't exist and file has a conflict`() {
-        val fileWithConflict = OC_FILE_WITH_SPACE_ID.copy(
-            etagInConflict = "5efb0c13c688i"
-        )
         every {
             remoteFileDataSource.checkPathExistence(
                 path = expectedRemotePath,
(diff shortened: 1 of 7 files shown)


owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCFile.kt
```diff
@@ -77,6 +125,27 @@ val OC_FILE_AVAILABLE_OFFLINE = OCFile(
     availableOfflineStatus = AvailableOfflineStatus.AVAILABLE_OFFLINE
 )
 
+val OC_FILE_DOWNLOADED = OC_FILE.copy(
+    storagePath = "/local/storage/path/username@demo.owncloud.com/Photos/image.jpt"
```
joragua, line 129: Its `jpg` extension instead of `jpt`, right?

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCFile.kt
```diff
@@ -77,6 +125,27 @@ val OC_FILE_AVAILABLE_OFFLINE = OCFile(
...
+)
+
+val OC_FILE_WITH_SPACE_ID = OCFile(
+    id = 126,
+    parentId = 123,
+    remotePath = "/Folder/image2.jpt",
```
joragua, line 135: The same as before

joragua (changes requested):
Only one comment here! 😄 @JuancaG05

owncloudTestUtil/src/main/java/com/owncloud/android/testutil/OCFile.kt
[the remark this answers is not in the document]
JuancaG05: If you check `testutil/OCFile.kt` file, you can see that historically for the example objects it has been always like that, with `.jpt` extension. Just kept this for consistence

joragua (approved):
LGTM 👍🏻

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
```diff
@@ -25,382 +26,2010 @@ import com.owncloud.android.data.files.datasources.LocalFileDataSource
...
+                accountName = commonAccountName
+            )
+        } returns OC_SPACE_PERSONAL.root.webDavUrl
+    }
+
+    @Test
+    fun `createFolder creates a new folder and saves it`() {
+        every {
+            localSpacesDataSource.getWebDavUrlForSpace(
+                spaceId = OC_PARENT_FOLDER_WITH_SPACE_ID.spaceId,
```
jabarros, line 104: This is used on different places, what do you think to create a generic structure as 

private fun mockWebDavUrlForSpace(spaceId: String, owner: String, url: String) {
    every {
        localSpacesDataSource.getWebDavUrlForSpace(spaceId, owner)
    } returns url
}

and reuse it on different tests?

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
```diff
@@ -25,382 +26,2010 @@ import com.owncloud.android.data.files.datasources.LocalFileDataSource
...
+                folder = OC_PARENT_FOLDER_WITH_SPACE_ID,
+                listOfFiles = withArg<List<OCFile>> {
+                    assertEquals(1, it.size)
+                    assertEquals(OC_FOLDER_WITH_SPACE_ID.remotePath, it[0].remotePath)
+                    assertEquals(OC_PARENT_FOLDER_WITH_SPACE_ID.owner, it[0].owner)
+                    assertEquals(0, it[0].length)
+                    assertEquals(MIME_DIR, it[0].mimeType)
+                    assertEquals(OC_PARENT_FOLDER_WITH_SPACE_ID.spaceId, it[0].spaceId)
+                    assertEquals("CK", it[0].permissions)
+                }
```
jabarros, line 135: Don't you include a check to verify if there are conflicts when creating a folder?

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
```diff
@@ -25,382 +26,2010 @@ import com.owncloud.android.data.files.datasources.LocalFileDataSource
...
+
+        ocFileRepository.saveDownloadWorkerUuid(OC_FILE_WITH_SPACE_ID.id!!, workerUuid)
+
+        verify(exactly = 1) {
+            localFileDataSource.saveDownloadWorkerUuid(
+                fileId = OC_FILE_WITH_SPACE_ID.id!!,
+                workerUuid = workerUuid
+            )
+        }
+    }
```
jabarros, line 2025: Maybe it is good to include test cases about network problems, for example (that I have not seen, or I am wrong)

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
```diff
@@ -25,382 +26,2010 @@ import com.owncloud.android.data.files.datasources.LocalFileDataSource
...
+            localStorageProvider.deleteLocalFolderIfItHasNoFilesInside(OC_PARENT_FOLDER_WITH_SPACE_ID)
+            localFileDataSource.deleteFile(OC_PARENT_FOLDER_WITH_SPACE_ID.id!!)
+        }
+    }
+
+    @Test
+    fun `renameFile renames a file correctly`() {
+        val newName = "image3.jpt"
+        val newRemotePath = "/Folder/image3.jpt"
+        val newStoragePath = "/local/storage/path/username@demo.owncloud.com/Folder/image3.jpt"
```
jabarros, line 1830: A global constant better?

jabarros (commented):
It is a huge and awesome work, @JuancaG05 !!

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
JuancaG05: As long as I can see, this is only used twice in this class in `every` blocks (one in `setUp()` method and another one here), the other uses are in `verify` blocks, which need to contain the method itself. But, it's true that since both uses in `every` blocks get the same parameters, this one can be removed and keep just the one in `setUp()`, so that we don't need to do a generic method 👍

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
JuancaG05: That check is in another class, so that will be tested in the corresponding tests. In this case, in the repository we just stick to the code we have, that is calling to remote datasource and local datasource afterwards, that's why it's done like this

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
JuancaG05: We were discussing about this, and we reached the conclusion that we're only going to do tests for non-valid cases when we explicitly catch an exception in the method being tested. That is the case with network errors: we can assume they are caught in upper or lower layers, but this method of this class has nothing to do with it (it will just send it up in case it receives an exception). Moreover, we don't consider it a good practice because we would have to mock an object (for example the datasource) to throw an exception, and the test will check just that, that a mocked object is throwing an exception, nothing to do with the real implementation of the method

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
JuancaG05: Didn't make it a global variable because we have 3 different cases (although they are very similar):

Case 1: `/local/storage/path/username@demo.owncloud.com/Folder/Photos/image2.jpt`
Case 2: `/local/storage/path/username@demo.owncloud.com/Folder/Photos/image2 (1).jpt`
Case 3: `/local/storage/path/username@demo.owncloud.com/Folder/image3.jpt`

Each of them used for a different purpose 😀

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
jabarros: 👍

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
jabarros: 👍

owncloudData/src/test/java/com/owncloud/android/data/files/repository/OCFileRepositoryTest.kt
jabarros: You are right, they are different. I thought they were the same :)

jabarros (approved):

PR #4393, opened 2024-04-24
[FEATURE REQUEST] Add search functionality to spaces list

joragua:
## Related Issues
App:

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -88,8 +88,9 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
 
     private fun subscribeToViewModels() {
         collectLatestLifecycleFlow(spacesListViewModel.spacesList) { uiState ->
-            val spacesToListFiltered = uiState.spaces.filter { it.name.toLowerCase().contains(uiState.searchFilter.toLowerCase()) }
             if (uiState.searchFilter != "") {
+                val spacesToListFiltered =
+                    uiState.spaces.filter { it.name.toLowerCase().contains(uiState.searchFilter.toLowerCase()) && !it.isPersonal }
                 val personalSpace = uiState.spaces.find { it.isPersonal }
                 val spacesToList = spacesToListFiltered.toMutableList().apply {
                     if (personalSpace != null) {
@@ -99,8 +100,8 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
                 showOrHideEmptyView(spacesToList)
                 spacesListAdapter.setData(spacesToList)
             } else {
-                showOrHideEmptyView(spacesToListFiltered)
-                spacesListAdapter.setData(spacesToListFiltered)
+                showOrHideEmptyView(uiState.spaces)
+                spacesListAdapter.setData(uiState.spaces)
             }
             binding.swipeRefreshSpacesList.isRefreshing = uiState.refreshing
             uiState.error?.let { showErrorInSnackbar(R.string.spaces_sync_failed, it) }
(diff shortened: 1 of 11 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -118,9 +129,25 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         super.onCreateOptionsMenu(menu, inflater)
+        (menu.findItem(R.id.action_search).actionView as SearchView).run {
+            maxWidth = Int.MAX_VALUE
+            queryHint = resources.getString(R.string.actionbar_search)
+            setOnQueryTextListener(this@SpacesListFragment)
+        }
```
JuancaG05, line 136: Here we're setting a `OnQueryTextListener` to the element with ID `action_search`, and then in `FileDisplayActivity`, we're setting another similar listener to the element with ID `root_toolbar_search_view`. Those aren't the same, and maybe we just need one of them. This deserves a check to avoid unnecessary code 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -118,9 +129,25 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
...
+    override fun onQueryTextSubmit(query: String?): Boolean = false
+
+    override fun onQueryTextChange(newText: String?): Boolean {
+        newText?.let { spacesListViewModel.updateSearchFilter(it) }
+        return true
+    }
+
+    fun setSearchListener(searchView: SearchView) {
+        searchView.setOnQueryTextListener(this)
+    }
```
JuancaG05, line 149: If we finally conclude that the call from `FileDisplayActivity` is not needed, this method is not necessary anymore 👍

JuancaG05 (changes requested):
Good job here @joragua! Some comments here 😸

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -87,16 +88,26 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
...
-            spacesListAdapter.setData(onlyEnabledSpaces)
+            if (uiState.searchFilter != "") {
+                val spacesToListFiltered =
+                    uiState.spaces.filter { it.name.toLowerCase().contains(uiState.searchFilter.toLowerCase()) && !it.isPersonal }
+                val personalSpace = uiState.spaces.find { it.isPersonal }
+                val spacesToList = spacesToListFiltered.toMutableList().apply {
+                    if (personalSpace != null) {
+                        add(0, personalSpace)
+                    }
+                }
```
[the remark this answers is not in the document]
joragua: Changed ✅

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Check previous comments ⬆️

JuancaG05 (approved):
LGTM!!

jesmrec:
The calens entry of the current PR should overwrite the entry https://github.com/owncloud/android/issues/4311. With the work done here, that other one is out of scope and its meaning is the opposite.

jesmrec:
QA checks:

- [X] List of spaces -> Personal never there.
- [X] Folder picker in Copy operation -> Personal always there.
- [X] Camera uploads picker -> Personal always there.
- [x] Share from 3rd party app picker -> Personal always there.

jesmrec:
I noticed that the spaces' thumbnails mixed after closing the searching option. Something related with recycler list?

Given the following list of spaces in server:

Screenshot 2024-04-29 at 14 46 09

Check the following videos out:

https://github.com/owncloud/android/assets/14894746/d6161a10-e8ed-4a63-8439-487bba945e7a

https://github.com/owncloud/android/assets/14894746/f3725e39-ed1d-4ff1-9d8c-c8c93f0e3fc1

first video: the space called `New space` show a repeated thumbnail, whereas it doesn't have any thumbnail
second video: the space called `New space` show a repeated thumbnail, whereas it doesn't have any thumbnail

it's like a glitch that can create some hassle

could you take a look?

joragua:
The problem here was that the image of `New space` couldn't load correctly because it was setting the image of the space that it was previously in the same position in `RecyclerView`. The solution here is reset the element with `dispose()`  before setting the default image. I've  fixed it and now it should work. ✅ 

Also I removed the calens about https://github.com/owncloud/android/issues/4311

jesmrec:
Glitch is fixed. This is ready to go 🚀

Issue #4353, opened 2024-04-02
[BUG] Av. offline files removed locally with the `Local only` option


jesmrec:
### Steps to reproduce

1.  Set as av. offline any file that is inside other folder (recursive)
2. Long press over the container folder
3. Select `Remove`
4. Select `Local Only`

### Actual behaviour

The local copy(s) of the av. offline file(s) are removed from internal storage. Recovered when browsing or when the av. offline worker runs

### Expected behaviour

Local copies of av. offline files should be never removed. Available offline means "downloaded + synced", so, we can not get rid of local copies because a lack of connection before a sync could lead users to a non-reachable status, breaking the av. offline condition. 

After developing a solution for https://github.com/owncloud/android/issues/3936, the option `Local Only` is only displayed in case of folders, if any file inside is downloaded. An extra condition should be added, so that, this is the new expected result:

`Local only` displayed:

- For files: When downloaded but not available offline (already done) ✅ 
- For folders: When contain downloaded files that are not av. offline

PR #4399, opened 2024-05-07
[BUG] Av. offline files removed locally with the `Local only` option

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4353

_____

## QA

Reports:

- [x] (1) Sync operation does not sync files in a subfolder that were locally removed beforehand https://github.com/owncloud/android/pull/4399#issuecomment-2164911723 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -525,7 +525,13 @@ class MainFileListFragment : Fragment(),
                                 FileMenuOption.SET_AV_OFFLINE -> {
                                     fileOperationsViewModel.performOperation(FileOperation.SetFilesAsAvailableOffline(listOf(file)))
                                     if (file.isFolder) {
-                                        fileOperationsViewModel.performOperation(FileOperation.SynchronizeFolderOperation(file, file.owner))
+                                        fileOperationsViewModel.performOperation(
+                                            FileOperation.SynchronizeFolderOperation(
+                                                folderToSync = file,
+                                                accountName = file.owner,
+                                                isActionSetFolderAvailableOffline = true,
+                                            )
+                                        )
                                     } else {
                                         fileOperationsViewModel.performOperation(FileOperation.SynchronizeFileOperation(file, file.owner))
                                     }
@@ -1129,7 +1135,13 @@ class MainFileListFragment : Fragment(),
                 R.id.action_set_available_offline -> {
                     fileOperationsViewModel.performOperation(FileOperation.SetFilesAsAvailableOffline(listOf(singleFile)))
                     if (singleFile.isFolder) {
-                        fileOperationsViewModel.performOperation(FileOperation.SynchronizeFolderOperation(singleFile, singleFile.owner))
+                        fileOperationsViewModel.performOperation(
+                            FileOperation.SynchronizeFolderOperation(
+                                folderToSync = singleFile,
+                                accountName = singleFile.owner,
+                                isActionSetFolderAvailableOffline = true,
+                            )
+                        )
                     } else {
                         fileOperationsViewModel.performOperation(FileOperation.SynchronizeFileOperation(singleFile, singleFile.owner))
                     }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -529,7 +529,7 @@ class MainFileListFragment : Fragment(),
                                             FileOperation.SynchronizeFolderOperation(
                                                 folderToSync = file,
                                                 accountName = file.owner,
-                                                isActionSetFolderAvailableOffline = true,
+                                                isActionSetFolderAvailableOfflineOrSynchronize = true,
                                             )
                                         )
                                     } else {
@@ -1139,7 +1139,7 @@ class MainFileListFragment : Fragment(),
                             FileOperation.SynchronizeFolderOperation(
                                 folderToSync = singleFile,
                                 accountName = singleFile.owner,
-                                isActionSetFolderAvailableOffline = true,
+                                isActionSetFolderAvailableOfflineOrSynchronize = true,
                             )
                         )
                     } else {
@@ -1391,7 +1391,13 @@ class MainFileListFragment : Fragment(),
     private fun syncFiles(files: List<OCFile>) {
         for (file in files) {
             if (file.isFolder) {
-                fileOperationsViewModel.performOperation(FileOperation.SynchronizeFolderOperation(folderToSync = file, accountName = file.owner))
+                fileOperationsViewModel.performOperation(
+                    FileOperation.SynchronizeFolderOperation(
+                        folderToSync = file,
+                        accountName = file.owner,
+                        isActionSetFolderAvailableOfflineOrSynchronize = true,
+                    )
+                )
             } else {
                 fileOperationsViewModel.performOperation(FileOperation.SynchronizeFileOperation(fileToSync = file, accountName = file.owner))
             }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -653,7 +653,7 @@ class MainFileListFragment : Fragment(),
             }
         }
 
-        collectLatestLifecycleFlow(fileOperationsViewModel.checkIfFileLocalSharedFlow) {
+        collectLatestLifecycleFlow(fileOperationsViewModel.checkIfFileIsLocalAndNotAvailableOfflineSharedFlow) {
             val fileActivity = (requireActivity() as FileActivity)
             when (it) {
                 is UIResult.Loading -> fileActivity.showLoadingDialog(R.string.common_loading)
@@ -996,8 +996,8 @@ class MainFileListFragment : Fragment(),
         }
     }
 
-    private fun onShowRemoveDialog(filesToRemove: List<OCFile>, isLocal: Boolean) {
-        val dialog = RemoveFilesDialogFragment.newInstance(ArrayList(filesToRemove), isLocal)
+    private fun onShowRemoveDialog(filesToRemove: List<OCFile>, isAvailableLocallyAndNotAvailableOffline: Boolean) {
+        val dialog = RemoveFilesDialogFragment.newInstance(ArrayList(filesToRemove), isAvailableLocallyAndNotAvailableOffline)
         dialog.show(requireActivity().supportFragmentManager, ConfirmationDialogFragment.FTAG_CONFIRMATION)
         fileListAdapter.clearSelection()
         updateActionModeAfterTogglingSelected()
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/removefile/RemoveFilesDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/removefile/RemoveFilesDialogFragment.kt
@@ -3,7 +3,9 @@
  *
  * @author David A. Velasco
  * @author Abel García de Prada
- * Copyright (C) 2021 ownCloud GmbH.
+ * @author Aitor Ballesteros Pavón
+ *
+ * Copyright (C) 2024 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -77,16 +79,12 @@ class RemoveFilesDialogFragment : ConfirmationDialogFragment(), ConfirmationDial
          * @return Dialog ready to show.
          */
         @JvmStatic
-        fun newInstance(files: ArrayList<OCFile>, isAvailableLocally: Boolean): RemoveFilesDialogFragment {
+        fun newInstance(files: ArrayList<OCFile>, isAvailableLocallyAndNotAvailableOffline: Boolean): RemoveFilesDialogFragment {
             val messageStringId: Int
             var containsFolder = false
-            var containsAvailableOffline = false
             for (file in files) {
                 if (file.isFolder) {
                     containsFolder = true
-                }
-                if (file.isAvailableOffline) {
-                    containsAvailableOffline = true
                 }
             }
 
@@ -106,7 +104,7 @@ class RemoveFilesDialogFragment : ConfirmationDialogFragment(), ConfirmationDial
                     R.string.confirmation_remove_files_alert
                 }
             }
-            val localRemoveButton = if (!containsAvailableOffline && isAvailableLocally) {
+            val localRemoveButton = if (isAvailableLocallyAndNotAvailableOffline) {
                 R.string.confirmation_remove_local
             } else {
                 -1
--- a/owncloudApp/src/main/java/com/owncloud/android/syncadapter/FileSyncAdapter.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/syncadapter/FileSyncAdapter.java
@@ -4,8 +4,9 @@
  * @author Bartek Przybylski
  * @author David A. Velasco
  * @author David González Verdugo
- * Copyright (C) 2011  Bartek Przybylski
- * Copyright (C) 2020 ownCloud GmbH.
+ * @author Aitor Ballesteros Pavón
+ *
+ * Copyright (C) 2024 ownCloud GmbH.
  * <p>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -231,7 +232,8 @@ public class FileSyncAdapter extends AbstractOwnCloudSyncAdapter {
                 folder.getRemotePath(),
                 folder.getOwner(),
                 folder.getSpaceId(),
-                SynchronizeFolderUseCase.SyncFolderMode.REFRESH_FOLDER_RECURSIVELY);
+                SynchronizeFolderUseCase.SyncFolderMode.REFRESH_FOLDER_RECURSIVELY,
+                false);
         UseCaseResult<Unit> useCaseResult;
 
         useCaseResult = synchronizeFolderUseCase.getValue().invoke(params);
--- a/owncloudApp/src/main/java/com/owncloud/android/syncadapter/FileSyncAdapter.java
+++ b/owncloudApp/src/main/java/com/owncloud/android/syncadapter/FileSyncAdapter.java
@@ -6,6 +6,7 @@
  * @author David González Verdugo
  * @author Aitor Ballesteros Pavón
  *
+ * Copyright (C) 2011  Bartek Przybylski
  * Copyright (C) 2024 ownCloud GmbH.
  * <p>
  * This program is free software: you can redistribute it and/or modify
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -763,6 +763,8 @@
     <string name="release_notes_4_3_0_subtitle_accessibility_improvements">Some improvements to make the application more accessible</string>
     <string name="release_notes_4_3_0_title_show_app_provider_icon_from_endpoint">New icons in "Open in (web)" option on the operations menu</string>
     <string name="release_notes_4_3_0_subtitle_show_app_provider_icon_from_endpoint">More appropriate icons have been added to the "Open in (web)" option on the operations menu</string>
+    <string name="release_notes_4_3_0_title_av_offline_files_removed_locally_with_local_only_option">Displayed \"Local only\" option appropriately in remove dialog</string>
+    <string name="release_notes_4_3_0_subtitle_av_offline_files_removed_locally_with_local_only_option">\"Local only\" option in remove dialog will be shown when the selected folder contains at least one downloaded file, ignoring those available offline</string>
 
     <!-- Open in web -->
     <string name="ic_action_open_in_web">Open in web</string>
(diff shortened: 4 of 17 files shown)


JuancaG05 (changes requested):
Some changes and comments here for you to review @Aitorbp!

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -584,19 +588,30 @@ class OCFileRepository(
...
+                } else {
+                    deleteLocalFile(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+                }
             }
         }
 
-        // 2. Remove the folder itself
-        deleteLocalFile(ocFile = ocFile, onlyFromLocalStorage = onlyFromLocalStorage)
+        // 2. Remove the folder itself if it has files
+        deleteFolderIfHasNoFilesInside(ocFile = ocFile, onlyFromLocalStorage = onlyFromLocalStorage)
```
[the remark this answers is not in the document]
Aitorbp: Previously, the app deleted the folder and its contents, whether it was empty or not, which I haven't checked but it should always be empty.
With the new implementation, the filesInFolder variable will not be empty because we have set a condition for the files to remain available offline. So this new method should only delete the folder if it has no content inside.
You can do a brief test with Android Studio to check that these files are not deleted in the Device Explorer.
With the following file status:

Folder1
       File1 ->Downloaded
       File2 -> Av. offline

Step 1: Check that we have these files in the Device Manager
Step 2: Click on the parent folder and do a delete Local only.
Step 3: Check the Device Manager again and observe that only `File2` remains in `Folder1`, the only one we had available offline.

Now if you want to check what would happen with the previous functionality, just change this `deleteLocalFolderIfItHasNoFilesInside` method to `deleteLocalFile`. Following the same steps you will see that the `Folder1` folder is completely deleted and this is not the result we want.

owncloudApp/src/main/java/com/owncloud/android/syncadapter/FileSyncAdapter.java
```diff
@@ -4,8 +4,9 @@
  * @author Bartek Przybylski
  * @author David A. Velasco
  * @author David González Verdugo
- * Copyright (C) 2011  Bartek Przybylski
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -584,19 +588,30 @@ class OCFileRepository(
...
+    }
+
+    private fun deleteFolderIfHasNoFilesInside(ocFile: OCFile, onlyFromLocalStorage: Boolean) {
```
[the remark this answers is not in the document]
Aitorbp: done

owncloudData/src/main/java/com/owncloud/android/data/providers/LocalStorageProvider.kt
```diff
@@ -211,6 +212,17 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
         return fileToDelete.deleteRecursively()
     }
 
+    fun deleteFolderIfHasNoFilesInside(ocFile: OCFile) {
+        val safeStoragePath = ocFile.getStoragePathOrExpectedPathForFile()
+        val folder = File(safeStoragePath)
+
+        val filesInFolder = folder.listFiles()
+        if (filesInFolder.isNullOrEmpty()) { // Verify if the folder is empty
+            // If it is empty, delete it
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/removefile/RemoveFilesDialogFragment.kt
```diff
@@ -106,7 +104,7 @@ class RemoveFilesDialogFragment : ConfirmationDialogFragment(), ConfirmationDial
                     R.string.confirmation_remove_files_alert
                 }
             }
-            val localRemoveButton = if (!containsAvailableOffline && isAvailableLocally) {
+            val localRemoveButton = if (isAvailableLocallyAndNotAvailableOffline) {
```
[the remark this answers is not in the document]
Aitorbp: The situation is that what is checked in that if has already been checked previously in this Use Case `IsAnyFileAvailableLocallyAndNotAvailableOfflineUseCase`, so it would no longer be necessary to keep that check there. I am talking about this one:
`val localRemoveButton = if (!containsAvailableOffline && isAvailableLocally) {`
Therefore, since we are already bringing the variable that has validated these two variables previously, it would be redundant to keep this condition as it was. We are saving code and not repeating a check that was already done previously.
This check is also not necessary because it was already done previously:
```
if (file.isAvailableOffline) {
                    containsAvailableOffline = true
                }
```

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -763,6 +763,8 @@
     <string name="release_notes_4_3_0_subtitle_accessibility_improvements">Some improvements to make the application more accessible</string>
     <string name="release_notes_4_3_0_title_show_app_provider_icon_from_endpoint">New icons in "Open in (web)" option on the operations menu</string>
     <string name="release_notes_4_3_0_subtitle_show_app_provider_icon_from_endpoint">More appropriate icons have been added to the "Open in (web)" option on the operations menu</string>
+    <string name="release_notes_4_3_0_title_av_offline_files_removed_locally_with_local_only_option">Displayed \"Local only\" option appropriately in remove dialog</string>
+    <string name="release_notes_4_3_0_subtitle_av_offline_files_removed_locally_with_local_only_option">\"Local only\" option in remove dialog will be shown when the selected folder contains at least one downloaded file, ignoring those available offline</string>
```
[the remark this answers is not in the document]
Aitorbp: Added

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -584,19 +588,30 @@ class OCFileRepository(
...
+        localStorageProvider.deleteFolderIfHasNoFilesInside(ocFile = ocFile)
+        deleteFromLocalDatabase(ocFile, onlyFromLocalStorage)
     }
 
     private fun deleteLocalFile(ocFile: OCFile, onlyFromLocalStorage: Boolean) {
         localStorageProvider.deleteLocalFile(ocFile)
+        deleteFromLocalDatabase(ocFile, onlyFromLocalStorage)
+    }
+
+    private fun deleteFromLocalDatabase(ocFile: OCFile, onlyFromLocalStorage: Boolean) {
```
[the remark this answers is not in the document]
Aitorbp: Done

Aitorbp:
Checks functionalities of `deleteLocalFolderRecursively`:

Case Delete files only local:
1. Create a folder with an available offline file and another downloaded file inside.
2. Delete the folder only local.
3. Check in Device Explorer that only the downloaded file has been deleted and both the parent folder and the available offline file are kept.

Case refresh on a folder that no longer exists:
1. Create a folder with available offline and downloaded files inside.
2. Delete the folder from the web.
3. Do a refresh in the main list.
4. Check that all files and also the available offline files have been deleted on that folder in Device Explorer

Case move folder with a source folder that no longer exists:
1. Create a folder with downloaded and available offline files.
2. Start the process of moving the folder.
3. Go to the web and delete the source folder.
4. End the move process.
5. Check Device Explorer to see if the entire source folder has been deleted.

If you move a folder to a destination folder that no longer exists:
1. Move a folder or file to a folder with available offline and downloaded files inside.
2. In the process of moving it, we delete the folder from the web.
3. We finish the process of moving the folder.
4. Check Device Explorer to see if the entire destination folder has been deleted

If you copy a folder from a source folder that no longer exists:
1. Create a folder with download and available offline files.
2. Start the process of copying the folder.
3. Go to the web and delete the source folder.
4. Check Device Explorer to see if the entire source folder has been deleted when the process is finished

If you copy a folder to a destination folder that no longer exists:
1. Copy a folder or file to a folder with available offline and downloaded files inside.
2. In the process of moving it, we delete the folder from the web.
3. We finish the process of moving the folder.
4. Check in Device Explorer if the entire destination folder has been deleted.

I leave you these test cases in case you want to try them when doing QA @jesmrec

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -579,24 +583,49 @@ class OCFileRepository(
...
+                        onlyFromLocalStorage = onlyFromLocalStorage,
+                    )
+                }
             } else {
-                deleteLocalFile(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+                deleteLocalFolderRecursivelyOrLocalFile(
+                    file = file,
+                    onlyFromLocalStorage = onlyFromLocalStorage,
+                )
             }
```
JuancaG05, line 603: The content of the `if` and the `else` is the same. That doesn't make sense, take a look at the algorithm

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -579,24 +583,49 @@ class OCFileRepository(
...
+
+    private fun deleteLocalFolderRecursivelyOrLocalFile(file: OCFile, onlyFromLocalStorage: Boolean) {
+        if (file.isFolder) {
+            deleteLocalFolderRecursively(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+        } else {
+            deleteLocalFile(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+        }
+    }
+
+    private fun deleteFolderIfItHasNoFilesInside(ocFolder: OCFile, onlyFromLocalStorage: Boolean) {
```
JuancaG05, line 618:
```suggestion
    private fun deleteLocalFolderIfItHasNoFilesInside(ocFolder: OCFile, onlyFromLocalStorage: Boolean) {
```

JuancaG05 (changes requested):
Some more things to check here @Aitorbp 👀

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -579,24 +583,49 @@ class OCFileRepository(
         localFileDataSource.cleanWorkersUuid(fileId)
     }
 
-    private fun deleteLocalFolderRecursively(ocFile: OCFile, onlyFromLocalStorage: Boolean) {
+    private fun deleteLocalFolderRecursively(ocFile: OCFile, onlyFromLocalStorage: Boolean, keepAvailableOfflineFile: Boolean = false) {
```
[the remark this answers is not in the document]
Aitorbp: There is only one method where the behavior is different. But I'm going to rename the variable to `removeAvailableOfflineFile` and set the default value to true. Then in deleteFile I'm going to set it to false. This way I think this condition is more understandable:
`if (removeAvailableOfflineFile || !file.isAvailableOffline) {`

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -579,24 +583,49 @@ class OCFileRepository(
...
         val folderContent = localFileDataSource.getFolderContent(ocFile.id!!)
 
         // 1. Remove folder content recursively
         folderContent.forEach { file ->
-            if (file.isFolder) {
-                deleteLocalFolderRecursively(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+            if (keepAvailableOfflineFile) {
+                if (!file.isAvailableOffline) {
```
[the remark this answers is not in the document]
Aitorbp: Right, done

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
Aitorbp: Changed

owncloudData/src/main/java/com/owncloud/android/data/providers/LocalStorageProvider.kt
```diff
@@ -211,6 +212,16 @@ sealed class LocalStorageProvider(private val rootFolderName: String) {
...
     }
 
+    fun deleteLocalFolderIfItHasNoFilesInside(ocFolder: OCFile) {
+        val safeStoragePath = ocFolder.getStoragePathOrExpectedPathForFile()
+        val folder = File(safeStoragePath)
+
+        val filesInFolder = folder.listFiles()
+        if (filesInFolder.isNullOrEmpty()) {
+            folder.deleteRecursively()
+        }
```
jvillafanez, line 222: Calling `folder.delete()` directly could be fine.

Listing files could take some time of the folder contain too many files. We just need to know where the folder contains no files, or at least one; getting the whole list is irrelevant.
The `deleteRecursively` will also have some little overhead when we know that the folder is empty (because we've just checked it).

By calling directly the `delete()` method, we don't need to list the whole folder because the method will fail if the folder isn't empty. We might need to ensure that the `ocFolder` is actually a folder so we don't try to remove a file (the check might still be needed regardless of the proposed changes, unless there are guarantees that the `ocFolder` will NEVER be a file)

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
```diff
@@ -1129,7 +1135,13 @@ class MainFileListFragment : Fragment(),
                 R.id.action_set_available_offline -> {
                     fileOperationsViewModel.performOperation(FileOperation.SetFilesAsAvailableOffline(listOf(singleFile)))
                     if (singleFile.isFolder) {
-                        fileOperationsViewModel.performOperation(FileOperation.SynchronizeFolderOperation(singleFile, singleFile.owner))
+                        fileOperationsViewModel.performOperation(
+                            FileOperation.SynchronizeFolderOperation(
+                                folderToSync = singleFile,
+                                accountName = singleFile.owner,
+                                isActionSetFolderAvailableOffline = true,
+                            )
```
jvillafanez, line 1143: Nitpicking here, I'd prefer to keep consistency with the rest of the code.

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -579,24 +583,35 @@ class OCFileRepository(
...
+    private fun deleteLocalFolderRecursively(ocFile: OCFile, onlyFromLocalStorage: Boolean, removeAvailableOfflineFile: Boolean = true) {
         val folderContent = localFileDataSource.getFolderContent(ocFile.id!!)
 
         // 1. Remove folder content recursively
         folderContent.forEach { file ->
-            if (file.isFolder) {
-                deleteLocalFolderRecursively(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
-            } else {
-                deleteLocalFile(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+            if (removeAvailableOfflineFile || !file.isAvailableOffline) {
```
JuancaG05, line 591: This is not equivalent to the condition you wanted to write before

JuancaG05 (changes requested):
One more change, and review as well the non-resolved comment of the previous CR round @Aitorbp

owncloudData/src/main/java/com/owncloud/android/data/providers/LocalStorageProvider.kt
JuancaG05: This method will only be invoked for folders. You can see that its only use in `OCFileRepository` is surrounded by `if (file.isFolder) { }`.

OTOH, I'm not sure how the `delete()` method handles the failure if we try to remove a non-empty file. Looking at the documentation, it seems that it doesn't throw an exception because of this. The only signal we have is that it may return `false`, but documentation only specifies that it returns `false` if the folder couldn't be deleted correctly, could be that it returns false and not due to the folder not being empty (for example, due to permissions or whatever). So, I don't really know if using `delete()` is a good approach in this case

jvillafanez: > This method will only be invoked for folders. You can see that its only use in OCFileRepository is surrounded by if (file.isFolder) { }.

But this method is public and it could be called from anywhere. Furthermore, specific requirements to call this method, such as `ocFolder` being an actual folder and not a file, aren't documented, so there is a risk that new code will use this method incorrectly and will cause issues.

> but documentation only specifies that it returns false if the folder couldn't be deleted correctly, could be that it returns false and not due to the folder not being empty (for example, due to permissions or whatever). 

I haven't checked the docs for the `deleteRecursively` method, but it should have the same problem. The question is whether we can and / or need to distinguish any of the error cases. With the current code, it doesn't seem we're interested in the reason of the failure.

Assuming we have only 2 error conditions (folder not empty and permission issue), we could assume that everything is fine and try to delete the folder via `delete()`, then if the method fails we could check if the folder is empty in order to rule out which error case happened. This should perform faster assuming the success case is more predominant.

Aitorbp: We've looked at the Java documentation and the `delete()` method doesn't specify when the method receives a false or an exception. So to try to cover at least the case of deleting the folder when it's empty, we're going to keep this part of the code:
```
val filesInFolder = folder.listFiles()
        if (filesInFolder.isNullOrEmpty()) {
```
 and change the `folder.deleteRecursively()` to the `delete()`.

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
```diff
@@ -584,19 +588,30 @@ class OCFileRepository(
 
         // 1. Remove folder content recursively
         folderContent.forEach { file ->
-            if (file.isFolder) {
-                deleteLocalFolderRecursively(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
-            } else {
-                deleteLocalFile(ocFile = file, onlyFromLocalStorage = onlyFromLocalStorage)
+            if (!onlyFromLocalStorage || !file.isAvailableOffline) {
```
JuancaG05, line 591: I can't see why you used `onlyFromLocalStorage` here, could you explain it a bit?

JuancaG05 (commented):
@Aitorbp an explanation would be nice here, so that we can check it in the future if we forget

owncloudData/src/main/java/com/owncloud/android/data/files/repository/OCFileRepository.kt
Aitorbp: `onlyFromLocalStorage` is working the same as the `removeAvailableOfflineFile` variable we had previously created. That is, when this variable is `true`, the available offline files should not be deleted and the other files will be reset. If the variable is false, the files in the database will be deleted, included av.offline.

The value of the variable in the uses of the `deleteLocalFolderRecursively` function is the same as in the previously created variable. That is, in the cases of the `catch` it is set to `false`, and in the `deleteFiles` case it is set to whatever is given to us by parameter, which could be `true` or `false`.

In conclusion, it does the same job as the `removeAvailableOfflineFile` variable.

JuancaG05: The condition is clearer and equivalent like this: `if (!(onlyFromLocalStorage && file.isAvailableOffline))`

Change it by this and add a clarifying comment and we're done

JuancaG05 (approved):
Let's move this to QA 🚀

jesmrec:
First, checks over https://github.com/owncloud/android/issues/4402, that is merged into here

Folder with some files and a subfolders that also contains some files

Folder1
|_ File1
|_ File2
|_ Folder2
&nbsp;&nbsp;&nbsp;&nbsp;|_ File3
&nbsp;&nbsp;&nbsp;&nbsp;|_ File4

- [X] Nothing downloaded + Av. offline + Unset av. offline  -> Everything downloaded
- [X] File1 & File2 downloaded +  remove locally File2 + Av. offline + Unset av. offline  -> Everything downloaded
- [x] File3 & File4 downloaded +  remove locally File3 + Av. offline + Unset av. offline  -> Everything downloaded
- [x] File1 & File3 downloaded +  remove locally File3 + Av. offline + Unset av. offline  -> Everything downloaded
- [x] File1 & File3 downloaded +  remove locally File3 & File1 + Av. offline + Unset av. offline  -> Everything downloaded
- [x] Everything downloaded +  remove everything locally  +  Av. offline + Unset av. offline  -> Everything downloaded

jesmrec:
Basic behaviour:

Remove folder dialog when...:

- [X] Nothing is downloaded inside -> `Local only` not displayed in dialog
- [X] Some files are dowloaded and others are not downloaded -> `Local only` displayed in dialog
- [X] Some files are downloaded and others are av. offline -> `Local only` displayed in dialog
- [X] Some files are av. offline and others are not downloaded -> `Local only` not displayed in dialog
- [X] Everything is downloaded -> `Local only` displayed in dialog
- [X] Everything is av. offline -> `Local only` not displayed in dialog

All those tests were repeated with files inside a subfolder.

Local copies of the items:

jesmrec:
### (1)

I noticed the bad behaviour when a sync is performed (not an av. offline);

With the following structure:

Folder with some files and a subfolders that also contains some files

Folder1
|_ File1
|_ File2
|_ Folder2
&nbsp;&nbsp;&nbsp;&nbsp;|_ File3
&nbsp;&nbsp;&nbsp;&nbsp;|_ File4

1. Download File3
2. Remove locally File3
3. Sync Folder1
4. Browse to Folder2

**Current:**

File3 and File 4 are not downloaded.

**Expected:**

Both File3 and File4 are downloaded, because their parent folder was synced in the previous step. 

That was the problem fixed in https://github.com/owncloud/android/issues/4402 from the av. offline files, that seems to be reproducible with sync.

Pixel 2, Android 11
`62a76c55`

jesmrec:
Approved on my side.

Issue #3874, opened 2023-01-26
[SPACES] Get Personal spaces quota from GraphAPI


jesmrec:
Quota is displayed at the bottom of the side menu. Legacy and current behaviour is showing the personal space quota. So, we should always show the personal space quota in the drawer (the general quota for oC10).

Now, this value should be fetched from GraphAPI's drives endpoint in case of oCIS, and keeping the current behaviour for 0C10.

```
"quota" : {
            "remaining" : 2655551488,
            "state" : "normal",
            "total" : 0,
            "used" : 66351210
},
```

Docs: https://owncloud.dev/libre-graph-api/#/me.drives/ListMyDrives

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/personal_space_quota
- [x] Development tasks
  - [x] Implement logic for oC10
  - [x] Implement logic for oCIS
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

abelgardep:
I understand the idea, but oCIS has compatibility with this AFAIT: 
Propfind Request`https://server/remote.php/dav/files/user`
```
<?xml version='1.0' encoding='UTF-8' ?>
<propfind xmlns="DAV:" xmlns:CAL="urn:ietf:params:xml:ns:caldav" xmlns:CARD="urn:ietf:params:xml:ns:carddav" xmlns:SABRE="http://sabredav.org/ns" xmlns:OC="http://owncloud.org/ns">
    <prop>
        <quota-used-bytes />
        <quota-available-bytes />
    </prop>
</propfind>
```
Response:
```
<d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
    <d:response>
        <d:href>/remote.php/dav/files/user/</d:href>
        <d:propstat>
            <d:prop>
                <d:quota-used-bytes>16996007</d:quota-used-bytes>
                <d:quota-available-bytes>-2</d:quota-available-bytes>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
    </d:response>
</d:multistatus>
```

Is it planned to remove the compatibility?
Do we need to show something different here?

michaelstingl:
Spaces have quota information in the Graph endpoint. Better align quota handling across Personal Space and Project Spaces?

PR #4401, opened 2024-05-09
[FEATURE] Get Personal spaces quota from GraphAPI

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/3874

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.3/Quota.md

Reports:

- [x] (1) Quota updated https://github.com/owncloud/android/pull/4401#issuecomment-2112894204 [FIXED]
- [x] (2) "No restriction" value shows a limit https://github.com/owncloud/android/pull/4401#issuecomment-2122914804 [FIXED]
- [X] (3) Improvements https://github.com/owncloud/android/pull/4401#issuecomment-2122995186 [WONT FIX]
- [x] (4) Crash on fresh install  https://github.com/owncloud/android/pull/4401#issuecomment-2127026972 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/operations/SyncProfileOperation.kt
```diff
@@ -60,30 +60,33 @@ class SyncProfileOperation(
...
+                        if (it.isSpacesAllowed()) {
+                            val refreshUserQuotaFromServerAsyncUseCase: RefreshUserQuotaFromServerAsyncUseCase by inject()
+                            val userQuotaResult =
+                                refreshUserQuotaFromServerAsyncUseCase(
+                                    RefreshUserQuotaFromServerAsyncUseCase.Params(
+                                        account.name
+                                    )
+                                )
+                            userQuotaResult.getDataOrNull()?.let {
+                                Timber.d("User quota synchronized for account ${account.name}")
```
JuancaG05, line 76: Just to make clear the difference in the logs, I would add an oC10 mention here:
```suggestion
                                Timber.d("User quota synchronized for oC10 account ${account.name}")
```

owncloudApp/src/main/java/com/owncloud/android/operations/SyncProfileOperation.kt
```diff
@@ -60,30 +60,33 @@ class SyncProfileOperation(
...
+                                val shouldFetchAvatar = storedCapabilities?.isFetchingAvatarAllowed() ?: true
```
JuancaG05, line 77: Let's put the avatar stuff out of the `let` block of the capabilities. In case they are null, this won't be executed, and we want it to be executed as it was before

owncloudData/src/main/java/com/owncloud/android/data/spaces/repository/OCSpacesRepository.kt
```diff
@@ -23,16 +24,25 @@ package com.owncloud.android.data.spaces.repository
...
     private val localSpacesDataSource: LocalSpacesDataSource,
+    private val localUserDataSource: LocalUserDataSource,
     private val remoteSpacesDataSource: RemoteSpacesDataSource,
 ) : SpacesRepository {
     override fun refreshSpacesForAccount(accountName: String) {
         remoteSpacesDataSource.refreshSpacesForAccount(accountName).also { listOfSpaces ->
             localSpacesDataSource.saveSpacesForAccount(listOfSpaces)
+            val personalSpace = listOfSpaces.find { it.isPersonal }
+            personalSpace?.let {
+                val userQuota = UserQuota(accountName, it.quota?.total!!, it.quota?.used!!)
```
JuancaG05, line 42: Is the  `available` property in `UserQuota` (the second parameter here) the equivalent to the `total` property in `SpaceQuota`? I have doubts whether that `available` corresponds to the `total` or to the `remaining`... This deserves a check 👍

JuancaG05 (changes requested):
Some suggestions and questions here @joragua 👍

owncloudData/src/main/java/com/owncloud/android/data/spaces/repository/OCSpacesRepository.kt
joragua: I've checked how the legacy version was and the `available` value is the `remaining` one, so I changed it ✅  Also the `DrawerActivty` calculates the total with `used` and `available` values.

owncloudApp/src/main/java/com/owncloud/android/operations/SyncProfileOperation.kt
joragua: Done 😃

JuancaG05: I meant, not only out of the user-quota `let` but out of the whole capabilities `let` 👍

JuancaG05 (changes requested):
Check the reply to one of the comments of the previous CR round and we're done 😉 @joragua

owncloudApp/src/main/java/com/owncloud/android/operations/SyncProfileOperation.kt
joragua: Ok, It's done

JuancaG05 (approved):
Good job! 💯

jesmrec:
### (1)

How or when is the quota checked? i mean, after adding an account, qouta data is retrieved and displayed. But, if quota changes later, is it rechecked and updated? i wasn't able to.

jesmrec:
Let's do a quick look over the current behaviour

|| current branch | current `master` | `v4.2.1` |
|:-|:-:|:-:|:-:|
| **oC10** | Correct after login. After a couple of reopenings, it is refreshed ✅ | Correct after login. After a couple of reopenings, it is refreshed ✅ |  Correct after login. After a couple of reopenings, it is refreshed ✅ | 
| **oCIS** | No updates beyond the login ❌  | Correct after login. After a couple of reopenings, it is refreshed ✅  | Correct after login. If quota changes, it is refreshed but not immediately. Need some reopens (don't know why) ✅  |

that means, i see changes in the quota in both master and 4.2.1 for both backends. The new current scenario is where i don't see updates.

jesmrec:
### (2)

1. In oCIS account, set quota to "No restriction"
2. Refresh quota in Android app

Current: 

This is what GraphAPI returns:

```
            "quota": {
                "remaining": 26283307008,
                "state": "normal",
                "total": 0,
                "used": 382726391
            },
```

and this is displayed in app: 

Screenshot 2024-05-21 at 17 33 29

Expected:

If "total" == 0 , there is no limit, so the amount to display would be only the "used" (365MB in the example) 

Pixel 2,  Android 11
`484062c4c`

jesmrec:
### (3)[IMPROVEMENT]

In case the quota is exceeded, we are showing a message `No storage usage information available`

In this case, GraphAPI returns something like:

```
            "quota": {
                "remaining": 0,
                "state": "exceeded",
                "total": 1000000000,
                "used": 1423007244
            },
```

so, we have information about it (state == exceeded)... options that come to my mind:

- Showing the same information as the non-exceeded quota with an extra message "Quota exceeded" or similar. Using red color (i think that text is not brandable)

- Showing only "Quota exceeded"

As extra ball.. apart of "normal" and "exceeded", there is another value for the status that is "nearing" when the quota is close to the limit.

For sure, these are improvements and could be addressed to another issue.

joragua:
### FIXES

About **(1)**: It has been added a refresh for spaces when `FileDisplayActivity` is created.


About **(2)** : At first, the value of `total` quota is checked. If it's 0 then `available` value is set to -3 (the same as oC10). If that value it's no 0 then `available` value is the `remaining` one from GraphAPI.

JuancaG05 (approved):

jesmrec:
(1) and (2) fixed.

will create a new issue for (3)

jesmrec:
### (4)

App crashes with frecuency after fresh install.. not always (unfortunately). I was diving a little bit, and i fount the following stacktrace

```
...
 Caused by: java.lang.NullPointerException: Attempt to read from field 'java.lang.String android.accounts.Account.name' on a null object reference
                 	at com.owncloud.android.ui.activity.FileDisplayActivity$spacesListViewModel$2.invoke(FileDisplayActivity.kt:180)
                 	at com.owncloud.android.ui.activity.FileDisplayActivity$spacesListViewModel$2.invoke(FileDisplayActivity.kt:179)
                 	at org.koin.core.scope.Scope.resolveInstance(Scope.kt:223)
                 	at org.koin.core.scope.Scope.get(Scope.kt:210)
...
```

pointing to `FileDisplayActivity` lines 179 and 180, that just were added in he current PR:

```
 private val spacesListViewModel: SpacesListViewModel by viewModel {
        parametersOf(account.name, false)
    }
```

that comes to my mind together with https://github.com/owncloud/android/pull/4408, where the null account is protected.

joragua:
**(4)** can be fixed refreshing all spaces when the account is set (`onAccountSet`) and not in `onCreate`. This solution would be:


```
private val spacesListViewModel: SpacesListViewModel by viewModel {
            parametersOf(account.name, false)
}
        
spacesListViewModel.refreshSpacesFromServer()
```

jesmrec:
(4) fixed, and PR approved

Merge blocked till 4.2.2 is out

Issue #4405, opened 2024-05-15
[BUG] Disabled spaces are listed


jesmrec:
### Steps to reproduce

1. In a oCIS account with some spaces, open the spaces list -> all spaces listed
2. In web, disable a space
3. Refresh spaces list in the ap

### Actual behaviour

Disabled space is listed, but thumbnail changes

### Expected behaviour

Disabled spaces should not be listed

In this short video, the change of thumbnails is visible. After the second refresh, the "AAAA" space is disabled in web:
 
https://github.com/owncloud/android/assets/14894746/218f7b38-0389-4bec-87e9-c6b2ef1060b9

Pixel 2, Android 11

PR #4406, opened 2024-05-17
[FIX] Disabled spaces are listed

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4405

_____


changelog/unreleased/4406
JuancaG05, line 1: This is regressive so we don't need a calens here 😉

JuancaG05 (changes requested):
Nice! We just need to remove calens here @joragua 🚀

changelog/unreleased/4406
joragua: Done ✅

JuancaG05 (approved):
LGTM!!

jesmrec:
Checked in the three main cases for spaces listing:

- List of spaces
- Folder picker in oCIS for copy operation
- Sharing from 3rd party app to oCIS account

In every of them, checking listing and filtering with some disabled spaced.

That's OK from my side

Issue #4402, opened 2024-05-10
[BUG] Av. offline operation does not work correctly when run from a parent folder


Aitorbp:
### Actual behaviour
If we have a folder with downloaded files and a subfolder with more downloaded files, if we locally delete one of them and click on Available offline on the parent folder, not all the files will be set as Available offline and this is not correct.

### Expected behaviour
When we click `Set available offline` on a folder, all its children must be set as available offline.
 
### Steps to reproduce
1. If we have the following tree of files in our account:
Folder1 
	File1 -> downloaded
	Folder2
		FIle2 -> downloaded 
		File3 -> downloaded
2. We delete the file file2 locally
3. We go to the parent folder and set it to av.offline 
4. We wait for the process to finish and click on `Unset available offline`.
5. We navigate to folder2 and observe that file2 has not been downloaded.

I think the problem comes from this [pull request](https://github.com/owncloud/android/issues/4197), now our code will only do a sync if the local and remote etags are different. When we click on `Set available offline`, the files contained within folder2 will have the same etag because they have not been modified remotely, therefore they will not be synchronized.
This bug also blocks this [pull request](https://github.com/owncloud/android/pull/4399).

Possible solution:
Create a new parameter in `SynchronizeFolderUseCase` called `isActionSetFolderAvailableOffline` set by default to false. When we perform the operation it will be set to true and a refresh of all the contents of that folder will be performed.

This issue is a regressive, so no need to add calens and release notes.

PR #4407, opened 2024-05-20
Fix Issue #4402 : Ensure all files are synchronized when setting folder to availab…

praveen7512:
…le offline

- Added a new parameter `isActionSetFolderAvailableOffline` to the `SynchronizeFileUseCase.Params` data class to handle forced synchronization when setting a folder to available offline.

- Updated `SynchronizeFileUseCase` to check for the `isActionSetFolderAvailableOffline` parameter and force a refresh of the folder's contents.

- Implemented the `forceRefresh` method to initiate a download of the file regardless of the etag comparison when `isActionSetFolderAvailableOffline` is true.

- Ensured that locally deleted files are re-downloaded if they still exist on the remote server when the parent folder is set to available offline.
- Added appropriate logging to track the synchronization process and forced refresh actions.

This fix addresses the issue where setting a folder to available offline did not correctly synchronize all its contents, especially when some files were deleted locally. Now, all files within the folder will be synchronized and set to available offline as expected.

## Related Issues
App:

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/synchronization/SynchronizeFileUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/synchronization/SynchronizeFileUseCase.kt
@@ -61,25 +61,31 @@ class SynchronizeFileUseCase(
                 return SyncType.FileNotFound
             }
 
-            // 2. File not downloaded -> Download it
+            // 2. Check if we need to force a refresh (isActionSetFolderAvailableOffline is true)
+            if (params.isActionSetFolderAvailableOffline) {
+                Timber.i("Forced refresh due to folder offline availability action")
+                return forceRefresh(accountName, fileToSynchronize)
+            }
+
+            // 3. File not downloaded -> Download it
             if (!fileToSynchronize.isAvailableLocally) {
                 Timber.i("File ${fileToSynchronize.fileName} is not downloaded. Let's download it")
                 val uuid = requestForDownload(accountName = accountName, ocFile = fileToSynchronize)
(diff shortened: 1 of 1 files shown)


jesmrec:
Thanks for your contribution. We have another PR that fixes #4402, that is #4407... we will check yours as well.

Aitorbp:
Hello @praveen7512!  Thank you very much for sending us this PR. I advise you that before you start working on an issue, you see if there is already someone else working on it with another solution. I recommend you take a look at this [PR](https://github.com/owncloud/android/pull/4399) where the solution is more accurate than the one you propose.

I'm afraid I'm going to close this PR because it's invalid, due to the following points:

1. First of all, you've implemented the development logic in the wrong class. Before the code goes to `SynchronizeFileUseCase`, it has to go through `SynchronizeFolderUseCase`. It's in the `fileRepository.refreshFolder` function where you have to check if we've clicked on the available offline option. If you go back to the case I proposed in the issue, with the implementation you've developed it's not going to work.
2. On the other hand, the variable `isActionSetFolderAvailable` Offline is not set to true in any case. Therefore it will always be set to `false`. Also in the `FileOperationsHelper.java` class you need to set the value of the parameter `isActionSetFolderAvailable`, otherwise Android Studio won't compile.
3. Finally, the `forceRefresh` function is not necessary, since the problem is not that the data is not downloaded and synchronized, but that the files that are to be made available offline are not being transferred correctly.

So, due to all these errors and that it implies the majority of code in this PR, I'm closing it. In any case, thanks for your PR and do not hesitate to open more PRs, we will be happy to review your code again. 😃

PR #4408, opened 2024-05-22
[FIX] Unwanted DELETE operations when synchronization in single file fails

JuancaG05:
When performing a read remote file operation (synchronization over a single file), if the account in the `OwncloudClient` is null, we'll throw an `AccountNotFoundException`. This will be handled in 3 different views:

- In the main file list view
- In the file details view
- In the image preview view

In the 3 cases, we'll show a Snackbar telling `Sync failed, you need to log in again`, suggesting that by repeating the login process, the problem can be solved (since we'll save again the account, which might have been lost in some migration process from a very old version).

Also, we changed the way to handle the 404 error of a `PROPFIND`. Previously, we deleted locally and remotely the file that we requested in the `PROPFIND`, now we just remove it locally (since if we receive a 404, the file shouldn't exist in remote), avoiding sending `DELETE` requests.

## Related Issues
App: https://github.com/owncloud/enterprise/issues/6638

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ ownCloud admins and users.
 * Bugfix - Video streaming in spaces: [#4328](https://github.com/owncloud/android/issues/4328)
 * Bugfix - Retried successful uploads are cleaned up from the temporary folder: [#4335](https://github.com/owncloud/android/issues/4335)
 * Bugfix - Resolve incorrect truncation of long display names in Manage Accounts: [#4351](https://github.com/owncloud/android/issues/4351)
+* Bugfix - Unwanted DELETE operations when synchronization in single file fails: [#6638](https://github.com/owncloud/enterprise/issues/6638)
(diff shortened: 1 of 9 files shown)


joragua (approved):
LGTM! 😄

jesmrec:
Just a question here:

> In the 3 cases, we'll show a Snackbar telling Sync failed, you need to log in again

When auth is lost due to an invalid token, we use to show a snackbar with the message and a button/link that prompts the user to the login view. How difficult is to add that button to the snackbar in the same way? like:

Screenshot 2024-05-24 at 12 48 38

JuancaG05:
> Just a question here:
> 
> > In the 3 cases, we'll show a Snackbar telling Sync failed, you need to log in again
> 
> When auth is lost due to an invalid token, we use to show a snackbar with the message and a button/link that prompts the user to the login view. How difficult is to add that button to the snackbar in the same way? like:
> 
> Screenshot 2024-05-24 at 12 48 38

Should be added now 😃

jesmrec:
> Should be added now 😃

Correctly added and working in list of files, details view and image preview. Ready to go.

PR #4412, opened 2024-05-25
Fix progress bar visibility issue

lakshay6907:
Fixes the issue where progress bars are not properly shown on devices, particularly when Battery Saver mode is enabled. This is due to animations being disabled, which affects the visibility of progress bars.

### DIFF
--- a/owncloudApp/src/debug/java/com/owncloud/android/testing/SingleFragmentActivity.kt
+++ b/owncloudApp/src/debug/java/com/owncloud/android/testing/SingleFragmentActivity.kt
@@ -33,7 +33,7 @@ open class SingleFragmentActivity : BaseActivity() {
                 ViewGroup.LayoutParams.MATCH_PARENT,
                 ViewGroup.LayoutParams.MATCH_PARENT
             )
-            id = R.id.container
+//            id = R.id.container
         }
         setContentView(content)
     }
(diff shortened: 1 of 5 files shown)


jesmrec:
thanks for your contribution, we'll take a look. Please keep tuned in case the code review or QA stages requieres some changes in the code.

JuancaG05:
Hi @lakshay6907! I'm afraid I'm closing this PR because it is not valid, due to the next points:

- The code related to the functionality does not work over progress bars as you indicate wrongly in your PR's comment (also wrongly referred in the issue's description), but on the spinning circle of the login view. In any case, the one you chose to substitute it is the one already set in the XML file (which seems not to be shown ever because it is always replaced programatically), so that `else` shouldn't be necessary
- There are several commented lines in random files you added that shouldn't be there
- One of the things you commented (in `FileDisplayActivity.kt`) is about touches with other visible windows. That shouldn't be modified, it's part of a preference!
- In that very same class, you added a log and removed a String, which is needed in that line. That doesn't make sense now
- You added an empty line in another random class
- You added a XML drawable resource that is not used at all in the code, so it is unnecessary
- The code is not correctly formatted, so that our lint system won't approve those changes. The good point here is that Android Studio can reformat your code for you 😄
- (No hard requirement but nice to have) Conventional commits: as explained in [CONTRIBUTING.md](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md), we're using this convention for the commits on our repo. In this PR, 2 out of the 4 commits are not following that convention

So, due to all these errors and that it implies the majority of code in this PR, I'm closing it. In any case, don't hesistate to open a new one with all the improvements I detailed in this message and we'll review it again 😃

Issue #4352, opened 2024-03-27
[BUG] Text (.txt) file won't download / open in Android App


geobzmm:
### Actual behaviour
Creating a .txt file on another device containing a few words or some notes in a accessible folder then syncs as expected to all devices - server, laptop, desktop and Mobile / Cell. the .txt file can be opened / viewed on server, laptop (win 10), desktop (win 11) but Android App displays message "Download Failed". All other "document" formats appear to work fine (eg. .pdf, .docx, .xlsx, .doc etc...).

### Expected behaviour
The .txt file should open and show the text in the file or be downloadable to the Android device (neither of which is possible).
 
### Steps to reproduce
1. Create .txt file containing text such as "Hello World!".
2. Allow to sync to OwnCloud Server - this works when file is originally generated on PC.
3. Refresh folder on Android device and attempt to open / download file.
4. Message appears onscreen "Download Failed".

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)
Not tested on Official Server

### Environment data
Android version: v13

Device model: Samsung Galaxy S20+5G

Stock or customized system: Stock One UI v5.1

ownCloud app version: v4.2.1

ownCloud server version: Infinite Scale Community v4.0.3 running webclient v7.1.2

#### ownCloud App Log (data/owncloud.log)
```
D: 2024-03-27 20:28:32:819(SingleSessionManager.java:98)getClientFor starting 
V: 2024-03-27 20:28:32:819(SingleSessionManager.java:125)reusing client for account [OwnCloudServer]
D: 2024-03-27 20:28:32:820(SingleSessionManager.java:180)getClientFor finishing 
D: 2024-03-27 20:28:32:869(OwnCloudClient.java:129)Executing in request with id ff931731-xxx-etc
I: 2024-03-27 20:28:32:869(ReadRemoteFileOperation.kt:72)Read remote file /Basic Files/Another Text File.txt with status 207
I: 2024-03-27 20:28:32:871(SynchronizeFileUseCase.kt:66)File Another Text File.txt is not downloaded. Let's download it
I: 2024-03-27 20:28:32:875(DownloadFileUseCase.kt:103)Download of Another Text File.txt has been enqueued.
D: 2024-03-27 20:28:32:876(ViewModelExt.kt:68)Use case executed: SynchronizeFileUseCase with result: Success(data=DownloadEnqueued(workerId=d04d033c-xxxx-4833-b2c2-xxx))
[...]

PR #4415, opened 2024-05-29
[FIX] Downloads not working when `Content-Length` is not sent by the server

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4352
Also solves: https://github.com/owncloud/enterprise/issues/6697

_____


owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
```diff
@@ -448,9 +448,15 @@ class FileDetailsFragment : FileFragment() {
         } else { // Transfer is upload (?)
             getString(R.string.uploader_upload_in_progress_ticker)
         }
+        val workProgress = workInfo.progress.getInt(DownloadFileWorker.WORKER_KEY_PROGRESS, -1)
         binding.fdProgressBar.apply {
-            isIndeterminate = false
-            progress = workInfo.progress.getInt(DownloadFileWorker.WORKER_KEY_PROGRESS, -1)
+            if (workProgress == -1) {
+                isIndeterminate = true
```
jabarros, line 454: what do you think about avoiding the global if structure?

binding.fdProgressBar.apply {
        isIndeterminate = (workProgress == -1)
        progress = if (workProgress == -1) progress else workProgress
        invalidate()
    }

owncloudApp/src/main/java/com/owncloud/android/ui/preview/FileDownloadFragment.kt
```diff
@@ -220,11 +223,15 @@ class FileDownloadFragment : FileFragment() {
...
-                onWorkRunning = { progress ->
+                onWorkRunning = { workProgress ->
                     progressBar?.apply {
-                        isIndeterminate = false
-                        setProgress(progress)
-                        invalidate()
+                        if (workProgress == -1) {
+                            isIndeterminate = true
+                        } else {
+                            isIndeterminate = false
```
jabarros, line 231: same here?

jabarros (approved):
LGTM

owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
JuancaG05: I just made it like that because progress shouldn't be defined in case `workProgress == -1`, and invalidate as well, those 2 lines should be only executed when `workProgress != -1`. That's why I decided englobe everything in a `if else` structure, to avoid adding the same `if` in every line

owncloudApp/src/main/java/com/owncloud/android/ui/preview/FileDownloadFragment.kt
JuancaG05: Replied in previous comment 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
jabarros: Right, invalidate is inside a condition :)

jesmrec:
Works for me. Ready to go in 4.2.2

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/files/DownloadRemoteFileOperation.kt
```diff
@@ -131,7 +131,7 @@ class DownloadRemoteFileOperation(
                     }
                 }
 
-                if (transferred == totalToTransfer) {  // Check if the file is completed
+                if (totalToTransfer == -1L || transferred == totalToTransfer) {  // Check if the file is completed
```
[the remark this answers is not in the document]
JuancaG05: Hey @appiekap653!

No, we're not checking if the download finishes with all the bytes for these cases in which `Content-Length` is not received. That's why we show an indeterminate progress bar 😄. In any case, this is a hotfix for the problem, the elegant solution will come later. Our idea is not relying on `Content-Length` anymore and check the total size (so that we're able to check the progress) from the metadata of the file in some way, since `Content-Length` might not express correctly the size of the file (for example, if the response was compressed in some way).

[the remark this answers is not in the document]
appiekap653: Does there excists no situation where `savedfile` is set `True` while it isn't saved at all, with this modification? and the rest of the code about the modification date is still executed then.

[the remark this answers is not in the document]
JuancaG05: Those cases are covered some lines before this. If there is a crash in the server, a cancellation will be requested and so an `OperationCancelledException` will be thrown (line 120), and if disk space is full, the `write` method will throw an exception (line 123)

PR #4419, opened 2024-06-10
Add new step to release template

jesmrec:
Send a ping when release is done, in order to keep https://owncloud.com/mobile-apps/ updated 

No calens or release notes required here

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -17,7 +17,7 @@ For OEM releases, keep the OEM Release template and remove the Open and Patch Re
 
 ### TASKS:
 
- - [ ] [DOC] Ping in #documentation-internal about the new release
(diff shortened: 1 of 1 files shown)


JuancaG05 (commented):
We can take advantage of this PR to introduce some more changes in the release template that are needed. For example:

- `[COM] Inform #updates in internal chat that release is out` does not apply anymore, we can change #updates by #general, and mention the chat website since it is public
- Add the step for pinging @mmattel about docs before starting the process and once it's about to finish
- In the case of patch releases, clarify the merge step. Last time, I did it locally (not via GitHub web as the step suggests), I don't think complex conflicts that can happen can be resolved easily by the web

Any more change that I'm missing?

jesmrec:
> Add the step for pinging @mmattel about docs before starting the process and once it's about to finish

These steps already exist, right? we could change them to adapt to new channel names

jesmrec:
Some changes applied. Take another look pleae.

JuancaG05 (approved):
LGTM

Issue #4413, opened 2024-05-28
[FEATURE REQUEST] Open and create .url shortcut files


TheOneRing:
Please implement support for creating and opening .url files.
The feature should behave similarly to https://github.com/owncloud/ios-app/pull/1344 but follow the normal android UX.
### TASKS

- [x] Research (if needed)
- [x] Create branch feature/open_create_url_shortcut_files
- [x] Development tasks
  - [x] Create dialog fragment with url and name file inputs
  - [x] Upload the file to the server
  - [x] Click on file url and download
  - [x] Open alert dialog to navigate to browser
  - [x] Redirect to the browser
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/open_create_url_shortcut_files into master

jesmrec:
Some Android specific details:

- `Shortcuts` should be a new option in the FAB menu

- We need an icon, similar to the iOS one, for the `.url` files

- As first iteration, only two input fields: name (taking care of name conflicts) and url (with no validation, but error handling). Later, we can take advantage of deeplinks as iOS does (open to discussion)

quick info about .url files

https://fileinfo.com/extension/url
https://www.lyberty.com/encyc/articles/tech/dot_url_format_-_an_unofficial_guide.html

PR #4420, opened 2024-06-10
[FEATURE REQUEST] Open and create .url shortcut files

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4413

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.3/Shortcuts.md

Reports and improvements:

- [x] (1) Alignment with other platforms https://github.com/owncloud/android/pull/4420#issuecomment-2167753175 [FIXED]
- [x] (2) .url next to the edit field https://github.com/owncloud/android/pull/4420#issuecomment-2167760303 [FIXED]
- [x] (3) Limit to long names https://github.com/owncloud/android/pull/4420#issuecomment-2167813103 [FIXED]
- [x] (4) Default protocol for shortcut https://github.com/owncloud/android/pull/4420#issuecomment-2167841517 [FIXED]
- [x] (5) Icon in uploads view https://github.com/owncloud/android/pull/4420#issuecomment-2167859184 [FIXED]
- [x] (6) Long names handling https://github.com/owncloud/android/pull/4420#issuecomment-2167913068 [FIXED]
- [x] (7) use `blanks` instead of `spaces` in error message https://github.com/owncloud/android/pull/4420#issuecomment-2176019091 [FIXED]

### DIFF
--- a/changelog/unreleased/4420
+++ b/changelog/unreleased/4420
@@ -1,0 +1,6 @@
+Enhancement: Support for URL shortcut files
+
+A new option has been added in the FAB to create a shortcut file with a .url extension. When the file is clicked, the URL will open in the browser.
+
+https://github.com/owncloud/android/issues/4413
+https://github.com/owncloud/android/pull/4420
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -114,19 +114,6 @@ fun Activity.goToUrl(
     }
 }
 
-fun extractUrlFromFile(filePath: String): String? {
-    val file = File(filePath)
-    if (file.exists()) {
-        val lines = file.readLines()
-        for (line in lines) {
-            if (line.startsWith("URL=")) {
-                return line.substringAfter("URL=").trim('"')
-            }
-        }
-    }
-    return null
-}
-
 fun Activity.openPrivacyPolicy() {
     val urlPrivacyPolicy = getString(R.string.url_privacy_policy)
 
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -114,6 +114,19 @@ fun Activity.goToUrl(
     }
 }
 
+fun Activity.extractUrlFromFile(filePath: String): String? {
+    val file = File(filePath)
+    if (file.exists()) {
+        val lines = file.readLines()
+        for (line in lines) {
+            if (line.startsWith("URL=")) {
+                return line.substringAfter("URL=").trim('"')
+            }
+        }
+    }
+    return null
+}
+
 fun Activity.openPrivacyPolicy() {
     val urlPrivacyPolicy = getString(R.string.url_privacy_policy)
 
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -114,7 +114,7 @@ fun Activity.goToUrl(
     }
 }
 
-fun Activity.extractUrlFromFile(filePath: String): String? {
+fun extractUrlFromFile(filePath: String): String? {
     val file = File(filePath)
     if (file.exists()) {
         val lines = file.readLines()
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
@@ -24,7 +24,6 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.core.view.isVisible
 import androidx.core.widget.doOnTextChanged
 import androidx.fragment.app.DialogFragment
 import com.owncloud.android.R
@@ -40,7 +39,7 @@ class CreateShortcutDialogFragment : DialogFragment() {
     private var isValidFileName = false
     private var hasForbiddenCharacters = false
     private var hasMaxCharacters = false
-    private var hasSpace = false
+    private var hasEmptyValue = false
     private val binding get() = _binding!!
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         _binding = CreateShortcutDialogBinding.inflate(inflater, container, false)
@@ -86,9 +85,9 @@ class CreateShortcutDialogFragment : DialogFragment() {
         }
         binding.createShortcutDialogUrlValue.doOnTextChanged { urlValue, _, _, _ ->
             urlValue?.let {
-                hasSpace = urlValue.contains(" ")
-                isValidUrl = urlValue.isNotBlank() && !hasSpace
-                handleUrlRequirements(hasSpace)
+                hasEmptyValue = urlValue.contains(" ")
+                isValidUrl = urlValue.isNotBlank() && !hasEmptyValue
+                handleUrlRequirements(hasEmptyValue)
                 updateCreateShortcutButtonState()
             }
         }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
@@ -105,7 +105,7 @@ class CreateShortcutDialogFragment : DialogFragment() {
     private fun handleUrlRequirements(hasSpace: Boolean) {
         binding.createShortcutDialogUrlLayout.apply {
             if (hasSpace) {
-                error = getString(R.string.create_shortcut_dialog_url_error_no_spaces)
+                error = getString(R.string.create_shortcut_dialog_url_error_no_blanks)
             } else {
                 error = null
             }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
@@ -1,0 +1,155 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Aitor Ballesteros Pavón
+ *
+ * Copyright (C) 2024 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.presentation.files.createshortcut
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.core.view.isVisible
+import androidx.core.widget.doOnTextChanged
+import androidx.fragment.app.DialogFragment
+import com.owncloud.android.R
+import com.owncloud.android.databinding.CreateShortcutDialogBinding
+import com.owncloud.android.domain.files.model.OCFile
+
+class CreateShortcutDialogFragment : DialogFragment() {
+    private lateinit var parentFolder: OCFile
+    private lateinit var createShortcutListener: CreateShortcutListener
+    private var _binding: CreateShortcutDialogBinding? = null
+    private var isCreateShortcutButtonEnable = false
+    private var isValidUrl = false
+    private var isValidFileName = false
+    private var hasForbiddenCharacters = false
+    private var hasMaxCharacters = false
+    private var hasSpace = false
+    private val binding get() = _binding!!
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        _binding = CreateShortcutDialogBinding.inflate(inflater, container, false)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.apply {
+            handleInputsUrlAndFileName()
+            yesButton.setOnClickListener {
+                if (isCreateShortcutButtonEnable) {
+                    createShortcutListener.createShortcutFileFromApp(
+                        fileName = createShortcutDialogNameFileValue.text.toString(),
+                        url = formatUrl(createShortcutDialogUrlValue.text.toString()),
+                    )
+                    dialog?.dismiss()
+                }
+            }
+            cancelButton.setOnClickListener {
+                dialog?.dismiss()
+            }
+        }
+    }
+
+    private fun formatUrl(url: String): String {
+        var formattedUrl = url
+        if (!url.startsWith("http://") && !url.startsWith("https://")) {
+            formattedUrl = "http://$url"
+        }
+        return formattedUrl
+    }
+
+    private fun handleInputsUrlAndFileName() {
+        binding.createShortcutDialogNameFileValue.doOnTextChanged { fileNameValue, _, _, _ ->
+            fileNameValue?.let {
+                hasForbiddenCharacters = FORBIDDEN_CHARACTERS.any { fileNameValue.contains(it) }
+                hasMaxCharacters = fileNameValue.length >= MAX_FILE_NAME
+                isValidFileName = fileNameValue.isNotBlank() && !hasForbiddenCharacters && !hasMaxCharacters
+                handleNameRequirements(hasForbiddenCharacters, hasMaxCharacters)
+                updateCreateShortcutButtonState()
+            }
+        }
+        binding.createShortcutDialogUrlValue.doOnTextChanged { urlValue, _, _, _ ->
+            urlValue?.let {
+                hasSpace = urlValue.contains(" ")
+                isValidUrl = urlValue.isNotBlank() && !hasSpace
+                handleUrlRequirements(hasSpace)
+                updateCreateShortcutButtonState()
+            }
+        }
+    }
+
+    private fun updateCreateShortcutButtonState() {
+        isCreateShortcutButtonEnable = isValidFileName && isValidUrl
+        enableYesButton(isCreateShortcutButtonEnable)
+    }
+
+    private fun handleNameRequirements(hasForbiddenCharacters: Boolean, hasMaxCharacters: Boolean) {
+        binding.createShortcutDialogNameFileLayout.apply {
+            error = when {
+                hasMaxCharacters -> getString(R.string.uploader_upload_text_dialog_filename_error_length_max, MAX_FILE_NAME)
+                hasForbiddenCharacters -> getString(R.string.filename_forbidden_characters)
+                else -> null
+            }
+        }
+    }
+
+    private fun handleUrlRequirements(hasSpace: Boolean) {
+        binding.createShortcutDialogUrlLayout.apply {
+            if (hasSpace) {
+                error = getString(R.string.create_shortcut_dialog_url_error_no_spaces)
+            } else {
+                error = null
+            }
+        }
+    }
+
+    private fun enableYesButton(enable: Boolean) {
+        binding.yesButton.apply {
+            isEnabled = enable
+            setTextColor(
+                if (enable) resources.getColor(R.color.primary_button_background_color, null)
+                else resources.getColor(R.color.grey, null)
+            )
+        }
+    }
+
+    interface CreateShortcutListener {
+        fun createShortcutFileFromApp(fileName: String, url: String)
+    }
+
+    companion object {
+
+        private const val FORBIDDEN_CHARACTERS = "/\\"
+        private const val MAX_FILE_NAME = 256
+
+        /**
+         * Public factory method to create new CreateShortcutDialogFragment instances.
+         *
+         * @param parentFile File to create
+         * @return Dialog ready to show.
+         */
+        @JvmStatic
+        fun newInstance(parent: OCFile, listener: CreateShortcutListener): CreateShortcutDialogFragment {
+            return CreateShortcutDialogFragment().apply {
+                createShortcutListener = listener
+                parentFolder = parent
+            }
+        }
+    }
+}
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
@@ -34,13 +34,9 @@ class CreateShortcutDialogFragment : DialogFragment() {
     private lateinit var parentFolder: OCFile
     private lateinit var createShortcutListener: CreateShortcutListener
     private var _binding: CreateShortcutDialogBinding? = null
-    private var isCreateShortcutButtonEnable = false
-    private var isValidUrl = false
-    private var isValidFileName = false
-    private var hasForbiddenCharacters = false
-    private var hasMaxCharacters = false
-    private var hasEmptyValue = false
     private val binding get() = _binding!!
+    private var isCreateShortcutButtonEnabled = false
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         _binding = CreateShortcutDialogBinding.inflate(inflater, container, false)
         return binding.root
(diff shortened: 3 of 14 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+        private const val FORBIDDEN_CHARACTERS = "/\\"
+        private const val MAX_FILE_NAME = 256
+
+        /**
+         * Public factory method to create new CreateShortcutDialogFragment instances.
+         *
+         * @param parentFile File to create
+         * @return Dialog ready to show.
+         */
+        @JvmStatic
```
JuancaG05, line 146: Do you think this label is necessary?

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+    companion object {
+
+        private const val FORBIDDEN_CHARACTERS = "/\\"
+        private const val MAX_FILE_NAME = 256
+
+        /**
+         * Public factory method to create new CreateShortcutDialogFragment instances.
+         *
+         * @param parentFile File to create
+         * @return Dialog ready to show.
```
JuancaG05, line 144: If you are going to add docs, add them properly:
- The name of the `@param parentFile` is not the same as the actual parameter, which is `parent`. I would go with `parentFolder`
- You didn't mention the second parameter, `listener`

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+import androidx.fragment.app.DialogFragment
+import com.owncloud.android.R
+import com.owncloud.android.databinding.CreateShortcutDialogBinding
+import com.owncloud.android.domain.files.model.OCFile
+
+class CreateShortcutDialogFragment : DialogFragment() {
+    private lateinit var parentFolder: OCFile
+    private lateinit var createShortcutListener: CreateShortcutListener
+    private var _binding: CreateShortcutDialogBinding? = null
+    private var isCreateShortcutButtonEnable = false
```
JuancaG05, line 37:
```suggestion
    private var isCreateShortcutButtonEnabled = false
```

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+    private var isValidUrl = false
+    private var isValidFileName = false
+    private var hasForbiddenCharacters = false
+    private var hasMaxCharacters = false
+    private var hasEmptyValue = false
+    private val binding get() = _binding!!
```
JuancaG05, line 43: Better place this one next to `_binding`, so that everything related is together

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+        _binding = CreateShortcutDialogBinding.inflate(inflater, container, false)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.apply {
+            handleInputsUrlAndFileName()
+            yesButton.setOnClickListener {
+                if (isCreateShortcutButtonEnable) {
```
JuancaG05, line 54: At this point, this condition is always going to be false, right? `isCreateShortcutButtonEnable` is set to false by default and this lifecycle method is one of the first ones in being executed

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+        binding.createShortcutDialogUrlLayout.apply {
+            if (hasSpace) {
+                error = getString(R.string.create_shortcut_dialog_url_error_no_spaces)
+            } else {
+                error = null
+            }
+        }
+    }
+
+    private fun enableYesButton(enable: Boolean) {
```
JuancaG05, line 121: This method should set the listener as well when `enable` is true, or set it to `null` when `enable` is false

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+class CreateShortcutDialogFragment : DialogFragment() {
+    private lateinit var parentFolder: OCFile
+    private lateinit var createShortcutListener: CreateShortcutListener
+    private var _binding: CreateShortcutDialogBinding? = null
+    private var isCreateShortcutButtonEnable = false
+    private var isValidUrl = false
+    private var isValidFileName = false
+    private var hasForbiddenCharacters = false
+    private var hasMaxCharacters = false
+    private var hasEmptyValue = false
```
JuancaG05, line 42: Do we really need all of these variables to be class properties? Couldn't they be local variables in the scope of the methods where they are needed?

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+            cancelButton.setOnClickListener {
+                dialog?.dismiss()
+            }
+        }
+    }
+
+    private fun formatUrl(url: String): String {
+        var formattedUrl = url
+        if (!url.startsWith("http://") && !url.startsWith("https://")) {
+            formattedUrl = "http://$url"
```
JuancaG05, line 71: I would format it by default with `https`, not with `http`

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    interface CreateShortcutListener {
+        fun createShortcutFileFromApp(fileName: String, url: String)
+    }
+
+    companion object {
+
+        private const val FORBIDDEN_CHARACTERS = "/\\"
+        private const val MAX_FILE_NAME = 256
```
JuancaG05, line 138: Which is the source of this? Maybe we can use these constants from another class if they are already defined, to keep consistency

owncloudApp/src/main/res/layout/create_shortcut_dialog.xml
```diff
@@ -0,0 +1,115 @@
...
+                    android:layout_height="wrap_content"
+                    android:layout_weight="1"
+                    style="@style/Button.Borderless"
+                    android:text="@string/share_cancel_public_link_button" />
+
+                <androidx.appcompat.widget.AppCompatButton
+                    android:id="@+id/yesButton"
+                    android:enabled="false"
+                    android:textColor="@color/grey"
+                    style="?android:attr/borderlessButtonStyle"
```
JuancaG05, line 105: The style of this and the previous button is not the same, but they look the same. Let's unify it

owncloudApp/src/main/res/layout/create_shortcut_dialog.xml
```diff
@@ -0,0 +1,115 @@
...
+                android:layout_marginTop="@dimen/standard_margin"
+                android:orientation="horizontal">
+
+                <androidx.appcompat.widget.AppCompatButton
+                    android:id="@+id/cancelButton"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:layout_weight="1"
+                    style="@style/Button.Borderless"
+                    android:text="@string/share_cancel_public_link_button" />
```
JuancaG05, line 99: Isn't there any other String for "Cancel" that is not the public link button String? Maybe an Android default one?

owncloudApp/src/main/res/layout/create_shortcut_dialog.xml
```diff
@@ -0,0 +1,115 @@
...
+        android:padding="@dimen/standard_half_padding"
+        android:text="@string/create_shortcut_dialog_title"
+        android:textAppearance="@android:style/TextAppearance.Material.DialogWindowTitle" />
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:orientation="vertical">
+
+            <RelativeLayout
```
JuancaG05, line 21: Well, `RelativeLayout` is not recommended anymore, it's better to use `ConstraintLayout`. But you can leave it like this for this case, just take it into account for the future

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1361,6 +1365,37 @@ class FileDisplayActivity : FileActivity(),
...
+                view.dismiss()
+            }
+            .setCancelable(true)
+            .create()
+        dialog.show()
+    }
+
+    private fun String.truncateWithEllipsis(maxLength: Int): String {
+        return if (this.length > maxLength) {
+            "${this.substring(0, maxLength)}..."
```
JuancaG05, line 1393: I don't know if in this case it would be better to ellipsize the URL at the end or in the middle, let's take a look at similar apps and see how they display it

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -304,6 +304,8 @@ class FileListAdapter(
             if (file.isFolder) {
                 // Folder
                 fileIcon.setImageResource(R.drawable.ic_menu_archive)
+            } else if (file.mimeType == "text/uri-list") {
```
JuancaG05, line 307: Better to use a constant as you did in `FileDisplayActivity` to keep consistency

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -836,4 +839,11 @@
     <string name="content_description_delete_share">Delete share</string>
     <string name="content_description_file_operations">%1$s operations</string>
 
+    <string name="create_shortcut_dialog_title">Create new shortcut</string>
+    <string name="create_shortcut_dialog_url">URL</string>
+    <string name="create_shortcut_dialog_url_information">The URL must start with https:// or http:// so that it can be opened in the browser.</string>
```
JuancaG05, line 844:
```suggestion
    <string name="create_shortcut_dialog_url_information">The URL must start with https:// or http:// so that it can be opened in the browser</string>
```

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -836,4 +839,11 @@
...
+    <string name="create_shortcut_dialog_url_error_no_spaces">The URL cannot have spaces</string>
```
JuancaG05, line 845:
```suggestion
    <string name="create_shortcut_dialog_url_error_no_spaces">The URL cannot contain spaces</string>
```
We should be careful with the concept "spaces" due to the starring oCIS feature 😄. Let's be as specific as possible when using this term

JuancaG05 (changes requested):
Some questions and changes requested here @Aitorbp!

owncloudApp/src/main/res/values/strings.xml
Aitorbp: Done

owncloudApp/src/main/res/values/strings.xml
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
Aitorbp: Done

owncloudApp/src/main/res/layout/create_shortcut_dialog.xml
Aitorbp: I haven´t seen another one in strings resources

owncloudApp/src/main/res/layout/create_shortcut_dialog.xml
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: I format it by default with `http` because if the websites do not support `https`, the websites will say: `The site can't be reached`. For example this one: `http://www.edu4java.com/es/web/web30.html`.
On the contrary, if we write a website with `http`, but it supports `https`, it will redirect us directly to `https`.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -114,6 +114,19 @@ fun Activity.goToUrl(
     }
 }
 
+fun Activity.extractUrlFromFile(filePath: String): String? {
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Removed

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Removed the condition.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
Aitorbp: Changed `parent` variable to `parentFolder`.

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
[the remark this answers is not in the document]
Aitorbp: Removed Activity

changelog/unreleased/4420
```diff
@@ -0,0 +1,6 @@
+Enhancement: New option to create and open a shortcut file
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
JuancaG05, line 1393: What about this, any news?

JuancaG05 (changes requested):
Some more changes here @Aitorbp. Pay attention to the comments in previous CR, you skipped some of them

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    interface CreateShortcutListener {
+        fun createShortcutFileFromApp(fileName: String, url: String)
+    }
+
+    companion object {
+
+        @JvmStatic
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/files/createshortcut/CreateShortcutDialogFragment.kt
```diff
@@ -0,0 +1,150 @@
...
+            isEnabled = enable
+            setTextColor(
+                if (enable) {
+                    setOnClickListener {
+                        createShortcutListener.createShortcutFileFromApp(
+                            fileName = binding.createShortcutDialogNameFileValue.text.toString(),
+                            url = formatUrl(binding.createShortcutDialogUrlValue.text.toString()),
+                        )
+                        dialog?.dismiss()
+                    }
```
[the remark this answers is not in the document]
Aitorbp: Fixed

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/FileListAdapter.kt
```diff
@@ -60,6 +60,7 @@ class FileListAdapter(
     var files = mutableListOf<Any>()
     private var account: Account? = AccountUtils.getCurrentOwnCloudAccount(context)
     private var fileListOption: FileListOption = FileListOption.ALL_FILES
+    private val MIMETYPE_TEXT_URI_LIST = "text/uri-list"
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
Aitorbp: I have removed the truncateWithEllipsis function and created two textviews inside a LinearLayout in the AlertDialog. The ellipsize is truncated to the center.

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -114,6 +114,19 @@ fun Activity.goToUrl(
...
+    if (file.exists()) {
+        val lines = file.readLines()
+        for (line in lines) {
+            if (line.startsWith("URL=")) {
+                return line.substringAfter("URL=").trim('"')
+            }
+        }
+    }
+    return null
+}
```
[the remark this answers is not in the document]
Aitorbp: Moved to FileDisplayActivity

JuancaG05 (approved):
LGTM now!

jesmrec:
### (1)

This is our current implementation together with iOS and web:

|Android|iOS|Web|
|:-:|:-:|:-:|
|Screenshot 2024-06-14 at 12 39 14|Screenshot 2024-06-14 at 12 36 03|Screenshot 2024-06-14 at 12 36 25|

- iOS and web put the URL field above, and the shortcut name below, just the opposite as we do. We should align behaviour across platforms.

- iOS and web call the submit button as `Create` or `Create shortcut`, whereas we call it `YES` (i guess, this is an Android default). We should go for `Create` as well.

Pixel 2, Android11
`87d3a1d2`

jesmrec:
### (2)

As improvement, i'd be cool to have the file name in web-wise:

Screenshot 2024-06-14 at 12 44 57

the `.url` at the end makes clear that the file name is the requested field, and the extension must not be typed by the user. 

Pixel 2, Android11
`87d3a1d2`

jesmrec:
### (3)

Tiny detail for long names:

error message is: `File name must not be longer than 223 characters`

By typing exactly 223 characters, the message is there. It shouldn't or changing the value to 222

Pixel 2, Android11
`87d3a1d2`

jesmrec:
### (4)

About this comment:

```
I format it by default with http because if the websites do not support https, the websites will say: The site can't be reached. For example this one: http://www.edu4java.com/es/web/web30.html.
On the contrary, if we write a website with http, but it supports https, it will redirect us directly to https.
```

but, what happens if it supports `http` and not `https`?  you'll get the error as well. You can check this with a basic oC10 server. The fact of supporting `http`, `https` or both is a matter of the hosting server, each one has its own setup and we can not assume an specific behaviour based in only one example.

Both web and iOS use `https` as default. I'd go for this.

jesmrec:
### (6)

I'm not pretty sure to ellipsize the URL when it is very long. The URL to open is sensible information that is hidden to the user. Let me know if i'm wrong, but it is being ellipsized if the length is longer than 1 line. That means, almost every URL that points to a subfolder/subdomain is going to be ellipsized.

I'd remove that, or expand to a longer size

Pixel 2, Android11
`87d3a1d2`

Aitorbp:
> ### (6)
> I'm not pretty sure to ellipsize the URL when it is very long. The URL to open is sensible information that is hidden to the user. Let me know if i'm wrong, but it is being ellipsized if the length is longer than 1 line. That means, almost every URL that points to a subfolder/subdomain is going to be ellipsized.
> 
> I'd remove that, or expand to a longer size
> 
> Pixel 2, Android11 `87d3a1d2`

Regarding the ellipsize of the URL when it is very long, the following occurs: if we want to put a maximum of three lines and set a truncated ellipsize in the middle, it will not work. Android does not have this functionality automatically, it would have to be programmed. Therefore the following options are considered:

- Set the ellipsize at the end of the text, keeping a maximum of three lines.
- Set the ellipsize in the middle with a maximum of three lines, making a method programmatically. A calculation would have to be made to put the ellipsize in the center of the second line. I don't know how this would work...
- Set a `HorizontalScrollLayout` so that it scrolls if the text exceeds the maximum of one line. With this component it only detects one line, I understand that it could also be done programmatically.
- The last option would be to leave it as iOS has it, without any type of restriction and that the URL occupies all it needs.

@jesmrec @JuancaG05

JuancaG05:
> * Set a `HorizontalScrollLayout` so that it scrolls if the text exceeds the maximum of one line. With this component it only detects one line, I understand that it could also be done programmatically so that it detects more lines.

I would discard this option directly. A scroll in a dialog is no good practice.

> * The last option would be to leave it as iOS has it, without any type of restriction and that the URL occupies all it needs.

This is not a good option for me. What if the URL is so so long (not strange case currently) that the dialog gets too big and we cannot see its buttons? We need to set a limit somewhere

Why is it that you can ellipsize at the end a text of more than 1 line but not in the middle?

jesmrec:
> This is not a good option for me. What if the URL is so so long (not strange case currently) that the dialog gets too big and we cannot see its buttons? We need to set a limit somewhere

Does not the dialog (including the buttons) expand? i mean, if the text is too long, the buttons are on the bottom in the same way, just the central is expanded.

JuancaG05:
> Does not the dialog (including the buttons) expand? i mean, if the text is too long, the buttons are on the bottom in the same way, just the central is expanded.

Yes, the center will expand, but... what if it expands so much that the buttons get out of the screen?

Aitorbp:
To sum up, to align with iOS behavior, we're going to leave the url unrestricted to the maximum number of lines, but add a scrollview so that when the text takes up more space than the screen, the url element will scroll.

Aitorbp:
All reports are ready to be tested again @jesmrec

jesmrec:
everything is fixed. Now, i will face a 2nd QA round.

jesmrec:
### (7)

Minor change

Typing a URL with blanks, the error `The URL cannot contain spaces` is displayed.

This is correct, but i'd use `blanks` instead of `spaces`, since "spaces" has a different meaning... it may lead to confusion

Aitorbp:
> The URL cannot contain space

Changed, ready to test @jesmrec

jesmrec:
Related https://github.com/owncloud/web/issues/11063

I noticed that the shorcuts created in the web client (oCIS) are not open in mobile clients if they lack of protocol prefix. If i'm not wrong, the current approach in Android app is adding the default protocol prefix (https) when the shortcut is open (not in creation), right @Aitorbp ?

So, we have two options here:

- Getting the current approach as correct for the moment and going ahead. If the prefix default protocol should be added in creation phase as before, it could be changed in a new PR

- Wait till a solution across the clients is agreed in the web issue mentioned above.

which one do you like the most?

Aitorbp:
> Related [owncloud/web#11063](https://github.com/owncloud/web/issues/11063)
> 
> I noticed that the shorcuts created in the web client (oCIS) are not open in mobile clients if they lack of protocol prefix. If i'm not wrong, the current approach in Android app is adding the default protocol prefix (https) when the shortcut is open (not in creation), right @Aitorbp ?
> 
> So, we have two options here:
> 
> * Getting the current approach as correct for the moment and going ahead. If the prefix default protocol should be added in creation phase as before, it could be changed in a new PR
> * Wait till a solution across the clients is agreed in the web issue mentioned above.
> 
> which one do you like the most?

Right we add the default protocol prefix when the shortcut is open. In my opinion, I would go for the current approach, even if the web were to set the protocol, the application would not be affected.

jesmrec:
https://github.com/owncloud/web/issues/11063#issuecomment-2178859637

Aitorbp:
Added back the protocol in shortcut creation.

jesmrec:
That's fixed, now adding the protocol in both creation and opening.

jesmrec:
Approved, ready to go! 💯

PR #4433, opened 2024-07-03
[a11y] 11.1.4.11 Non-text contrast

Aitorbp:
## Related Issues
App:

_____

## QA

QA checks:

https://github.com/owncloud/android/pull/4433#issuecomment-2233195218

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -3,6 +3,7 @@
  *
  * @author Juan Carlos Garrote Gascón
  * @author Jorge Aguado Recio
+ * @author Aitor Ballesteros Pavón
  *
  * Copyright (C) 2024 ownCloud GmbH.
  *
@@ -58,7 +59,6 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
 
     private lateinit var spacesListAdapter: SpacesListAdapter
 
-
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
@@ -84,6 +84,7 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
         binding.swipeRefreshSpacesList.setOnRefreshListener {
             spacesListViewModel.refreshSpacesFromServer()
         }
+        setTextHintRootToolbar()
     }
 
     private fun subscribeToViewModels() {
@@ -131,6 +132,7 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
         super.onCreateOptionsMenu(menu, inflater)
         (menu.findItem(R.id.action_search).actionView as SearchView).run {
             setOnQueryTextListener(this@SpacesListFragment)
+            queryHint = resources.getString(R.string.actionbar_search_space)
         }
         menu.findItem(R.id.action_share_current_folder)?.itemId?.let { menu.removeItem(it) }
     }
@@ -144,6 +146,11 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
 
     fun setSearchListener(searchView: SearchView) {
         searchView.setOnQueryTextListener(this)
+    }
+
+    private fun setTextHintRootToolbar() {
+        val searchViewRootToolbar = requireActivity().findViewById<SearchView>(R.id.root_toolbar_search_view)
+        searchViewRootToolbar.queryHint = getString(R.string.actionbar_search_space)
     }
 
     companion object {
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -130,8 +130,6 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         super.onCreateOptionsMenu(menu, inflater)
         (menu.findItem(R.id.action_search).actionView as SearchView).run {
-            maxWidth = Int.MAX_VALUE
-            queryHint = resources.getString(R.string.actionbar_search)
             setOnQueryTextListener(this@SpacesListFragment)
         }
         menu.findItem(R.id.action_share_current_folder)?.itemId?.let { menu.removeItem(it) }
--- a/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -44,15 +44,12 @@ import android.util.TypedValue
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
-import android.widget.EditText
 import android.widget.FrameLayout
-import android.widget.ImageView
 import android.widget.LinearLayout
 import android.widget.ScrollView
 import android.widget.TextView
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.appcompat.app.AlertDialog
-import androidx.appcompat.widget.SearchView
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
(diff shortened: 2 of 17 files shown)


Aitorbp:
Changes to SearchView have been made on the following screens:
- Search file main list
- Search space in Space view
- Search space in Select folder
- Picture upload path -> Choose destination folder
- Video upload path -> Choose destination folder
- Receive external files  --> choose destination folder

No changes will be made to the Users and group SearchView in the Share view.

Now, the searchbar includes a new resource with rounded edges, using the same background color (brandable) as the containing toolbar: `rounded_search_view.xml`

The new color `#B4BAC0` of the hint text and the search cross meets the contrast requirements of the issue:

Captura de pantalla 2024-07-03 a las 14 10 33

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -130,8 +130,6 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         super.onCreateOptionsMenu(menu, inflater)
         (menu.findItem(R.id.action_search).actionView as SearchView).run {
-            maxWidth = Int.MAX_VALUE
-            queryHint = resources.getString(R.string.actionbar_search)
```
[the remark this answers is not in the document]
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -552,6 +555,19 @@ class FileDisplayActivity : FileActivity(),
 
         inflater.inflate(R.menu.main_menu, menu)
 
+        (menu.findItem(R.id.action_search).actionView as SearchView).run {
+            maxWidth = Int.MAX_VALUE
+            queryHint = resources.getString(R.string.actionbar_search)
```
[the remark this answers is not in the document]
Aitorbp: I added actionbar_search in the `MainFileListFragment` and actionbar_search_space in the `SpacesListFragment`

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -552,6 +555,19 @@ class FileDisplayActivity : FileActivity(),
...
+
+            val textHint = findViewById<EditText>(androidx.appcompat.R.id.search_src_text)
```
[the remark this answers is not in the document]
Aitorbp: Changed

owncloudApp/src/main/res/layout/owncloud_toolbar.xml
```diff
@@ -48,6 +48,7 @@
             app:layout_constraintStart_toEndOf="@id/root_toolbar_left_icon"
             app:layout_constraintTop_toTopOf="parent"
             app:queryHint="@string/actionbar_search"
+            android:background="@drawable/rounded_search_view"
```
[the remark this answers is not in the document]
Aitorbp: I've kept it here for the `rootToolbar`, it's done programmatically for the `standartToolbar`, which can't be done via xml.

owncloudApp/src/main/res/values/setup.xml
```diff
@@ -46,6 +46,7 @@
     <color name="secondary_button_text_color">@color/black</color>
     <color name="drawer_header_color">@color/owncloud_blue_accent</color>
     <color name="spaces_card_background_color">#edf3fa</color>
+    <color name="search_view_background_color">@color/owncloud_blue</color>
```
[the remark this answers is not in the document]
Aitorbp: Changed

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
```diff
@@ -97,6 +101,10 @@ abstract class ToolbarActivity : BaseActivity() {
                 toolbarTitle.visibility = VISIBLE
                 false
             }
+            val textSearchView = findViewById<EditText>(androidx.appcompat.R.id.search_src_text)
+            val closeButton = searchView.findViewById<ImageView>(androidx.appcompat.R.id.search_close_btn)
```
JuancaG05, line 105: If this is inside the `apply` block, we don't need reference to `searchView`

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
```diff
@@ -137,4 +145,21 @@ abstract class ToolbarActivity : BaseActivity() {
...
+
+    override fun onPrepareOptionsMenu(menu: Menu): Boolean {
+        (menu.findItem(R.id.action_search).actionView as SearchView).run {
+            val searchText = findViewById<EditText>(androidx.appcompat.R.id.search_src_text)
+            val closeButton = findViewById<ImageView>(androidx.appcompat.R.id.search_close_btn)
+            val searchButton = findViewById<ImageView>(androidx.appcompat.R.id.search_button)
+
+            maxWidth = Int.MAX_VALUE
+
+            searchButton.setBackgroundColor(getColor(R.color.actionbar_start_color))
```
JuancaG05, line 157: Why is this needed just for the standard toolbar and not the root toolbar? Or maybe it is not needed at all?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
```diff
@@ -137,4 +145,21 @@ abstract class ToolbarActivity : BaseActivity() {
     private fun getRootToolbar(): ConstraintLayout = findViewById(R.id.root_toolbar)
 
     private fun getStandardToolbar(): Toolbar = findViewById(R.id.standard_toolbar)
+
+    override fun onPrepareOptionsMenu(menu: Menu): Boolean {
```
JuancaG05, line 149: This should be done better in `onCreateOptionsMenu` or you will have problems with classes that extend this one and use that method

JuancaG05 (changes requested):
Some more changes requested here @Aitorbp

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
Aitorbp: In this case, the the background color of the searchButton get the color of the background SearchView, so we have to set the color of the toolbar.

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
Aitorbp: Changed to `onCreateOptionsMenu`

owncloudApp/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.kt
Aitorbp: Done

JuancaG05 (approved):
LGTM now

Aitorbp:
Research:
First of all, we need to know that there are two toolbars currently working in the application: `RootToolbar` and `StandardToolbar`. 

The `RootToolbar` is displayed in the following views: 
- `Root List File`
- `Spaces View`
- `Upload View`
- `Shares View`
- `Available offline View`

The `Standard toolbar` appears in the following views: 
- Set target path for automatic uploads
- List files (not root)
- Receiving external files

In this PR we are trying to unify the behaviour of the toolbar to the `onCreateOptionsMenu` method in `ToolbarActivity`. We are moving the general behaviour of the toolbar from the child classes to the `ToolbarActivity`, removing the useless code in the child classes. So everything related to colors is now handled in `ToolbarActivity` and the corner cases in its classes override the `onCreateOptionsMenu` function.

This way, the following classes don't need to display the main_menu, so we override this method to return `false` (won't be displayed):
- `UploadListActivity`
- `ShareActivity`
- `PreviewImageActivity`
- `PreviewVideoActivity`
We need to do this now because these classes extend from `ToolbarActivity` and `main_menu` is attached to it. So we have to do this behaviour.

On the other hand we have fixed a bug. Now when we are in the spaces view, the text hint of the SearchView is `Search spaces`.

jesmrec:
About (1), i notice that the regression behaviour is that one. The point is, if the new bubble makes it a bit weird. With the same color it looks good, with two colors... maybe not. Open to discussion.

jesmrec:
There is a non-regressive problem in the search bar: it stops working after changing device orientation. It will be addressed in a separate issue

jesmrec:
Problem with filtering after rotating: https://github.com/owncloud/android/issues/4441
Problem with bubble taking place in the top bar:

jesmrec:
About the search hint color

<color name="search_view_hint_text">#BDBDBD</color>

That default color was added to the color's list, but is not brandable for the search hint. Is that right? i'd go for the first approach, but we should not forget that the bubble background and the top bar are brandable.

Aitorbp:
> About the search hint color
> 
> `<color name="search_view_hint_text">#BDBDBD</color>`
> 
> That default color was added to the color's list, but is not brandable for the search hint. Is that right? i'd go for the first approach, but we should not forget that the bubble background and the top bar are brandable.

We've left it in the `search_view_background_color`. There's no reason to make the color of the text hitn brandable as well. We could do that, so we'd give customers more freedom to play with color contrasts. But by adjusting the `search_view_background_color` they could do it.

jesmrec:
Approved on my side.

A problem was detected, as no regression was moved to another place: https://github.com/owncloud/android/issues/4441

Issue #4432, opened 2024-07-01
[BUG] Shares in non-root are not displayed


jesmrec:
### Steps to reproduce

1. In any item that is not in root folder, create a public link 

### Actual behaviour

Item is not listed in the shares view. By closing the shares view and opening again, it is listed

https://github.com/owncloud/android/assets/14894746/aca5f28f-b653-44fc-a9e5-234dde27bed1

### Expected behaviour

Item is listed just after being created.

NOTE: same problem with edition, and with "share with users and groups"

Could be related: https://github.com/owncloud/android/issues/3016

Pixel 2, Android 11

parneet-guraya:
Can I work on it :hand:

JuancaG05:
Hi @parneet-guraya! 👋  Nice to see you again!

Of course! Go for it 🚀

parneet-guraya:
Hi @JuancaG05 I pushed the PR  #4435 have a look.

JuancaG05:
Hi @parneet-guraya! Cool! I'll have a look whenever I can, since right now we have other priorities, but I hope I can do it ASAP 😀

Issue #3016, opened 2020-11-10
Links in non-root are not refreshed in Share by Link view


jesmrec:
This one comes form https://github.com/owncloud/android/pull/3007#issuecomment-719426005

If any link of an item in non-root is removed from the `Share by Link` view, such view is not refreshed and the removed link stays there. 

The reason is that the view is a filter from file list and needs to be refreshed to show the updated information. The problem can be fixed when the database migration is merged into `master`

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

### PR

- App
- Library (if needed)

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent — closing as resolved.

This was reported in 2020 against the old `Share by Link` view, which was a non-reactive *filter over the file list* and so did not refresh when a link on a non-root item was removed (the issue itself noted it could only be fixed "when the database migration is merged into `master`").

That migration landed long ago, and the view has since been rewritten. The current implementation (`MainFileListViewModel` → `retrieveFlowForShareByLink` → `GetSharedByLinkForAccountAsStreamUseCase`) is backed by a reactive Room stream: removing a public link updates the database, and the change propagates through the `Flow` so the `Share by Link` list updates automatically — for items at any depth, not just root. The original "view is a stale filter" cause no longer exists.

Closing as outdated/fixed. If you can still reproduce a missing refresh on a current ownCloud Android release, please **reopen** (or open a fresh issue) with the version and exact steps — thank you\!

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4435, opened 2024-07-10
[BUG] Shares in non-root are not displayed

parneet-guraya:
Fixes: #4432 

Demo -->

https://github.com/owncloud/android/assets/111801812/623c9c3d-32bd-4465-bb45-3669f8d28af9

The reason for this issue was whenever we create/edit a link. After `POST` request we make a `GET` call to retrieve the share info because it was mentioned [here](https://github.com/owncloud/android/blob/bba83debbb09160065a39fdf67961bd334885687/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.kt#L131) that the POST request does not return the response that we need.

But the `GET` request is returning wrong path i.e for a file that exists at `/Folder/file` it returns the path `/file` . So, it always returns a path that a file exists in the root even though it is inside some folder. 

Screenshot from 2024-07-10 13-29-27

Above screenshot shows two responses. First one is the `POST` which has path /Fol/file and the second one is `GET` response which has path /file (which is incorrect)

Screenshot from 2024-07-10 13-30-00

**Also the incorrect response from `GET` lead to duplicate entries in DB -->**
- First entry is stored from the response with path as /file(incorrect)
- Second entry is stored when the `ShareView` is refreshed and upon sync (from remote), instead of getting replaced the new entry gets added because there's path ##difference.

**Solution -->**

Turns out `POST` request does return same response as `GET` does and also with correct `filePath`. So, now the need for `GET` doesn't exist.


parneet-guraya:
@jesmrec does it also fix #3016?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareOperation.kt
Aitorbp, line 1: Why did you delete this class? What was its function and why is it not necessary? @parneet-guraya

parneet-guraya: Initially we do `POST` request to create/update link. Then a `GET` request to get the additional info. The reason for making `GET` request was because (mentioned in code's [comment](https://github.com/owncloud/android/blob/bba83debbb09160065a39fdf67961bd334885687/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/CreateRemoteShareOperation.kt#L131)) the `POST` request does not return the info that we need. 

There are two reasons to remove this file:

- Turns out `POST` request does return all the info we need same as `GET`. So, no reason to call another `GET` request. Hence, removed the file.

- Also `GET` request returns the `filePath` as if the file exists in the root. Meaning even if a file exist inside a folder, it still would show as if it's in the root.

jesmrec:
> @jesmrec does it also fix https://github.com/owncloud/android/issues/3016?

i don't think so, are different things. The current one affects the list in the shares view, and the #3016 affects the list of files. Anyway, i will take a look in QA phase.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/shares/GetRemoteShareOperation.kt
parneet-guraya: Okay, I wondered the same myself as I said my only clue is that the second call is for purpose of getting additional info that according to the comment wasn't provided by the first request we do. I tried looking back in the history it was a part of very huge change so couldn't track it back further. 

But, as you can see in my PR's comment we're getting the same response as the second request does, in fact the second one returns the wrong path. So, to answer your question I think it could be necessary to make a second request in the past that the `POST` req doesn't give us the info but at sometime later it was added into the API to return the info too? 

Let me know if this clarify your doubt or you have any questions.

Aitorbp: Ok, I think the same like you

Aitorbp (commented):
For my part the work is fine. Only one thing is missing, could you add the calens file? Here you have some information to do it: https://github.com/owncloud/android/blob/master/changelog/README.md

parneet-guraya:
@Aitorbp , Added the changelog :+1:

changelog/unreleased/4435
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix the Shares screen not updating.
```
Aitorbp, line 1: I would change the title: `Shares in non-root are updated correctly`
In the title we don't need the final point.

changelog/unreleased/4435
```diff
@@ -0,0 +1,6 @@
...
+
+Fixed the Shares screen not updating instantly when create/edit a link.
```
Aitorbp, line 3: And here I would write: 
`The items of the "Share" view are updated instantly when create/edit a link or share with users or groups in a non-root file.`

changelog/unreleased/4435
```diff
@@ -0,0 +1,6 @@
+Bugfix: Shares in non-root are updated correctly.
```
Aitorbp, line 1: Please, remove the final point `.`

Aitorbp (changes requested):

Aitorbp (approved):
@parneet-guraya very good job! I will move to QA!

jesmrec:
Let's QA this

jesmrec:
@parneet-guraya first, thanks for the contribution.

I noticed that the fix applies to the public links, but for sharing with users and groups (also reproducible and reported in the issue as a note). Could you take a look?

parneet-guraya:
> I noticed that the fix applies to the public links, but for sharing with users and groups (also reproducible and reported in the issue as a note). Could you take a look?

@jesmrec , you're right. However, I'm not able to add any shares, it seems the demo account does not have privileges? 

Screenshot from 2024-07-24 02-47-24

This is what I get as a response (even on master) but I'm able to do the same from web just fine. Could you give it a look?

jesmrec:
@parneet-guraya this should not be a matter of permissions because web allows to share. Something is not correct in the app there. I will try to check.

jesmrec:
@parneet-guraya which server instance are you using to check it?

jesmrec:
I was able to reproduce the problem with `ocis.ocis.master.owncloud.works` but not with the oCIS stable 5.0.6. I created an issue in oCIS repo: https://github.com/owncloud/ocis/issues/9681 because the problem is also reproducible with iOS client. Web client works fine. In order not to block the current issue, i'd recommend to use an instance with stable 5.0.6

parneet-guraya:
Hi @jesmrec, I'm using demo instance at `demo.owncloud.com` this could be running on latest master? 

So, to run different version i.e 5.0.6, I guess I have to run it locally?

jesmrec:
You can use this instance with the stable version: `ocis.ocis.production.owncloud.works` (i just found it). With same usernames and pwds.

parneet-guraya:
@jesmrec, Would you mind directing me to the resource to how to do it?

jesmrec:
> @jesmrec, Would you mind directing me to the resource to how to do it?

Sorry, what do you mean? you can access [https://ocis.ocis.production.owncloud.works](https://ocis.ocis.production.owncloud.works) just with the URL, as any other instance. Is that right? if not, let me know.

parneet-guraya:
Okay, I get it now. Thanks :+1:

jesmrec:
@parneet-guraya were you able to finally reproduce the "share with users" problem? let us know if you need some extra help or clarification. thanks!!

parneet-guraya:
> @parneet-guraya were you able to finally reproduce the "share with users" problem? let us know if you need some extra help or clarification. thanks!!

Hi, turns out the `POST` request that we make to share with users and groups returns the wrong path (as if the file always is in root folder). This is same as we create public shares where first a `POST` followed by a `GET` request was made but in that case the second request was returning the wrong path so I removed the request because first request was enough according to the data we need.

But, now in this case the only request that we make to share, returns the wrong path. So, it is a backend issue. I read from the issue created by @jesmrec [comment](https://github.com/owncloud/ocis/issues/9681#issuecomment-2247831945) that mobile clients are using deprecated API so could it be the bug exists in the deprecated api?

A quick fix could be just to make a refresh call from remote after doing the `POST` request for adding the share. Which triggers when a dialog appears for adding new shares and that is the reason we start seeing the items that were not showing earlier. 

Do let me know what you think? @JuancaG05 @Aitorbp

jesmrec:
> But, now in this case the only request that we make to share, returns the wrong path. So, it is a backend issue. I read from the issue created by @jesmrec https://github.com/owncloud/ocis/issues/9681#issuecomment-2247831945 that mobile clients are using deprecated API so could it be the bug exists in the deprecated api?

Don't worry about this, ocs is not deprecrated yet. Problem in the oCIS issue finally comes from other side.

Aitorbp:
> > @parneet-guraya were you able to finally reproduce the "share with users" problem? let us know if you need some extra help or clarification. thanks!!
> 
> Hi, turns out the `POST` request that we make to share with users and groups returns the wrong path (as if the file always is in root folder). This is same as we create public shares where first a `POST` followed by a `GET` request was made but in that case the second request was returning the wrong path so I removed the request because first request was enough according to the data we need.
> 
> But, now in this case the only request that we make to share, returns the wrong path. So, it is a backend issue. I read from the issue created by @jesmrec [comment](https://github.com/owncloud/ocis/issues/9681#issuecomment-2247831945) that mobile clients are using deprecated API so could it be the bug exists in the deprecated api?
> 
> A quick fix could be just to make a refresh call from remote after doing the `POST` request for adding the share. Which triggers when a dialog appears for adding new shares and that is the reason we start seeing the items that were not showing earlier.
> 
> Do let me know what you think? @JuancaG05 @Aitorbp

Hello @parneet-guraya , very good research. As you know, we are using a deprecated API and we don't know if someone could change something in this API. So maybe the fast way at the moment is try to do to refresh as you comment. So, yes, go for it 👍

parneet-guraya:
Hi, pushed the fix :+1:

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
```diff
@@ -154,7 +154,10 @@ class ShareActivity : FileActivity(), ShareFragmentListener {
                     is UIResult.Loading -> {
                         showLoadingDialog(R.string.common_loading)
                     }
-                    is UIResult.Success -> {}
+                    is UIResult.Success -> {
+                        // needs to trigger refresh after creation because creation request returns wrong path [issue: https://github.com/owncloud/ocis/issues/9746] resulting list not getting updated.
```
Aitorbp, line 158: Please could you remove the link in the comment. We will link in the pull request. Also the first word of the comment should be in capital letter. Fix this and ready to go 🥇

Aitorbp (changes requested):
@parneet-guraya a few changes more and ready to go 💯

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
jesmrec: And please, fix indentation.

parneet-guraya: @jesmrec I'm using reformat code (Ctrl+Alt+L) , is there any other I should be formatting this?

jesmrec: that's right @parneet-guraya , not i see it correct 👍

jesmrec:
Now, everything is fixed. @Aitorbp please approve if you think this is OK as well and we can merge it

Aitorbp (approved):
LGTM 🥇  Good job!

PR #4437, opened 2024-07-16
[FEATURE REQUEST] Improve the `removed from original folder` option in Auto Uploads

Aitorbp:
## Related Issues
App:

_____

## QA

Checks:

https://github.com/owncloud/android/pull/4437#issuecomment-2257762959

Reports:

- [x] (1) Server down and retry does not remove the local copies with "remove" behaviour https://github.com/owncloud/android/pull/4437#issuecomment-2257831123 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
@@ -73,8 +73,7 @@ class UploadFileFromContentUriUseCase(
                 .then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
                 .enqueue()
         } else {
-            workManager.beginWith(uploadFileFromContentUriWorker)
-                .enqueue()
+            workManager.enqueue(uploadFileFromContentUriWorker)
         }
 
         Timber.i("Plain upload of ${params.contentUri.path} has been enqueued.")
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
@@ -30,7 +30,7 @@ import androidx.work.WorkManager
 import androidx.work.workDataOf
 import com.owncloud.android.domain.BaseUseCase
 import com.owncloud.android.domain.camerauploads.model.UploadBehavior
-import com.owncloud.android.workers.RemoveLocalFileWorker
+import com.owncloud.android.workers.RemoveSourceFileWorker
 import com.owncloud.android.workers.UploadFileFromContentUriWorker
 import timber.log.Timber
 
@@ -47,7 +47,7 @@ class UploadFileFromContentUriUseCase(
             UploadFileFromContentUriWorker.KEY_PARAM_UPLOAD_PATH to params.uploadPath,
             UploadFileFromContentUriWorker.KEY_PARAM_UPLOAD_ID to params.uploadIdInStorageManager
         )
-        val inputDataRemoveLocalFileWorker = workDataOf(
+        val inputDataRemoveSourceFileWorker = workDataOf(
             UploadFileFromContentUriWorker.KEY_PARAM_CONTENT_URI to params.contentUri.toString(),
         )
 
@@ -66,11 +66,11 @@ class UploadFileFromContentUriUseCase(
 
         val behavior = UploadBehavior.fromString(params.behavior)
         if (behavior == UploadBehavior.MOVE) {
-            val removeLocalFileWorker = OneTimeWorkRequestBuilder<RemoveLocalFileWorker>()
-                .setInputData(inputDataRemoveLocalFileWorker)
+            val removeSourceFileWorker = OneTimeWorkRequestBuilder<RemoveSourceFileWorker>()
+                .setInputData(inputDataRemoveSourceFileWorker)
                 .build()
             workManager.beginWith(uploadFileFromContentUriWorker)
-                .then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+                .then(removeSourceFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
                 .enqueue()
         } else {
             workManager.enqueue(uploadFileFromContentUriWorker)
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
@@ -64,15 +64,18 @@ class UploadFileFromContentUriUseCase(
             .addTag(params.uploadIdInStorageManager.toString())
             .build()
 
-        var workContinuation = workManager.beginWith(uploadFileFromContentUriWorker)
         val behavior = UploadBehavior.fromString(params.behavior)
         if (behavior == UploadBehavior.MOVE) {
             val removeLocalFileWorker = OneTimeWorkRequestBuilder<RemoveLocalFileWorker>()
                 .setInputData(inputDataRemoveLocalFileWorker)
                 .build()
-            workContinuation = workContinuation.then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+            workManager.beginWith(uploadFileFromContentUriWorker)
+                .then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+                .enqueue()
+        } else {
+            workManager.beginWith(uploadFileFromContentUriWorker)
+                .enqueue()
         }
-        workContinuation.enqueue()
 
         Timber.i("Plain upload of ${params.contentUri.path} has been enqueued.")
     }
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
@@ -3,8 +3,9 @@
  *
  * @author Abel García de Prada
  * @author Juan Carlos Garrote Gascón
+ * @author Aitor Ballesteros Pavón
  *
- * Copyright (C) 2022 ownCloud GmbH.
+ * Copyright (C) 2024 ownCloud GmbH.
  * <p>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -28,6 +29,7 @@ import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.workDataOf
 import com.owncloud.android.domain.BaseUseCase
+import com.owncloud.android.workers.RemoveLocalFileWorker
 import com.owncloud.android.workers.UploadFileFromContentUriWorker
 import timber.log.Timber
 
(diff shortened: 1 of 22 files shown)


owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
```diff
@@ -52,13 +58,20 @@ class UploadFileFromContentUriUseCase(
...
             .build()
 
-        workManager.enqueue(uploadFileFromContentUriWorker)
+        val removeLocalFileWorker = OneTimeWorkRequestBuilder<RemoveLocalFileWorker>()
+            .setInputData(inputDataRemoveLocalFileWorker)
+            .build()
+
+        workManager.beginWith(uploadFileFromContentUriWorker)
+            .then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+            .enqueue()
```
JuancaG05, line 73: Instead of starting a new worker, we could save it when not needed by checking here the behaviour instead of sending it to the new worker

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveLocalFileWorker.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        contentUri = paramContentUri?.toUri() ?: return false
+        behavior = paramBehavior?.let { UploadBehavior.fromString(it) } ?: return false
+
+        return true
+    }
+
+    private fun removeLocalFile() {
+        val documentFile = DocumentFile.fromSingleUri(appContext, contentUri)
+        documentFile?.delete()
```
JuancaG05, line 71: No need to create a new method for 2 lines, that does exactly what the worker's name indicates

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveLocalFileWorker.kt
```diff
@@ -0,0 +1,73 @@
...
+            }
+            Result.success()
+        } catch (throwable: Throwable) {
+            Timber.e(throwable)
+            Result.failure()
+        }
+    }
+
+    private fun areParametersValid(): Boolean {
+        val paramBehavior = workerParameters.inputData.getString(KEY_PARAM_BEHAVIOR)
```
JuancaG05, line 60: If you follow the first comment I wrote, you won't need this anymore

JuancaG05 (changes requested):

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
Aitorbp: Every invocation of the `WorkContinuation.then(...)` returns a new instance of `WorkContinuation`. So we need to make sure we are calling `enqueue() `on the correct instance. 
`workContinuation` represents a continuation of work and should be modified to include the new task (in this case, removeLocalFileWorker) only if a specific condition is met (behavior == UploadBehavior.MOVE). Therefore, we update `workContinuation` with the new instance:
`workContinuation = workContinuation.then(removeLocalFileWorker)`

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
```diff
@@ -52,13 +58,22 @@ class UploadFileFromContentUriUseCase(
...
-        workManager.enqueue(uploadFileFromContentUriWorker)
+        var workContinuation = workManager.beginWith(uploadFileFromContentUriWorker)
+        val behavior = UploadBehavior.fromString(params.behavior)
+        if (behavior == UploadBehavior.MOVE) {
+            val removeLocalFileWorker = OneTimeWorkRequestBuilder<RemoveLocalFileWorker>()
+                .setInputData(inputDataRemoveLocalFileWorker)
+                .build()
+            workContinuation = workContinuation.then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+        }
+        workContinuation.enqueue()
```
JuancaG05, line 75: Is it really that deep? I mean, I think this gets too complex, I'm not even sure if making a `workContinuation` with only 1 work and enqueueing it will work. Wouldn't it be easier if you write inside the `if` block the enqueue of both works and in the `else` block what you had before?

JuancaG05 (changes requested):
One comment here @Aitorbp

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
Aitorbp: Ok, the reason I did it is because I didn't want to repeat code and I thought was better option to make a `workContinuation`. That was working. Also we didn't need it to add an `else`. In any case, let's go with your approach. Could be easy to understand and less complex as you say.

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
```diff
@@ -52,13 +58,25 @@ class UploadFileFromContentUriUseCase(
...
+        if (behavior == UploadBehavior.MOVE) {
+            val removeLocalFileWorker = OneTimeWorkRequestBuilder<RemoveLocalFileWorker>()
+                .setInputData(inputDataRemoveLocalFileWorker)
+                .build()
+            workManager.beginWith(uploadFileFromContentUriWorker)
+                .then(removeLocalFileWorker) // File is already uploaded, so the original one can be removed if the behaviour is MOVE
+                .enqueue()
+        } else {
+            workManager.beginWith(uploadFileFromContentUriWorker)
+                .enqueue()
```
JuancaG05, line 77: But why `beginWith` and `enqueue` when you know it's going to be just 1 work here? Make it easier, just do it as it was before!

```suggestion
            workManager.enqueue(uploadFileFromContentUriWorker)
```

JuancaG05 (changes requested):

owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/uploads/UploadFileFromContentUriUseCase.kt
Aitorbp: Done

JuancaG05 (approved):
LGTM

Aitorbp:
The solution that has been attempted in this issue is the following:
We need the file deletion to be done after that file has been uploaded to the server. 
For this we have a worker `UploadFileFromContentUriWorker`, which performs the upload task. So we must create another worker in the background that runs when `UploadFileFromContentUriWorker` has been successful. If it has not been successful, the file will not be deleted, avoiding data loss.
Technically, to do this, the workManager's `then()` function has been used. The content of this function is executed when the first worker has been successful.

jesmrec:
Let's QA this

jesmrec:
### (1)

i did the following steps:

1. Enable Auto uploads for pictures with `removed from original folder` behaviour
2. Take some pictures
3. Switch server off (it will make the uploads fail)
4. Wait till pictures are enqueued/failed
5. Switch server on
6. Retry failed uploads

**Current:** uploads are finally uploaded, but they are still in local
**Expected:** uploads are finally uploaded, but they are removed from local because the selected behaviour in `Settings`

NOTE: with connection loss, it works fine.

Pixel 2, Android11
`41c2016a0`

Aitorbp:
> ### (1)
> i did the following steps:
> 
> 1. Enable Auto uploads for pictures with `removed from original folder` behaviour
> 2. Take some pictures
> 3. Switch server off (it will make the uploads fail)
> 4. Wait till pictures are enqueued/failed
> 5. Switch server on
> 6. Retry failed uploads
> 
> **Current:** uploads are finally uploaded, but they are still in local **Expected:** uploads are finally uploaded, but they are removed from local because the selected behaviour in `Settings`
> 
> NOTE: with connection loss, it works fine.
> 
> Pixel 2, Android11 `41c2016a0`

To put this problem into context, first of all we need to know the different behaviors between how the worker works when there is no connection and when the server is off.

**No connection**: The worker is not executed, therefore the upload remains queued until the connection is recovered.

**Server off**: The worker is executed, but the upload ends up failing because the server is not operational and the upload gives a `Connection error`. We will need to click `Retry` to try the upload again.

The problem is that within the `UploadFileFromContentUriWorker` worker the following `copyFileToLocalStorage()` function is being executed. This method updates the `localPath` to `tmp`. Therefore when we do the `Retry` it will delete the file that we have in the `tmp` folder and not the file that is on the device.

As a solution, we have created a new column (`sourcePath`) in the `Transfers` table, where we save the path of the file on the device, since we have no way to recover it once the upload is ready to do a `Retry`. In this way, when we click on `Retry`, the file we have on the device will be deleted when the upload is finished.

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveSourceFileWorker.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.owncloud.android.workers.UploadFileFromContentUriWorker.Companion.KEY_PARAM_CONTENT_URI
+import org.koin.core.component.KoinComponent
+import timber.log.Timber
+
+class RemoveSourceFileWorker(
+    private val appContext: Context,
+    private val workerParameters: WorkerParameters
+) : CoroutineWorker(
+    appContext,
+    workerParameters
```
mplazaspalacio, line 38: Add a coma here

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveSourceFileWorker.kt
```diff
@@ -0,0 +1,62 @@
...
+), KoinComponent {
+
+    private lateinit var contentUri: Uri
+
+    override suspend fun doWork(): Result {
+        if (!areParametersValid()) return Result.failure()
```
mplazaspalacio, line 44: I think is more clear here to put only one return and add an if else. 
Something like this:
`return if(a==b){a}else{b}`

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveSourceFileWorker.kt
```diff
@@ -0,0 +1,62 @@
...
+            Result.failure()
+        }
+    }
+
+    private fun areParametersValid(): Boolean {
+        val paramContentUri = workerParameters.inputData.getString(KEY_PARAM_CONTENT_URI)
+
+        contentUri = paramContentUri?.toUri() ?: return false
+
+        return true
```
mplazaspalacio, line 60: Put only one return.  `return contentUri != null`

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromFileSystemWorker.kt
```diff
@@ -250,8 +250,8 @@ class UploadFileFromFileSystemWorker(
 
         val result = executeRemoteOperation { uploadFileOperation.execute(client) }
 
-        if (result == Unit && behavior == UploadBehavior.MOVE) {
-            removeLocalFile()
+        if (result == Unit) {
```
mplazaspalacio, line 253: There isn't a behaviour related with this result.

mplazaspalacio (changes requested):
Some comments

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveSourceFileWorker.kt
Aitorbp: Well, in this case it seems clear to me, we usually follow this structure when we pass parameters to a worker. We only want it to fail if the parameters are not valid. If they are valid we want it to continue with the following lines of code.

owncloudApp/src/main/java/com/owncloud/android/workers/RemoveSourceFileWorker.kt
Aitorbp: The problem is the `paramContentUri` could be `null`, so we need to use the elvis operator in case this value is `null`. If I set you proposal `return contentUri != null` and remove the elvis operator, Kotlin is going to tell me we need a `!!` operator or `?: return false`.

owncloudApp/src/main/java/com/owncloud/android/workers/UploadFileFromFileSystemWorker.kt
Aitorbp: Yes, here it is not necessary to handle this condition `behavior == UploadBehavior.MOVE` because the file we are going to delete is in the temporal folder. We this file should always be deleted because if not we will have a lot files in temporal folder.  So, to delete the file in the `sourcePath` (new parameter in database) will be done in the `RemoveSourceFileWorker`.

mplazaspalacio (approved):
LGTM

Issue #4369, opened 2024-04-05
[a11y] 11.2.4.7 Focus visible


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

Note: If the focus is highlighted exclusively by a color change,
the contrast between the focused and unfocused state should be sufficient.
It is recommended that the contrast ratio is at least 3:1.

### Personal Space

- [ ] All interactive elements on the page have a poorly recognizable keyboard focus.
keyboard focus. The contrast between the focused and unfocused state
is well below the recommended contrast ratio of 3:1.
the contrast of the labels on the buttons deteriorates when focusing.
buttons deteriorate.
![image](https://github.com/owncloud/android/assets/132444655/20fb5283-33bc-4899-9c92-b044f7735036)
![image](https://github.com/owncloud/android/assets/132444655/95fff6ba-a34d-4afd-a79e-ed3431ee3cd7)

### Menu

- [ ] All interactive elements on the page have a poorly recognizable keyboard focus. 
keyboard focus. The contrast between the focused and unfocused state 
is well below the recommended contrast ratio of 3:1.
![image](https://github.com/owncloud/android/assets/132444655/0a6d8257-3fd1-4c39-8a6d-b256a678b1c1)

### Share

- [ ] All interactive elements on the page have a poorly recognizable keyboard focus. 
keyboard focus. The contrast between the focused and unfocused state is well below the recommended contrast ratio of 3:1. 
is well below the recommended contrast ratio of 3:1. 
the contrast of the buttons themselves deteriorates when focusing.
![image](https://github.com/owncloud/android/assets/132444655/bdff042a-a10c-4a44-a1d6-bcd7bc383ec0)

### Edit shared Link

- [ ] All interactive elements on the page, with the exception of the input field, are given a 
a poorly recognizable keyboard focus.
![image](https://github.com/owncloud/android/assets/132444655/e8bd8cec-8eba-48e8-967c-d4436b983d07)

### Spaces

- [ ] All interactive elements on the page have a poorly recognizable keyboard focus. 
keyboard focus. The contrast between the focused and unfocused state 
is well below the recommended contrast ratio of 3:1.
e.g:
[...]

PR #4446, opened 2024-07-31
[a11y] 11.2.4.7 Focus visible

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4369

_____


Aitorbp:
This issue is blocked until a migration to Material 3 is done. This new approach has been proposed because there is a conflict between the colors when they are focused/unfocused. In addition, there is also a conflict between the focus color and the texts/icons that are inside the elements. If we have a Material 3 palette, all these conflicts should be solved.

Below are the things that have been taken into account in this issue, ignoring the further development of Material.
According to the issue, the color contrast requirements between the focus and unfocus colors must be met `3:01`. This is the minimum recommended ratio. This ratio is not possible to meet.

### 
Points to keep in mind:

**List of files**. The current focus of the default application (`E5E5E5`) does not meet the contrast with white (unfocus state). The contrast is `1.25:1`, far from the minimum recommended.
This focus also does not meet the text (`707575`) that appears in the item. The ratio is `3.71:1` and would not meet the minimum ratio with respect to the text, which is `4.5:1`. So it is not possible to put a color that meets both ratios at the same time.

**Toolbar**: This element presents a similar problem. If we put a white `FFFFFF` focus, the contrast ratio with respect to the blue background (`2D5177`) would be `2.78:1` (the color that would appear in the spotlight is `7C95B1`). It would not meet the minimum recommended ratio, remember that it was `3:01`.

At this point another doubt arises that is more visible in the create `share dialog buttons`. Should we put a dark or light focus? In this case the `save button` is disabled and has a lighter tone than usual. So if we also put a light focus, both elements may be visually confused. Therefore, it may be better to put the focus in a dark color. Open to discussion.

Another thing to keep in mind is that in this issue the color of the press element will not be taken into account. The only thing to know here is that the press element sets the elements to a darker shade.

On the other hand, we must also take into account the brandable colors. As the application is now, two brandable colors should be created for the focus. One for the toolbar and bottom navigation elements. And another for the buttons with the primary style (check style.xml) and the floating button.

This pull request has taken into account these considerations mentioned above. In addition, for buttons with borderless, a new style has been created that can fit with the focus (`custom_borderless_button`).

As a last piece of information to add, this branch has been taken from the keyboard navigator branch.

The work here is far from finished, but it is a good starting point to continue what has been started if Material is not finally implemented before this issue.

For the collection of colors the following page has been used: https://pickcoloronline.com/
For the contrasts this other one: https://webaim.org/resources/contrastchecker/

JuancaG05:
Closing this due to technical complexity, which is out of scope right now.

More info: https://github.com/owncloud/android/issues/4369#issuecomment-2783265260

Issue #4371, opened 2024-04-05
[a11y] 11.2.5.3 Label in the name


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

### Edit shared Link

- [x] The "Cancel" and "Save" buttons have visual labels that are not included in the accessible name. Instead, variable names appear to be used as accessible names.
![image](https://github.com/owncloud/android/assets/132444655/212f1528-edd5-47bd-9422-52e4df89681f)

Issue #4363, opened 2024-04-05
[a11y] 11.1.3.1d Info and relationships (form elements)


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

### Edit shared link

- [x] Neither the input field nor the two switches are programmatically linked to the corresponding labels.
linked to the corresponding labels. When operating with swipe gestures
and activated screen reader, form elements and labels should ideally only
should ideally only represent one focus point. The screen reader would then
automatically read out the associated caption when the form element is focused
which should either be highlighted or linked.

jesmrec:
- Check first if XML-feasible? -> may need jetpack compose. As alternative: add in alternative text in the switch, same as the label... not ideal. 

- Text field -> not feasible in XML the link name. Other option... remove label and just the text field with a hint (compose style)

"add screenshot"

jesmrec:
Check if switches have any property to indicate a label/name. It could replace the existing label, indeed.

Aitorbp:
> audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV
> 
> https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de
> 
> ### Edit shared link
> * [ ]  Neither the input field nor the two switches are programmatically linked to the corresponding labels.
>   linked to the corresponding labels. When operating with swipe gestures
>   and activated screen reader, form elements and labels should ideally only
>   should ideally only represent one focus point. The screen reader would then
>   automatically read out the associated caption when the form element is focused
>   which should either be highlighted or linked.

the input field and the switches have been linked to the corresponding labels.

Issue #4362, opened 2024-04-05
[a11y] 11.1.3.1a Info and relationships (headings)


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

### Shares

- [x] There are several visually clearly recognizable headings on the page, but they are not
labeled as such. These headings include "Sharing", "Users
and groups" and "Public links".

### Edit shared link

- [x] There is a clearly visually recognizable heading on the page "Edit shared link
Edit shared link" on the page, but it has not been labeled as such.

### Spaces

- [x] There is a clearly visually recognizable heading "Spaces" on the page, but it has not been
not labeled as such.

Note: The heading "Spaces" also functions as a button, but this is not visually recognizable.
visually recognizable. It would be better to provide a separate button for editing
button.

### Manage accounts

- [x] There is a clearly visually recognizable heading "Manage accounts" on the page, which
not labeled as such.

### Image Viewer

- [x] The file name could be labeled as a heading.

jesmrec:
Need research: how to set text as heading in Android. Don't seem to be something native in the platform. 

As posible plan B: add "heading" prefix in alternative text.

- Translatable strings

JuancaG05:
Check together with https://github.com/owncloud/android/issues/4363, https://github.com/owncloud/android/issues/4371 and https://github.com/owncloud/android/issues/4373

Aitorbp:
> audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV
> 
> https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de
> 
> ### Shares
> * [ ]  There are several visually clearly recognizable headings on the page, but they are not
>   labeled as such. These headings include "Sharing", "Users
>   and groups" and "Public links".

The heading has been included in the element of the Shares: "toolbar" title, "Users and groups" and "Public links".
 
> ### Edit shared link
> * [ ]  There is a clearly visually recognizable heading on the page "Edit shared link
>   Edit shared link" on the page, but it has not been labeled as such.

[...]

PR #4448, opened 2024-08-07
[a11y] 11.1.3.1a Info and relationships

Aitorbp:
## Related Issues
App:

_____

## QA

Review: https://github.com/owncloud/android/pull/4448#issuecomment-2277274597

Reports:

- [x] (1) X unlabelled in manage accounts https://github.com/owncloud/android/pull/4448#issuecomment-2277380359 [FIXED]
- [x] (2) Actions in lists of shares https://github.com/owncloud/android/pull/4448#issuecomment-2277397616 [FIXED]

### DIFF
--- a/changelog/unreleased/4448
+++ b/changelog/unreleased/4448
@@ -1,0 +1,10 @@
+Enhancement: Improved accessibility of information and relationships
+
+Headings have been added to the following views: Share, Edit/Create Share Link, Standard Toolbar and Manage Accounts.
+The filename input field and the two switches are now linked to their labels.
+The 'contentDescription' attributes of the buttons in the Edit/Create Share Link view have also been updated.
+
+https://github.com/owncloud/android/issues/4362
+https://github.com/owncloud/android/issues/4363
+https://github.com/owncloud/android/issues/4371
+https://github.com/owncloud/android/pull/4448
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,11 +32,24 @@ ownCloud admins and users.
 
 ## Summary
 
+* Enhancement - Improved accessibility of information and relationships: [#4362](https://github.com/owncloud/android/issues/4362)
 * Enhancement - Changed the color of some elements to improve accessibility: [#4364](https://github.com/owncloud/android/issues/4364)
 * Enhancement - Improved SearchView accessibility: [#4365](https://github.com/owncloud/android/issues/4365)
 * Enhancement - Hardware keyboard support: [#4438](https://github.com/owncloud/android/pull/4438)
 
 ## Details
+
+* Enhancement - Improved accessibility of information and relationships: [#4362](https://github.com/owncloud/android/issues/4362)
+
+   Headings have been added to the following views: Share, Edit/Create Share Link,
+   Standard Toolbar and Manage Accounts. The filename input field and the two
+   switches are now linked to their labels. The 'contentDescription' attributes of
+   the buttons in the Edit/Create Share Link view have also been updated.
+
+   https://github.com/owncloud/android/issues/4362
+   https://github.com/owncloud/android/issues/4363
+   https://github.com/owncloud/android/issues/4371
+   https://github.com/owncloud/android/pull/4448
 
 * Enhancement - Changed the color of some elements to improve accessibility: [#4364](https://github.com/owncloud/android/issues/4364)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,12 +32,21 @@ ownCloud admins and users.
 
 ## Summary
 
+* Bugfix - Shares in non-root are updated correctly: [#4432](https://github.com/owncloud/android/issues/4432)
 * Enhancement - Improved accessibility of information and relationships: [#4362](https://github.com/owncloud/android/issues/4362)
 * Enhancement - Changed the color of some elements to improve accessibility: [#4364](https://github.com/owncloud/android/issues/4364)
 * Enhancement - Improved SearchView accessibility: [#4365](https://github.com/owncloud/android/issues/4365)
 * Enhancement - Hardware keyboard support: [#4438](https://github.com/owncloud/android/pull/4438)
 
 ## Details
+
+* Bugfix - Shares in non-root are updated correctly: [#4432](https://github.com/owncloud/android/issues/4432)
+
+   The items of the "Share" view are updated instantly when create/edit a link or
+   share with users or groups in a non-root file.
+
+   https://github.com/owncloud/android/issues/4432
+   https://github.com/owncloud/android/pull/4435
 
 * Enhancement - Improved accessibility of information and relationships: [#4362](https://github.com/owncloud/android/issues/4362)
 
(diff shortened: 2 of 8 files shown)


Aitorbp:
In this pull request, we will choose to use the `android:accessibilityHeading` element in the xml to tell the Talkback that we are dealing with a Heading, that is, that they are not elements that have an assigned action, but rather their function is to summarize groups of text that appear on the screen. Below is some documentation offered by Android on this topic: https://developer.android.com/guide/topics/ui/accessibility/principles#headings_within_text

Just for you to know, we should avoid to use adding a `contentDescription` to the elements because the word heading must not appear on the screen. This is anti-pattern behavior. 

There is a corner case where `android:accessibilityHeading` is not working as usual. It is the `toolbar` in share view. For this attribute to work we need to add the attributes `android:focusable="true"` and `android:focusableInTouchMode="false"`. These attributes are needed to:

- Make the Toolbar recognizable and focusable by accessibility services.
- Ensure that it does not interfere with the normal touch interaction of the user.

**A little more theory about these two attributes:**
`android:focusable="true"` says that this view should be focusable, while `android:focusableInTouchMode="false"` makes it so it’s not focusable in touch mode. This means you can focus it when using a screen reader, but not otherwise by touch. You do this because this parent view isn’t clickable.

As for the file name of the `image viewer`, it is not going to be labeled as a heading, because we consider it to be a specific element within the toolbar and there are a lot of elements in that view (five exactly). The same occurs with the name of the `Spaces`, when clicking on it a `SearchView` opens. The Search view could be changed to only open by clicking the search view icon, but that is outside the scope of the issue. Therefore, it will not be changed either.

Aitorbp:
In this pull request also we have fixed the following issues:
https://github.com/owncloud/android/issues/4363: Used the attribute `labelFor` to link the elements
https://github.com/owncloud/android/issues/4371: Changed the name of `contentDescription`

jabarros:
LGTM

jabarros (approved):

changelog/unreleased/4448
```diff
@@ -0,0 +1,10 @@
+Enhancement: Improved accessibility of information and relationships
+
+Headings have been added to the following views: Share, Edit/Create Share Link, Standard Toolbar and Manage Accounts.
+The filename input field and the two switches are now linked to the their labels.
+The 'contentDescription' attributes of the buttons in the edit/create share link have also been changed.
```
jesmrec, line 5: edit/create must be both in capitals or not.

jesmrec:
## https://github.com/owncloud/android/issues/4371

- [ ] The "Cancel" and "Save" buttons have visual labels that are not included in the accessible name. Instead, variable names appear to be used as accessible names.

## https://github.com/owncloud/android/issues/4362

### Shares

- [ ] There are several visually clearly recognizable headings on the page, but they are not
labeled as such. These headings include "Sharing", "Users
and groups" and "Public links".

### Edit shared link

- [ ] There is a clearly visually recognizable heading on the page "Edit shared link
Edit shared link" on the page, but it has not been labeled as such.

### Spaces

- [ ] There is a clearly visually recognizable heading "Spaces" on the page, but it has not been
not labeled as such.

Note: The heading "Spaces" also functions as a button, but this is not visually recognizable.
visually recognizable. It would be better to provide a separate button for editing

### Manage accounts
- [ ] There is a clearly visually recognizable heading "Manage accounts" on the page, which
not labeled as such.

### Image Viewer

- [ ] The file name could be labeled as a heading.

## https://github.com/owncloud/android/issues/4363

### Edit shared link
 
- [ ] Neither the input field nor the two switches are programmatically linked to the corresponding labels.
linked to the corresponding labels. When operating with swipe gestures
and activated screen reader, form elements and labels should ideally only
should ideally only represent one focus point. The screen reader would then
automatically read out the associated caption when the form element is focused
which should either be highlighted or linked.

jesmrec:
### (1)

A pretty simple fix, out of the scope indeed.

Opening the "Manage accounts" view, the X button appears as "unlabelled". That element is a new one from latest sprints, so, the problem wasn't reported because it didn't exist in that moment.

jesmrec:
### (2)

Just a question:

When selecting any item in the public list or in the share with users and groups list, the dictation tells the name correctly, but finishes with  `Double tap and hold to long press`. Long press is not available for those items, unlike to the items in the file list that can be long-pressed.

changelog/unreleased/4448
Aitorbp: For me yes, these are different views. Open to discussion.

jesmrec: I mean, in the line 3 you wrote `Edit/Create Share Link` (everything in capitals). In the line 5 you wrote `edit/create share link` (everything in non-capitals). No discussion, both in capitals or both in non-capitals.

Aitorbp:
> ### (2) Ready to text @jesmrec
> Just a question:
> 
> When selecting any item in the public list or in the share with users and groups list, the dictation tells the name correctly, but finishes with `Double tap and hold to long press`. Long press is not available for those items, unlike to the items in the file list that can be long-pressed.

To resolve this we have deleted the `longClickable` attribute, so the talkback won't tell us the next sentence: `Double tap and hold to long press`

changelog/unreleased/4448
jesmrec: still differences:

Edit/Create Share Link
Edit/Create share link

Aitorbp: Ups... done

jesmrec:
>  There is a clearly visually recognizable heading "Spaces" on the page, but it has not been
not labeled as such.
Note: The heading "Spaces" also functions as a button, but this is not visually recognizable.
visually recognizable. It would be better to provide a separate button for editing

That needs an extra work and research beyond accessibility. The way to open the search bar should be only the lens icon, not the whole cell. The point is, whether it will turn the text into a heading, as @Aitorbp states [here](https://github.com/owncloud/android/issues/4362#issuecomment-2277392489). Same for the images preview

jesmrec:
@TheOneRing @tbsbdr should the name of the file be handled as a heading in this case?: 

Screenshot 2024-08-12 at 09 18 31

It seems to, but, it's on the top together with other icons/options in the same hierarchy level. So, it's not totally clear for us. Any ideas?

TheOneRing:
An alternative approach would be a subtitle, but please compare with other android apps.

Aitorbp:
> @TheOneRing @tbsbdr should the name of the file be handled as a heading in this case?:
> 
> Screenshot 2024-08-12 at 09 18 31
> 
> It seems to, but, it's on the top together with other icons/options in the same hierarchy level. So, it's not totally clear for us. Any ideas?

For example, in the Google Drive application, talkback does not detect the title of the text on the image detail screen as a heading, it simply reads the content of the title.
drive_title_detail_view_talkback_example

Aitorbp:
> > There is a clearly visually recognizable heading "Spaces" on the page, but it has not been
> > not labeled as such.
> > Note: The heading "Spaces" also functions as a button, but this is not visually recognizable.
> > visually recognizable. It would be better to provide a separate button for editing
> 
> That needs an extra work and research beyond accessibility. The way to open the search bar should be only the lens icon, not the whole cell. The point is, whether it will turn the text into a heading, as @Aitorbp states [here](https://github.com/owncloud/android/issues/4362#issuecomment-2277392489). Same for the images preview

This report will be further worked on in the following issue: https://github.com/owncloud/android/issues/4452

jesmrec (approved):

jesmrec:
All reports are correctly fixed, explained or moved to other issues. Let's move this one forward and iterate over it.

Issue #4373, opened 2024-04-05
[a11y] 11.4.1.2 Name-role value


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

### Personal Space

- [ ] Several buttons have been detected on the page that do not have a role
assigned. 
![image](https://github.com/owncloud/android/assets/132444655/e317e6e5-ead7-4bb3-8770-501f8bcee1a9)

- [ ] If the elements on the page are sorted, the status of the sorting is only visually
sorting status is only visually represented by arrows. However, this visual representation
visual representation is a limitation for non-visual users, as they do not receive any
information about the current sorting. 
![image](https://github.com/owncloud/android/assets/132444655/2a4953e2-b6d7-4374-ac73-ebdd9e1828cd)

- [ ] The button with the "+" does have an accessible name, 
but this is not meaningful. It would make sense to use the word "Add" as an accessible 
the word "Add", for example, to give users a clear description of the function. 
clear description of the function. 
There is also a problem for non-visual users, as opening the menu is not 
opening the menu is not clearly signaled. It is important that the status 
of the menu is accessible to all users.
![image](https://github.com/owncloud/android/assets/132444655/a47e38e7-d80d-40f2-9fd7-d7af986db6db)

### Menu

- [ ] The link "...-Portal" was not correctly marked as a link 
correctly. The role "Link" is missing, which should be used for the semantically correct 
should be used for the semantically correct labeling of links.

- [ ] The logo (see screenshot) was erroneously labeled as a button 
although clicking on the element opens a new page in the browser. 
opens a new page in the browser. Instead, the logo should be provided with a link element 
in order to be semantically correct.
![image](https://github.com/owncloud/android/assets/132444655/b463abd0-5b33-4115-a105-3e5243c55d0b)

- [ ] All elements of the menu are missing the correct role, either 
"Button" or "Link".
![image](https://github.com/owncloud/android/assets/132444655/b6dd55ec-3db8-4a54-8efc-8f836550eb53)

- [ ] The button with the account data does not have a corresponding role. 
In addition, the labeling is too long and difficult to understand. It is 
It is also unclear which view of the menu users are currently in.
[...]

PR #4454, opened 2024-08-14
[a11y] 11.4.1.2 Name-role value

Aitorbp:
## Related Issues
App:https://github.com/owncloud/android/issues/4373

_____

## QA

QA checks: https://github.com/owncloud/android/pull/4454#issuecomment-2302306468

Reports:

- [x] (1) `Sort By` always dictated with "Name" option https://github.com/owncloud/android/pull/4454#issuecomment-2302329383 [FIXED]
- [x] (2) Bad dictation of `Upload` option in FAB https://github.com/owncloud/android/pull/4454#issuecomment-2302346986 [FIXED]
- [x] (3) List in drawer with 7 elements https://github.com/owncloud/android/pull/4454#issuecomment-2302388937  [FIXED]
- [x] (4) Expiration date with no role https://github.com/owncloud/android/pull/4454#issuecomment-2302483076  [FIXED]
- [x] (5) Role in space cards dictated first https://github.com/owncloud/android/pull/4454#issuecomment-2302493232 [WONT FIX]
- [x] (6) Account name role dictated first https://github.com/owncloud/android/pull/4454#issuecomment-2302499191 [WONT FIX]


Aitorbp:
For more details on how we've handled certain elements, please refer to the comment in the [issue](https://github.com/owncloud/android/issues/4373#issuecomment-2286065992).

You'll notice that we haven't consistently used the same attribute to define a role. For example:

`className`:
We've used `className` because it's the most commonly used attribute for assigning a role, and it also supports multilingual roles. We use className when we need to specify or override the type of widget that a view represents. This is especially useful if you're creating a custom component or if a non-standard view needs to be identified as a specific type of widget.

`roleDescription`:
We've used this attribute when it's necessary to describe the purpose or function of a view that isn't clear from its class type alone. This is particularly helpful for custom controls or when you want to provide additional context to assistive technology users. Specifically, we've used it to define the "Link" role.

`contentDescription`:
Finally, we've used this option when, for technical reasons, it wasn't possible to work with `className` or `roleDescription`. This applies to cases like `MenuItems`, which we have already discussed in the [issue comment](https://github.com/owncloud/android/issues/4373#issuecomment-2286065992).

mplazaspalacio (approved):
LGTM

jesmrec:
### (2)

Not sure if this one is part of the current PR, but probably a quick one to fix:

The mini FAB option for `Upload` is dictated as `New folder`

Let me know if it should be moved to a different issue.

Pixel 2, Android 11
`c0c62bed`

jesmrec:
### (3)

The elements on the list of items in the drawer is dictated as "x of 7", as  7 elements are in the list. But, there are only 5. Any reason? something fixable?

Pixel 2, Android 11
`c0c62bed`

jesmrec:
### (4)

About the expiration date, you stated: 

> These are not buttons, they work as informational text. In any case, these elements have been labeled with the corresponding switches in this pr https://github.com/owncloud/android/pull/4448 related to this issue https://github.com/owncloud/android/issues/4363

the "expiration" text is not a button, so that's correct

but, the expiration date is a button used to open the date picker in case of updating the expiration date. It should have a role.

Pixel 2, Android 11
`c0c62bed`

jesmrec:
### (6)

Same as (5), but for items in the list of accounts inside the `Manage accounts` view. 

Pixel 2, Android 11
`c0c62bed`

Aitorbp:
> ### (3)
> The elements on the list of items in the drawer is dictated as "x of 7", as 7 elements are in the list. But, there are only 5. Any reason? something fixable?
> 
> Pixel 2, Android 11 `c0c62bed`

Here the problem is there are two items that never use. I don't know why. The specific items are: `footer_spacer_1` and `footer_spacer_2`. This cause there are 7 items. I have added a new attribute `android:visible="false"` in the items to avoid them from accessibility.

jesmrec:
> Here the problem is there are two items that never use. I don't know why. The specific items are: footer_spacer_1 and footer_spacer_2. This cause there are 7 items. I have added a new attribute android:visible="false" in the items to avoid them from accessibility.

These two `footer_space` seem to be free slots to fill up if more options are required for any reason. If there is no branding option affecting, i'd get rid of them. If new options are required at any moment, they could be added again, right?

jesmrec:
About report (1). Now the correct method of sorting is dictated correctly. But, i realised that an `ownCloud` word is dictated when a sorting method is selected from the bottom sheet, then it says the method and the asc/desc.

jesmrec:
https://github.com/owncloud/android/pull/4454#issuecomment-2304062615

is that checked and correct?

Aitorbp:
> [#4454 (comment)](https://github.com/owncloud/android/pull/4454#issuecomment-2304062615)
> 
> is that checked and correct?

Yes, it is checked, drawer_menu.xml is the only place where these items are referenced.

Aitorbp:
> About report (1). Now the correct method of sorting is dictated correctly. But, i realised that an `ownCloud` word is dictated when a sorting method is selected from the bottom sheet, then it says the method and the asc/desc. @Aitorbp

The problem is that when you click on one of the `sortOptionsMenu` options and go back to the main screen, Talkback thinks you've started the app again and it says `Owncloud`.
This same test can be done if you delete the app from recent views and open it again, it will also say "ownCloud".
It also happens, for example, if you're debugging, a snackbar appears saying: `ownCloud isn't responding`. And click on `Wait`. The talkback will say `Owncloud` again.

jesmrec:
> The problem is that when you click on one of the sortOptionsMenu options and go back to the main screen, Talkback thinks you've started the app again and it says Owncloud.

is there any doc or something that proves that? or just a guess?

> This same test can be done if you delete the app from recent views and open it again, it will also say "ownCloud".
It also happens, for example, if you're debugging, a snackbar appears saying: ownCloud isn't responding. And click on Wait. The talkback will say Owncloud again.

this is a completely different scenario. And also, saying "ownCloud" when you click on "Wait" does not sounds as something correct.

jesmrec:
The `android:label` in the `AndroidManifest.xml` file is causing that. Talkback says the labelled name of the activity every time the focus changes. It could be prevented but the solution might not be a good one.

jesmrec:
> The android:label in the AndroidManifest.xml file is causing that. Talkback says the labelled name of the activity every time the focus changes. It could be prevented but the solution might not be a good one.

I noticed that `android:label` is available for `application` entry as a "generical" label. It can be attached to every single activity, so that, when the activity is loaded, the specific label of the activity is dictated instead of a generical "ownCloud". The point here is, how to handle fragments. Any idea?

jesmrec:
> So, the solution that is working is to modify the label, either in the Activity or in the Manifest itself. But for me it is not a good solution.

why? the label will help the visual impaired person to know where he/she is after an activity change. If the part of the fragments work fine (i don't know at this point), it could be useful.

Aitorbp:
> > > The android:label in the AndroidManifest.xml file is causing that. Talkback says the labelled name of the activity every time the focus changes. It could be prevented but the solution might not be a good one.
> > 
> > 
> > I noticed that `android:label` is available for `application` entry as a "generical" label. It can be attached to every single activity, so that, when the activity is loaded, the specific label of the activity is dictated instead of a generical "ownCloud". The point here is, how to handle fragments. Any idea?
> 
> We could try to handle the fragments using for example this code: `view.importantForAccessibility = View.IMPORTANT_FOR_ACCESSIBILITY_NO`. But it is not working, I think it has to be manage from the Activity which has the responsibility for the fragments.
> 
> I have tried several things. For example, using the `dispatchPopulateAccessibilityEvent` override function. But it is not working neither, I think it was something Android used some years ago. I don't see current documentation about that and accessibility could be changed the criterial.
> 
> So, the solution that is working is to modify the `label`, either in the Activity or in the Manifest itself. But for me it is not a good solution.

Everything relate to the label in activities and fragments will be done in this issue https://github.com/owncloud/android/issues/4458

jesmrec (approved):

Issue #4366, opened 2024-04-05
[a11y] 11.2.1.1 Keyboard


tschiebel:
audit: https://infinite.owncloud.com/s/QtwuiGHikwFjozV

https://www.etsi.org/deliver/etsi_en/301500_301599/301549/03.02.01_60/en_301549v030201p.pdf
https://github.com/BIK-BITV/BIK-App-Test/tree/main/Pr%C3%BCfschritte/de

### Personal Space 

- [x] If the search at the top is not activated, only the buttons at the top of the page (see screenshot) can be accessed using the keyboard.
![image](https://github.com/owncloud/android/assets/132444655/ee189e3a-5b09-4c55-8939-46176440f577)
To access the other interactive elements on the page, you must activate the "X" button (see screenshot). However, you can then return to the top elements of the page, which is a limitation for users who use the keyboard. 
![image](https://github.com/owncloud/android/assets/132444655/c086bf68-f30b-4007-a9df-20fb15dacce8)

- [x] If you tap on a directory or file entry for a long time, a new page opens on which you have the option of selecting directories and files. Unfortunately, this option is not available to keyboard users. Furthermore, it is not immediately apparent to all users that that this option for selecting elements exists.
![image](https://github.com/owncloud/android/assets/132444655/16e24679-01b6-485d-8fe6-31a47ea86309)

### Share

- [x] The "Back" button cannot be reached using the keyboard.
![image](https://github.com/owncloud/android/assets/132444655/54d5b120-2782-4e65-b241-a86f68d6fe26)

### Spaces

- [x] If the search at the top is not activated, only the top buttons on the page (see screenshot) can be accessed using the keyboard.
![image](https://github.com/owncloud/android/assets/132444655/54c11d40-6ea0-4ec7-b0da-855364f1d9df)
To access the other interactive elements on the page, you must activate the "X" button (see screenshot). However, you can then no longer return to the upper elements of the page, which is a limitation for users who use the keyboard.
![image](https://github.com/owncloud/android/assets/132444655/6860ba14-90cf-4d09-a871-247b6db1698a)

### Manage accounts

- [x] The "Back" button cannot be reached using the keyboard.
![image](https://github.com/owncloud/android/assets/132444655/2d75255c-e49e-45f2-9c25-5d5f8d007169)

### Image Viewer

- [x] The buttons at the top of the page cannot be focused or activated using the keyboard. They are invisible to keyboard users and only become visible when tapped on the screen.
[...]

PR #4455, opened 2024-08-20
[ally] Keyboard navigation in security passcode view

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4447

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
@@ -120,6 +120,7 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
                 binding.explanation.visibility = View.INVISIBLE
                 supportActionBar?.setDisplayHomeAsUpEnabled(false) //Don´t show the back arrow
             }
+
             ACTION_CREATE -> { //Create a new password
                 if (confirmingPassCode) {
                     //the app was in the passcode confirmation
@@ -138,15 +139,18 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
                         intent.extras?.getBoolean(EXTRAS_MIGRATION) == true -> {
                             supportActionBar?.setDisplayHomeAsUpEnabled(false)
                         }
+
                         intent.extras?.getBoolean(EXTRAS_LOCK_ENFORCED) == true -> {
                             supportActionBar?.setDisplayHomeAsUpEnabled(false)
                         }
+
                         else -> {
                             supportActionBar?.setDisplayHomeAsUpEnabled(true)
                         }
                     }
                 }
             }
+
             ACTION_REMOVE -> { // Remove password
                 // pass code preference has just been disabled in Preferences;
                 // will confirm user knows pass code, then remove it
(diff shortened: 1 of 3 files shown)


Aitorbp:
The toolbar is not accessible by the keyboard because of technical restriction. See this [comment](https://github.com/owncloud/android/issues/4366#issuecomment-2236411592) to more details.

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
```diff
@@ -434,15 +444,33 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
...
             }
         }
     }
 
+    override fun onKeyUp(keyCode: Int, event: KeyEvent?): Boolean {
+        return when (keyCode) {
+            KeyEvent.KEYCODE_TAB -> {
+                findViewById<View>(R.id.lock_time).requestFocus()
+                true
+            }
```
jesmrec, line 463: Why don't you add here a blank line in between the `when` options? in the code above 👆  , you added blanks in every line of existing code

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
```diff
@@ -434,15 +444,33 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
...
+            KeyEvent.KEYCODE_DPAD_DOWN -> {
+                if (!findViewById<View>(R.id.numberKeyboard).hasFocus()) {
+                    findViewById<View>(R.id.numberKeyboard).requestFocus()
+                }
+                true
+            }
```
jesmrec, line 469: Same here

jesmrec (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
Aitorbp: Changed

jesmrec:
Things i am missing here:

- Navigation through buttons is correct, but, numerical keyboard should work as well. User should be able to type/delete the passcode with the physical keyboard in the same way as clicking buttons. 

- Navigation through buttons is a keyboard trap in the settings. I did not find the way to jump from the buttons to the topbar, in case i want to cancel the view

- Are the focused buttons with the keyboard fulfilling contrast requirements?

Aitorbp:
> Things i am missing here:
> 
> * Navigation through buttons is correct, but, numerical keyboard should work as well. User should be able to type/delete the passcode with the physical keyboard in the same way as clicking buttons.

Now It should work.

> * Navigation through buttons is a keyboard trap in the settings. I did not find the way to jump from the buttons to the topbar, in case i want to cancel the view

Try again. With my external keyboard the ESC key is working. But I added something else. Let´s see.
> * Are the focused buttons with the keyboard fulfilling contrast requirements? 

No, It is not meeting the contrast requirements
@jesmrec

jesmrec:
After last commit:

- typing numbers with physical keyboard is now posible. Numbers are displayed while typing. It'd be better not to be shown but this is not a big problem

- `Esc` works to dismiss the view 💯

jesmrec:
Let's move the contrast thing to the global topic, then...

this one is approved, at least for a first iteration.

PR #4459, opened 2024-08-31
Fix/list filtering not working after rotating device 4441

AwaisKhan128:
## Related Issues
App:

_____

### DIFF
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -48,6 +48,10 @@ android {
         ignoreWarnings true
     }
 
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
+
     testOptions {
         unitTests {
             includeAndroidResources = true
--- a/owncloudComLibrary/build.gradle
+++ b/owncloudComLibrary/build.gradle
@@ -35,6 +35,13 @@ android {
         minSdkVersion sdkMinVersion
         targetSdkVersion sdkTargetVersion
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
+    }
+    kotlinOptions {
+        jvmTarget = "17"
+    }
 
     lint {
         abortOnError false
--- a/owncloudData/build.gradle
+++ b/owncloudData/build.gradle
@@ -26,14 +26,15 @@ android {
         }
     }
 
-    kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
 
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_17.toString()
     }
+
 
     sourceSets {
         androidTest.java.srcDirs += "src/test-common/java"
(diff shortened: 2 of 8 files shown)


owncloudComLibrary/build.gradle
```diff
@@ -35,12 +35,23 @@ android {
...
+    }
 
     lint {
         abortOnError false
         ignoreWarnings true
     }
 
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
```
JuancaG05, line 53: Same as before, this is a strong change and shouldn't be carried out here

owncloudData/build.gradle
```diff
@@ -26,15 +26,16 @@ android {
...
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
 
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_17.toString()
     }
```
JuancaG05, line 36: And same here

owncloudTestUtil/build.gradle
```diff
@@ -9,12 +9,19 @@ android {
...
+    }
     buildTypes {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+    kotlinOptions {
+        jvmTarget = "17"
+    }
```
JuancaG05, line 24: Idem

JuancaG05:
Hi @AwaisKhan128! Thanks a lot for your contribution! 🚀 

I'm closing this PR since fixing it would imply changing most part of the code you committed. I left you a code review here, but the most important points are: we shouldn't do strong changes (such as the Java version), at least in a PR like this -addressed to a little bug-, we shouldn't leave unused code and imports, and shouldn't add unnecessary code either. In any case, I would take a look into the activity or fragments code to try fixing the search bar stuff, instead of adding the `configChanges` property in the activity in the Manifest. `FileDisplayActivity` is the main activity in the app and hosts many fragments, this could have undesired collateral effects.

Also, we use "Conventional Commits" to name our commits, so that we can keep all of them aligned. You can check more info here: https://github.com/owncloud/android/pull/4459/files

I encourage you to open a new PR and send a new contribution, just take into account all that I posted here! 💪

AwaisKhan128:
@JuancaG05 Thanks for taking a review. I will open a new PR and send new contribution while taking account all above.
The thing I want to make you notice that updating of Java version was made to build app in new gradle and new android studio version. Old versions are no more supported to build the app. How it is possible to check and fix the bug if I can not build the app properly?

Anyways, Thanks for your review.

JuancaG05:
Hi @AwaisKhan128. Which Android Studio version are you using? I use `Android Studio Koala | 2024.1.1`, which is relatively recent, and with the Java and Gradle versions specified in the project, it runs correctly

AwaisKhan128:
@JuancaG05. Yes it works. But how. Once you clone the project. It shows an error then you upgrade its java version and sync/build it. It successfully builds up. Once it builds up now if you downgrade it again it won't show you an error.

AwaisKhan128:
@JuancaG05 also can you help me installing and using calens? I am new to it.

jesmrec:
This is a matter of scope. This issue is intended to fix a problem in the filtering. Please, put the focus in the problem, if you think other parts of the app could be improved, this is welcome, but, in a different place. Fill up a new issue to address the fix there. Thanks.

JuancaG05:
> @JuancaG05 also can you help me installing and using calens? I am new to it.

Hi @AwaisKhan128! No worries, there's no need to install nothing with Calens 😄. Calens is the plugin that the repo uses internally to generate the changelog based on the new entry files we create (as the one you created). Just read the main brief instructions (https://github.com/owncloud/android/blob/master/changelog/README.md) to create a new entry file and that's it. You did it correctly except for that couple of details I commented in the review 👍

Issue #4442, opened 2024-07-18
[BUG] Some colors are not correctly set up


jesmrec:
1. The new manage accounts menu takes the header color from an incorrect place. Correct color should be the top bar one. Just an example:

![Screenshot 2024-07-18 at 10 58 04](https://github.com/user-attachments/assets/998e9db3-b2b9-4f1f-88fa-75dd88f0f710)

2. The notifications bar on the top of the screen is not brandable. 

Example in: https://github.com/owncloud/enterprise/issues/6839#issuecomment-2278782162

PR #4463, opened 2024-09-05
[BUG] Some colors are not correctly set up

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4442

_____


JuancaG05 (approved):
Since we're not using `colorPrimaryDark` for other purposes, let's go with this approach for the moment

jesmrec:
Test done here:

in file `setup.xml`, the field `actionbar_start_color` points to `@color/primary`

I changed the value of `@color/primary` to other values to check that the manage accounts dialog bar, and the notifications bar are tinted with such color. 

It worked fine. 

You can move it forward.

PR #4466, opened 2024-09-09
[TECHNICAL] Changes in accessibility role management

Aitorbp:
## Related Issues
App:

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -34,6 +34,7 @@ import android.net.Uri
 import android.text.method.LinkMovementMethod
 import android.util.TypedValue
 import android.view.ContextThemeWrapper
+import android.view.inputmethod.InputMethodManager
 import android.webkit.MimeTypeMap
 import android.widget.Button
 import android.widget.LinearLayout
@@ -271,6 +272,17 @@ private fun makeIntent(file: File?, context: Context): Intent {
     }
     sendIntent.putExtra(Intent.ACTION_SEND, true) // Send Action
     return sendIntent
+}
+
+fun Activity.hideSoftKeyboard() {
+    val focusedView = currentFocus
+    focusedView?.let {
+        val inputMethodManager = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+        inputMethodManager.hideSoftInputFromWindow(
+            focusedView.windowToken,
+            0
+        )
+    }
 }
 
 fun Activity.checkPasscodeEnforced(securityEnforced: SecurityEnforced) {
--- a/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
@@ -34,7 +34,6 @@ import android.net.Uri
 import android.text.method.LinkMovementMethod
 import android.util.TypedValue
 import android.view.ContextThemeWrapper
-import android.view.inputmethod.InputMethodManager
 import android.webkit.MimeTypeMap
 import android.widget.Button
 import android.widget.LinearLayout
@@ -272,17 +271,6 @@ private fun makeIntent(file: File?, context: Context): Intent {
     }
     sendIntent.putExtra(Intent.ACTION_SEND, true) // Send Action
     return sendIntent
-}
-
-fun Activity.hideSoftKeyboard() {
-    val focusedView = currentFocus
-    focusedView?.let {
-        val inputMethodManager = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-        inputMethodManager.hideSoftInputFromWindow(
-            focusedView.windowToken,
-            0
-        )
-    }
 }
 
 fun Activity.checkPasscodeEnforced(securityEnforced: SecurityEnforced) {
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
@@ -35,7 +35,6 @@ import android.text.Editable
 import android.view.KeyEvent
 import android.view.MenuItem
 import android.view.View
-import android.view.View.OnFocusChangeListener
 import android.view.WindowManager
 import android.widget.EditText
 import android.widget.LinearLayout
@@ -162,8 +161,6 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
                 throw IllegalArgumentException(R.string.illegal_argument_exception_message.toString() + " ")
             }
         }
-
-        setTextListeners()
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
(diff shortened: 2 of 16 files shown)


Aitorbp:
**Report 1:**
` Include contentDescription as another parameter in the setAccessibilityRole extension function, so that we avoid implementing AccessibilityDelegateCompat again, like in SortOptionsView`

As I said earlier in the pr: Assigning an `AccessibilityDelegate` to a MenuItem is not as straightforward as assigning it to a normal view because MenuItem is not a View itself. Instead, MenuItem has an associated View via `actionView`, which can sometimes be null if it hasn't been manually configured.

So if we try to access an `itemMenu` view following, for example, this code: 

`view?.findViewById<View>(R.id.action_see_details)?.setAccessibilityRole( contentDescription = "${getString(R.string.actionbar_see_details)} ${getString(R.string.button_role_accessibility)}" )`

Not working. Therefore I would leave it as it is and until `contentDescription` had no more use I would not put it in the extension function.

Likewise, we cannot remove the `AccessibilityDelegateCompat` from the `SortOptionsView` because what this method does is detect the clicks that the talkback makes and the changes that have been made at that moment. If we put this functionality in the `setAccessibilityRole` extension function pointing to the corresponding view, it would not detect ascending/descending state changes.

**Report 2:**
`Check if onCreateOptionsMenu can actually be moved to FileFragment and remove it from its subclasses`

I don't see a problem that now the oncreateView is created in the parent and not in the children. To give more detail to the structure, leave here what the menu would look like. In `FileFragment` there would be all those options that are common to all and in the children their particular cases:

`FileFragment`:
  1. Open with item
  2. Send item
  3. Set as available offline item
  4. Unset as available offline item

FileFragment inheritances:

  `FileDetailsFragment:`
    1. Rename item
    2. Remove item

  `PreviewAudioFragment`
    1. Details item

  `PreviewTextFragment`
    1. Details item

  `PreviewImageFragment`
    1. Details item

  `FileDownloadFragment`: is not affected.

Aitorbp:
**Research passcode visibility when introducing numbers from the keyboard**

First of all, we need to mention that the operating system has internal mechanisms to manage text input interactions, both from physical and virtual keyboards. In the case of physical keyboards, when typing into a password `EditText`, the character is briefly displayed and then hidden. This behavior can be restricted from the Android settings.

Therefore, in our code, two functionalities coexist: first, the system behavior is executed, followed by our own implementation.

The most optimal solution we have found is to set the `EditText`'s focusable attribute to `false`. This way, the system will not briefly display the digit on the screen.

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
```diff
@@ -194,7 +194,24 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
     private fun setTextListeners() {
         val numberOfPasscodeDigits = (passCodeViewModel.getPassCode()?.length ?: passCodeViewModel.getNumberOfPassCodeDigits())
         for (i in 0 until numberOfPasscodeDigits) {
-            passCodeEditTexts[i]?.setOnClickListener { hideSoftKeyboard() }
+
+            passCodeEditTexts[i]?.apply {
+                hideSoftKeyboard()
```
JuancaG05, line 199: This doesn't make sense here. Previously this line was inside the `setOnClickListener` method for each of the `EditText`, but we cannot apply it to a `EditText` directly

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
```diff
@@ -194,7 +194,24 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
...
+                            View.FOCUS_DOWN
+                        } else {
+                            View.FOCUS_LEFT
+                        }
+                        v.focusSearch(focusDirection)?.requestFocus()
+                        true
+                    } else {
+                        false
+                    }
+                }
```
JuancaG05, line 213: Why is all of this inside an `apply` block for each of the `EditText`s? Do you really use `passCodeEditTexts[i]` here? Am I missing something?

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
```diff
@@ -194,7 +194,24 @@ class PassCodeActivity : AppCompatActivity(), NumberKeyboardListener, EnableBiom
...
+
+            passCodeEditTexts[i]?.apply {
+                hideSoftKeyboard()
+
+                setOnKeyListener { v, keyCode, event ->
+                    if (keyCode == KeyEvent.KEYCODE_TAB && event.action == KeyEvent.ACTION_DOWN) {
+                        val focusDirection = if (passCodeEditTexts.indexOf(v) == 0) {
+                            View.FOCUS_DOWN
+                        } else {
+                            View.FOCUS_LEFT
```
JuancaG05, line 206: Why `FOCUS_LEFT` here? Wouldn't it be to the right?

JuancaG05 (changes requested):
Some comments here for you to check @Aitorbp

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
Aitorbp: Ok, I notice that when I added the `focus = false` to the `EditTest` the softkeyboard is not shown. So maybe it is better to remove this function because is not useful anymore if we want keep the same behaviour between custom keyboard and external keyboard.

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
Aitorbp: Yeah, I did it because of the last comment you told me: to unify the `hideSoftKeyboard()` and the `setOnKeyListener` in the same block. But if you agree we the first approach I will remove the `hideSoftKeyboard()` and leave just the `setOnKeyListener` without the `apply`.

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
Aitorbp: Yes, in the beginning I set to `FOCUS_LEFT`, because the navigation it wasn't working well with `FOCUS_RIGHT`. I don't know why. But with the `focus` to `false` it isn't affected. So i will change to the correct one.

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
```diff
@@ -274,17 +273,6 @@ private fun makeIntent(file: File?, context: Context): Intent {
...
-fun Activity.hideSoftKeyboard() {
-    val focusedView = currentFocus
-    focusedView?.let {
-        val inputMethodManager = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-        inputMethodManager.hideSoftInputFromWindow(
-            focusedView.windowToken,
-            0
-        )
-    }
-}
```
JuancaG05, line 286: Maybe this is not used anymore, but I wouldn't delete it since it can be a useful utils function in a future

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
JuancaG05: But what are these lines used for? If those `EditText`s are not focusable anymore, the `setOnKeyListener`s will never run, right?

JuancaG05 (changes requested):
One more change, and check the previous CR comments

owncloudApp/src/main/java/com/owncloud/android/presentation/security/passcode/PassCodeActivity.kt
JuancaG05: But what are these lines used for? If those `EditText`s are not focusable anymore, the `setOnKeyListener`s will never run, right?

Aitorbp: Yeah, it is not useful, deleted. Also the `onFocusChangeListener` have been deleted

owncloudApp/src/main/java/com/owncloud/android/extensions/ActivityExt.kt
Aitorbp: ok

JuancaG05 (approved):
LGTM, let's move this to QA 🚀

JuancaG05:
Here the `EditText`s management in the passcode view changed, so I recommend to put a special focus on the passcode feature: check that it is correctly introduced, even when we remove some characters, etc... FYI @jesmrec 😀

jesmrec:
This task is a refactor over other [task](https://github.com/owncloud/android/issues/4373) in the same milestone that already has an entry in Calens file. That means, we don't need two entries for the same feature in the same release/milestone. Please, join all together the roles management stuff in just one entry in the Calens.

Aitorbp:
> This task is a refactor over other [task](https://github.com/owncloud/android/issues/4373) in the same milestone that already has an entry in Calens file. That means, we don't need two entries for the same feature in the same release/milestone. Please, join all together the roles management stuff in just one entry in the Calens.

Ok, maybe removing this one is ok because this issue handle two differents issues. What do you thing ? @jesmrec

jesmrec:
> Ok, maybe removing this one is ok because this issue handle two differents issues. What do you thing ? @jesmrec

it depends. If the stuff in the current PR adds something new and relevant to the work in the other, it should be added. If it is just refactor, corrections and bugfixing you are right, we could get rid of the current entry.

jesmrec:
### [WIP]

Some checks over navegability in passcode view:

Open create view and ...
- click on tab + enter -> 1 clicked
- click on down arrow key + enter  -> 1 clicked
- Click on enter -> moved to 1 and another enter submit 1

Open unlock view and ...
- click on tab + enter -> 1 clicked
- click on down arrow + enter -> 1 clicked
- click on enter -> moved to 1 and another enter submit 1

jesmrec:
> Ok, maybe removing this one is ok because this issue handle two differents issues. What do you thing ? @jesmrec

And anyway, we can remove this one but adding the PR reference to the existing one @Aitorbp

Aitorbp:
> > Ok, maybe removing this one is ok because this issue handle two differents issues. What do you thing ? @jesmrec
> 
> And anyway, we can remove this one but adding the PR reference to the existing one @Aitorbp

Ok I did it.

jesmrec:
Approved on my side

Issue #4458, opened 2024-08-26
[a11y] Correct label for each activities and fragments


Aitorbp:
**Is your feature request related to a problem? Please describe.**

In relation to this issue https://github.com/owncloud/android/issues/4373, we have noticed that when we have talkback enabled and we are switching between activities/fragments, talkback reads the word `ownCloud` for each change.
This is not accessible to users, since the word `ownCloud` does not provide much information to define which view the user is on.

**Describe the solution you'd like**
A possible solution would be to add the `android:label` in the `AndroidManifest.xml` in each activity to the correct name. 

In the case of fragments, we must change the label through the `setTitle`. Example below:
`activity?.setTitle("The new label here")`
Doc: https://stackoverflow.com/questions/27918701/android-fragment-change-title

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/correct_label_for_each_activities_fragments
- [x] Development tasks
  - [x] Added label to activities
  - [x] Added setTitle to fragment
- [x] Code review and apply changes requested
- [x] Design test plan
- [ ] QA
- [ ] Merge branch feature/correct_label_for_each_activities_fragments

PR #4470, opened 2024-09-18
[a11y] Correct label for each activities and fragments

Aitorbp:
## Related Issues
App: https://github.com/owncloud/android/issues/4458

_____

### DIFF
--- a/changelog/unreleased/4470
+++ b/changelog/unreleased/4470
@@ -1,0 +1,6 @@
+Enhancement: TalkBack announces the view label correctly
+
+TalkBack no longer announces "ownCloud" every time the screen changes. Now, it correctly dictates the name of the current view.
+
+https://github.com/owncloud/android/issues/4458
+https://github.com/owncloud/android/pull/4470
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
@@ -217,7 +217,7 @@ class FileDetailsFragment : FileFragment() {
         }
         startListeningToOngoingTransfers()
         fileDetailsViewModel.checkOnGoingTransfersWhenOpening()
-        activity?.title = getString(R.string.details_label)
+        requireActivity().title = getString(R.string.details_label)
     }
 
 
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
@@ -217,6 +217,7 @@ class FileDetailsFragment : FileFragment() {
         }
         startListeningToOngoingTransfers()
         fileDetailsViewModel.checkOnGoingTransfersWhenOpening()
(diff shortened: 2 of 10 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/settings/SettingsActivity.kt
```diff
@@ -70,7 +71,7 @@ class SettingsActivity : AppCompatActivity() {
             is SettingsMoreFragment -> R.string.prefs_subsection_more
             else -> R.string.actionbar_settings
         }
-
+        setTitle(titleId)
```
JuancaG05, line 74: Maybe with this, that changes the title dynamically depending on the section you are at, the label in the Manifest is not needed anymore for this activity

owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
```diff
@@ -217,6 +217,7 @@ class FileDetailsFragment : FileFragment() {
         }
         startListeningToOngoingTransfers()
         fileDetailsViewModel.checkOnGoingTransfersWhenOpening()
+        activity?.title = getString(R.string.details_label)
```
JuancaG05, line 220: In these cases I'd say is better to use `requireActivity()` so that we avoid handling nullability, since we know this fragment should be always attached to an activity

changelog/unreleased/4470
```diff
@@ -0,0 +1,6 @@
+Enhancement: Talkback announces the view label correctly
```
JuancaG05, line 1:
```suggestion
Enhancement: TalkBack announces the view label correctly
```
You wrote it correctly in the description but not in the title

JuancaG05 (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/files/details/FileDetailsFragment.kt
Aitorbp: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/settings/SettingsActivity.kt
Aitorbp: Yeah, I tested and it is not needed. 
Also I added a label to `LogsListActivity`, because when we navigate to list of logs, the TalkBack said: owncloud. I also remove the `setTitle(R.string.prefs_log_open_logs_list_view)` in the activity, because is not needed anymore.

changelog/unreleased/4470
Aitorbp: Done

JuancaG05 (approved):
LGTM, moving to QA

jesmrec:
As far a i can see, these are the affected views:

Folder picker ✅ 
Passcode ✅ 
Pattern ✅ 
Receive external files ✅ 
Video preview ✅ 
Release notes ✅ 
Image preview ✅ 
Login ✅ 
What's new ✅ 
Audio preview - detected problem in playing -> https://github.com/owncloud/android/issues/4474
Details ✅ 
Text preview ✅

PR #4476, opened 2024-09-23
Sync stops after Network switch

### LINKED ISSUE #4381: Sync stops after Network switch
## Describe the bug
When the network switches from WiFi to mobile while syncing, the syncing process will stop and needs to be manually started again to continue. This is not appreciated; it should resume automatically.

## Steps to reproduce
1. Start syncing.
2. Switch networks.
3. Realize it has stopped.

## Expected behavior
If the network I switched to is set/allowed to sync, it should resume automatically.

## Actual behavior
It stops and needs to be resumed manually.

<details>
<p>

```console
$ /usr/bin/ocis version
Version: 5.1.0-prealpha+52b00b116
Compiled: 2024-04-10 00:00:00 +0000 UTC
```

</p>
</details>

Aitorbp:
## Related Issues
App:

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/downloads/DownloadFileUseCase.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/usecases/transfers/downloads/DownloadFileUseCase.kt
@@ -18,6 +18,7 @@
  */
 package com.owncloud.android.usecases.transfers.downloads
 
+import androidx.work.BackoffPolicy
 import androidx.work.OneTimeWorkRequestBuilder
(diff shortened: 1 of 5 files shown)


Aitorbp:
Things to checks:
- Stranges behaviour when changes the connections.
- Sometimes we receive the `NetworkErrorException` or `NoConnectionWithServerException`. Should be `NetworkErrorException`.
- I don't know if the `BackoffCriteria` is working good because when the upload  is retry it doesn't take 10 seconds, but less.
- Download behaviour is not tested.

jesmrec:
Will try another ideas.

PR #4487, opened 2024-10-04
[TECHNICAL] Detekt: static code analyzer

jesmrec:
## Related Issues
App: https://github.com/owncloud/android/issues/4506

_____


gradle/libs.versions.toml
```diff
@@ -22,6 +22,7 @@ androidxTestMonitor = "1.6.1"
 androidxTestUiAutomator ="2.2.0"
 androidxWork = "2.8.1"
 coil = "2.2.2"
+detekt="1.23.3"
```
JuancaG05, line 25:
```suggestion
detekt = "1.23.3"
```
To keep the same formatting across the whole file 👍

.github/workflows/detekt.yml
```diff
@@ -0,0 +1,62 @@
...
+        # Default value if no value is explicitly provided
+        default: 'World'
+        # Input has to be provided for the workflow to run
+        required: true
+        # The data type of the input
+        type: string
+  
+  push:
+    branches:
+      - feature/detekt
```
JuancaG05, line 23: We'll need to review this before merging, and probably remove it

.github/workflows/detekt.yml
```diff
@@ -0,0 +1,62 @@
...
+  pull_request:
+    branches:
+      - "master"
```
JuancaG05, line 26: Probably we want to run Detekt in every PR we do, not only in master (for which we cannot create PR), so this needs to be reviewed as well

.github/workflows/detekt.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Set up JDK 17
+      uses: actions/setup-java@v4
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+
+    # Configure Gradle for optimal use in GitHub Actions, including caching of downloaded dependencies.
+    # See: https://github.com/gradle/actions/blob/main/setup-gradle/README.md
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.0.0
```
JuancaG05, line 50: What is this `v4.0.0` referred to? I guess it's not the Gradle or Android Gradle Plugin version, otherwise this won't be working 😄

.github/workflows/detekt.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Set base line
+      run: ./gradlew detektBaseline
```
JuancaG05, line 52: This is something to review as well, the idea is removing the baseline before we merge so that when a new error is introduced, make the work fail, right?

JuancaG05 (changes requested):
Some comments to take into account before we merge 😃 @jesmrec

.github/workflows/detekt.yml
jesmrec: This is totally intended at this point. That makes detekt to execute only in the current branch because the work is not finished. This will not be removed but replaced for a correct policy that we will agree in.

.github/workflows/detekt.yml
jesmrec: Same as before. Intended and provisional.

.github/workflows/detekt.yml
jesmrec: Same as before, intended and provisional.

gradle/libs.versions.toml
jesmrec: Done

JuancaG05:
`MaximumLineLength`, from the "Formatting" section, is currently enabled. In the docs it says [it overlaps](https://detekt.dev/docs/1.23.3/rules/formatting#maximumlinelength) with `MaxLineLength`, which we are relying on, and they have different values. Needs to be checked.

JuancaG05:
Just checked (tested empirically) with the `LongMethod` rule that Detekt counts LLOC when talking about lines of code, and so I guess this happens with every rule that has to do with lines of code. That means, when trying to reduce lines of code, don't play with blank lines or comments since that won't make effect 😉

owncloudData/src/test/java/com/owncloud/android/data/files/datasources/implementation/OCLocalFileDataSourceTest.kt
```diff
@@ -428,7 +436,8 @@ class OCLocalFileDataSourceTest {
     }
 
     @Test
-    fun `getFilesWithSyncInfoAvailableOfflineFromAccountAsFlow returns a Flow with an empty list when DAO returns a Flow with an empty list`() = runTest {
+    fun `getFilesWithSyncInfoAvailableOfflineFromAccountAsFlow returns a Flow with an empty list when DAO returns a Flow with an empty list`()
+    = runTest {
```
joragua, line 440: Bitrise doesn't like this, so `=` can be moved at the end of the previous line

JuancaG05: Done

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
```diff
@@ -294,11 +294,12 @@ class ShareActivity : FileActivity(), ShareFragmentListener {
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         var retval = true
-        when (item.itemId) {
-            android.R.id.home -> if (!supportFragmentManager.popBackStackImmediate()) {
+        if (item.itemId == android.R.id.home) {
+            if (!supportFragmentManager.popBackStackImmediate()) {
```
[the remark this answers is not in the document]
JuancaG05: Hm, the rule I was talking about is `CollapsibleIfStatements`, which is already done, but seems you didn't change this. I keep thinking they can be joined, did you check if that rule reports this case after these changes?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1517,26 +1510,7 @@ class FileDisplayActivity : FileActivity(),
...
-     * associated shares), but not their contents. Only the contents of files marked to be kept-in-sync are
-     * synchronized too.
-     *
-     * @param folder     Folder to refresh.
-     * @param ignoreETag If 'true', the data from the server will be fetched and synced even if the eTag
-     * didn't change.
-     */
-    fun startSyncFolderOperation(folder: OCFile?, ignoreETag: Boolean) {
-        // TODO: SYNC FOLDER
+        // Nothing to do
```
JuancaG05, line 1513: So, maybe we can omit this, if the implementation is not compulsory (we can use parent's implementation). Try to remove it 👍

JuancaG05 (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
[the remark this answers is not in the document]
joragua: Done ✅

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
joragua: Removed!

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListViewModel.kt
```diff
@@ -368,7 +369,7 @@ class MainFileListViewModel(
             FileListOption.ALL_FILES -> retrieveFlowForAllFiles(currentFolderDisplayed, currentFolderDisplayed.owner)
             FileListOption.SHARED_BY_LINK -> retrieveFlowForShareByLink(currentFolderDisplayed, currentFolderDisplayed.owner)
             FileListOption.AV_OFFLINE -> retrieveFlowForAvailableOffline(currentFolderDisplayed, currentFolderDisplayed.owner)
-            FileListOption.SPACES_LIST -> TODO()
+            FileListOption.SPACES_LIST -> flow {}
```
JuancaG05, line 372: Hm, not sure if `flow {}` or ´flowOf()` is a better option

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1351,8 +1351,8 @@ class FileDisplayActivity : FileActivity(),
                     }
 
                     is SynchronizeFileUseCase.SyncType.UploadEnqueued -> showSnackMessage(getString(R.string.upload_enqueued_msg))
-
-                    null -> TODO()
+                    null -> { // Nothing to do
+                    }
```
JuancaG05, line 1355: If it is just 1 line, better placing it all together:
`null -> { // Nothing to do }`

joragua: Done 👍🏻 If we want to have it in a single line, it is necessary to use `/* comments */` instead of `// comments`

owncloudApp/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.kt
```diff
@@ -80,7 +80,7 @@ class FileDataStorageManager(
         result
     }
 
-    // TODO: New_arch: Remove this and call usecase inside FilesViewModel
+    // To do: New_arch: Remove this and call usecase inside FilesViewModel
```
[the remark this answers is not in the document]
joragua: I forgot it. 😅 Not a specific reason. I've added `To do` in all comments.

owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListViewModel.kt
joragua: Replaced `flow{}` by `flowOf()`. In the project, we always use the second option.

JuancaG05 (approved):

jesmrec:
Will test:

- [ ] Code smell in ownCloudApp
- [ ] Code smell in ownCloudDomain
- [ ] Code smell in ownCloudData
- [ ] Code smell in ownCloudTestUtil
- [ ] Code smell in ownCloudComLibrary
- [ ] Code smell in all modules

jesmrec:
This is OK to go. This first iteration:

- Will have a complete set of code smells fixed, but others are pending. New issue to be created

- The execution of detekt is sequential through the app modules. If in one module are detected code smells, the execution will stop. We will check if this behaviour could be improved somehow.

Issue #4252, opened 2023-12-11
[FEATURE REQUEST] `Auto-uploads` string in  Settings


jesmrec:
Opening the `Settings` view:

![Screenshot 2023-12-11 at 09 44 21](https://github.com/owncloud/android/assets/14894746/ee45ec03-2a1f-4002-ad3a-6bd8f63e8a7b)

i don't get the point of `Auto-uploads` instead of `Auto uploads` 

I would get rid of the dash inbetween.

Any reason for this???

Checking the git-blame, @EParzefall is the author. 

Could you clarify the reason? thanks

EParzefall:
"auto" alone is ambiguous, but of course nobody would think you could upload automobiles from your Android phone to ownCloud. Still, "auto" as short form of automatic is usually directly attached to the word it refers (as in autobiography)  to or used with a hyphen if the combination is not so common yet as in auto-suggest or auto-complete.
Check out the definition in Webster: https://www.merriam-webster.com/dictionary/aut-

jesmrec:
would it be a good option to call them `Automatic uploads` instead of `Auto uploads`? 

In the past, we used to call them `Instant uploads` because pictures from camera were uploaded straight from the camera to the cloud. With the Android restrictions starting in Android6, the 15 minutes gap forced us to change the wording.

CC @michaelstingl @Aitorbp @JuancaG05 @TheOneRing

EParzefall:
> would it be a good option to call them `Automatic uploads` instead of `Auto uploads`?

Absolutely!

JuancaG05:
Whatever is the option taken, I think we should align with the iOS app, which is the other client that has the automatic uploads feature. AFAIK, in the iOS app it's `Auto Uploads` currently.

michaelstingl:
> Whatever is the option taken, I think we should align with the iOS app, which is the other client that has the automatic uploads feature. AFAIK, in the iOS app it's `Auto Uploads` currently.

@hosy could you help align this?

/cc @tbsbdr

jesmrec:
Options to mind:

- Auto uploads
- Autouploads
- Automatic uploads
- Auto-upload
- Other ... ¿?

mmattel:
From a doc pov and for translations, I propose to use `Automatic uploads`.

mmattel:
@jesmrec Note, as the text is now more precise about what it is for...

`Automatic upload pictures` --> `Automatic picture upload`
`Automatic upload videos` --> `Automatic video upload`

[...]

PR #4492, opened 2024-10-08
[FEATURE REQUEST] Replace "auto-uploads" with "automatic uploads"

jesmrec:
## Related Issues

App: https://github.com/owncloud/android/issues/4252

_____

### DIFF
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -58,9 +58,9 @@
     <string name="prefs_touches_with_other_visible_windows_summary">Allow touches when the view is obscured by another visible window. Enable it to use light filtering apps.</string>
     <string name="confirmation_touches_with_other_windows_title">Are you sure you want to enable this feature?</string>
     <string name="confirmation_touches_with_other_windows_message">Use this feature at your own risk. A malicious application could try to spoof you into unknowingly performing some actions, using other views.</string>
-    <string name="prefs_subsection_picture_uploads">Auto-upload pictures</string>
+    <string name="prefs_subsection_picture_uploads">Automatic picture uploads</string>
     <string name="prefs_subsection_picture_uploads_summary">Manage location and behavior of the pictures uploaded automatically</string>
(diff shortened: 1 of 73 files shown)


jesmrec:
Replace all "Camera uploads" with "Automatic uploads", not only in the UI but also in the code. 

Just one exception: names in file providers and database references were not change to prevent unnecessary migrations. Let me know if this is a good or a bad decision.

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -58,9 +58,9 @@
     <string name="prefs_touches_with_other_visible_windows_summary">Allow touches when the view is obscured by another visible window. Enable it to use light filtering apps.</string>
     <string name="confirmation_touches_with_other_windows_title">Are you sure you want to enable this feature?</string>
     <string name="confirmation_touches_with_other_windows_message">Use this feature at your own risk. A malicious application could try to spoof you into unknowingly performing some actions, using other views.</string>
-    <string name="prefs_subsection_picture_uploads">Auto-upload pictures</string>
+    <string name="prefs_subsection_picture_uploads">Automatic picture uploads</string>
     <string name="prefs_subsection_picture_uploads_summary">Manage location and behavior of the pictures uploaded automatically</string>
-    <string name="prefs_subsection_video_uploads">Auto-upload videos</string>
+    <string name="prefs_subsection_video_uploads">Automatic video uploads</string>
```
JuancaG05, line 63: I think you need to remove all the translations with the translations editor, othwerise Transifex won't detect that this is a new String and won't ask to translate it

JuancaG05 (changes requested):
Just one detail @jesmrec 👍

owncloudApp/src/main/res/values/strings.xml
jesmrec: You're right, done!

JuancaG05 (approved):
LGTM!

joragua:
Approved from my side. All works correctly 🚀

PR #4497, opened 2024-10-11
Feature/add text labels bottom nav bar

Yogeshjindal:
Added text labels to the bottom nav bar
Used default activeStateIndicator which will show which label is selected 

Now, all the items in nav bar shows the labels and upon selection the activeIndicator is enabled which shows that which item is selected.
Fix#4484


changelog/unreleased/4484
```diff
@@ -0,0 +1,4 @@
+Enhancement - Added text labels and active indicator for BottomNavigationView
```
JuancaG05, line 1: To fulfill with the format, it should be like `Enhancement: Added...`

JuancaG05, line 1: Actually, the active indicator was already added, so let's just mention the labels in this case 😄

JuancaG05 (changes requested):
Hey @Yogeshjindal! Thanks for this new PR! Let's take a look at the points I highlighted in the previous one:
- Branch naming is perfect! ✅ 
- Conventional commits are not well used here. There are 4 commits, of which 3 are not useful. We only need the second one, that should be named `feat: added text labels and active indicator to BottomNavigationView` (for example) to follow the convention. I recommend removing those 3 non-necessary commits to avoid confusions (see the comment below), seems you were importing some Material stuff that you didn't need in the end 👍.
- Check the comments I added about the Calens file and fix it 😄
- There's also a release note to add. To do this, you just have to add a new `ReleaseNote` object to the list in `ReleaseNoteViewModel.kt`, creating new strings for that

Feel free to open a new PR or correcting it in this same branch. In the next comment I'll show you how to remove commits 👍

JuancaG05:
A quick explanation about removing commits so that you learn for a future too:

1. Using the `Terminal` from Android Studio, run the command:
```
git rebase -i HEAD~5
```
The number that goes together with `HEAD` is the number of commits you want to work with, but since this PR has 4, 5 is enough

2. Replace `pick` by `drop` in those commits you want to discard (remember this is a vim editor, you have to use arrows to move, `i` to enter writing mode and `Esc` to exit the writing mode)
4. Exit the writing mode, write `:wq` and press `Enter`
5. Run the command to push it to GitHub (`-f` because otherwise you won't be able to push it since it will detect conflicts):
```
git push -f
```

After this, you should have in this PR only the commits we want and I'll be able to review it again 😊

Yogeshjindal:
@JuancaG05 Sir,Thanks a lot for sharing this valuable information.

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -832,5 +832,7 @@
     <string name="audio_preview_label">Audio preview</string>
     <string name="details_label">Details</string>
     <string name="text_preview_label">Text preview</string>
+    <string name="release_notes_title_enhanced_bottom_nav_bar">Added text labels on bottom nav bar</string>
```
JuancaG05, line 835: This is going to be read by all users, so let's not use technical language. Bottom bar is OK 👍 
```suggestion
    <string name="release_notes_title_enhanced_bottom_nav_bar">Added text labels on bottom bar</string>
```

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -832,5 +832,7 @@
...
+    <string name="release_notes_subtitle_bottom_nav_bar">Text labels were added and default active indicator is used to show which label is selected on the bottom nav bar</string>
```
JuancaG05, line 836:
```suggestion
    <string name="release_notes_subtitle_bottom_nav_bar">Text labels were added and default active indicator is used to show which section is selected on the bottom bar</string>
```

owncloudApp/build.gradle
```diff
@@ -34,7 +34,7 @@ dependencies {
     implementation libs.androidx.work.runtime.ktx
     implementation(libs.androidx.browser) { because "CustomTabs required for OAuth2 and OIDC" }
     implementation(libs.androidx.enterprise.feedback) { because "MDM feedback" }
-
+
```
JuancaG05, line 37: You are keeping this change, it's not needed

JuancaG05 (changes requested):
Hey @Yogeshjindal, check this new CR round I made. Conventional commits are not correctly used yet, I insist on following the instructions I posted in the message above to remove commits that are not needed 👍.
Also, this needs a rebase against `master`, there was new content added there.

JuancaG05:
Hi @Yogeshjindal, I see there are more and more commits, but none of them are following conventional commits guidelines as I indicated. Also, there are lots of commits and most of them are not useful. Rebase is not properly done as well, you merged `master` into this branch and that's not what's supposed to happen 😅.

Please, refer to this comment https://github.com/owncloud/android/pull/4497#issuecomment-2407332140 where I explain how to remove commits, and leave this PR with just the needed commits written in conventional commits style. Otherwise, I will be forced to close this PR since it doesn't follow the required guidelines 😞.

Issue #4478, opened 2024-09-24
[FEATURE REQUEST] Update gradle version


jesmrec:
We are currently using gradle plugin `v7.4.2`

This is pretty much outdated and we could try to put it in a newer version. 

https://developer.android.com/build/releases/gradle-plugin

I wouldn't set the newest one but also, i'd keep such old one.

I guess, after the update, if app builds correctly we can give that as correct. 

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
  - [ ] Implement unit tests (if needed)
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

AwaisKhan128:
@jesmrec May I move on with it to upgrade gradle version?

jesmrec:
@AwaisKhan128 sorry, it was already assigned, i have just added it.

JuancaG05:
Just to make clear: Gradle and Android Gradle Plugin are different things.

- **Gradle:** it's the base build system to compile the whole project. It can be set in Android Studio going to `File > Project Structure`. Currently we're using the **7.5 version**.
- **Android Gradle Plugin:** it's an extension for Gradle that provides the necessary tools to build Android apps. It can be set in `libs.version.toml` in the `androidGradlePlugin` variable. Currently we're using the **7.4.2 version**.

When we update the Android Gradle Plugin version, we also need to update the Gradle version ([see compatibilities](https://developer.android.com/build/releases/gradle-plugin?buildsystem=ndk-build#updating-gradle))

JuancaG05:
There is also an inter-dependency with the Java JDK version used.
- **JDK:** it's the Java version used to build and create the Android app. It can be set going to `Android Studio > Settings > Build, Execution, Deployment > Build Tools > Gradle > Gradle JDK`. Currently we're using **JBR 17**, which stands for JetBrains Runtime and is an enhanced JDK of the corresponding version which includes optimizations for Android Studio.

JuancaG05:
There's a fantastic tool embedded in Android Studio that allows an assisted upgrade of the AGP version, with everything it implies (breaking changes, updating other necessary dependencies...). It lets the user decide about several questions introduced in newer AGP versions, and then performs automatically the necessary changes. This tool can be accessed by going to `Tools > AGP Upgrade Assistant...`.

[...]

PR #4507, opened 2024-11-04
[TECHNICAL] Update Android Gradle Plugin version

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4478

_____

## QA

https://github.com/owncloud/android/pull/4507#issuecomment-2454737622

### DIFF
--- a/changelog/unreleased/4507
+++ b/changelog/unreleased/4507
@@ -1,0 +1,7 @@
+Enhancement: Upgraded AGP version to 8.7.2
+
+The Android Gradle Plugin version has been upgraded to 8.7.2, together with Gradle version (updated to 8.9) and
+JDK version (updated to JBR 17).
+
+https://github.com/owncloud/android/issues/4478
+https://github.com/owncloud/android/pull/4507
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ ownCloud admins and users.
 
 * Change - Replace auto-uploads with automatic uploads: [#4252](https://github.com/owncloud/android/issues/4252)
 * Enhancement - Add status message when (un)setting av. offline from preview: [#4382](https://github.com/owncloud/android/issues/4382)
+* Enhancement - Upgraded AGP version to 8.7.2: [#4478](https://github.com/owncloud/android/issues/4478)
 * Enhancement - Added text labels for BottomNavigationView: [#4484](https://github.com/owncloud/android/issues/4484)
 
 ## Details
@@ -56,6 +57,14 @@ ownCloud admins and users.
 
    https://github.com/owncloud/android/issues/4382
    https://github.com/owncloud/android/pull/4482
+
+* Enhancement - Upgraded AGP version to 8.7.2: [#4478](https://github.com/owncloud/android/issues/4478)
+
+   The Android Gradle Plugin version has been upgraded to 8.7.2, together with
+   Gradle version (updated to 8.9) and JDK version (updated to JBR 17).
+
+   https://github.com/owncloud/android/issues/4478
+   https://github.com/owncloud/android/pull/4507
 
 * Enhancement - Added text labels for BottomNavigationView: [#4484](https://github.com/owncloud/android/issues/4484)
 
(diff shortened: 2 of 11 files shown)


changelog/unreleased/4507
```diff
@@ -0,0 +1,7 @@
+Enhancement: Upgraded AGP version to 8.7.2
+
+The Android Gradle Plugin version was upgraded to 8.7.2, together with Gradle version (updated to 8.9) and
```
joragua, line 3:
```suggestion
The Android Gradle Plugin version has been updated to 8.7.2, together with Gradle version (updated to 8.9) and
```
This should be in present perfect passive, right?

joragua (changes requested):
A tiny comment to check here @JuancaG05! 🚀

changelog/unreleased/4507
JuancaG05: Done

joragua (approved):
LGTM! Ready for QA ✅

DeepDiver1975 (approved):
:+1:

jesmrec:
### (1)

Using Android Studio Ladybug to generate a signed APK:

Lint is returning errors that don't allow to build:

```
Lint found fatal errors while assembling a release target.

Fix the issues identified by lint, or create a baseline to see only new errors.
To create a baseline, run `gradlew updateLintBaseline` after adding the following to the module's build.gradle file:

android {
    lint {
        baseline = file("lint-baseline.xml")
    }
}

For more details, see https://developer.android.com/studio/write/lint#snapshot

owncloudApp/src/main/java/com/owncloud/android/workers/DownloadFileWorker.kt:308: Error: Missing PendingIntent mutability flag [UnspecifiedImmutableFlag]
            PendingIntent.FLAG_ONE_SHOT
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~
owncloudApp/src/main/java/com/owncloud/android/workers/DownloadFileWorker.kt:329: Error: Missing PendingIntent mutability flag [UnspecifiedImmutableFlag]
            0
            ~

   Explanation for issues of type "UnspecifiedImmutableFlag":
   Apps targeting Android 12 and higher must specify either FLAG_IMMUTABLE or
   FLAG_MUTABLE when constructing a PendingIntent.

   FLAG_IMMUTABLE is available since target SDK 23, and is almost always the
   best choice. See
   https://developer.android.com/guide/components/intents-filters#CreateImmuta
   blePendingIntents for a list of common exceptions to this rule.

   https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability
   https://goo.gle/UnspecifiedImmutableFlag

2 errors, 0 warnings
```

the problem seems to be more related with the android and SDK versions than the gradle one.. but `v4.4.1 ` was correctly built with the same method.

Debug builds (play button) work fine

DeepDiver1975:
so apk build is not part of ci ... otherwise this would have been detected by ci - or am I missing anything? THX

DeepDiver1975:
Is the difference signed vs not signed or debug vs release?

jesmrec:
the difference in that case is signed vs not signed. It's posible to build release unsigned artifacts

JuancaG05:
(1) should be fixed now @jesmrec

jesmrec:
(1) is fixed

jesmrec:
### (2)

The branding build system seems to need kind of update. It fails with the following trace:

```
* What went wrong:
A problem occurred evaluating project ':owncloudApp'.
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 17 to run. You are currently using Java 11.
      Your current JDK is located in /usr/lib/jvm/java-11-openjdk-amd64
      You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

we'd need someone with access to the machine to update Java version to 17. Otherwise, it will not be posible to build branded artifacts anymore. Who is in change of this, @DeepDiver1975 ?

jesmrec:
(2) will be fixed in the time till next release is done. Without a fix, the branding system will not work anymore. But, the app will go on with the new build setup. 

Let's move

PR #4524, opened 2024-12-10
Main

Franlexa:
## Related Issues
App:

_____

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+  - repo: local
+    hooks:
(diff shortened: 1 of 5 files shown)


JuancaG05:
Hey @Franlexa! I see you're playing a bit with Detekt 😀, we're already working to introduce the tool in our CI system, so I'm closing this PR since I can see no clear aim with it. Also, when opening a new PR try to be descriptive in its title, so that we understand better what it is about. If you want to submit any contribution this way (always better if related to an issue), you will be welcome! 🍻

Issue #4521, opened 2024-12-04
[TECHNICAL] Technical improvements for user quota


joragua:
Right now, the user quota is fetched (`getStoredQuota` method) from the database multiple times in `DrawerActivity` because it is not get as a `Flow`. In this issue we are going to improve performance by fetching the user quota as `Flow` and remove unnecessary calls from the activity.

### TASKS

- [x] Research (if needed)
- [x] Create branch technical/technical_improvements_for_quota
- [x] Development tasks
  - [x] Create a new use case (`GetStoredQuotaAsStreamUseCase`)
  - [x] Remove unnecessary calls from `DrawerActivity`
  - [x] Fix some elements from drawer
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch technical/technical_improvements_for_quota

joragua:
Apart from technical work, the progress bar in the drawer has been removed when the quota is unlimited

<img width="278" alt="Captura de pantalla 2025-01-09 a las 10 40 25" src="https://github.com/user-attachments/assets/7a9740a7-7358-463a-ab42-c8d218aadb7b" />

PR #4525, opened 2024-12-11
[TECHNICAL] Technical improvements for user quota

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4521

_____

## QA

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.5/Quota%20Refactor%20(4525).md

- [X] (1) Color of the bar in manage accounts, when it is `Storage nearly full` https://github.com/owncloud/android/pull/4525#issuecomment-2580262783 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
@@ -32,11 +32,15 @@ import com.owncloud.android.data.providers.LocalStorageProvider
 import com.owncloud.android.domain.user.model.UserQuota
 import com.owncloud.android.domain.user.usecases.GetStoredQuotaAsStreamUseCase
 import com.owncloud.android.domain.user.usecases.GetUserQuotasUseCase
+import com.owncloud.android.domain.utils.Event
+import com.owncloud.android.extensions.ViewModelExt.runUseCaseWithResult
 import com.owncloud.android.presentation.authentication.AccountUtils
 import com.owncloud.android.providers.ContextProvider
 import com.owncloud.android.providers.CoroutinesDispatcherProvider
 import com.owncloud.android.usecases.accounts.RemoveAccountUseCase
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
@@ -50,7 +54,19 @@ class DrawerViewModel(
     accountName: String,
 ) : ViewModel() {
 
-    val userQuota: Flow<UserQuota> = getStoredQuotaAsStreamUseCase(GetStoredQuotaAsStreamUseCase.Params(accountName))
+    private val _userQuota = MutableStateFlow<Event<UIResult<Flow<UserQuota?>>>?>(null)
+    val userQuota: StateFlow<Event<UIResult<Flow<UserQuota?>>>?> = _userQuota
+
+    init {
+        runUseCaseWithResult(
+            coroutineDispatcher = coroutinesDispatcherProvider.io,
+            requiresConnection = false,
+            showLoading = true,
+            flow = _userQuota,
+            useCase = getStoredQuotaAsStreamUseCase,
+            useCaseParams = GetStoredQuotaAsStreamUseCase.Params(accountName = accountName),
+        )
+    }
 
     fun getAccounts(context: Context): List<Account> {
         return AccountUtils.getAccounts(context).asList()
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
@@ -45,19 +45,18 @@ import kotlinx.coroutines.launch
 import timber.log.Timber
 
 class DrawerViewModel(
-    getStoredQuotaAsStreamUseCase: GetStoredQuotaAsStreamUseCase,
+    private val getStoredQuotaAsStreamUseCase: GetStoredQuotaAsStreamUseCase,
     private val removeAccountUseCase: RemoveAccountUseCase,
     private val getUserQuotasUseCase: GetUserQuotasUseCase,
     private val localStorageProvider: LocalStorageProvider,
     private val coroutinesDispatcherProvider: CoroutinesDispatcherProvider,
     private val contextProvider: ContextProvider,
-    accountName: String,
 ) : ViewModel() {
 
     private val _userQuota = MutableStateFlow<Event<UIResult<Flow<UserQuota?>>>?>(null)
     val userQuota: StateFlow<Event<UIResult<Flow<UserQuota?>>>?> = _userQuota
 
-    init {
+    fun getUserQuota(accountName: String) {
         runUseCaseWithResult(
             coroutineDispatcher = coroutinesDispatcherProvider.io,
             requiresConnection = false,
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
@@ -25,53 +25,35 @@ package com.owncloud.android.presentation.common
 
 import android.accounts.Account
 import android.content.Context
-import androidx.lifecycle.LiveData
-import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.owncloud.android.R
 import com.owncloud.android.data.providers.LocalStorageProvider
 import com.owncloud.android.domain.user.model.UserQuota
-import com.owncloud.android.domain.user.usecases.GetStoredQuotaUseCase
+import com.owncloud.android.domain.user.usecases.GetStoredQuotaAsStreamUseCase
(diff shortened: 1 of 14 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
```diff
@@ -25,65 +25,43 @@ package com.owncloud.android.presentation.common
...
-        accountName: String
-    ) = runUseCaseWithResult(
-        coroutineDispatcher = coroutinesDispatcherProvider.io,
-        requiresConnection = false,
-        showLoading = true,
-        liveData = _userQuota,
-        useCase = getStoredQuotaUseCase,
-        useCaseParams = GetStoredQuotaUseCase.Params(accountName = accountName)
-    )
+    val userQuota: Flow<UserQuota> = getStoredQuotaAsStreamUseCase(GetStoredQuotaAsStreamUseCase.Params(accountName))
```
JuancaG05, line 53: Hm, not sure if I like this... I think it would be better being encapsulated with a private variable. Otherwise everytime we access `userQuota`, the use case might be executed, and that's not what we want. Check for example `SettingsPictureUploadsViewModel` to see how this is done with the `pictureUploads` variable 👍

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
```diff
@@ -82,7 +83,11 @@ import timber.log.Timber
  */
 abstract class DrawerActivity : ToolbarActivity() {
 
-    private val drawerViewModel by viewModel<DrawerViewModel>()
+    private val drawerViewModel by viewModel<DrawerViewModel> {
+        parametersOf(
+            account?.name
```
JuancaG05, line 88: What if `account` is null? Did you check where this value is set in `BaseActivity`?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
```diff
@@ -305,115 +309,105 @@ abstract class DrawerActivity : ToolbarActivity() {
...
+                            getString(R.string.drawer_quota),
+                            DisplayUtils.bytesToHumanReadable(userQuota.used, this, true),
+                            DisplayUtils.bytesToHumanReadable(userQuota.getTotal(), this, true),
+                            userQuota.getRelative()
+                        )
+                        getAccountQuotaStatusText()?.visibility = View.GONE
                     }
                 }
-                is UIResult.Loading -> getAccountQuotaText()?.text = getString(R.string.drawer_loading_quota)
-                is UIResult.Error -> getAccountQuotaText()?.text = getString(R.string.drawer_unavailable_used_storage)
```
JuancaG05, line 416: We're losing these 2 cases!! We shouldn't, they are so necessary

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
```diff
@@ -305,115 +309,105 @@ abstract class DrawerActivity : ToolbarActivity() {
...
-                                    getAccountQuotaText()?.text = String.format(
-                                        getString(R.string.drawer_quota),
-                                        DisplayUtils.bytesToHumanReadable(userQuota.used, this, true),
-                                        DisplayUtils.bytesToHumanReadable(userQuota.getTotal(), this, true),
-                                        userQuota.getRelative()
-                                    )
-                                    getAccountQuotaStatusText()?.visibility = View.GONE
-                                }
-                            }
+        collectLatestLifecycleFlow(drawerViewModel.userQuota) { userQuota ->
```
JuancaG05, line 312: I did some changes in all these lines in the Detekt branch, I think just for the `MaxLineLength` rule. Try to keep them in the new place for these lines so that we don't duplicate work

owncloudData/src/main/java/com/owncloud/android/data/user/db/UserDao.kt
```diff
@@ -33,6 +35,11 @@ interface UserDao {
         accountName: String
     ): UserQuotaEntity?
 
+    @Query(SELECT_QUOTA)
+    fun getQuotaForAccountAsFlow(
+        accountName: String
+    ): Flow<UserQuotaEntity>
```
JuancaG05, line 41: Why with the same SQL query you have a nullable result in the previous method but not here?

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCLocalUserDataSourceTest.kt
```diff
@@ -82,12 +86,16 @@ class OCLocalUserDataSourceTest {
...
     @Test
-    fun `deleteQuotaForAccount removes user quota correctly`() {
+    fun `getQuotaForAccountAsFlow returns a Flow with an UserQuota`() = runTest {
+        every {
+            ocUserQuotaDao.getQuotaForAccountAsFlow(OC_ACCOUNT_NAME)
+        } returns flowOf(userQuotaEntity)
 
-        ocLocalUserDataSource.deleteQuotaForAccount(OC_ACCOUNT_NAME)
+        val userQuota = ocLocalUserDataSource.getQuotaForAccountAsFlow(OC_ACCOUNT_NAME).first()
+        assertEquals(userQuotaEntity.toModel(), userQuota)
```
JuancaG05, line 95:
```suggestion
        assertEquals(OC_USER_QUOTA, userQuota)
```

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCLocalUserDataSourceTest.kt
```diff
@@ -104,4 +112,29 @@ class OCLocalUserDataSourceTest {
...
     }
+
+    @Test
+    fun `getAllUserQuotasAsFlow returns a Flow with a list of UserQuota`() = runTest {
+        every {
+            ocUserQuotaDao.getAllUserQuotasAsFlow()
+        } returns flowOf(listOf(userQuotaEntity))
+
+        val listOfUserQuotas = ocLocalUserDataSource.getAllUserQuotasAsFlow().first()
+        assertEquals(listOf(userQuotaEntity.toModel()), listOfUserQuotas)
```
JuancaG05, line 123:
```suggestion
        assertEquals(listOf(OC_USER_QUOTA), listOfUserQuotas)
```

JuancaG05 (changes requested):
Good @joragua!! Let's do some changes here 👍

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCLocalUserDataSourceTest.kt
joragua: Done ✅

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCLocalUserDataSourceTest.kt
joragua: Same as previous comment 😄

owncloudData/src/main/java/com/owncloud/android/data/user/db/UserDao.kt
joragua: Added nullability to keep the same structure in both SQL queries

owncloudApp/src/main/java/com/owncloud/android/presentation/common/DrawerViewModel.kt
joragua: Done! 😄

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
joragua: Done ✅

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
joragua: In this case, `account` variable cannot be null because is set in `swapToDefaultAccount` method from `BaseActivity` before creating the `DrawerViewModel`. This is the same with other view models like `CapabilityViewModel` or `ShareViewModel`

owncloudApp/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.kt
joragua: I've replace `BaseUseCase` by `BaseUseCaseWithResult` and now, we have same cases than before

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCLocalUserDataSourceTest.kt
JuancaG05: Not done actually 😬

JuancaG05 (changes requested):
Just one thing that you didn't apply from the previous round of CR and we're done @joragua 👍. Also, you can rebase the branch so that we have it ready for QA

JuancaG05 (approved):
LGTM!

joragua:
**(1)** should be done with the last commit @jesmrec

jesmrec:
Approved on my side.

Issue #4340, opened 2024-03-14
[BUG] Up button in navigation in folder picker for auto-uploads not appearing


joragua:
### Actual behaviour
- Up button in navigation for auto-uploads (video/pictures) settings doesn't appear on the screen. This bug it's only in the root screen.

https://github.com/owncloud/android/assets/160487817/2b55161e-d52f-41f5-9415-14a9d5923d28

### Expected behaviour
- The app should show the up arrow navigation for auto-uploads settings. 
### Steps to reproduce
1. Open the app
2. Settings
3. Auto-uploads (pictures/video)
4. Activate pictures/video uploads (if it's not activated)
5. Picture/video upload path

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test) 

Yes, but in an OCIS Server.

### Environment data
Android version: 13

Device model: Motorola Moto G14

Stock or customized system: Stock

ownCloud app version: 4.2.1

ownCloud server version: Doesn't matter

jesmrec:
Reopening, not completed, QA pending

jesmrec:
Now approved. Closing.

PR #4535, opened 2025-01-23
[FIX] Up button in navigation in folder picker for auto-uploads not appearing

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4340

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ ownCloud admins and users.
 
 ## Summary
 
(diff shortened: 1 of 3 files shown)


jesmrec:
Checking folder picker navigation from root to spaces in:

- [X] Camera uploads
- [X] Copy (regression)
- [X] Share from 3rd party app (regression)
- [X] Share text (regression)
- [x] Share image (regression)
- [X] Move (regression, no arrow in this case) 

Approved.

PR #4543, opened 2025-02-06
[TECHNICAL] Send just one Detekt command against all modules

jesmrec:
After latest check, just triggering `./gradlew detekt` is enough to check all the modules at the time.

If any smell raises up in a module, the execution will stop. The only way to make parallel execution is creating 5 different jobs.


JuancaG05 (approved):
👍

PR #4555, opened 2025-03-05
Update strings.xml - Fix typo in values-it

andymeneguzzo:
Fix typo in values-it
_____


DeepDiver1975:
We maintain translations on transifex. I will fix this string there.

Issue #4568, opened 2025-04-09
[a11y] Accessibility reports in 4.5.1


JuancaG05:
Running an accessibility analysis for the current version 4.5.1, using the official Google tool [Accessibility Scanner](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor), we found there are some reports newly added in some of the screens that the original report covers:

- [x] **File list:** the "Close" button represented by a cross in the bottom sheet of the 3-dot menu doesn't have a content description
- [x] **Share:** when there is more than 1 row in the list of private shares and public links, all the equal buttons among rows have the same content description
- [x] **Manage accounts:** when there is more than 1 row in the list of accounts, all the equal buttons among rows have the same content description

The other screens checked (**Drawer menu**, **Spaces list** and **Image preview**) are OK in terms of accessibility.

PR #4573, opened 2025-04-15
[a11y] Accessibility reports in 4.5.1

JuancaG05:
## Related Issues
App: https://github.com/owncloud/android/issues/4568

_____


joragua (approved):
LGTM! 👍🏻

jesmrec:
>  File list: the "Close" button represented by a cross in the bottom sheet of the 3-dot menu doesn't have a content description

Fixed ✅ 

>  Share: when there is more than 1 row in the list of private shares and public links, all the equal buttons among rows have the same content description

Fixed ✅ 

> Manage accounts: when there is more than 1 row in the list of accounts, all the equal buttons among rows have the same content description

Fixed ✅ 


Let's go with this one

Issue #4513, opened 2024-11-14
[BUG] Side menu collapses info in landscape


jesmrec:
### Actual behaviour

![Screenshot 2024-11-14 at 12 06 54](https://github.com/user-attachments/assets/c1a4e6bc-35ff-47ef-8f7b-bbd90338bccb)

### Expected behaviour

Any way to scroll and show the information there correctly
 
### Steps to reproduce

1. Open the drawer and set the device in landscape orientation 

`v4.4.1`
Xiaomi Redmi 13, Android 14

joragua:
Check previous versions

joragua:
This bug comes from two different PR:

1) https://github.com/owncloud/android/pull/4454. In this PR, the last two items from `drawer_menu.xml` were removed in order to exclude them from Talkback. With this change, the drawer (opened from the beginning in landscape mode) is not set correctly and the scroll doesn't work. We've decided to revert this change and have the last two empty items in the menu (only for UI/UX purposes) with a new property `android:importantForAccessibility="no"`
(https://developer.android.com/guide/topics/ui/accessibility/principles#decorative)

2) https://github.com/owncloud/android/pull/4467. At the same time, in this PR a new property was added to the `manifest` in order not to restart the `FileDisplayActivity` when the device orientation changed. As a result, the activity doesn't use a different `nav_drawer_footer.xml` depending on the orientation because it is not restarted. So, at this point, we've decided to keep this behavior because the drawer is not collapsed anymore with the first fix.

PR #4580, opened 2025-04-29
[FIX] Side menu collapses info in landscape

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4513

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ ownCloud admins and users.
 
 ## Summary
 
+* Bugfix - Side menu collapses info in landscape: [#4513](https://github.com/owncloud/android/issues/4513)
 * Bugfix - Content in Spaces not shown from third-party apps: [#4522](https://github.com/owncloud/android/issues/4522)
 * Bugfix - Add bottom margin for used quota in account dialog: [#4566](https://github.com/owncloud/android/issues/4566)
 * Change - Bump target SDK to 35: [#4529](https://github.com/owncloud/android/issues/4529)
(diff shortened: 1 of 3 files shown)


jesmrec (approved):

jesmrec:
LGTM 👍

jesmrec:
QA checks:

- [ ] Drawer in portrait
- [ ] Drawer in landscape
- [ ] Drawer open and changing orientation
- [ ] Talkback in the drawer
- [ ] Keyboard moving through the drawer
- [ ] Logo + Link in portrait
- [ ] Logo + Link in landscape
- [ ] Logo + Link in drawer open and changing orientation

jesmrec:
Since there is no a totally fittable solution, we'll go for the first approach:

> Keep the bad numbering dictation that allows the UI glitch to be removed

there are other risky solutions that can lead to other bigger problems and bad UI/UX situations that are not worthy at this point.

Finally:

- We assume that numeration of items in the drawer is not totally correct and excedes the number of options. It could lead to confusion to visual impaired users, but, it does not hurt the UX.
- Drawer's look improved in both orientations, letting users to scroll options without overlapped text.

As usual, new feedback to receive is always welcome and we are open to new improvements and iterations in the feature. 

Agreed @joragua?

joragua:
Yeah, I agree. There could be another compromise solution but I think that, at the moment we are, is not feasible: modify the drawer layout and have an empty block between the menu and the footer (for example). I've tried it but the drawer collapses many times so, in my opinion, it is not worthy to work on it. 🤔 In this case, I would prefer to keep the bad numbering dictation from Talkback and fix the drawer glitch.

jesmrec:
Approved, then.

Issue #4080, opened 2023-06-19
[BUG] Token request with `Bearer` returns error


jesmrec:
### Steps to reproduce

1. Add an oCIS account
2. Add another oCIS account over the same server
3. Repeat operation 2. 

### Actual behaviour

After adding >1 accounts on the same oCIS server, error `Unsuccessful authorization` in login view. This is happening because `/konnect/v1/token` request sends the `Authorization` header with a `Bearer` token instead of `Basic`.  Token sent in request header is not valid yet because a valid token will come from the subsequent response.

### Expected behaviour
Users can add as many accounts as they wish over the same or different server with no errors.
 
[owncloud.2023-06-19.log](https://github.com/owncloud/android/files/11788021/owncloud.2023-06-19.log)

Pixel 2
Android 11
`b86a19ea9`

joragua:
### Things to keep in mind 🤔

- All operations/requests have an `Authorization` header that is set in the client (`OwncloudClient`) 
- `TokenRequestRemoteOperation` is the only operation in which  an `Authorization` header is added before running this method: `val status = client.executeHttpMethod(...)` 

***
The bug is located in these lines (`OwncloudClient`): 
```
if (mCredentials.getHeaderAuth() != null && !mCredentials.getHeaderAuth().isEmpty()) {
       method.setRequestHeader(AUTHORIZATION_HEADER, mCredentials.getHeaderAuth());
}
```
On the first attempt to log in with an account, a `Basic` token will be sent in the `Authorization` header of the request. This is because the credentials have not been stored yet in the client (`OwncloudClient`). If you try to log in with another account that belongs to the same server, the credentials will be already stored and a `Bearer` token will be sent in the `Authorization` header of the request. That is wrong! The token should be `Basic` like the first attempt. The solution is: check in the client if there isn't an `Authorization` header in the request before adding it.

```
if (mCredentials.getHeaderAuth() != null && !mCredentials.getHeaderAuth().isEmpty() && method.getRequest().header(AUTHORIZATION_HEADER) == null) {
       method.setRequestHeader(AUTHORIZATION_HEADER, mCredentials.getHeaderAuth());
}
```

In case the method cames from `TokenRequestRemoteOperation` the header will not be replaced (`Basic` token will be used). In other case, the `Authorization` header will be set with a `Bearer` token (necessary to execute remote operations)

Issue #4332, opened 2024-02-29
[BUG]


individual-it:
### Actual behaviour
- After not using the app for a while I get the message that the access token is expired
![Screenshot_20240229-085550](https://github.com/owncloud/android/assets/2425577/a9eb6e39-4a29-43e0-8701-9b15550b1ec6)
- after clicking on "Anmelden" (login) and going through the login process on the webUI first time an "unknown error" was displayed and then again the same one about the expired token. But still it seemed to got some data, because the file list is updated:
![Screenshot_20240229-085605](https://github.com/owncloud/android/assets/2425577/beec4877-58a8-4c45-adb0-6cb825d686d6)
- The only way to get the app working is to "resync" through the accounts list

### Expected behaviour
- If the token has expired, login again should work
 
### Steps to reproduce
1. don't use the app for a couple of weeks
2. try to use it when the token expired
3. go through the login process

Can this problem be reproduced with the official owncloud server?
(url: https://demo.owncloud.org, user: test, password: test)

### Environment data
Android version:
13
Device model: 
Fairphone 3
Stock or customized system:
stock
ownCloud app version:
4.2.1
ownCloud server version:
ocis 5.0.0-rc5

### Logs
Happy to share the android log if it can be done not in public

JuancaG05:
Hi @individual-it! Thanks for opening a new issue! 🍻 

Sure, you can share the logs privately if you send them to android-app@owncloud.com, they will be very helpful to us 👍

kjeldahl:
I am also seeing this error. 
ownCloud release version 4.5.0 (7bef18e82)

jesmrec:
We are aware of the problem, it's kind of random to reproduce (indeed, i can't). Would it be posible to share logs with the problem? we will provide you a safe way to do it, so that, we can research about it.

TermeHansen:
@jesmrec I just did it now with debugging enabled, and then it failed in first auth, but actually came fully through second time. Do you wan to have a look at this log? How should I share?

jesmrec:
The first failed attemp is the problem you reproduced, right?

You can push the log file [here](https://infinite.owncloud.com/s/VJEBBlrSlPiKUST) pwd: "I54<aV~f)nU0" without quotation marks

TermeHansen:
@jesmrec I just uploaded my log. Yes there was one failed attempt first, and then the second time my browser had the login from last time and it went through. Sometimes this has ended in an endless loop of trials with no success.

[...]

PR #4586, opened 2025-05-22
[FIX] Token request with `Bearer` returns error

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4080

_____

## QA

Checks done: https://github.com/owncloud/android/pull/4586#issuecomment-2909025118

Reports:

- [X] (1) https://github.com/owncloud/android/pull/4586#issuecomment-2912202109 [FIXED]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ ownCloud admins and users.
 
 ## Summary
 
+* Bugfix - Token request with Bearer returns error: [#4080](https://github.com/owncloud/android/issues/4080)
 * Bugfix - Side menu collapses info in landscape: [#4513](https://github.com/owncloud/android/issues/4513)
 * Bugfix - Content in Spaces not shown from third-party apps: [#4522](https://github.com/owncloud/android/issues/4522)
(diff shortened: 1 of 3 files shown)


jesmrec:
I noticed the following effect in the current branch:

When the token expires and `/konnect/v1/token` is called to renew, the process is correct but the following request uses the expired token, causing un `401` and showing in the screen the snackbar dialog to re-authenticate. Just the following request, the following ones are OK and use the correct token. That means, user sees authentication is lost, bit it's not lost.

I realised that this `401` request is always a `PROPFIND` to `/dav/spaces/...` , that is expected but not sure...

After checking with `master` , it's not so reproducible as in the current branch.  Also checked with two devices.

Maybe related with https://github.com/owncloud/android/issues/4332

joragua:
When the tokes expires, the next request after `/konnect/v1/token` includes an `Authorization` header with the expired token. In this case, the header must be replaced with the new token. Therefore, the header replacement should only be applied if the request path is not `/konnect/v1/token`. I will update the code and  it should be fixed 🤔

jesmrec:
Last update works pretty fine! 💯 . No more fake authirization lost messages. Checked that the following request to the `token` one is already using the new token. So, (1) is fixed.

Issue #4588, opened 2025-05-29
[FEATURE REQUEST] Connect vanilla app to Kiteworks servers using no client secret


joragua:
Try to integrate ownCloud vanilla app to Kiteworks servers using no client secret. 

In this case, `PKCE` has to be used as a security measurement because the client secret is not available.

- iOS reference: https://github.com/owncloud/ios-sdk/tree/feature/kiteworks-drive-name

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/kiteworks_integration_without_client_secret
- [x] Development tasks
  - [x] Check if `/status.php` request is not reachable and use `/kwdav/status.php` instead
  - [x] Check if `productName` has the value `kiteworks`
  - [x] Add `/kwdav` prefix before each request except `/.well-known` and `/oauth/token` requests
  - [x] Use `PKCE` when `client-secret` is not available
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/kiteworks_integration_without_client_secret into master

joragua:
- `PKCE` is used by default in oC Android app
- `setup.xml` file must be modified in order to connect to Kiteworks servers
- There are some file operations that are not supported yet (we will open a new issue to integrate them)

___

Steps to establish connection to Kiteworks servers:

1) Check if `/status.php` fails and use `/kwdav/status.php` instead
2) Check if `productname`, included in the response, has the value `kiteworks`
3) Set `isKiteworksServer` variable to true in `OwnCloudClient` (the network client)
4) Add `/kwdav` prefix before each request except `/.well-known` and `/oauth/token` requests

When the token expires there are some additional steps to take into account:

1) If the token has expired, the request returns `401`. That error is handled in `ConnectionValidator.kt`
2) Check if `/kwdav/status.php` fails 
3) Check if `/kwdav/dav/files` fails. If it is a Kiteworks server, this request always returns `404` because there is no personal space. So in this case, there is an additional condition to refresh the token: check if the server is from Kiteworks instead of checking if the status request is `unauthorized`.

NOTE: In oC servers, when the token has expired, all requests return `401`.

PR #4589, opened 2025-06-04
[FEATURE REQUEST] Connect vanilla app to Kws servers using no client secret

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4588

_____

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
@@ -54,6 +54,7 @@ public class OwnCloudClient extends HttpClient {
     public static final String STATUS_PATH = "/status.php";
     private static final String WEBDAV_UPLOADS_PATH_4_0 = "/remote.php/dav/uploads/";
     private static final String KONNECT_V1_TOKEN_PATH = "/konnect/v1/token";
+    public static final String KWDAV_PATH = "/kwdav";
     private static final int MAX_RETRY_COUNT = 2;
 
     private static int sIntanceCounter = 0;
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/StatusRequester.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/status/StatusRequester.kt
@@ -76,11 +76,17 @@ internal class StatusRequester {
 
     fun request(baseLocation: String, client: OwnCloudClient): RequestResult {
         val currentLocation = baseLocation + OwnCloudClient.STATUS_PATH
+        val fallbackLocation = baseLocation + OwnCloudClient.KWDAV_PATH + OwnCloudClient.STATUS_PATH
         var status: Int
-        val getMethod = getGetMethod(currentLocation)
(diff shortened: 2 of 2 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
```diff
@@ -54,6 +54,9 @@ public class OwnCloudClient extends HttpClient {
     public static final String STATUS_PATH = "/status.php";
     private static final String WEBDAV_UPLOADS_PATH_4_0 = "/remote.php/dav/uploads/";
     private static final String KONNECT_V1_TOKEN_PATH = "/konnect/v1/token";
+    public static final String KWDAV_PATH = "/kwdav";
```
jesmrec, line 57: This is just customization, caused by the bridge connection. I'd move this one to `setup.xml` if posible. What do you think?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
```diff
@@ -84,6 +84,7 @@ abstract class HttpBaseMethod constructor(url: URL) {
     }
 
     open fun setUrl(url: HttpUrl) {
+        httpUrl = url
```
jesmrec, line 87: why this assignment?

jesmrec (changes requested):

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/methods/HttpBaseMethod.kt
joragua: When the request is a `DAV` method, the property `httpUrl` is used in the constructor. If `/kwdav` prefix has been added (the request is a new one) it is necessary to refresh that variable. In other case, the prefix will not be added and the request will be malformed.

jesmrec: 👍

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/OwnCloudClient.java
jesmrec: That will be static, so, no customization needed. Nothing to do here.

joragua: `/kwdav` is not configurable at this moment, so it is not necessary to include it in `setup.xml`

jesmrec:
QA checks:

- [x] Basic auth
- [x] OAuth2 with secret id (token granted)
- [x] OAuth2 with secret id (token renewed)
- [x] OAuth2 without secret id (token granted) 
- [x] OAuth2 without secret id (token renewed)
- [x] OIDC (token granted)
- [x] OIDC (token renewed)
- [x] Webfinger
- [x] User light

jesmrec:
Great job! 💯 

Approved on my side

jesmrec (approved):

Issue #4595, opened 2025-06-11
[TECHNICAL] Integrate instrumented tests into GitHub Actions


joragua:
Since Bitrise is KO, try to integrate instrumented tests into GitHub Actions. Unit tests and E2E tests are already integrated into GitHub Actions.

- Unit tests workflow file reference: https://github.com/owncloud/android/blob/master/.github/workflows/android-unit-tests.yml

- E2E tests workflow file reference: https://github.com/jesmrec/ownCloud-E2E-Android-Cucumber-Appium/blob/master/.github/workflows/e2e.yml

### TASKS

- [x] Research (if needed)
- [x] Create branch technical/integration_of_instrumented_tests
- [x] Development tasks
  - [x] Create a yaml file for instrumented tests
  - [x] Check if the new workflow is working well in CI
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch technical/integration_of_instrumented_tests

PR #4602, opened 2025-06-17
[TECHNICAL] Integrate instrumented tests into GitHub Actions 

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4595

_____

### DIFF
--- a/.github/workflows/android-instrumented-tests.yml
+++ b/.github/workflows/android-instrumented-tests.yml
@@ -1,0 +1,41 @@
+name: Android Instrumented Tests
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+
+jobs:
+  instrumented-tests:
+    name: Run Android Instrumented Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Run Instrumented Tests with emulator
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 33
+          target: google_apis
+          arch: x86_64
+          profile: pixel
+          avd-name: instrumented-tests-avd
+          force-avd-creation: true
+          disable-animations: true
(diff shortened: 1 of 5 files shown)


.github/workflows/android-unit-tests.yml
```diff
@@ -2,11 +2,10 @@ name: Android Unit Tests
 
 on:
```
jesmrec, line 6: Over the `on:` clause, you can add the read permission on the flow, as well as in the instrumented tests, in order to prevent warnings from the CI system. Not in the scope of this PR, but, you already did some changes in the yaml file 😄

.github/workflows/android-instrumented-tests.yml
```diff
@@ -0,0 +1,41 @@
+name: Android Instrumented Tests
+
```
jesmrec: I'd add that these tests ron over the DB. UI tests are also instrumented and they don't run here. Suggestions: 

`name: Android Database Instrumented Tests`
`name: Android DB Instrumented Tests`

CHANGELOG.md
```diff
@@ -187,6 +188,15 @@ ownCloud admins and users.
    https://github.com/owncloud/android/issues/4588
    https://github.com/owncloud/android/pull/4589
 
+* Enhancement - Integration of instrumented tests in GitHub Actions: [#4595](https://github.com/owncloud/android/issues/4595)
+
+   A new yaml workflow has been added to run instrumented tests in GitHub Actions
+   in order to have a more consistent CI pipeline in the project since Bitrise is
+   not available.
```
jesmrec: I wouldn't mention BitRise here, just the part of the integration. The fact of integrating tests here does not have to do with other external CI systems, which are not available anymore in this case, but they could be up un running in other cases. Also, `yaml` word is not needed since all workflows here are .yaml ;)

.github/workflows/android-unit-tests.yml
joragua: Added 👍🏻

CHANGELOG.md
joragua: Removed BitRise reference and `yaml` from calens file ✅

.github/workflows/android-instrumented-tests.yml
joragua: In this case, there is a test (`OCLocalAuthenticationDataSourceTest`) that doesn't run over the DB, so I think those names are not appropiated. I'd use `Android Instrumented Data Tests` instead... 🤔

.github/workflows/android-unit-tests.yml
jesmrec: 💯

.github/workflows/android-instrumented-tests.yml
jesmrec: 👍

CHANGELOG.md
jesmrec: :100

jesmrec (approved):

jesmrec:
Everything fine, CR passed and checked the job result. Go ahead!

PR #4621, opened 2025-07-02
[FEATURE REQUEST] Modify sbom workflow to push to the repo

jesmrec:
Changes: 

- Replaced the `read` permission for `write`, since we have to push
- `sbom.json` to be created when pushing to `master`. Before: in every PR, does not matter the target branch.
- `sbom.json` file in root folder of the current repository

- Added a step that will compare the `sbom.json` in `master` with the generated one. If they match, no push (no changes since the latest time).

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -1,21 +1,27 @@
-name: SBOM 
-
-permissions:
-  contents: read
+name: SBOM
 
 on:
   workflow_dispatch:
-  pull_request:
+  push:
+    branches:
+      - master
+      - main
+
+permissions:
+  contents: write
 
 jobs:
   sbom:
     runs-on: ubuntu-latest
-   
+
     steps:
+      # Checkout the full repository history (required to access origin/master)
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          ssh-key: ${{ secrets.DEPLOYMENT_SSH_KEY }}
 
-      # Caches Gradle dependencies to avoid downloading them on every run
+      # Cache Gradle dependencies for faster builds
       - name: Cache Gradle dependencies
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/sbom.yml
```diff
@@ -1,21 +1,27 @@
...
-  contents: read
+name: SBOM
 
 on:
   workflow_dispatch:
-  pull_request:
+  push:
+    branches:
+      - master
+      - main
```
joragua: Is it necessary to add the `main` branch here? Currently, our main branch is called `master`. We might change it in the future, but that would require additional changes in the repository (including this one)

.github/workflows/sbom.yml
```diff
@@ -27,48 +33,55 @@ jobs:
...
+      # Commit and push the new SBOM only if it differs from master
+      - name: Commit files
+        if: steps.diff.outputs.no_changes == 'false'
+        uses: GuillaumeFalourd/git-commit-push@v1.3
         with:
-          name: ${{ steps.vars.outputs.artifact_name }}
-          path: ${{ steps.vars.outputs.artifact_name }}.zip
+          email: devops@owncloud.com
+          name: ownClouders
+          commit_message: "docs: SBOM updated [skip ci]"
```
joragua, line 84: `[skip ci]` is used to avoid infinite loops in CI, right? Since `github.token` is being used here, is it safe to remove it?  🤔

joragua (changes requested):
Some comments here @jesmrec!

.github/workflows/sbom.yml
jesmrec: it could be removed, yes. I added that as an extra... it does not hurt either. Will remove

.github/workflows/sbom.yml
jesmrec: yep. Some (not every) job are not triggered when that tag is detected... we will know when it's merged.

Issue #4622, opened 2025-07-02
[FEATURE REQUEST] Create new set of configurations for Kiteworks servers in setup.xml


joragua:
In order to support connection to Kiteworks servers, we need to define a new set of configurations entries in `setup.xml`. These values should be separate from the default ownCloud settings to avoid conflicts and allow independent configuration.

 ownCloud configurations:

`oauth2_client_id`, `oauth2_client_secret`, `oauth2_redirect_uri_scheme`, `oauth2_redirect_uri_host`, `oauth2_redirect_uri_path`, `oauth2_openid_scope`  and `oauth2_openid_prompt`

Candidate names for Kiteworks:

`kiteworks_client_id`, `kiteworks_client_secret`,`kiteworks_redirect_uri_scheme`, `kiteworks_redirect_uri-host`, `kiteworks_redirect_uri_path`,`kiteworks_openid_scope` and `kiteworks_openid_prompt`

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/new_setup_for_kiteworks_servers
- [x] Development tasks
  - [x] Create new set of configurations for Kiteworks servers in `setup.xml`
  - [x] Adapt the authentication flow for the new setup
  - [x] Fix the logic in `OwnCloudClient` when `/kwdav` prefix is added and the account is attached
  - [x] Adapt the token renewal process for the new setup
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/new_setup_for_kiteworks_servers into master

jesmrec:
Add also `oauth2_openid_prompt`

joragua:
## Summary

- A new set of configurations in `setup.xml` has been created to connect to Kiteworks servers. The names of these variables are:  `kiteworks_client_id`, `kiteworks_client_secret`, `kiteworks_redirect_uri_path`, `kiteworks_redirect_uri_scheme`, `kiteworks_redirect_uri_host`, `kiteworks_openid_scope` and `kiteworks_openid_prompt`

- When connecting to a Kiteworks server, an `OIDCServerConfiguration` is always returned with the property `isKiteworksServer` set to true. This variable is used to determine which configuration should be selected from `setup.xml` file.

- After a succesful login, the property `isKiteworksServer` is attached to the account in the `AccountManager` and then it is used to create the new network client.

[...]

PR #4632, opened 2025-07-08
[FEATURE REQUEST] Create new set of configurations for Kiteworks servers in setup.xml

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4622

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
@@ -53,6 +53,7 @@ import com.owncloud.android.data.authentication.KEY_USER_ID
 import com.owncloud.android.databinding.AccountSetupBinding
 import com.owncloud.android.domain.authentication.oauth.model.ResponseType
 import com.owncloud.android.domain.authentication.oauth.model.TokenRequest
+import com.owncloud.android.domain.authentication.oauth.model.TokenResponse
 import com.owncloud.android.domain.exceptions.NoNetworkConnectionException
 import com.owncloud.android.domain.exceptions.OwncloudVersionNotSupportedException
 import com.owncloud.android.domain.exceptions.SSLErrorCode
@@ -107,6 +108,7 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
     private lateinit var serverBaseUrl: String
 
     private var oidcSupported = false
+    private var isKiteworksServer = false
 
     private lateinit var binding: AccountSetupBinding
 
@@ -548,13 +550,16 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
         val customTabsBuilder: CustomTabsIntent.Builder = CustomTabsIntent.Builder()
         val customTabsIntent: CustomTabsIntent = customTabsBuilder.build()
 
+        val serverInfo = authenticationViewModel.serverInfo.value?.peekContent()?.getStoredData()
+        isKiteworksServer = serverInfo is ServerInfo.OIDCServer && serverInfo.oidcServerConfiguration.isKiteworksServer
+
         val authorizationEndpointUri = OAuthUtils.buildAuthorizationRequest(
             authorizationEndpoint = authorizationEndpoint,
-            redirectUri = OAuthUtils.buildRedirectUri(applicationContext).toString(),
-            clientId = clientId,
+            redirectUri = OAuthUtils.buildRedirectUri(applicationContext, isKiteworksServer).toString(),
+            clientId = if (isKiteworksServer) getString(R.string.kiteworks_client_id) else clientId,
             responseType = ResponseType.CODE.string,
-            scope = if (oidcSupported) mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_SCOPE, R.string.oauth2_openid_scope) else "",
-            prompt = if (oidcSupported) mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_PROMPT, R.string.oauth2_openid_prompt) else "",
+            scope = getScope(),
+            prompt = getPrompt(),
             codeChallenge = authenticationViewModel.codeChallenge,
             state = authenticationViewModel.oidcState,
             username = username,
@@ -573,6 +578,33 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
             Timber.e("No Activity found to handle Intent")
         }
     }
+
+    private fun getScope(): String =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_scope)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_SCOPE, R.string.oauth2_openid_scope)
+        } else {
+            ""
+        }
+
+    private fun getScopeForLogin(tokenResponse: TokenResponse): String? =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_scope)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_SCOPE, R.string.oauth2_openid_scope)
+        } else {
+            tokenResponse.scope
+        }
+
+    private fun getPrompt(): String =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_prompt)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_PROMPT, R.string.oauth2_openid_prompt)
+        } else {
+            ""
+        }
 
     override fun onNewIntent(intent: Intent?) {
         super.onNewIntent(intent)
@@ -618,7 +650,11 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
             OAuthUtils.getClientAuth(clientRegistrationInfo.clientSecret as String, clientRegistrationInfo.clientId)
 
         } else {
-            OAuthUtils.getClientAuth(getString(R.string.oauth2_client_secret), getString(R.string.oauth2_client_id))
+            if (isKiteworksServer) {
+                OAuthUtils.getClientAuth(getString(R.string.kiteworks_client_id), getString(R.string.kiteworks_client_secret))
+            } else {
+                OAuthUtils.getClientAuth(getString(R.string.oauth2_client_secret), getString(R.string.oauth2_client_id))
+            }
         }
 
         // Use oidc discovery one, or build an oauth endpoint using serverBaseUrl + Setup string.
@@ -631,24 +667,27 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
         if (serverInfo is ServerInfo.OIDCServer) {
             tokenEndPoint = serverInfo.oidcServerConfiguration.tokenEndpoint
             if (serverInfo.oidcServerConfiguration.isTokenEndpointAuthMethodSupportedClientSecretPost()) {
-                clientId = clientRegistrationInfo?.clientId ?: contextProvider.getString(R.string.oauth2_client_id)
-                clientSecret = clientRegistrationInfo?.clientSecret ?: contextProvider.getString(R.string.oauth2_client_secret)
+                if (isKiteworksServer) {
+                    clientId = contextProvider.getString(R.string.kiteworks_client_id)
+                    clientSecret = contextProvider.getString(R.string.kiteworks_client_secret)
+                } else {
+                    clientId = clientRegistrationInfo?.clientId ?: contextProvider.getString(R.string.oauth2_client_id)
+                    clientSecret = clientRegistrationInfo?.clientSecret ?: contextProvider.getString(R.string.oauth2_client_secret)
+                }
             }
         } else {
             tokenEndPoint = "$serverBaseUrl${File.separator}${contextProvider.getString(R.string.oauth2_url_endpoint_access)}"
         }
 
-        val scope = resources.getString(R.string.oauth2_openid_scope)
-
         val requestToken = TokenRequest.AccessToken(
             baseUrl = serverBaseUrl,
             tokenEndpoint = tokenEndPoint,
             clientAuth = clientAuth,
-            scope = scope,
+            scope = getScope(),
             clientId = clientId,
             clientSecret = clientSecret,
             authorizationCode = authorizationCode,
-            redirectUri = OAuthUtils.buildRedirectUri(applicationContext).toString(),
+            redirectUri = OAuthUtils.buildRedirectUri(applicationContext, isKiteworksServer).toString(),
             codeVerifier = authenticationViewModel.codeVerifier
         )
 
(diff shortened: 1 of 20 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -618,7 +650,11 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
             OAuthUtils.getClientAuth(clientRegistrationInfo.clientSecret as String, clientRegistrationInfo.clientId)
 
         } else {
-            OAuthUtils.getClientAuth(getString(R.string.oauth2_client_secret), getString(R.string.oauth2_client_id))
+            if (isKiteworksServer) {
+                OAuthUtils.getClientAuth(getString(R.string.kiteworks_client_id), getString(R.string.kiteworks_client_secret))
+            } else {
+                OAuthUtils.getClientAuth(getString(R.string.oauth2_client_secret), getString(R.string.oauth2_client_id))
+            }
```
jesmrec, line 657: Parameters of `getClientAuth` are reversed in these calls.

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -631,24 +667,27 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
             tokenEndPoint = serverInfo.oidcServerConfiguration.tokenEndpoint
             if (serverInfo.oidcServerConfiguration.isTokenEndpointAuthMethodSupportedClientSecretPost()) {
-                clientId = clientRegistrationInfo?.clientId ?: contextProvider.getString(R.string.oauth2_client_id)
-                clientSecret = clientRegistrationInfo?.clientSecret ?: contextProvider.getString(R.string.oauth2_client_secret)
+                if (isKiteworksServer) {
+                    clientId = contextProvider.getString(R.string.kiteworks_client_id)
+                    clientSecret = contextProvider.getString(R.string.kiteworks_client_secret)
+                } else {
+                    clientId = clientRegistrationInfo?.clientId ?: contextProvider.getString(R.string.oauth2_client_id)
+                    clientSecret = clientRegistrationInfo?.clientSecret ?: contextProvider.getString(R.string.oauth2_client_secret)
```
jesmrec: Why don't you get `client_id` and `secret` straight from the contextProvider?

owncloudApp/src/main/res/values/setup.xml
```diff
@@ -104,6 +104,12 @@
     <string name="oauth2_client_id">e4rAsNUSIUs0lF4nbv9FmCeUkTlV9GdgTLDH1b5uie7syb90SzEVrbN7HIpmWJeD</string>
     <string name="oauth2_client_secret">dInFYGV33xKzhbRmpqQltYNdfLdJIfJ9L5ISoKhNoT9qZftpdWSP71VrpGR9pmoD</string>
 
+    <string name="kiteworks_client_id">163d022d-f2bf-55a7-9cdb-45507fa21cd5</string>
+    <string name="kiteworks_client_secret"></string>
+    <string name="kiteworks_redirect_uri_path">/</string>
```
jesmrec, line 112: Aren't `kiteworks_redirect_uri_scheme` and `kiteworks_redirect_uri_host` also required?

owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCRemoteAuthenticationDataSource.kt
```diff
@@ -52,7 +52,7 @@ class OCRemoteAuthenticationDataSource(
         val getBaseUrlRemoteOperation = GetBaseUrlRemoteOperation()
         val rawBaseUrl = executeRemoteOperation { getBaseUrlRemoteOperation.execute(client) }
 
-        val userBaseUri = rawBaseUrl?.replace(WEBDAV_FILES_PATH_4_0, "")
+        val userBaseUri = rawBaseUrl?.takeIf { !client.isKiteworksServer }?.replace(WEBDAV_FILES_PATH_4_0, "")
```
jesmrec, line 55: WEBDAV_FILES_PATH_4_0 = `/remote.php/dav/files/` , right? 

why is not being replaced in case of kiteworks server?

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasources/implementation/OCRemoteOAuthDataSource.kt
```diff
@@ -48,7 +49,13 @@ class OCRemoteOAuthDataSource(
...
         }
 
-        return serverConfiguration.toModel()
+        var serverConfigurationModel = serverConfiguration.toModel()
+
+        if (ownCloudClient.isKiteworksServer) {
+            serverConfigurationModel = serverConfigurationModel.copy(isKiteworksServer = true)
+        }
+
+        return serverConfigurationModel
```
jesmrec: could this block be more compact?. Like:

```
return serverConfiguration.toModel().run {
    if (ownCloudClient.isKiteworksServer) copy(isKiteworksServer = true) else this
}
```

(not sure if this is totally correct in Kotlin, or if is the same as the original block)

owncloudData/src/test/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSourceTest.kt
```diff
@@ -98,6 +100,25 @@ class OCRemoteServerInfoDataSourceTest {
...
+        val expectedValue = AuthenticationMethod.BEARER_TOKEN
+        prepareAuthorizationMethodToBeRetrieved(expectedValue)
+
+        every {
+            ocClientMocked.isKiteworksServer
+        } returns true
+
+        val currentValue = ocRemoteServerInfoDatasource.getAuthenticationMethod(OC_SECURE_SERVER_INFO_OIDC_AUTH.baseUrl)
+
+        assertNotNull(expectedValue)
```
jesmrec: You declared above: 

`val expectedValue = AuthenticationMethod.BEARER_TOKEN`

so, it will be never null. Is that right? In that case, such assertion is empty.

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -574,6 +579,33 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
     }
 
+    private fun getScope(): String =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_scope)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_SCOPE, R.string.oauth2_openid_scope)
+        } else {
+            ""
+        }
```
jesmrec, line 589: The `if...else if...` sequence, would it be clearer by using a `when` clause?

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -574,6 +579,33 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+
+    private fun getScopeForLogin(tokenResponse: TokenResponse): String? =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_scope)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_SCOPE, R.string.oauth2_openid_scope)
+        } else {
+            tokenResponse.scope
+        }
```
jesmrec, line 598: same here

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
```diff
@@ -574,6 +579,33 @@ class LoginActivity : AppCompatActivity(), SslUntrustedCertDialog.OnSslUntrusted
...
+
+    private fun getPrompt(): String =
+        if (isKiteworksServer) {
+            getString(R.string.kiteworks_openid_prompt)
+        } else if (oidcSupported) {
+            mdmProvider.getBrandingString(CONFIGURATION_OAUTH2_OPEN_ID_PROMPT, R.string.oauth2_openid_prompt)
+        } else {
+            ""
+        }
```
jesmrec, line 607: same here

jesmrec:
Some comments @joragua , tiny details. The PR is good-shaped 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: That's right, my fault... 😅 Changed! The client_secret is the first parameter and client_id is the second one.

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: The problem here is that the logic I've added is not correct. We need to keep the old logic (which uses the `clientRegistrationInfo` variable) and add the new one. First, we have to check if `clientRegistrationInfo` is not null to retrieve the values. If it is null, then we have to check if the server is from Kiteworks or not and use the correct configuration. I will change it 😄

owncloudApp/src/main/res/values/setup.xml
joragua: Yeah, they are required but in this case the values are the same as those used for ownCloud instances. Anyway, I can add two more fields as you said because in the future they can change and we have to keep both configurations (oC and Kiteworks) at the same time...

owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCRemoteAuthenticationDataSource.kt
joragua: Because in Kiteworks the variable `rawBaseUrl` has the value `<server_address>/kwdav/remote.php/dav/files` and if we only remove `WEBDAV_FILES_PATH_4_0`, the prefix `/kwdav` is still included in the `baseUri`. Therefore, for Kiteworks servers we have to use the `baseUri` supplied by the client.

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasources/implementation/OCRemoteOAuthDataSource.kt
joragua: Your suggestion is more Kotlinized and shorter than the original one. Changed! 👍🏻

owncloudData/src/test/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSourceTest.kt
joragua: Yes, that check is useless in this case. Removed! ✅

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: We usually use the `when` clause when all cases depend on the same variable. In this case, since we are checking two different variables: `isKiteworksServer` and `oidcSupported`, I think the `if...else if...` sequence is more appropiate.

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: Check last comment

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
joragua: Check last comment

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
jesmrec: 👍  you're right, makes sense.

owncloudData/src/test/java/com/owncloud/android/data/server/datasources/implementation/OCRemoteServerInfoDataSourceTest.kt
jesmrec: 💯

owncloudData/src/main/java/com/owncloud/android/data/authentication/datasources/implementation/OCRemoteAuthenticationDataSource.kt
jesmrec: understood!

owncloudApp/src/main/res/values/setup.xml
jesmrec: Cool!, it makes sense... it doesn't hurt and we are ready, just in case they are required somewhen.

owncloudApp/src/main/res/values/setup.xml
```diff
@@ -104,6 +104,14 @@
     <string name="oauth2_client_id">e4rAsNUSIUs0lF4nbv9FmCeUkTlV9GdgTLDH1b5uie7syb90SzEVrbN7HIpmWJeD</string>
     <string name="oauth2_client_secret">dInFYGV33xKzhbRmpqQltYNdfLdJIfJ9L5ISoKhNoT9qZftpdWSP71VrpGR9pmoD</string>
 
+    <string name="kiteworks_client_id">163d022d-f2bf-55a7-9cdb-45507fa21cd5</string>
```
jesmrec, line 108: I miss a comment heading the parameters block, like other blocks in the `setup.xml` file

owncloudApp/src/main/java/com/owncloud/android/presentation/authentication/LoginActivity.kt
jesmrec: 👍

owncloudData/src/main/java/com/owncloud/android/data/oauth/datasources/implementation/OCRemoteOAuthDataSource.kt
jesmrec: 👍

jesmrec (changes requested):

owncloudApp/src/main/res/values/setup.xml
joragua: Added 👍🏻

jesmrec: 👍

jesmrec:
CR passed ✅ !! good job, @joragua. Let's go for the QA.

jesmrec (approved):

jesmrec:
QA checks

- [x] oC
  - [x] Basic 
  - [x] OAuth2 - authorization
  - [x] OAuth2 - renewal
  - [x] OIDC - authorization
  - [x] OIDC - renewal
- [x] Kws
  - [x] OAuth2 - authorization
  - [x] OAuth2 - renewal

jesmrec:
Approved on my side. Ready to support both backends at the time!

PR #4636, opened 2025-07-08
[FIX] Push sbom into master after merging via ssh

jesmrec:
Previous version tried to push the `sbom.json` file generated by CI to the `master` branch, but the branch protection rules prevented to perform such action. Current PR would fix that.

Included:

- SSH is used to perform the commit. To push the commit, an SSH public key and private key must be configured in the repository settings.

- The current version of the SBOM is normalized and compared with the previous one. A commit is only made if differences are found.

- If a commit is made, it will trigger another job, which will be skipped due to a condition defined in the job configuration.

@DeepDiver1975 you added a `DEPLOYMENT_SSH_KEY` in Deploy Keys of Settings. That's the public key, it would be required to add the private key as repo secret as well. Other option: I created a pair of keys for testing and i could re-use them as keys by replacing yours.

I think this solution will work, but, we'll not know until the PR is merged (SBOM commit is created).

An alternative considered was pushing the sbom.json file to each development branch, but it was ultimately discarded due to lower overall performance.

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-
+      # Checkout the full repository history (required to access origin/master)
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
@@ -66,7 +66,7 @@ jobs:
           java-version: '17'
           distribution: 'temurin'
 
-      # Generate the Software Bill of Materials using CycloneDX Gradle plugin
+      # Generate the Software Bill of Materials (SBOM) using CycloneDX Gradle plugin
       - name: Generate SBOM (CycloneDX)
         run: ./gradlew --no-daemon cyclonedxBom
 
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -20,14 +20,14 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          ssh-key: ${{ secrets.DEPLOYMENT_SSH_KEY }}
+          ssh-key: ${{ secrets.DEPLOYMENT_SSH_KEY_SBOM }}
           persist-credentials: false
 
       # Start SSH agent and add the SSH key to authenticate Git operations
       - name: Start SSH agent and add key
         run: |
           mkdir -p ~/.ssh
-          echo "${{ secrets.DEPLOYMENT_SSH_KEY }}" > ~/.ssh/id_rsa
+          echo "${{ secrets.DEPLOYMENT_SSH_KEY_SBOM }}" > ~/.ssh/id_rsa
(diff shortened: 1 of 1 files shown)


.github/workflows/sbom.yml
```diff
@@ -11,16 +11,43 @@ permissions:
 
 jobs:
   sbom:
+    # Skip if the job was triggered by the SBOM commit
+    if: "!contains(github.event.head_commit.message, 'SBOM updated')"
     runs-on: ubuntu-latest
 
     steps:
-      # Checkout the full repository history (required to access origin/master)
```
joragua, line 17: Why did you remove this comment? I would keep this one in order to have comments in all steps... 🤔

.github/workflows/sbom.yml
```diff
@@ -32,54 +59,70 @@ jobs:
...
-      # Set up Java 17 for the Gradle build
+      # Set up Java 17 (required by Gradle and CycloneDX plugin)
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
 
-      # Generate the SBOM file using the CycloneDX plugin
+      # Generate the Software Bill of Materials using CycloneDX Gradle plugin
```
joragua:
```suggestion
      # Generate the Software Bill of Materials (SBOM) using CycloneDX Gradle plugin
```

A tiny suggestion: I would add the abbreviation here, but it's just my opinion... Up to you! 😄

.github/workflows/sbom.yml
```diff
@@ -11,16 +11,43 @@ permissions:
 
 jobs:
   sbom:
+    # Skip if the job was triggered by the SBOM commit
+    if: "!contains(github.event.head_commit.message, 'SBOM updated')"
```
joragua, line 15: The commit message should be `docs: SBOM updated`, right? This is the message that you have used for the last step (commit and push)

joragua (changes requested):
Some comments here @jesmrec! 🙌🏻

.github/workflows/sbom.yml
jesmrec: 👍 I will recover it, it was probably removed in any of the multiple refactors that were done to achieve the solution

.github/workflows/sbom.yml
jesmrec: 👍  let's go for it

.github/workflows/sbom.yml
jesmrec: `SBOM updated` is the message on the SBOM commit, right. If the latest commit contains the string `SBOM updated` means that we have just updated the SBOM and the job run was triggered for the correspondent commit.

joragua (approved):
Approved on my side! 👍🏻

Issue #4623, opened 2025-07-03
[FEATURE REQUEST] Add badges to README


jesmrec:
Our current [README](https://github.com/owncloud/android/blob/master/README.md) file points to the former BitRise status to display the last run of the tests.

We should replace that for the new jobs running over GitHub Actions

### TASKS

- [X] Research (if needed)
- [X] Create branch feature/feature_name
- [X] Development tasks
  - [X] Remove BitRise badges
  - [X] Add Actions badges
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [ ] Merge branch feature/feature_name into master

PR #4637, opened 2025-07-09
[FEATURE REQUEST] Add GitHub Actions badges to README removing the BitRise ones

jesmrec:
Implements: https://github.com/owncloud/android/issues/4623

Removes the status from BitRise and adds three badges from GitHub Actions

- Unit tests
- Data instrumented tests
- Detekt

Other badges could be added in the future. This is how it looks like:

Screenshot 2025-07-09 at 10 10 25

Badges shows the status of the latest execution.

## Related Issues
App:

_____

### DIFF
--- a/changelog/unreleased/4637
+++ b/changelog/unreleased/4637
@@ -1,0 +1,6 @@
+Change: Update CI badges in README file
+
+Bitrise badges in README file have been replaced by GitHub Actions badges.
+
+https://github.com/owncloud/android/issues/4623
+https://github.com/owncloud/android/pull/4637
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ ownCloud admins and users.
 * Change - Bump target SDK to 35: [#4529](https://github.com/owncloud/android/issues/4529)
 * Change - Replace dav4android location: [#4536](https://github.com/owncloud/android/issues/4536)
 * Change - Modify biometrics fail source string: [#4572](https://github.com/owncloud/android/issues/4572)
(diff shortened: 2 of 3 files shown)


changelog/unreleased/4637
```diff
@@ -0,0 +1,6 @@
+Change: Badges in README file
```
joragua:
```suggestion
Change: New badges in README file
```
I would change the title, but I'm not sure  if it's appropiate! What do you think? 😄

changelog/unreleased/4637
```diff
@@ -0,0 +1,6 @@
...
+
+Removed BitRise badges and added GitHub Actions ones.
```
joragua:
```suggestion
Bitrise badges in README file have been replaced by GitHub Actions badges.
```
I would use the present perfect passive tense in Calens but your version is correct, so... up to you!  Anyway, we have to polish the [TEMPLATE.md](https://github.com/owncloud/android/blob/master/changelog/TEMPLATE.md) file in order to have consistent rules for these files 🙃

changelog/unreleased/4637
jesmrec: If `New badges` is set, it might not make sense to set the  `Change` prefix. What could we use?

changelog/unreleased/4637
jesmrec: You're right, lets change it 👍

changelog/unreleased/4637
joragua: Hmm, `Change: Update CI badges in README file` maybe?

jesmrec: 👍

PR #4641, opened 2025-07-15
[FIX] Workflow to update sbom only if different than the latest one

jesmrec:
What does it do?:

- Creates the SBOM file over the feature branch code
- Compares with the latest commit SBOM, after normalizing both versions
  - If there are no differences, end
  - If there are differences, commit  
- If the commit is sent, new job triggered and skipped because the latest commit's message matches `SBOM updated`
- Commit will be created whenever changes in the newly generated sbom happen.

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/sbom.yml
+++ b/.github/workflows/sbom.yml
@@ -4,14 +4,19 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - master
+    - feature/*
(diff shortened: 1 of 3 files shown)


joragua (approved):
LGTM! 👍🏻 Good job @jesmrec

Issue #4584, opened 2025-05-14
[RELEASE] 4.6.0


jesmrec:
### TASKS:

 - [ ] [COM] Ping @mmattel about the new release
 - [ ] [GIT] Create branch `release/4.6.0` in owncloud/android from `master`
 - [ ] [DEV] Update version number and name in build.gradle in owncloudApp module
 - [ ] [DOC] Update [SBOM](https://infinite.owncloud.com/f/31e6d44f-f373-557c-9ab3-1748fc0c650d$4994cd9c-1c17-4254-829a-f5ef6e1ff7e3%215080be84-fbcc-4aca-956e-b278a7090418)
 - [ ] [DIS] Move Calens files from `unreleased` to a new folder like `M.m.p_YYYY-MM-DD` inside the `changelog` folder
 - [ ] [DEV] Check and reorder release notes in `ReleaseNotesViewModel.kt` to assure nothing important is missing there
 - [ ] [DEV] Code review
 - [ ] [DIS] Generate final bundle and APK from last commit in release branch
 - [ ] [COM] Prepare post in central.owncloud.org ([Category:News + Tag:android](https://central.owncloud.org/tags/c/news/5/android))
 - [ ] [DIS] Check for new screenshots in Play Store/GitHub/F-Droid and generate them
 - [ ] [QA] Design test plan
 - [ ] [QA] Regression test execution
 - [ ] [QA] QA approval
 - [ ] [DIS] Upload release APK and bundle to internal ownCloud instance
 - [ ] [COM] Ping @mmattel that we are close to sign the new tags
 - [ ] [DIS] Upload and publish release bundle and changelog in Play Store
 - [ ] [DIS] Update screenshots in Play Store/GitHub/F-Droid
 - [ ] [GIT] Create and sign tag `v4.6.0` in HEAD commit of release branch, in owncloud/android
 - [ ] [GIT] Move tag `latest` pointing the same commit as the release commit
 - [ ] [DIS] Publish a new [release](https://github.com/owncloud/android/releases) in owncloud/android
 - [ ] [DIS] Release published in Play Store
 - [ ] [COM] Publish post in central.owncloud.org ([Category:News + Tag:android](https://central.owncloud.org/tags/c/news/5/android))
 - [ ] [COM] Inform in "ownCloud General" and #general that release is out
 - [ ] [GIT] Merge `master` into `release/4.6.0`, fixing all the conflicts that could happen, in owncloud/android
[...]

Issue #4624, opened 2025-07-03
[BUG] Change spaces icon for folder in details view for the personal space


jesmrec:
![Image](https://github.com/user-attachments/assets/033c889f-d86f-4f47-8f90-9d772c1e3edf)

### Actual behaviour

The `Details` view shows the space icon for `Personal` spaces

### Expected behaviour

The `Details` view shows the folder icon for `Personal` spaces. That's the behaviour for other views like the uploads view and the av. offline view
 

Can this problem be reproduced with the official owncloud server?
(url: https://ocis.ocis.master.owncloud.works, user: einstein, password: relativity)

### Environment data
Android version:

Device model: 

Stock or customized system:

ownCloud app version:

ownCloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### ownCloud log (data/owncloud.log)
```
Insert your ownCloud log here
```

Issue #4582, opened 2025-05-08
[BUG] No message when uploading a file with no quota


jesmrec:
### Steps to reproduce
1.  Account with quota exceeded
2. Open the FAB > Upload from file system/Upload from camera/New shortcut
3. Select the file/Take a picture/Create a correct shortcut

### Actual behaviour

Server returns a `507` error (insufficient storage). Files are not uploaded because the lack of space, but no feedback is displayed. Nothing happens.

### Expected behaviour

An error or feedback message about the problem should be notified to the user via snackbar like, for example, when a new file is copied with no storage.

Xiaomi Redmi 13, Android 15

Issue #4576, opened 2025-04-21
[BUG] Infinite edges Android 15


jesmrec:
### Actual behaviour

After the [update to SDK 35](https://github.com/owncloud/android/pull/4556), the new feature "infinite edges" causes glitches in the upper and bottom bar of the app. Some examples:

| | | | |
|:-:|:-:|:-:|:-:|
|![Image](https://github.com/user-attachments/assets/aaf135ff-3b4a-44c9-bc46-d332bb540a52)|![Image](https://github.com/user-attachments/assets/50a1fee6-26a5-45d7-9e3c-c708cf8910a0)|![Image](https://github.com/user-attachments/assets/bb8c7cb6-e88f-4545-9016-95f373086915)|![Image](https://github.com/user-attachments/assets/174899b6-b581-4f1b-b0fd-e5c0159cda06)

### Expected behaviour

Bottom and upper bar does not collide with the Android system bars

Android 15

JuancaG05:
Indeed this is change that came with upgrading the target SDK to Android 15, as explained in https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge. There, they suggest to adapt component by component with insets, but due to the variety of components we have in the app, this may get a bit complex.

There are other solutions, as the ones suggested in the answers of https://stackoverflow.com/questions/79095037/android-15-edge-to-edge-mode-hide-a-top-view, which may be temporal but enough for the moment.

In any case, this needs deeper research, and one of the next steps could be checking the official [Android codelab about how to handle edge-to-edge enforcements](https://developer.android.com/codelabs/edge-to-edge).

joragua:
Since the target SDK was bumped to 36 (https://github.com/owncloud/android/issues/4743), the solution that fixed this issue is now deprecated and no longer available:
```
<item name="android:windowOptOutEdgeToEdgeEnforcement" tools:targetApi="35">true</item>
```

Reopening the issue because of this. It deserves another look.

PR #4643, opened 2025-07-16
[RELEASE] 4.6.0

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4584
_____

### DIFF
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -732,11 +732,11 @@
 
     <!-- Release Notes -->
     <string name="release_notes_4_6_0_title_shares_space_docs_provider">Shares space in Android native file explorer</string>
-    <string name="release_notes_4_6_0_subtitle_shares_space_docs_provider">Shares space now appears all together with the other spaces in Documents Provider in Infinite Scale accounts</string>
+    <string name="release_notes_4_6_0_subtitle_shares_space_docs_provider">Shares space now appears next to the other spaces in Documents Provider for Infinite Scale accounts</string>
     <string name="release_notes_4_6_0_title_duplicated_automatic_uploads">Automatic uploads improved</string>
     <string name="release_notes_4_6_0_subtitle_duplicated_automatic_uploads">Automatic uploads are more consistent against duplications</string>
-    <string name="release_notes_4_6_0_title_bearer_token_handling">Multiple accounts in oCIS</string>
-    <string name="release_notes_4_6_0_subtitle_bearer_token_handling">Multiple accounts can be logged in at the same time on the same oCIS server. Additionally, it is no longer necessary to log in again when a session expires.</string>
+    <string name="release_notes_4_6_0_title_bearer_token_handling">Multiple accounts on Infinite Scale servers</string>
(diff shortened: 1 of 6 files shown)


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -735,6 +735,8 @@
     <string name="release_notes_4_6_0_subtitle_shares_space_docs_provider">Shares space now appears all together with the other spaces in Documents Provider in Infinite Scale accounts</string>
     <string name="release_notes_4_6_0_title_duplicated_automatic_uploads">Automatic uploads improved</string>
     <string name="release_notes_4_6_0_subtitle_duplicated_automatic_uploads">Automatic uploads are more consistent against duplications</string>
+    <string name="release_notes_4_6_0_title_bearer_token_handling">Multiple accounts in oCIS</string>
+    <string name="release_notes_4_6_0_subtitle_bearer_token_handling">Multiple accounts can be logged in at the same time on the same oCIS server. Additionally, it is no longer necessary to log in again when a session expires.</string>
```
jesmrec: `Additionally, it is no longer necessary to log in again when a session expires`

It sounds like: it was necessary before, and now it isn't. That was a bug that happened from time to time (very cumbersome), but not a new improved behaviour. My suggestion:

`Additionally,  fixed a problem related with session handling`

(avoiding technical concepts like `token` or `renewal` , that non-techie users may not understand). That's more compact and clarifies that it's a fix over an existing problem, not a new feature. What do you think?

OTOH, about another release note, that was added before:

`Shares space now appears all together with the other spaces in Documents Provider in Infinite Scale accounts`

Suggestion: `all together with` is not incorrect, but does not sound correct to me in that context. I'd user `next to`  instead. Also `in Infinite Scale accounts` should be replaced by `for Infinite Scale accounts`

Be careful with the existing translations in case that last change is implemented.

joragua: Totally agree, let's change both release notes! 😄

jesmrec (approved):

owncloudApp/src/main/res/values/strings.xml
jesmrec: Perfect 💯

jesmrec:
QA:

Changelog checks:

- [x] Side menu collapses info in landscape: https://github.com/owncloud/android/issues/4513
- [x] Content in Spaces not shown from third-party apps: https://github.com/owncloud/android/issues/4522
- [x] Add bottom margin for used quota in account dialog: https://github.com/owncloud/android/issues/4566
- [x] Infinite edges in Android 15: https://github.com/owncloud/android/issues/4576
- [x] No message when uploading a file with no quota: https://github.com/owncloud/android/issues/4582
- [x] Change space icon in file details view for personal space: https://github.com/owncloud/android/issues/4624
- [x] Modify biometrics fail source string: https://github.com/owncloud/android/issues/4572
- [x] Shares space in Android native file explorer: https://github.com/owncloud/android/issues/4515
- [x] Support for Kiteworks

E2E tests:
[4.6.0_20250718_oCIS7.1.3.html.zip](https://github.com/user-attachments/files/21342673/4.6.0_20250718_oCIS7.1.3.html.zip)

Regression test: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.6/RegressionTest.md

Android Studio version: `Android Studio Meerkat Feature Drop | 2024.3.2 Patch 1`

Approved commit: `a71aa5220bb`

jesmrec:
Approved

Issue #4605, opened 2025-06-18
[FEATURE REQUEST] Add the role/account information to the user


jesmrec:
Update:

In the clients, the rol is not so important as the permissions that are granted to such role. Instead of getting the role (admin, space admin, user, user light) would be enough to know the permissions granted to the account. How to get them?

Requesting the following endpoint:

`graph/v1.0/me`

will return something like:

```
{
   "displayName" : "Joe Smith",
   "id" : "52037314-d64f-49cb-a0be-1db422d172e9",
   "identities" : [
      {
         "issuer" : "https://mymachine:9200",
         "issuerAssignedId" : "52037314-d64f-49cb-a0be-1db422d172e9"
      }
   ],
   "mail" : "joe@smith.com",
   "onPremisesSamAccountName" : "jsmith",
   "preferredLanguage" : "en",
   "userType" : "Member"
}
```
the `id` field is the account id, static and useful to get info about it

Then, requesting

`/api/v0/settings/permissions-list` 

with body:

```
{
    "account_uuid": "<id>"
}
```

will get something like:

```
{
    "permissions": [
        "Event.SpaceMembershipExpired.ReadWrite.own",
        "Event.SpaceDisabled.ReadWrite.own",
        "Event.SpaceDeleted.ReadWrite.own",
        "Groups.ReadWrite.all",
        "Drives.Create.all",
        "ReadOnlyPublicLinkPassword.Delete.all",
        "Event.ShareExpired.ReadWrite.own",
        "Event.SpaceShared.ReadWrite.own",
        "Drives.ReadWriteEnabled.all",
        "Drives.ReadWriteProjectQuota.all",
        "Drives.DeleteProject.all",
        "Event.ShareRemoved.ReadWrite.own",
        "Favorites.List.own",
        "Roles.ReadWrite.all",
        "AutoAcceptShares.ReadWriteDisabled.own",
        "Event.ShareCreated.ReadWrite.own",
        "Language.ReadWrite.all",
        "Drives.List.all",
        "PublicLink.Write.all",
        "Drives.ReadWritePersonalQuota.all",
        "Favorites.Write.own",
        "Settings.ReadWrite.all",
        "Shares.Write.all",
        "Drives.DeletePersonal.all",
        "EmailNotifications.ReadWriteDisabled.own",
        "Drives.ReadWrite.all",
        "Accounts.ReadWrite.all",
        "EmailSendingInterval.ReadWrite.own",
        "Logo.Write.all",
[...]

PR #4647, opened 2025-07-23
[FEATURE REQUEST] Add account information to the user

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4605

_____

### DIFF
--- a/changelog/unreleased/4647
+++ b/changelog/unreleased/4647
@@ -1,0 +1,7 @@
+Enhancement: Add account ID to the user information
+
+The account ID has been added to the Account Manager only for Infinite Scale users.
+This information will be used to fetch all permissions related to space management.
+
+https://github.com/owncloud/android/issues/4605
+https://github.com/owncloud/android/pull/4647
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserIdOperation.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserIdOperation.kt
@@ -1,0 +1,73 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author Jorge Aguado Recio
+ *
+ * Copyright (C) 2025 ownCloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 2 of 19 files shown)


changelog/unreleased/4647
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add account ID information to the user
```
jesmrec: My option here: `Add account ID to the user information` . User is confusing... it's the one that uses the app :D

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserIdOperation.kt
```diff
@@ -0,0 +1,73 @@
...
+
+                val moshi: Moshi = Moshi.Builder().build()
+                val adapter: JsonAdapter<GraphMeResponse> = moshi.adapter(GraphMeResponse::class.java)
+
+                result = RemoteOperationResult(ResultCode.OK)
+                result.data = getMethod.getResponseBodyAsString().let { adapter.fromJson(it)!!.id }
+
+                Timber.d("Get user id completed and parsed to ${result.data}")
+            } else {
+                result = RemoteOperationResult(getMethod)
```
jesmrec, line 56: why `getMethod` as parameter here? shouldn't be any error information like `ResultCode.xxx`?

changelog/unreleased/4647
joragua: I agree. Changed! 👍🏻

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/users/GetRemoteUserIdOperation.kt
joragua: The constructor used here takes a `HttpMethod` as its parameter. It is intended for cases  when the result needs to be interpreted from the response (You can see the implementation in the `RemoteOperationResult` class)

jesmrec: 👍 . `RemoteOperationResult` has a bunch of creators calling themselves... that confused me. That's OK

jesmrec:
CR approved

jesmrec:
@joragua, the process of adding the client ID works fine for new accounts. But, i guess that it won't work for migrations, since the id is saved in first instance.

PR #4649, opened 2025-07-25
[FIX] Document editing not available on ByCS

joragua:
## Related Issues
App: https://github.com/owncloud/enterprise/issues/7277

_____

### DIFF
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/responses/AppRegistryResponse.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/appregistry/responses/AppRegistryResponse.kt
@@ -53,6 +53,6 @@ data class AppRegistryMimeTypeResponse(
 data class AppRegistryProviderResponse(
     val name: String,
     @Json(name = "product_name")
-    val productName: String,
+    val productName: String?,
     val icon: String,
 )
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -27,6 +28,28 @@
...
+# Changelog for ownCloud Android Client [unreleased] (UNRELEASED)
+
+The following sections list the changes in ownCloud Android Client unreleased relevant to
+ownCloud admins and users.
+
+[unreleased]: https://github.com/owncloud/android/compare/v4.6.0...master
+
+## Summary
+
+* Bugfix - Document editing not available on ByCS: [#7277](https://github.com/owncloud/enterprise/issues/7277)
```
jesmrec: Not only editing, the problem caused that creation was not allowed either. I'd call it like... `Open with web not always available` or similar. What do you think?

joragua: Hmm... 🤔 I wouldn't say `Open with` because it's a different feature in the app and it is working well on ByCS. Other idea that comes to my mind: `Bugfix: Mini-fab for creating new documents not available on ByCS`?

jesmrec: `Open with` and `Open with web` are different, even when the names are similar, but it's ok to avoid both of them in case it could lead to confusion. On the other side, i'd not mention an specific customer. Another approach:  `Bugfix: Mini-fab for creating new documents not always available` . Within the description, the reasons are clear.

joragua: I will change it 👍🏻

jesmrec:
QA checks

- [x] Mini fab present in instances 1, 2 and 3
- [x] Create document in instance 1
- [x] Create document in instance 2
- [x] Create document in instance 3
- [x] Open and edit file in instance 1
- [x] Open and edit file in instance 2
- [x] Open and edit file in instance 3

jesmrec:
Approved from my side

Issue #4650, opened 2025-07-25
[BUG] Crash in TransferListFragment (Play Console)


jesmrec:
Crash reported in play console, this is the stack trace:

```
Exception java.lang.RuntimeException:
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:4170)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:4361)
  at android.app.ActivityThread.handleRelaunchActivityInner (ActivityThread.java:6446)
  at android.app.ActivityThread.handleRelaunchActivity (ActivityThread.java:6329)
  at android.app.servertransaction.ActivityRelaunchItem.execute (ActivityRelaunchItem.java:82)
  at android.app.servertransaction.ActivityTransactionItem.execute (ActivityTransactionItem.java:60)
  at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem (TransactionExecutor.java:174)
  at android.app.servertransaction.TransactionExecutor.executeTransactionItems (TransactionExecutor.java:109)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:81)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2747)
  at android.os.Handler.dispatchMessage (Handler.java:112)
  at android.os.Looper.loopOnce (Looper.java:268)
  at android.os.Looper.loop (Looper.java:384)
  at android.app.ActivityThread.main (ActivityThread.java:8921)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:580)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:907)
Caused by androidx.fragment.app.Fragment$InstantiationException:
  at androidx.fragment.app.Fragment.instantiate (Fragment.java:681)
  at androidx.fragment.app.FragmentContainer.instantiate (FragmentContainer.java:57)
  at androidx.fragment.app.FragmentManager$3.instantiate (FragmentManager.java:507)
  at androidx.fragment.app.FragmentState.instantiate (FragmentState.java:81)
  at androidx.fragment.app.FragmentStateManager.<init> (FragmentStateManager.java:85)
  at androidx.fragment.app.FragmentManager.restoreSaveStateInternal (FragmentManager.java:2507)
  at androidx.fragment.app.FragmentManager.attachController (FragmentManager.java:2667)
[...]

PR #4652, opened 2025-07-28
[FIX] Crash in TransferListFragment (Play Console)

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4650

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,19 @@ ownCloud admins and users.
 
 ## Summary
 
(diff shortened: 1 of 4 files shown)


jesmrec (approved):

jesmrec:
CR 👍

jesmrec:
The way to reproduce the problem is turning the device (landscape - portrait), and then, it crashes. The developed fix solved the problem and no crash anymore.

Approved

PR #4661, opened 2025-08-13
[FEATURE REQUEST] Add account id to the user

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4605

_____


owncloudData/src/main/java/com/owncloud/android/data/user/repository/OCUserRepository.kt
```diff
@@ -54,4 +56,11 @@ class OCUserRepository(
 
     override fun getUserAvatar(accountName: String): UserAvatar =
         remoteUserDataSource.getUserAvatar(accountName)
+
+    override fun getUserId(accountName: String): String =
```
jesmrec, line 60: Shouldn't be `getAccountId` a better name?

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCRemoteUserDataSourceTest.kt
```diff
@@ -123,5 +126,21 @@ class OCRemoteUserDataSourceTest {
...
+            ocUserService.getUserId()
+        } returns getUserIdResult
+
+        val userId = ocRemoteUserDataSource.getUserId(OC_ACCOUNT_NAME)
+        assertEquals(OC_USER_ID, userId)
+
+        verify(exactly = 1) {
+            ocUserService.getUserId()
+        }
+    }
```
jesmrec, line 144: About this test: what is it being really checked? assertion checks that the call retrieves the mock result, and it will always happen.

owncloudData/src/test/java/com/owncloud/android/data/user/repository/OCUserRepositoryTest.kt
```diff
@@ -155,4 +158,38 @@ class OCUserRepositoryTest {
...
+        } returns OC_USER_ID
+
+        val accountId = ocUserRepository.getUserId(OC_ACCOUNT_NAME)
+        assertEquals(OC_USER_ID, accountId)
+
+        verify(exactly = 1) {
+            localAuthenticationDataSource.getUserId(OC_ACCOUNT_NAME)
+        }
+    }
+
```
jesmrec, line 174: Same in this test

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -120,6 +122,22 @@ class SpacesListFragment : SpacesListAdapter.SpacesListAdapterListener, Fragment
...
+        collectLatestLifecycleFlow(spacesListViewModel.userId) { event ->
+            event?.let {
+                val accountName = requireArguments().getString(BUNDLE_ACCOUNT_NAME)
+                when (val uiResult = event.peekContent()) {
+                    is UIResult.Success -> {
+                        Timber.d ("The account id for $accountName is: ${uiResult.data}")
+                    }
+                    is UIResult.Loading -> { }
+                    is UIResult.Error -> {
+                        Timber.e(uiResult.error, "Failed to retrieve user id for account $accountName")
```
jesmrec, line 135: In this case, error means that the data could not be retrieved? or data is "null"?

owncloudData/src/main/java/com/owncloud/android/data/user/repository/OCUserRepository.kt
joragua: The suggested name is reasonable, but if we keep the current wording (`user`)  I'd avoid renaming it to ensure consistency with existing components such as packages, usecases, repositories... Do you agree? 🤔

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCRemoteUserDataSourceTest.kt
joragua: In this case, the method only contains a single line of code and it's true that the assertion is checking the mocked result but, there are a lot of test methods with the same structure. Anyway, we will need to review them and decide what to do in the future...

owncloudData/src/test/java/com/owncloud/android/data/user/repository/OCUserRepositoryTest.kt
joragua: Check previous comment

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: It means that the use case was unable to retrieve data for some reason (e.g., no network connection)

owncloudData/src/main/java/com/owncloud/android/data/user/repository/OCUserRepository.kt
jesmrec: Agreed, if the naming convention was set up. it should be followed. Let's keep it then

owncloudData/src/test/java/com/owncloud/android/data/user/datasources/implementation/OCRemoteUserDataSourceTest.kt
jesmrec: As commented before, let's keep the convention. But, we should review the way to develop tests. They should give valid information about the production logic. If not, they might not make sense.

owncloudData/src/test/java/com/owncloud/android/data/user/repository/OCUserRepositoryTest.kt
jesmrec: Same, check my previous comment :)

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: So, no matter whether the id is `null` or not `null` , the `when` clause will go to `UIResult.Success` if no external errors happen. OK!

jesmrec (approved):

jesmrec:
QA checks:

- [X] Installation from scratch
  - [X] oCIS -> shows the id in the logs
  - [X] oC10 -> does not show the id in the logs
- [x] Update from previous versions
  - [x] oCIS -> shows the id in the logs
  - [x] oC10 -> does not show the id in the logs
- [X] Check with no connection

jesmrec:
Approved on my side. Anyway, this is only a preliminary work that will help to other operations related with Space Management. In QA-phases of such features, we will really check if everything is correctly done.

Go ahead!

PR #4663, opened 2025-08-22
[TECHNICAL] Add workflow to test upgrade from "latest" to current

jesmrec:
## Related Issues

Update testing is required to every release pre-QA stage, to assure that the new build will migrate correctly over the existing version. The current pull request automates such process.

In this first version, this is the process:

- Build `latest` and `current` by using a matrix that will run both in parallel
- Once both are done, the test hosted [here](https://github.com/jesmrec/ownCloudUpdateTesting) will run:
   - Install `latest` version
   - Log in and enable security
   - Install `current` over that
   - Checks that the security is asked, the list of files is correct, and the commit hash is correctly displayed in `Settings`.

For sure, more assertions and checks could be added to have a more accurate testing. 

As artifacts:

- Both apks (retention of 1 day)
- Execution log, from test repository
- Crash log, if needed
- Recoding video

App:

_____


.github/workflows/update.yml
```diff
@@ -0,0 +1,223 @@
...
+    steps:
+      # Checkout the specific commit from matrix
+      - name: Checkout specific commit
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ matrix.commit }}
+
+      # Set up Java JDK required for Gradle
+      - name: Set up JDK
+        uses: actions/setup-java@v3
```
joragua: I would use the latest version (`v5`) for `setup-java` step. We already merged a PR that includes this update (https://github.com/owncloud/android/pull/4668)

.github/workflows/update.yml
```diff
@@ -0,0 +1,223 @@
...
+        run: ./gradlew assembleqaRelease --no-daemon --stacktrace --info
+
+      # Copy the built APK to a commit-specific name
+      - name: Get apk
+        run: cp owncloudApp/build/outputs/apk/qa/release/owncloud_*-qa-release*.apk owncloud-${{ matrix.commit }}.apk
+
+      # Decode keystore from secret for signing
+      - name: Restore keystore
+        run: |
+          echo "${{ secrets.TEST_KS_B64 }}" | base64 -d > ./test.keystore
```
joragua, line 70: Is `TEST_KS_B64` a suitable name for a final version? I would use `KEYSTORE_B64` or a similar name instead. What do you think? 🤔

.github/workflows/update.yml
```diff
@@ -0,0 +1,223 @@
...
+
+      # Align and sign the APK
+      - name: Sign APK
+        run: |
+          APK_INPUT="owncloud-${{ matrix.commit }}.apk"
+          APK_ALIGNED="owncloud-${{ matrix.commit }}-aligned.apk"
+          APK_SIGNED="owncloudSigned-${{ matrix.commit }}.apk"
+          KEYSTORE="./test.keystore"
+          KEY_ALIAS="${{ secrets.TEST_KS_ALIAS }}"
+          KEY_PASSWORD="${{ secrets.TEST_KS_KEY }}"
```
joragua, line 80: Same as previous comment

.github/workflows/update.yml
```diff
@@ -0,0 +1,223 @@
...
+            LOGCAT_PID=$!
+
+            ./gradlew clean test -Dserver="${{ secrets.OC_SERVER_URL }}" -Dusername="${{ secrets.OC_SERVER_USERNAME_TEST }}" -Dpassword="${{ secrets.OC_SERVER_PASSWORD_TEST }}" -Dcommit="$(echo $GITHUB_SHA | cut -c1-7)"
+
+            # Kill logcat to serve crash_log.txt
+            kill $LOGCAT_PID || true
+
+      # If crash happened, upload crash log
+      - name: Upload crash log artifact
+        if: always() && hashFiles('crash_log.txt') != ''
```
joragua, line 191: One question: the `always()` clause is used for executing steps even if the main job fails, right?

changelog/unreleased/4663
```diff
@@ -0,0 +1,5 @@
+Enhancement: Update test in GitHub Actions
+
+Workflow to check whether the upgrade from "latest" to "current" works flawless. Exports the execution log and the crash log if any crash happens.
```
joragua: I would re-write the calens file like this (or similar): `A new workflow to check wheter the upgrade from "latest" to "current" works flawless has been added to GitHub Actions. It also exports the execution log and the crash log if something goes wrong`. Do you agree? Of course, it's a personal opinion... 😄

.github/workflows/update.yml
```diff
@@ -0,0 +1,223 @@
...
+    name: Build APKs
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Run matrix for latest commit on branch and current commit SHA
+        commit: ["latest", "${{ github.sha }}"]
+
+    env:
+      BUILD_TOOLS_VERSION: "34.0.0"
+      ANDROID_API_LEVEL: 34
```
joragua: Is this variable (`ANDROID_API_LEVEL`) being used in the job?

joragua (changes requested):
Somme comments here @jesmrec

.github/workflows/update.yml
jesmrec: You're right, let's push it to the latest version.

.github/workflows/update.yml
jesmrec: well, i set the `TEST` prefix because that's a keystore just for testing purposes (not the official one). If any day we set the official one with deploying purposes, we'd need something to distinguish them

.github/workflows/update.yml
jesmrec: Check my comment as well

.github/workflows/update.yml
jesmrec: yes, that's right. The step that contains `always()` will be executed regardless the result of any previous step.

.github/workflows/update.yml
jesmrec: i used it firstly, but i removed the step in which it was included. So, it is totally useless. Good catch ;)

I'll finish it.

changelog/unreleased/4663
jesmrec: Let me catch your approach with  tiny changes: `A new Github Actions workflow has been added, in order to check whether the upgrade from "latest" to "current" works flawless. It also exports the execution log and the crash log if something goes wrong`.

joragua: Nice! Go ahead 😄

.github/workflows/update.yml
joragua: 👍🏻

.github/workflows/update.yml
joragua: Ok, it makes sense then! ✅

joragua (approved):
LGTM! Good job 💯

PR #4671, opened 2025-08-28
[FIX] Set authorization heador only when necessary

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4575

_____


changelog/unreleased/4671
```diff
@@ -0,0 +1,6 @@
+Bugfix: Set authorization header for token request only when necessary
+
+A new field has been added to the token request to indicate whether the authorization header should be applied.
```
jesmrec: I'd be more accurate here, like:

```
A new field has been added to the token request to indicate whether the authorization header should be applied in case `client_secret_post` method is supported or not.
```

jesmrec (changes requested):

jesmrec:
@joragua just a little comment about the calens entry

changelog/unreleased/4671
joragua: Changed!! 👍🏻

jesmrec:
LGTM now 👍

jesmrec:
Moving to QA

jesmrec:
The test class `FileDetailsFragmentTest` returns an error when building a debug apk from Android Studio. This seems to be a UI test class (not maintained ftm), but, we should take a look inside in order not have errors. Also, we may need a check in CI to warn about this issues (or improving one of the existing checks)

joragua:
In my opinion, I wouldn't create a new workflow to warn about these issues, since UI tests are not being maintained at the moment. We might review them in the future, but they are not part of our current plans.

On the other hand, the error that you mentioned is fixed in this PR (https://github.com/owncloud/android/pull/4682) so now, Android Studio should be able to build the APK without errors.

jesmrec:
Let's approve this, but the testing will continue...

PR #4675, opened 2025-09-01
[FIX] Add explicit permission to detekt workflow

jesmrec:
Internal code scanning returned the following issue (medium severity):

https://github.com/owncloud/android/security/code-scanning/7

as explanation:

```
If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write. Often these permissions do not adhere to the principle of least privilege and can be reduced to read-only, leaving the write permission only to a specific types as issues: write or pull-requests: write.
```

Fixed by adding the `permissions:` block with just `read` permission to the detekt workflow, avoiding the default write pemission.

## Related Issues
App: https://github.com/owncloud/android/security/code-scanning/7

_____


joragua (approved):
LGTM! 💯

PR #4681, opened 2025-09-05
[FEATURE] Workflow to build apk from latest, only in specific moments

jesmrec:
Generates a signed apk from `latest` in the following cases:

- When `latest` is pushed
- Manually
- Every 2 months

The idea is related with the `Update Test` workflow. In the current status, a build over `latest` is generated every push. `latest` changes from time to time (every new release), that means, every workflow execution the same apk is generated again and again. That's not efficient.

Thanks to the step [dawidd6/action-download-artifact](https://github.com/dawidd6/action-download-artifact), it's posible to recover other workflow's artifacts to use them in the current execution.

The maximum time retention for an artifact is 90 days. If we don't sign a new `latest` (new release) in such amount of time, the artifact would be no longer available, making "Update test" to fail. To avoid such effect, this workflow is automatically triggered every two months.

## Related Issues
App:

_____


.github/workflows/buildapkfromlatest.yml
```diff
@@ -0,0 +1,105 @@
...
+      # Set up Java JDK required for Gradle
+      - name: Set up JDK
+        uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      # Set up Android SDK and build tools
+      - name: Set up Android SDK
+        uses: android-actions/setup-android@v2
```
joragua: `v3` seems to be more appropiated than `v2`, since Dependabot already created a pull request with that update (https://github.com/owncloud/android/pull/4678)

changelog/unreleased/4681
```diff
@@ -0,0 +1,5 @@
+Change: Only CI-generate a build from "latest" tag on demand
+
+New workflow has been added to build signed apk from "latest" tag when the tag is pushed, or every 2 months. Also, manually can be triggered.
```
joragua: It's just my opinion... Up to you! 😄 

```suggestion
Enhancement: New workflow to generate a build from "latest" tag on demand

New workflow has been added to GitHub Actions to build signed apk from "latest" tag when the tag is pushed, every 2 months or manually triggered.
```

joragua (changes requested):
Somme comments here @jesmrec:

- I would use `feat:` for the type of the first commit, since the project follows conventional commits.

- I would rename the workflow to `build-apk-from-latest`, following the same convention that we use for unit and instrumented tests workflows.

jesmrec:
> I would use feat: for the type of the first commit, since the project follows conventional commits.

sure, that was my fault. Will fix

> I would rename the workflow to build-apk-from-latest, following the same convention that we use for unit and instrumented tests workflows.

Will do as well

.github/workflows/buildapkfromlatest.yml
jesmrec: sure, will do 👍

changelog/unreleased/4681
jesmrec: `Change` or `Enhancement` that's the question 😄 . We can use `Enhancement` here and `Change` in the other workflow, since it is already in production and will consume this one. Is this ok @joragua ?

joragua: Yeah, I totally agree. I would use `Enhancement` in this case (new workflow) and `Change` in the other one.

joragua (approved):
LGTM! Good job 👍🏻

Issue #4094, opened 2023-07-10
[BUG] Handling of HTTP response code 413


yunal:
### Actual behaviour

Android App version dated 25. of May 2023, version 4.0.

I try to upload a image to the OC server.  And get a "unknown error" message.  I google.  I set php upload limits.  The error message persists.

I enable server side debug logging.  No direct clues there.  I enable client side logging. Aha!

```
D: 2023-07-10 21:43:16:842(FileRequestBody.kt:70) .writeTo()File with name IMG_20230710_213046.jpg and size 2385151 written in request body
D: 2023-07-10 21:43:16:842(LogBuilder.kt:38) .logHttp()[Network, response] [info] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Method: PUT URL: https://oc.langfeldt.net/remote.php/dav/files/Nicolai/Photos/IMG_202307\
10_213046.jpg Code: 413 Message: Request Entity Too Large
D: 2023-07-10 21:43:16:843(LogBuilder.kt:38) .logHttp()[Network, response] [header] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Server: nginx/1.18.0 (Ubuntu)
D: 2023-07-10 21:43:16:844(LogBuilder.kt:38) .logHttp()[Network, response] [header] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Date: Mon, 10 Jul 2023 19:42:48 GMT
D: 2023-07-10 21:43:16:844(LogBuilder.kt:38) .logHttp()[Network, response] [header] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Content-Type: text/html
D: 2023-07-10 21:43:16:845(LogBuilder.kt:38) .logHttp()[Network, response] [header] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Content-Length: 192
D: 2023-07-10 21:43:16:845(LogBuilder.kt:38) .logHttp()[Network, response] [header] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Connection: close
D: 2023-07-10 21:43:16:846(LogBuilder.kt:38) .logHttp()[Network, response] [body] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Length: 192 byte body
D: 2023-07-10 21:43:16:848(LogBuilder.kt:38) .logHttp()[Network, response] [body] [3651f353-788b-4a52-8caa-30f0faf8a0d2] Type: text/html
[...]

PR #4685, opened 2025-09-24
[FIX] Handling of HTTP response code 413

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4094

_____

## QA

https://github.com/owncloud/android/pull/4685#issuecomment-3365419895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ ownCloud admins and users.
 
 ## Summary
 
+* Bugfix - Handling of HTTP code response 413: [#4094](https://github.com/owncloud/android/issues/4094)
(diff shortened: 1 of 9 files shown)


jesmrec:
LGTM 👍 . Moving to QA

jesmrec:
About QA, there are two different cases:

- The mocked `413` shows the correct error message in both uploads view and notification. So, in case the server returns a `413`, the app behaviour is correct ✅ 
- Trying to reproduce with a real very big file, i got some problems due to hardware restrictions in test devices.

Since this specific case affects only oC10 (oCIS with no space restriction beyond the quotas or the disk available space) and the `413` handling is correct, we can assume that a server response with such code will be correctly managed.

Let's move this forward

Issue #4607, opened 2025-06-18
[FEATURE REQUEST] Edit space


jesmrec:
AC:

- Add an option in the bottom side menu: `Edit space`. 
- Only available if the user is space manager, admin, or space admin. 
- Same dialog/view of creation can be displayed to edit, applying the same rules (space not empty, no longer than 255, without forbidden chars)
- Quota:  only are allowed to modify it the follwing roles `admin` and `space admin`. Space manager not allowed.

Base request:

`PATCH https://<server>/graph/v1.0/drives/<drive-id>`

following the same lines as the creation

There is no information about error handling, so if any error would happen, it'd be enough showing a generic error message.

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
  - [ ] Implement unit tests (if needed)
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

jesmrec:
To check if a user has the permission to edit. Two ways:

## Crossing requests

1.- The role ids seems to be static. They are defined here in the iOS app:

https://github.com/owncloud/ios-sdk/blob/ceda5fae8e20b2324544560dd791ea16fe94cb57/ownCloudSDK/Share/Roles/OCShareRole.m#L133

There are V1 and V2. iOS has a method that transforms V1 in V2: https://github.com/owncloud/web/blob/6983ef727ea25430d57c7625ec53f0b249132246/packages/web-client/src/graph/drives/drives.ts#L8


2.- Check in `graph/v1.0/me/drives` the role of the user in the given space (`editor` , `manager` , `viewer`)

3.- Get the id of the permission (from step 1.-)

4.- Look for it in a request to `graph/v1beta1/roleManagement/permissions/roleDefinitions` (field `id`)

5.- If the match contains `libre.graph/driveItem/permissions/update` in the "allowedResourceActions", the edit operation is allowed. Otherwise, it is not.

## Straight from permissions

Request to `graph/v1beta1/drives/<drive-id>/root/permissions`

That authenticated request will return the permissions of the user in the given space. In the response, the array `@libre.graph.permissions.actions.allowedValues"` should include `"libre.graph/driveItem/permissions/update"`to get the permission. There are more information in the response, but ftm, i think it's no relevant

PR #4687, opened 2025-09-26
[FEATURE REQUEST] Edit space

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4607

_____

## QA

Test Plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.7/Spaces%20-%20Edit.md

Reports

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListDiffUtil.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListDiffUtil.kt
@@ -2,8 +2,9 @@
  * ownCloud Android client application
  *
  * @author Juan Carlos Garrote Gascón
+ * @author Jorge Aguado Recio
  *
- * Copyright (C) 2023 ownCloud GmbH.
+ * Copyright (C) 2025 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -42,7 +43,7 @@ class SpacesListDiffUtil(
         val oldItem = oldList[oldItemPosition]
         val newItem = newList[newItemPosition]
 
-        if ((oldItem.name != newItem.name) || (oldItem.description != newItem.description) ||
+        if ((oldItem.name != newItem.name) || (oldItem.description != newItem.description) || (oldItem.quota != newItem.quota) ||
             (oldItem.getSpaceSpecialImage()?.id != newItem.getSpaceSpecialImage()?.id)) {
             return false
         }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -71,7 +71,8 @@ class SpacesListFragment :
     private val binding get() = _binding!!
 
     private var isMultiPersonal = false
-    private var editSpacesPermission: Boolean = false
+    private var editSpacesPermission = false
+    private var editQuotaPermission = false
     private lateinit var currentSpace: OCSpace
 
     private val spacesListViewModel: SpacesListViewModel by viewModel {
@@ -119,7 +120,12 @@ class SpacesListFragment :
         }
 
         binding.fabCreateSpace.setOnClickListener {
-            val dialog = CreateSpaceDialogFragment.newInstance(requireArguments().getString(BUNDLE_ACCOUNT_NAME), this)
+            val dialog = CreateSpaceDialogFragment.newInstance(
+                isEditMode = false,
+                canEditQuota = false,
+                currentSpace = null,
+                listener = this
+            )
             dialog.show(requireActivity().supportFragmentManager, DIALOG_CREATE_SPACE)
             binding.fabCreateSpace.isFocusable = false
         }
@@ -176,6 +182,7 @@ class SpacesListFragment :
                         uiResult.data?.let {
                             binding.fabCreateSpace.isVisible = it.contains(DRIVES_CREATE_ALL_PERMISSION)
                             editSpacesPermission = it.contains(DRIVES_READ_WRITE_ALL_PERMISSION)
+                            editQuotaPermission = it.contains(DRIVES_READ_WRITE_PROJECT_QUOTA_ALL_PERMISSION)
                         }
                     }
                     is UIResult.Loading -> { }
@@ -304,8 +311,12 @@ class SpacesListFragment :
                 dialog.dismiss()
                 when(menuOption) {
                     SpaceMenuOption.EDIT -> {
-                        val accountName = requireArguments().getString(BUNDLE_ACCOUNT_NAME)
-                        val editDialog = CreateSpaceDialogFragment.newInstance(accountName, this@SpacesListFragment)
+                        val editDialog = CreateSpaceDialogFragment.newInstance(
+                            isEditMode = true,
+                            canEditQuota = editQuotaPermission,
+                            currentSpace = currentSpace,
+                            listener = this@SpacesListFragment
+                        )
                         editDialog.show(requireActivity().supportFragmentManager, DIALOG_CREATE_SPACE)
                     }
                 }
@@ -321,6 +332,7 @@ class SpacesListFragment :
         const val BUNDLE_ACCOUNT_NAME = "accountName"
         const val DRIVES_CREATE_ALL_PERMISSION = "Drives.Create.all"
         const val DRIVES_READ_WRITE_ALL_PERMISSION = "Drives.ReadWrite.all"
+        const val DRIVES_READ_WRITE_PROJECT_QUOTA_ALL_PERMISSION = "Drives.ReadWriteProjectQuota.all"
 
         private const val DIALOG_CREATE_SPACE = "DIALOG_CREATE_SPACE"
 
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -204,45 +204,18 @@ class SpacesListFragment :
             }
         }
 
-        collectLatestLifecycleFlow(spacesListViewModel.menuOptions) { menuOptions ->
-            val spaceOptionsBottomSheetBinding = FileOptionsBottomSheetFragmentBinding.inflate(layoutInflater)
-            val dialog = BottomSheetDialog(requireContext())
-            dialog.setContentView(spaceOptionsBottomSheetBinding.root)
-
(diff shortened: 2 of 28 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListDiffUtil.kt
```diff
@@ -42,7 +43,7 @@ class SpacesListDiffUtil(
         val oldItem = oldList[oldItemPosition]
         val newItem = newList[newItemPosition]
 
-        if ((oldItem.name != newItem.name) || (oldItem.description != newItem.description) ||
+        if ((oldItem.name != newItem.name) || (oldItem.description != newItem.description) || (oldItem.quota != newItem.quota) ||
```
jesmrec, line 46: What is this for?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -211,6 +242,44 @@ class SpacesListFragment :
     }
 
     override fun onThreeDotButtonClick(ocSpace: OCSpace) {
+        currentSpace = ocSpace
+        spacesListViewModel.filterMenuOptions(ocSpace, editSpacesPermission)
```
jesmrec, line 246: Question: will you call `filterMenuOptions` with every permission individually? would it be more compact to use a data structure that contains all the permissions?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -247,33 +320,27 @@ class SpacesListFragment :
...
+                    SpaceMenuOption.EDIT -> {
+                        val editDialog = CreateSpaceDialogFragment.newInstance(
+                            isEditMode = true,
+                            canEditQuota = editQuotaPermission,
+                            currentSpace = currentSpace,
+                            listener = this@SpacesListFragment
+                        )
+                        editDialog.show(requireActivity().supportFragmentManager, DIALOG_CREATE_SPACE)
+                    }
+                }
```
jesmrec, line 340: When the `menuOption` grows with more operations, would it be interesting to handle it in another fun?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
```diff
@@ -44,19 +46,51 @@ class CreateSpaceDialogFragment : DialogFragment() {
...
-                    spaceSubtitle = binding.createSpaceDialogSubtitleValue.text.toString(),
-                    spaceQuota = spaceQuota
-                )
+                val spaceName = createSpaceDialogNameValue.text.toString()
+                val spaceSubtitle = createSpaceDialogSubtitleValue.text.toString()
+                val spaceQuota = convertToBytes(createSpaceDialogQuotaUnit.selectedItem.toString())
+
+                if (isEditMode) {
+                    createSpaceListener.editSpace(
+                        spaceId = currentSpace!!.id,
```
jesmrec: Could that lead to NPE somehow?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
```diff
@@ -44,19 +46,51 @@ class CreateSpaceDialogFragment : DialogFragment() {
...
             createSpaceButton.setOnClickListener {
-                val spaceQuota = convertToBytes(binding.createSpaceDialogQuotaUnit.selectedItem.toString())
-                createSpaceListener.createSpace(
-                    spaceName = binding.createSpaceDialogNameValue.text.toString(),
-                    spaceSubtitle = binding.createSpaceDialogSubtitleValue.text.toString(),
-                    spaceQuota = spaceQuota
-                )
+                val spaceName = createSpaceDialogNameValue.text.toString()
+                val spaceSubtitle = createSpaceDialogSubtitleValue.text.toString()
+                val spaceQuota = convertToBytes(createSpaceDialogQuotaUnit.selectedItem.toString())
```
jesmrec, line 78: Should this conversion be done in the view model? since the fragments responsibility should be only displaying the data.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
```diff
@@ -89,17 +123,41 @@ class CreateSpaceDialogFragment : DialogFragment() {
         return quotaNumber * 1_000_000_000L
     }
 
+    private fun getQuotaValueForSpinner(spaceQuota: Long): Int {
+        val totalGB = spaceQuota / 1_000_000_000.0
+        return when (totalGB) {
+            1.0 -> 0
+            2.0-> 1
```
jesmrec: blank missing in between the `2.0` and `->`

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
```diff
@@ -84,20 +93,24 @@ data class SpaceSpecial(
...
 data class SpaceFile(
     val mimeType: String
-)
+) : Parcelable
 
+@Parcelize
 data class SpaceSpecialFolder(
     val name: String
-)
+) : Parcelable
```
jesmrec, line 116: Why are now all these data classes Parcelable?

jesmrec (changes requested):

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
joragua: Added 👍🏻

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
joragua: It shouldn't because when `isEditMode` is true, `currentSpace` can't be null. Anyway, I can wrap it in a `.let { }` block as a safeguard, just like in the previous lines.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListDiffUtil.kt
joragua: This is used to refresh the adapter list when a changed is detected. Without this clause, the list is not refreshed when the quota is modified and it's not displayed correctly in the dialog.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Yes, we can move all the logic to a separate function but right now there is only one option in the bottom menu. We can refactor it in future issues when we add more options or we feel the function is overloaded. Do you agree? 🤔

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
joragua: Because one of the arguments in `newInstance()` function of `CreateSpaceDialogFragment` is an `OCSpace`, which must implement `Parcelable`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Good catch! In the future, we will probably need to handle more permissions in the `filterMenuOptions` function so it might be interesting to have a data structure with all necessary information. I can create it in a future issue, once additional permissions are needed or in this PR, but it might delay other tasks like QA. What do you think? 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
joragua: Yeah, that's right. The fragment should only display data and the view model should handle the logic but in this case, this fragment doesn't have a linked view model because it is using an interface for editing/creating a space. So, I think we can make an exception here and skip instantiating the view model to avoid executing all the methods in the `init` block of `SpaceListViewModel`.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListDiffUtil.kt
jesmrec: 👍  Thanks!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: It's OK to follow that path, since the current PR just has one filter. When it encloses more parameters, will do the refactor

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: Sure, same as above

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
jesmrec: ok, it will be totally safe, then. Thanks 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
jesmrec: You're right. Let's let it be :D

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/createspace/CreateSpaceDialogFragment.kt
jesmrec: 💯

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
jesmrec: 👍

jesmrec:
LGTM 👍

jesmrec:
Approved on my side, everything perfect! 💯

PR #4690, opened 2025-10-08
[FIX] Improve some emulator and gradle parameters 

jesmrec:
The update test executed in GitHub actions has some flaky executions. In order to make it more robust and deterministic, some changes were done in the workflow:

- Adding `-no-snapshot` to the emulator options: That options will assure that no leftovers from other executions affect to the current one. It will work together with the `force-avd-creation` parameter.
- Switching `auto` for `on` in the `-accel` parameter: if hardware acceleration is available, it will be used making the test more robust. If not available, test will fail. GitHub actions' runners have the option available, so, we can assume that "con"
- Increasing memory assigned to the emulator, so, it will be less likely to hang (from 2048 to 4096 MB)
- Added `--no-daemon` to `gradlew` command. That should not affect because there is only one `gradlew` command, but in the local executions could make the execution more performant.

In addition, added `dependabot/*` to the `on:` section, so the test will be executed in PRs opened by dependabot.

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -9,7 +9,7 @@ on:
     - improvement/*
     - release/*
     - technical/*
-    - dependabot/*
+    - 'dependabot/**'
 
 permissions:
   # Only need read access to repository contents
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -9,6 +9,7 @@ on:
     - improvement/*
     - release/*
     - technical/*
+    - dependabot/*
 
 permissions:
   # Only need read access to repository contents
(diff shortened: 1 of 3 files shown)


.github/workflows/update.yml
```diff
@@ -9,6 +9,7 @@ on:
     - improvement/*
     - release/*
     - technical/*
+    - dependabot/*
```
joragua:
```suggestion
    - 'dependabot/**'
```

It should be like this, right? Since `dependabot` branches contain multiple slashes

jesmrec: sure, you're right 👍

Issue #4611, opened 2025-06-25
[FEATURE REQUEST] Disable/Remove space


jesmrec:
An space could be disabled, so, no one can interact with it (like blocked). A disabled space:

- could be finally removed
- could be enabled again

AC:

- Add a new option to the bottom menu: `Disable space`. Only available if the user is  `Space Admin`, `Admin` (platform permission: `Drives.DeleteProject.all`) or `Space manager` (space permission: `libre.graph/driveItem/permissions/delete`)

- If space is disabled:
  - Add two new options to the bottom menu: `Enable space` (re-enable the space) and `Delete space` (remove the space after asking for confirmation). Only available if the user is `Space manager`

- Refresh list after every operation

- Extra ball: showing an option in the view to show/hide disabled spaces. By default, disabled spaces are hidden, but web and iOS have an option to list them (as disabled). This one could be moved to another issue

#### Disable | Delete
 DELETE `https://<server>/graph/v1.0/drives/<drive-id>` (with header `Purge: T` for delete operation)

#### Enable 
PATCH `https://<server>/graph/v1.0/drives/<drive-id>` (with header `Restore: T` and empty body)
			
persist in DB the status (enabled/disabled) -> new field?
remove from DB if deleted (automatically when removed from backend)

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/space_lifecycle_operations
- [x] Development tasks
  - [x] Add a menu option to **disable** a space
  - [x] Implement the required methods and operations to disable a space and refresh the list
  - [x] Add a new menu option to **enable** a space
  - [x] Implement the required methods and operations to enable a space and refresh the list
  - [x] Add a new menu option to **delete** a space
  - [x] Implement the required methods and operations to delete a space and refresh the list
  - [x] Display disabled spaces in the spaces list (only for space managers)
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/space_lifecycle_operations

jesmrec:
how to display the show / hide disabled spaces:

- Show in text like the web
- Show in minifabs
- three-dot-button on the top right, with the option. 
- option like grid/list in the list of files
- list all spaces (including disabled spaces)

PR #4696, opened 2025-10-16
[FEATURE REQUEST] Disable/Remove space

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4611

_____

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.7/Spaces%20-%20Disable%20Remove.md

Reports:

- [x] (1) User light does not refresh spaces view https://github.com/owncloud/android/pull/4696#issuecomment-3460266299 [WONT FIX]
- [X] (2) Error message when lacking permissions https://github.com/owncloud/android/pull/4696#issuecomment-3460588597 [SERVER ISSUE]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ ownCloud admins and users.
 * Enhancement - Add account ID to the user information: [#4605](https://github.com/owncloud/android/issues/4605)
 * Enhancement - Create a new space: [#4606](https://github.com/owncloud/android/issues/4606)
 * Enhancement - Edit a space: [#4607](https://github.com/owncloud/android/issues/4607)
+* Enhancement - Disable/Remove a space: [#4611](https://github.com/owncloud/android/issues/4611)
 * Enhancement - Update test in GitHub Actions: [#4663](https://github.com/owncloud/android/pull/4663)
 * Enhancement - New workflow to generate a build from "latest" tag on demand: [#4681](https://github.com/owncloud/android/pull/4681)
 * Enhancement - Make Update test more robust: [#4690](https://github.com/owncloud/android/pull/4690)
@@ -142,6 +143,15 @@ ownCloud admins and users.
    https://github.com/owncloud/android/issues/4688
    https://github.com/owncloud/android/pull/4687
    https://github.com/owncloud/android/pull/4694
+
+* Enhancement - Disable/Remove a space: [#4611](https://github.com/owncloud/android/issues/4611)
+
+   New options to disable, enable and remove spaces have been added to the bottom
+   sheet, available only to users with the required permissions when the three-dot
+   menu button is tapped.
+
+   https://github.com/owncloud/android/issues/4611
+   https://github.com/owncloud/android/pull/4696
 
 * Enhancement - Update test in GitHub Actions: [#4663](https://github.com/owncloud/android/pull/4663)
 


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListAdapter.kt
```diff
@@ -55,10 +56,13 @@ class SpacesListAdapter(
...
             val space = spacesList[position]
-
-            spacesListItemCard.setOnClickListener {
-                listener.onItemClick(space)
+            if (space.isDisabled) {
+                spacesListItemCard.isClickable = false
+            } else {
+                spacesListItemCard.setOnClickListener { listener.onItemClick(space) }
             }
+            spacesListItemImage.alpha = if (space.isDisabled) 0.5f else 1f
```
jesmrec, line 64: Which is the real effect here? afaik, alpha handles the image opacity, which is changed to 0.5 in case of disabled space. Can you elaborate?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -214,10 +227,35 @@ class SpacesListFragment :
...
+
+        collectLatestLifecycleFlow(spacesListViewModel.deleteSpaceFlow) { event ->
+            event?.let {
+                when (val uiResult = event.peekContent()) {
+                    is UIResult.Success -> { showMessageInSnackbar(getString(R.string.delete_space_correctly)) }
+                    is UIResult.Loading -> { }
+                    is UIResult.Error -> { showErrorInSnackbar(R.string.delete_space_failed, uiResult.error) }
+                }
+            }
+        }
```
jesmrec, line 224: That's ok in functional terms, but it seems like a code repetition for every operation (createSpaceFlow, editSpaceFlow, disableSpaceFlow, enableSpaceFlow, deleteSpaceFlow) where the only difference is the messages to show in the snackbar. Might it be refactored somehow?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListViewModel.kt
```diff
@@ -176,19 +190,49 @@ class SpacesListViewModel(
...
+
+    fun deleteSpace(spaceId: String){
+        viewModelScope.launch(coroutinesDispatcherProvider.io) {
+            when (val result = disableSpaceUseCase(DisableSpaceUseCase.Params(accountName, spaceId, true))) {
+                is UseCaseResult.Success -> _deleteSpaceFlow.emit(Event(UIResult.Success(result.getDataOrNull())))
+                is UseCaseResult.Error -> _deleteSpaceFlow.emit(Event(UIResult.Error(error = result.getThrowableOrNull())))
+            }
+            refreshSpacesFromServerAsyncUseCase(RefreshSpacesFromServerAsyncUseCase.Params(accountName))
+        }
+    }
```
jesmrec, line 239: Same here, any way to refactor to avoid repeated code?

owncloudApp/src/main/res/layout/spaces_list_item.xml
```diff
@@ -46,13 +46,41 @@
...
+                    android:visibility="gone"
+                    android:elevation="8dp"
+                    android:textStyle="bold"
+                    android:background="@drawable/disabled_space_label_background"
+                    android:paddingHorizontal="@dimen/standard_padding"
+                    android:paddingVertical="10dp"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent"
+                    app:layout_constraintBottom_toBottomOf="parent"
+                    app:layout_constraintEnd_toEndOf="parent" />
```
jesmrec, line 70: Do we have here accessibility options?

owncloudData/src/test/java/com/owncloud/android/data/spaces/datasources/implementation/OCRemoteSpacesDataSourceTest.kt
```diff
@@ -150,4 +150,72 @@ class OCRemoteSpacesDataSourceTest {
...
+            ocSpaceService.enableSpace(OC_SPACE_PROJECT_WITH_IMAGE.id)
+        } returns enableSpaceResult
+
+        ocRemoteSpacesDataSource.enableSpace(OC_ACCOUNT_NAME, OC_SPACE_PROJECT_WITH_IMAGE.id)
+
+        verify(exactly = 1) {
+            clientManager.getSpacesService(OC_ACCOUNT_NAME)
+            ocSpaceService.enableSpace(OC_SPACE_PROJECT_WITH_IMAGE.id)
+        }
+    }
```
jesmrec, line 219: Those tests just seem to check that exactly one call is done, but no extra assertions. As we have already discussed, we try to avoid asserts that checks the mockk result, but anyway, we did it by convention. Is here something missing?

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
```diff
@@ -350,4 +350,31 @@ class OCSpacesRepositoryTest {
...
+
+    @Test
+    fun `enableSpace enables a space correctly`() {
+        ocSpacesRepository.enableSpace(OC_ACCOUNT_NAME, OC_SPACE_PROJECT_WITH_IMAGE.id)
+
+        verify(exactly = 1) {
+            remoteSpacesDataSource.enableSpace(OC_ACCOUNT_NAME, OC_SPACE_PROJECT_WITH_IMAGE.id)
+        }
+    }
+
```
jesmrec, line 379: Same here

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListAdapter.kt
joragua: The idea is to change the image opacity for disabled spaces, so the default image (space icon) looks faded without applying a grayscale filter.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Good catch! 😄 I’ve added a generic method that observes a flow and displays the appropriate snackbar message.

owncloudData/src/test/java/com/owncloud/android/data/spaces/datasources/implementation/OCRemoteSpacesDataSourceTest.kt
joragua: These methods (`disableSpace` and `enableSpace`) return `Unit`, so it's not necessary to add assertions since there is no result to compare against.

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
joragua: Check previous comment 😄

owncloudApp/src/main/res/layout/spaces_list_item.xml
joragua: Nop, added! ✅

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListViewModel.kt
joragua: Refactored! 👍🏻 I also added a generic method to execute use cases and emit their results to the appropriate `Flow`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListAdapter.kt
jesmrec: The space icon is almost covered by the `Disable` space, right? that effect would be very hidden since a disabled space does not show the image.

owncloudApp/src/main/res/layout/spaces_list_item.xml
jesmrec: 👍

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
jesmrec: 👍

owncloudData/src/test/java/com/owncloud/android/data/spaces/datasources/implementation/OCRemoteSpacesDataSourceTest.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: 💯

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListViewModel.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListAdapter.kt
jesmrec: After comparing the images with and without the fade effect, it's better to keep the `0.5f` approach. So, no changes to do.

jesmrec:
CR approved!! 🚀 

Let's move to QA

jesmrec:
### (1) [WONT FIX]

1. Login with an existing user (with role `User`) that has one space available
2. Check the `Spaces` view -> space is listed 
3. In the web admin dashboard, change the user to `User Light` 
4. Refresh the view without browsing or killing the app

**Current:**

`server could not be reached` 

**Expected:**

Correct refresh. It could be related with [https://github.com/owncloud/android/pull/4698](https://github.com/owncloud/android/pull/4698) ?? seems to be a similar effect

Xiaomi Redmi 13, Android 15
`5695ca2c1e`

jesmrec:
(1) -> after discussion, it won't be fixed since we assume the role change user -> user light is not supported

jesmrec:
### (2) [SERVER ISSUE]

1. An space manager opens 3-dot-button over an enabled space
2. Before any operation, another space manager revokes `manager` permission and turns him to `editor`
3. In app, click on `Disable` and `YES`

**Current:**

Error `it was not found` in snackbar, but the space is visible and accessible

**Expected:**

An message like `Permission error` may fit better to that error.  `it was not found` makes more sense if the space is no longer available or visible.

Is that change feasible?

Xiaomi Redmi 13, Android 15
`5695ca2c1e`

joragua:
**(2)** -> The server returns `404` status code instead of `401`, so the snackbar message says `it was not found`. Based on that, I think the bug is on the server side and not in our implementation. 🤔

jesmrec:
Approved on my side 💯. Moving forward!

PR #4699, opened 2025-10-23
[TECHNICAL] Rewrite 4.7.0 release notes

jesmrec:
All space management features in only one note.

Existing translations removed.

Calens not needed for this.

## Related Issues
App:

_____

### DIFF
--- a/owncloudApp/src/main/res/values/strings.xml
+++ b/owncloudApp/src/main/res/values/strings.xml
@@ -741,10 +741,8 @@
     <string name="release_notes_bugfixes_subtitle">Some bugs were fixed to improve experience in the app</string>
     <string name="release_notes_4_7_0_title_new_layout_for_spaces">New design for spaces list</string>
     <string name="release_notes_4_7_0_subtitle_new_layout_for_spaces">A new design for Infinite Scale users in the spaces list, with a three-dot button on each space card that opens an options menu when clicked</string>
-    <string name="release_notes_4_7_0_title_create_new_spaces">Create project spaces</string>
-    <string name="release_notes_4_7_0_subtitle_create_new_spaces">Infinite Scale users with right permissions can now create project spaces directly from the app</string>
-    <string name="release_notes_4_7_0_title_edit_spaces">Edit project spaces</string>
-    <string name="release_notes_4_7_0_subtitle_edit_spaces">Infinite Scale users with right permissions can now edit project spaces directly from the app</string>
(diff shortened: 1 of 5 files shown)


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -741,10 +741,8 @@
...
     <string name="release_notes_4_7_0_title_new_layout_for_spaces">New design for spaces list</string>
     <string name="release_notes_4_7_0_subtitle_new_layout_for_spaces">A new design for Infinite Scale users in the spaces list, with a three-dot button on each space card that opens an options menu when clicked</string>
-    <string name="release_notes_4_7_0_title_create_new_spaces">Create project spaces</string>
-    <string name="release_notes_4_7_0_subtitle_create_new_spaces">Infinite Scale users with right permissions can now create project spaces directly from the app</string>
-    <string name="release_notes_4_7_0_title_edit_spaces">Edit project spaces</string>
-    <string name="release_notes_4_7_0_subtitle_edit_spaces">Infinite Scale users with right permissions can now edit project spaces directly from the app</string>
+    <string name="release_notes_4_7_0_title_space_management">Space management</string>
+    <string name="release_notes_4_7_0_subtitle_space_management">Infinite Scale users with right permissions can now manage spaces(create,edit,disable…) directly from the app. Membership not supported yet.</string>
```
joragua: Is it necessary to add `Membership not supported yet` in the release note? 🤔 If we list all the operations that will be included in the release, I think we can remove that part. I would rewrite the release note like this (just a suggestion):
```suggestion
    <string name="release_notes_4_7_0_subtitle_space_management">Infinite Scale users with right permissions can now manage spaces (create, edit, disable, enable and delete) directly from the app</string>
```

jesmrec: the point is: image upload or image change are also available operations (or we expect that). We can make a "long" list of operations, or simply tell users which ones are not available. Which one sounds the best?

joragua: We can assume both options are `Edit`, so we only would have: `Create, edit, disable, enable and delete` in the list. I don't know whether is too long but I wouldn't add more options there since they won't be completed in this release.

jesmrec: Let's get your approach 👍

joragua (approved):
LGTM! 💯 QA not needed here

Issue #4693, opened 2025-10-10
[FEATURE REQUEST] Show space quota


jesmrec:
The current space info shows:

<img width="326" height="68" alt="Image" src="https://github.com/user-attachments/assets/18c688af-aa49-423c-bfdf-840877be137b" />

We should add the assigned quota together with the used size like:

`5.9 MB of 1 GB used`

- In case there is no quota restriction (`"total": 0`), the current approach is valid.
- When a space is disabled, the `used` value doesn't exist -> `No storage information available` as string?

Quota of a space comes together with the size info as `total`, in the same json object from the `drives` endpoint:

```
            "quota": {
                "remaining": 999018577,
                "state": "normal",
                "total": 1000000000,
                "used": 981423
            },
```

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
  - [ ] Implement unit tests (if needed)
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

PR #4704, opened 2025-10-29
[FEATURE REQUEST] Show space quota

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4693

_____

## QA

Checks: https://github.com/owncloud/android/pull/4704#issuecomment-3462879567

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -306,17 +306,14 @@ class SpacesListFragment :
             val usedQuota = quota.used
             val totalQuota = quota.total
 
-            val quotaText = if (usedQuota == null) {
-                getString(R.string.drawer_unavailable_used_storage)
-            } else if (totalQuota == 0L) {
-                DisplayUtils.bytesToHumanReadable(usedQuota, requireContext(), true)
-            } else {
-                getString(
+            val quotaText = when {
+                usedQuota == null -> getString(R.string.drawer_unavailable_used_storage)
+                totalQuota == 0L -> DisplayUtils.bytesToHumanReadable(usedQuota, requireContext(), true)
+                else -> getString(
                     R.string.drawer_quota,
                     DisplayUtils.bytesToHumanReadable(usedQuota, requireContext(), true),
                     DisplayUtils.bytesToHumanReadable(totalQuota, requireContext(), true),
-                    quota.getRelative().toString()
-                )
+                    quota.getRelative().toString())
             }
             spaceSizeBottomSheet.text = quotaText
         }


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -302,7 +302,24 @@ class SpacesListFragment :
...
+            } else if (totalQuota == 0L) {
+                DisplayUtils.bytesToHumanReadable(usedQuota, requireContext(), true)
+            } else {
+                getString(
+                    R.string.drawer_quota,
+                    DisplayUtils.bytesToHumanReadable(usedQuota, requireContext(), true),
+                    DisplayUtils.bytesToHumanReadable(totalQuota, requireContext(), true),
+                    quota.getRelative().toString()
+                )
+            }
```
jesmrec: does it make sense to replace by a `when` clause?

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
```diff
@@ -73,7 +74,12 @@ data class SpaceQuota(
     val state: String?,
     val total: Long,
     val used: Long?,
-) : Parcelable
+) : Parcelable {
+
+    fun getRelative(): Double =
+        if (used == null) 0.0 else ((used * 100).toDouble() / total * 100).roundToLong() / 100.0
```
jesmrec, line 80: i don't get this way to calculate the % of quota used. Why don't:

`((used / total).roundToLong) * 100`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: In this case, it compares different properties (`used` and `total`) and `when` should be for just one parameter, just like the `switch` in Java. So, I prefer tu use `if | else if` here. What do you think? 🤔

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
joragua: This method is similar to the other one in `UserQuota` class (see it [here](https://github.com/owncloud/android/blob/master/owncloudDomain/src/main/java/com/owncloud/android/domain/user/model/UserQuota.kt)). The idea is to calculate the relative quota with only two decimals. By multiplying by 100 and then using `.roundToLong()`, we obtain the most accurate rounded value. Anyway, I can use `BigDecimal` in order to get the two decimals but I'm not sure if it is worth it here.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: as a question: is it not posible to use something like:

```
val quotaText = when {
  usedQuota == null -> ...
  totalQuota == 0L -> ...
  else -> ...
}
```

owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/model/OCSpace.kt
jesmrec: Let's keep consistence with other parts of the code. But, i think it makes more calculations than needed 😄 . In QA, i will test if the values displayed are correct (i expect so, will see)

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: After discussion, I refactored the code to use a `when` statement instead of an `if-else` chain, since it’s just a variable assignment. Nice catch! 😄

jesmrec: 💯

jesmrec:
LGTM 👍 

Moving to QA

jesmrec:
QA checks:

- [x] Space disabled -> `No storage usage information available`
- [X] Space enabled with no limit -> just the used space
- [x] Space enabled with limit -> show the used space, quota and % of used
   - [X] 0.002GB  
   - [X] 0.5GB 
   - [X] 1GB 
   - [x] 5GB 
   - [x] 100GB
- [X] Multiple spaces

jesmrec:
Moving on! 🚀 

Approved

Issue #4691, opened 2025-10-10
[FEATURE REQUEST] Update space image


jesmrec:
Every space has a representative image as presentation. The app should be able to offer users the ability of changing the space image.

AC:

- New option in the bottom menu: `Edit image`. Permission level required to perform the action: `"libre.graph/driveItem/permissions/update",`. Available in `graph/v1beta1/drives/<drive-id>/root/permissions` endpoint. As checked in the web, only space `manager` users seem to be able to edit the image, regardless the platform role.
- By choosing that option, the user can pick an image from the system to be set as space image. **Only one**

- The selected image will be pushed to the server to the folder `/dav/spaces/<drive-id>/.space/`, by using an authenticated `PUT` request, adding the image content to the payload.
- Once pushed, a `PATCH` operation to `graph/v1.0/drives/<drive-id>` needed to set the pushed image as space image. The `PATCH` request will include in the payload:

```
{
    "special": [
        {
            "id": "<file-id>",
            "specialFolder": {
                "name": "image"
            }
        }
    ]
}
```
- The accepted image formats are: `JPEG`, `JPG`, `PNG`, `BMP` and `GIF`
- Errors: regular errors (connection, server unreachable...).

Extra ball: set an emoji as image

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/update_space_image
- [x] Development tasks
  - [x] Add a menu option to **update the space image**
  - [x] Open SAF with some restrictions: only images and multi-selection not allowed
  - [x] Upload the chosen file to `/.space` directory
  - [x] Update the space image using the `PATCH` operation to `graph/v1.0/drives/<drive-id>`
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/update_space_image into master

PR #4705, opened 2025-10-30
[FEATURE REQUEST] Update space image

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4691

_____

## QA

Test plan:

https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.7/Spaces%20-%20Edit%20image.md

Reports:

- [x] (1) Lack of error message when no permissions https://github.com/owncloud/android/pull/4705#issuecomment-3490196619 [FIXED]
- [X] (2) Space image not updated after recovering connection https://github.com/owncloud/android/pull/4705#issuecomment-3490672287  [[WONT FIX]](https://github.com/owncloud/android/pull/4705#issuecomment-3492313716)

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
@@ -247,10 +247,17 @@ class SpacesListFragment :
             val remotePath = SPACE_CONFIG_DIR + selectedImageName
             val matchedTransfer = transfersWithSpace.map { it.first }.find { it.remotePath == remotePath }
 
-            if (matchedTransfer != null && matchedTransfer.status == TransferStatus.TRANSFER_SUCCEEDED &&
-                lastUpdatedRemotePath != matchedTransfer.remotePath) {
-                spacesListViewModel.editSpaceImage(currentSpace.id, matchedTransfer.remotePath)
-                lastUpdatedRemotePath = matchedTransfer.remotePath
+            if (matchedTransfer != null && lastUpdatedRemotePath != matchedTransfer.remotePath) {
+                when(matchedTransfer.status) {
+                    TransferStatus.TRANSFER_SUCCEEDED -> {
+                        spacesListViewModel.editSpaceImage(currentSpace.id, matchedTransfer.remotePath)
+                        lastUpdatedRemotePath = matchedTransfer.remotePath
+                    }
+                    TransferStatus.TRANSFER_FAILED -> {
+                        showMessageInSnackbar(getString(R.string.edit_space_image_failed))
+                    }
+                    else -> { }
+                }
             }
         }
 


owncloudApp/src/main/res/values/strings.xml
```diff
@@ -851,6 +851,9 @@
     <string name="edit_space">Edit space</string>
     <string name="edit_space_correctly">Space updated correctly</string>
     <string name="edit_space_failed">Space could not be updated</string>
+    <string name="edit_space_image">Edit space image</string>
+    <string name="edit_space_image_correctly">Space image updated correctly</string>
+    <string name="edit_space_image_failed">Space image could not be updated</string>
```
jesmrec: I'd go for a shorter one: `Edit image` . Web and iOS clients go for that approach. Then, if we go for it, the snackbar messages should be shortened as well

`Image updated correctly`
`Image could not be updated`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -210,6 +237,17 @@ class SpacesListFragment :
             showSpaceMenuOptionsDialog(menuOptions)
         }
 
+        collectLatestLifecycleFlow(transfersViewModel.transfersWithSpaceStateFlow) { transfersWithSpace->
```
jesmrec: missing blank between `transfersWithSpace` and `->` ?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -210,6 +237,17 @@ class SpacesListFragment :
...
+            val remotePath = SPACE_CONFIG_DIR + selectedImageName
+            val matchedTransfer = transfersWithSpace.map { it.first }.find { it.remotePath == remotePath }
+
+            if (matchedTransfer != null && matchedTransfer.status == TransferStatus.TRANSFER_SUCCEEDED &&
+                lastUpdatedRemotePath != matchedTransfer.remotePath && lastUpdatedRemotePath != null) {
```
jesmrec: 💯  to avoid repetitions

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -382,6 +420,14 @@ class SpacesListFragment :
                             negativeButtonText = getString(R.string.common_no)
                         )
                     }
+                    SpaceMenuOption.EDIT_IMAGE -> {
+                        val action = Intent(Intent.ACTION_OPEN_DOCUMENT).apply {
+                            addCategory(Intent.CATEGORY_OPENABLE)
+                            type = "image/*"
+                            putExtra(Intent.EXTRA_MIME_TYPES, arrayOf("image/jpeg", "image/png", "image/bmp", "image/x-ms-bmp", "image/gif"))
```
jesmrec: Should this static values be moved to a const or companion object?

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
```diff
@@ -350,6 +351,31 @@ class OCSpacesRepositoryTest {
...
+
+        verify(exactly = 1) {
+            remoteSpacesDataSource.editSpaceImage(
+                accountName = OC_ACCOUNT_NAME,
+                spaceId = OC_SPACE_PROJECT_WITH_IMAGE.id,
+                imageId = OC_SPACE_SPECIAL_IMAGE.id
+            )
+        }
+    }
+
```
jesmrec, line 378: Tests added are OK, just as a question: would it make sense to add negative tests? i mean, to check the correct response if the action failed (image editing returned an error, a null or so)

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
```diff
@@ -382,6 +420,14 @@ class SpacesListFragment :
                             negativeButtonText = getString(R.string.common_no)
                         )
                     }
+                    SpaceMenuOption.EDIT_IMAGE -> {
+                        val action = Intent(Intent.ACTION_OPEN_DOCUMENT).apply {
```
jesmrec, line 437: As a question: selecting just one file is the default behaviour?

jesmrec (changes requested):

owncloudApp/src/main/res/values/strings.xml
joragua: Ok! Let's go for it 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Added 👍🏻

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Yes. If you want to allow multiple selections, just add this flag: `putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true)`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
joragua: Yeah! Created new constants for these values ✅

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
joragua: In this case, the two methods covered by the new tests don’t return null or throw errors. The ones that can are the network operations and use cases, but their tests haven’t been updated yet.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: perfect 👍

owncloudData/src/test/java/com/owncloud/android/data/spaces/repository/OCSpacesRepositoryTest.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: Fixed

owncloudApp/src/main/res/values/strings.xml
jesmrec: Done 🥇

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/SpacesListFragment.kt
jesmrec: Cool! perfect!

jesmrec (approved):

jesmrec:
Moving to QA 🚀

jesmrec:
### (1) [FIXED]

1. User1 is manager of the space
2. User1 choose `Edit image`
3. Before choosing an image, another space manager downgrades User1 to `viewer`
4. User1 submit the image

**Current**:

No feedback in the app, there is no image change and no error message (just in the uploads view)

**Expected:**

As other handled errors, an snackbar would tell the user about the issue

NOTE: theoretically, it should work the same by changing to `editor`, but, backend allows the image update. IMO, this is a bug in the backend. I will create an issue there -> https://github.com/owncloud/ocis/issues/11794

Xiaomi Redmi 13, Android 15
`0749466fb`

jesmrec:
### (2) [[WONT FIX]](https://github.com/owncloud/android/pull/4705#issuecomment-3492313716)

1. A user with permissions click on `Edit image`
2. Remove device connection
3. Pick a image -> enqueued 
4. Recover device connection

**Current:**

Upload is done, image is uploaded, but space image is not updated. There is no `PATCH` request after the successful `PUT`

**Expected:**

Space image updated.

Xiaomi Redmi 13, Android 15
`0749466fb`

jesmrec:
Just as a comment: i detected a 7MB limit for the space image. A larger image will not update, but, a success is returned. Pointing to another server issue...

https://github.com/owncloud/ocis/issues/11832

joragua:
**(1)** should be fixed with the last commit @jesmrec

joragua:
About **[(2)](https://github.com/owncloud/android/pull/4705#issuecomment-3490672287)**, there are two different scenarios:

### 1st scenario

1) A user with permissions click on `Edit image`
2) Remove device connection
3) Pick a image -> `QUEUED`
4) Recover device connection
5) Stay in the spaces list

In this case, the space image will be updated automatically once the upload is completed after the connection is recovered.

### 2nd scenario
1) A user with permissions click on `Edit image`
2) Remove device connection
3) Pick a image -> `QUEUED`
4) Navigate to other view
5) Recover device connection

In this case, the image will be uploaded to the `.space` directory but the space image won't be updated. This happens because the `selectedImageName` variable (inside `SpacesListFragment`) is reset while navigating. As a result, once you return to the spaces list, the remote transfer path will not match with the selected image name and the `PATCH` operation isn't triggered. Due to technical limitations related to navigation, fragment lifecycle, and activity handling, I don’t have a clear solution for this at the moment. Anyway, it's a corner case and there is no data-loss, so probably we can assume this problem until we have a potential fix. What do you think, @jesmrec? 🤔 

#### Potential solutions discarded

1) When a transfer has the status `QUEUED` or `FAILED` the `PATCH` operation isn't performed. **Problem:** every transfer goes through the `QUEUED` status before `IN_PROGRESS` regardless of whether the connection is lost or not.

jesmrec:
About the message just above ⬆️ ⬆️ 

In order to reproduce the problem:

1. User has initially connection, that allows to click on `Edit image`
2. Then, user loses connection while the image is chosen
3. After submitting, user must change the view, otherwise the problem is not reproducible.

Problem: image not updated

User must repeat the operation under connection (low impact side-effect)

As @joragua said, there is no critical problem at all. But the solution carries a big complexity to achieve a small piece of improvement. So, i'd go to mark it as `known problem`.

jesmrec:
Approved from my side. 

Known problem: https://github.com/owncloud/android/pull/4705#issuecomment-3492313716

Issue #4707, opened 2025-11-04
[FEATURE REQUEST] Set emoji as space image


joragua:
In the web and iOS app it is possible to set an emoji as space image. The Android app should be able to offers the ability to change the space image using an emoji selected from the device’s system emoji picker. 

- New option in the bottom menu: `Set icon`. Permission level required to perform the action: `"libre.graph/driveItem/permissions/update"`. As checked in the web, only `space manager` users seem to be able to set icon, regardless the platform role.

### Steps

1) Select the emoji from emoji picker in a dialog
2) Convert the emoji (as a string) into an octet-stream or another valid image format.
3) Upload the file to `/.space` directory
4) Update the space image 

Steps 3 and 4 should be the same as for a regular image.

 ### ⚙️ References
Emoji picker: https://developer.android.com/develop/ui/views/text-and-emoji/emoji-picker

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/set_emoji_as_space_image
- [x] Development tasks
  - [x] Add a menu option to **set an icon/emoji** as space image
  - [x] Open a dialog with the emoji picker
  - [x] Convert emoji to an image file
  - [x] Upload the file to `/.space` directory 
  - [x] Update the space image using the `PATCH` operation to `graph/v1.0/drives/<drive-id>`
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/set_emoji_as_space_image into master

joragua:
When `androidx-emoji2-emojipicker` library is added, the dependency `kotlinx-coroutines-test` is updated automatically. 
Currently, we’re using version `1.6.4`, but starting from `1.8.0`, the `TestCoroutineDispatcher` is no longer supported, which causes errors. Therefore, if we want to include the emoji picker library, we’ll need to refactor some ViewModel tests.
I’ve created a separate issue (https://github.com/owncloud/android/issues/4710) for that refactor, and once it’s done, we can continue with this one.

joragua:
### Important notes 

- An `emoji.png` file is generated from selected emoji and stored in the cache directory so it can be uploaded to the `.space` directory. 
- The `forceOverwrite` property is set to `true` to ensure the existing emoji file in the space config directory is replaced.
[...]

PR #4708, opened 2025-11-05
[FEATURE REQUEST] Set emoji as space image

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4707

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.8/Spaces%20-%20Set%20icon.md

Reports:

- [ ] (1) Icon not alligned https://github.com/owncloud/android/pull/4708#issuecomment-3581394685
- [x] (2) Lack of connection error message https://github.com/owncloud/android/pull/4708#issuecomment-3582144299 [WONT FIX]
- [x] (3) Upload not cleared https://github.com/owncloud/android/pull/4708#issuecomment-3582177436 [WONT FIX]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
@@ -57,7 +57,7 @@ class SetSpaceIconDialogFragment : DialogFragment() {
 
     private fun convertEmojiToImageFile(emoji: String): File {
         val paint = Paint(Paint.ANTI_ALIAS_FLAG).apply {
-            textSize = 250f
+            textSize = EMOJI_SIZE
             textAlign = Paint.Align.CENTER
         }
         val bitmap = Bitmap.createBitmap(ICON_WIDTH, ICON_HEIGHT, Bitmap.Config.ARGB_8888)
@@ -83,6 +83,7 @@ class SetSpaceIconDialogFragment : DialogFragment() {
     companion object {
         private const val ICON_HEIGHT = 405
         private const val ICON_WIDTH = 720
+        private const val EMOJI_SIZE = 250f
         private const val EMOJI_FILE_NAME = "emoji.png"
 
         fun newInstance(


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
```diff
@@ -0,0 +1,96 @@
...
+        return file
+    }
+
+    interface SetIconListener {
+        fun setSpaceIcon(fileName: String, localPath: String)
+    }
+
+    companion object {
+        private const val ICON_HEIGHT = 405
+        private const val ICON_WIDTH = 720
```
jesmrec, line 85: This fixed sizes for the dialog may fit correctly for a phone, but a bigger screen like a tablet, it could seem very small, what do you think? is it feasible to make it "adaptable" to the current screen size?

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -745,6 +745,8 @@
     <string name="release_notes_4_7_0_subtitle_new_layout_for_spaces">A new design for Infinite Scale users in the spaces list, with a three-dot button on each space card that opens an options menu when clicked</string>
     <string name="release_notes_4_7_0_title_space_management">Space management</string>
     <string name="release_notes_4_7_0_subtitle_space_management">Infinite Scale users with right permissions can now manage spaces (create, edit, disable, enable and delete) directly from the app</string>
+    <string name="release_notes_4_8_0_title_set_emoji_as_space_image">Set emoji as space image</string>
+    <string name="release_notes_4_8_0_subtitle_set_emoji_as_space_image">Infinite Scale users with right permissions can now set an emoji as space image directly from the app</string>
```
jesmrec: i'd remove the last part `directly from the app`. It's clear that everything new is from the app 😉

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
```diff
@@ -0,0 +1,96 @@
...
+                val emojiFile = convertEmojiToImageFile(emojiItem.emoji)
+                setIconListener.setSpaceIcon(emojiFile.name, emojiFile.absolutePath)
+                dialog?.dismiss()
+            }
+        }
+    }
+
+    private fun convertEmojiToImageFile(emoji: String): File {
+        val paint = Paint(Paint.ANTI_ALIAS_FLAG).apply {
+            textSize = 250f
```
jesmrec: What is this for? should be something static?

owncloudApp/src/main/res/layout/set_space_icon_dialog.xml
```diff
@@ -0,0 +1,57 @@
...
+    android:padding="@dimen/standard_half_margin">
+
+    <TextView
+        android:id="@+id/set_space_icon_title"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginBottom="@dimen/standard_half_margin"
+        android:padding="@dimen/standard_half_padding"
+        android:text="@string/set_space_icon"
+        android:textAppearance="@android:style/TextAppearance.Material.DialogWindowTitle" />
```
jesmrec, line 37: Is here missing the `contentDesctiption` for accessibility?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
joragua: These dimensions are used only to create the `emoji.png` file. On iOS and the web, I believe the same size is used: `720×405`. When the image is displayed on a device (tablet, phone, PC, etc.), it is automatically resized according to the screen dimensions.

owncloudApp/src/main/res/values/strings.xml
joragua: Ok, done! ✅

owncloudApp/src/main/res/layout/set_space_icon_dialog.xml
joragua: It’s just a simple `TextView`, and tools like `TalkBack` read its text directly. Therefore, a `contentDescription` is not needed in my opinion 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
joragua: `Paint` object acts like a “brush.” In this function, it’s necessary because it determines how the emoji is rendered on the bitmap (alignment, size, etc.). In my opinion, it shouldn’t be static, as it’s similar to other mutable objects in the class, and there could be multiple instances of it. Additionally, creating a new `Paint` object is not expensive, so it can be instantiated each time it’s needed.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
jesmrec: 👍

owncloudApp/src/main/res/layout/set_space_icon_dialog.xml
jesmrec: 👍  anyway, it will be checked in QA stage as well

owncloudApp/src/main/res/values/strings.xml
jesmrec: 💯

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/setspaceicon/SetSpaceIconDialogFragment.kt
joragua: In this case, `textSize` property is the dimension (`250x250`) of the emoji on the bitmap. I will refactor the code and set it as a static value 👍🏻

jesmrec: ok, let's check how it looks like in QA phase

jesmrec:
Let's move to QA

jesmrec:
### (1)

Not sure if this is something on our side or not. Uploading an icon, looks like:

Screenshot 2025-11-26 at 14 40 47

it's not center-alliged in the cell. iOS and web do.

Xiaomi Redmi Note 13
Android 15
`7679bfc4`

jesmrec:
### (2) [WONT FIX]

1. Click on Spaces on the bottom bar
2. Click on 3-dot-button over a space with manager permissions
3. Click on `Set icon`
4. Remove connection from device
5. Choose image and submit, staying at spaces view

**Current:**

Nothing happens

**Expected:**

An snackbar, same as other cases, with info related to the lack of connection

Xiaomi Redmi Note 13
Android 15
`7679bfc4`

jesmrec:
### (3) [WONT FIX] - known problem

1. Click on Spaces on the bottom bar
2. Click on 3-dot-button over a space with manager permissions
3. Click on `Set icon`
4. Remove connection from device
5. Choose image and submit (here happens (2) )
6. Browse to other view like `Uploads` view
7. Recover connection

**Current**

Emoji is uploaded but not removed from upload list

**Expected**

Removed from uploads list

Does it have something to do with https://github.com/owncloud/android/pull/4705#issuecomment-3492313716?

Xiaomi Redmi Note 13
Android 15
`7679bfc4`

joragua:
Regarding **(2)**:

When there is no connection, all files stay in the `QUEUED` status until the connection is restored. Every transfer goes through the `QUEUED` status before `IN_PROGRESS` regardless of whether the connection is lost or not. Because of this technical restriction,  I think it's not possible to show a snackbar when there is no connection. The message can only be shown  when the transfer reaches the `FAILED` status.

Regarding **(3)**:

The problem is the same as in  https://github.com/owncloud/android/pull/4705#issuecomment-3492313716. If you navigate to another screen before the transfer has been successful, the `PATCH` operation will not be triggered, and as a result, the transfer won't be removed from the database. 

NOTE: If there is a `PATCH` operation, the transfer will be removed. If not, it will remain in the database and it will be shown in the uploads list.

jesmrec:
(2) is a won't fix, accepting the fact that an enqueued upload is still alive
(3) In the scope of https://github.com/owncloud/android/pull/4705#issuecomment-3492313716 will be marked as known problem

jesmrec:
About (1), it will be addressed in a separate issue since the problem comes from the way the icon is printed inside the card. Also, when setting an image to the space, the image is not totally visible. Will create an issue for that, and, this one is approved.

PR #4721, opened 2025-11-17
[FIX] Changes in release template because latest changes in the workflow

jesmrec:
Changes:

- Removed SBOM steps, since this is already automatic
- Change the channel to inform about the release
- BitRise changed for generic CI
- Remove "update doc" step (needs much more than an release update)

## Related Issues
App:

_____


joragua (approved):
LGTM! ✅

Issue #4742, opened 2025-12-18
[BUG] Edit space option available with no connection


joragua:
### Actual behaviour
`Edit space` option is available when the device has no connection. 

### Expected behaviour

`Edit space` option should not be available without connection. The bottom sheet dialog should display no options
### Steps to reproduce
1. Remove device connection
2. Open bottom sheet dialog in a user account with `Manager` role in that space or `Admin` / `Space Admin` platform role.

<img width="202" height="450" alt="Image" src="https://github.com/user-attachments/assets/e6b16ba5-3cf4-459c-a6b5-82646ca9a8ff" />

PR #4750, opened 2026-01-08
[FIX] Edit space option available with no connection 

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4742

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ ownCloud admins and users.
 
 ## Summary
 
+* Bugfix - Edit space option available when there is no connection: [#4742](https://github.com/owncloud/android/issues/4742)
 * Change - Migrate tests to the new kotlinx-coroutines-test API: [#4710](https://github.com/owncloud/android/issues/4710)
 * Change - Increase rating dialog delay: [#4744](https://github.com/owncloud/android/pull/4744)
 * Enhancement - Show members of a space: [#4612](https://github.com/owncloud/android/issues/4612)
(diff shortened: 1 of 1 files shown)


changelog/unreleased/4750
```diff
@@ -0,0 +1,7 @@
+Bugfix: Edit space option available when there is no connection
+
+A new condition has been added to prevent the edit space option
+from being shown when there is no available network connection.
```
jesmrec: I'd do like:

`A new condition has been added to hide the edit space option when no network connection is available.`

jesmrec (changes requested):

changelog/unreleased/4750
joragua: Changed! 👍🏻

jesmrec:
Two different scenarios here:

- Connection is already lost when `Spaces` is clicked -> works fine ✅ 
- Connection is up when `Spaces` is clicked and then, it is lost -> `Edit space` and `Disable space` are listed ❌ 

Steps:

1. Enter in oCIS account with spaces
2. Click on `Spaces` 
3. Remove device connection
4. Click on `...` of any space with permissions of editing

**Current:**
`Edit space` and `Disable space` are listed 

Xiaomi Redmi 13, Android 15
`809bf710`

joragua:
The second scenario should be fixed with the last commit @jesmrec

jesmrec:
> The second scenario should be fixed with the last commit @jesmrec

Fixed. I realised a side-effect: with the new request, when the `...` is clicked without connection, the FAB gets also hidden ✅ , but, does not get visible if the connection is recovered ❌ . I mean, everything inside the `Spaces` view, browsing through views makes everything perfect. 

Is it posible to check it when the view is refreshed? does it make sense? @joragua

joragua:
The view (in this case, the FAB) is refreshed when the three-dot menu button is opened, or when you navigate to another screen and then return to the spaces list.

jesmrec:
> The view (in this case, the FAB) is refreshed when the three-dot menu button is opened, or when you navigate to another screen and then return to the spaces list.

Let's get this as a good approach. 

Approved, moving forward!

PR #4751, opened 2026-01-09
[TECHNICAL] Workflow that builds an unsigned app with version and variant

jesmrec:
Reusable workflow to be called for any other workflow to get an apk artifact to be used in testing or any other purposes. Artifact is unsigned, since the keys and the keystore are responsibility on the calling workflow.

Inputs: build variant and commit/tag/branch

## Related Issues
App:

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ ownCloud admins and users.
 * Change - Increase rating dialog delay: [#4744](https://github.com/owncloud/android/pull/4744)
 * Enhancement - Show members of a space: [#4612](https://github.com/owncloud/android/issues/4612)
 * Enhancement - Set emoji as space image: [#4707](https://github.com/owncloud/android/issues/4707)
+* Enhancement - Workflow to build APK: [#4751](https://github.com/owncloud/android/pull/4751)
 
 ## Details
 
@@ -80,6 +81,13 @@ ownCloud admins and users.
 
    https://github.com/owncloud/android/issues/4707
    https://github.com/owncloud/android/pull/4708
+
+* Enhancement - Workflow to build APK: [#4751](https://github.com/owncloud/android/pull/4751)
+
+   A new workflow to build unsigned APKs has been added, using variant and version
+   as inputs and generating the artifact and commit hash as outputs.
+
+   https://github.com/owncloud/android/pull/4751
 
 # Changelog for ownCloud Android Client [4.7.0] (2025-11-17)
 


.github/workflows/buildapk.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Validate build variant
+        run: |
+          case "${{ inputs.build_variant }}" in
+            originalDebug|originalRelease|mdmDebug|mdmRelease|qaDebug|qaRelease) ;;
+            *) echo "Invalid build_variant: ${{ inputs.build_variant }}" && exit 1 ;;
+          esac
+
+      # Checkout the repo
+      - name: Checkout current repo
+        uses: actions/checkout@v4
```
joragua: I think the version we are using in the repo for the other workflows is `actions/checkout@v5`

.github/workflows/build-apk.yml
```diff
@@ -0,0 +1,140 @@
...
+          case "${{ inputs.build_variant }}" in
+            originalDebug|originalRelease|mdmDebug|mdmRelease|qaDebug|qaRelease) ;;
+            *) echo "Invalid build_variant: ${{ inputs.build_variant }}" && exit 1 ;;
+          esac
+
+      # Checkout the repo
+      - name: Checkout current repo
+        uses: actions/checkout@v4
+        with:
+          repository: owncloud/android
```
joragua, line 73: Is it necessary? This workflow is already located in the owncloud/android repository 🤔

.github/workflows/buildapk.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Get commit hash
+        id: commit
+        run: |
+          COMMIT_HASH=$(git rev-parse HEAD)
+          echo "Commit hash: $COMMIT_HASH"
+          echo "commit_hash=$COMMIT_HASH" >> $GITHUB_OUTPUT
+ 
+      # Set Java-JDK version
+      - name: Setup JDK
+        uses: actions/setup-java@v4
```
joragua: Is `v4` the latest version? I think we are already using `actions/setup-java@v5` in other workflows

changelog/unreleased/4751
```diff
@@ -0,0 +1,5 @@
+Enhancement: Workflow to build APK
+
+A workflow to build unsigned apks has been added. As inputs, variant and version. As outputs, artifact and commit hash.
```
joragua: I would rewrite the changelog (just a suggestion)

`A new workflow to build unsigned APKs has been added, using variant and version as inputs and generating the artifact and commit hash as outputs.`

.github/workflows/buildapk.yml
jesmrec: sure, i will change it to `v5`

.github/workflows/build-apk.yml
jesmrec: Yes, because as a reusable workflow it can be called from another repo. If the `repository` field is not there, the `checkout` will be applied to the caller repo and not to `owncloud/android`, breaking the flow.

.github/workflows/buildapk.yml
jesmrec: will check and change if there is a newer version

changelog/unreleased/4751
jesmrec: sounds good, will use your version.

jesmrec:
> NOTE: I would rename the workflow to build-apk.yml in order to follow the naming convention used for workflows in the repository.

will do, in order to keep consistency as well

.github/workflows/buildapk.yml
joragua: 👍🏻

.github/workflows/build-apk.yml
joragua: Perfect, understood 👍🏻

.github/workflows/buildapk.yml
joragua: 💯

Issue #4613, opened 2025-06-26
[FEATURE REQUEST] Add a member to a space


jesmrec:
AC:

- The option "Invite" will allow to add members to a space. To be placed just bellow the list of members
-  Only available for users with "Space manager" permission. 
- The "Invite" option will open a new view/dialog to perform the invitation. This is how it looks like in iOS in different steps, also the web is a valid reference. 

|Step 1|Step 2|Step 3|Step 4|
|:-:|:-:|:-:|:-:|
|<img width="333" alt="Image" src="https://github.com/user-attachments/assets/7ab3146d-b915-4ec6-9211-83f1a9337ccd" />|<img width="337" alt="Image" src="https://github.com/user-attachments/assets/3434a9e8-b19c-48e6-bb2f-7707f2838f05" />|<img width="338" alt="Image" src="https://github.com/user-attachments/assets/8ff6f5c5-2860-4379-889c-94c3c8184dc6" />|<img width="336" alt="Image" src="https://github.com/user-attachments/assets/2314b888-0de2-43f3-8b83-0487bf3742fe" />

**Step 1:** Shows the basic info of the space (icon, name, quota...), an info message (`Enter the user or group...`)
and a textbox in which the user will type the username.

**Step 2:**: As the user types the name the results are shown, refreshing after every character if the string is at least 3 characters long.

**Step 3:**: When the user chooses one user or group to share, the name it's shown as `share with` and the available list of permissions is shown underneath

**Step 4**: After choosing the permission level, the optional "Expiration Date" file picker is displayed, in case the share is removed after such date, and the `Invite` button is enabled. Share can not be created without a permission level. (this step could be merged into the step 3).

A `Cancel` button also available across the process.

Additionally, after clicking `Invite`, navigation to `Members` view with the new member included in the list.

If a target user is deleted before submitting the share, an error should be shown (as corner case)

Base request:

To search users/groups:

`https://...:9200/graph/v1.0/users?$search="John"&$orderby=displayName`
`https://...:9200/graph/v1.0/groups?$search="John"&$orderby=displayName`

it returns a list of results like:

```
{
    "value": [
        {
            "accountEnabled": true,
            "displayName": "John Smith",
            "id": "cbe0ff51-e1ef-4c7b-b74e-827bc3bcd7fa",
            "mail": "john@smith.com",
            "onPremisesSamAccountName": "jsmith",
            "surname": "jsmith",
            "userType": "Member"
        },
        ...
}
```

To get the available permissions:

[...]

PR #4754, opened 2026-01-13
[FEATURE REQUEST] Add a member to a space: Search user and show the list of matches

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4613

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.8/Spaces%20-%20Add%20member.md

Reports:

- [x] (1) Search results not refreshed https://github.com/owncloud/android/pull/4754#issuecomment-3778299732 [FIXED]
- [x] (2) `Unknown error` when special characters https://github.com/owncloud/android/pull/4754#issuecomment-3778309880 [FIXED]
- [x] (3) Message when no matches https://github.com/owncloud/android/pull/4754#issuecomment-3778392254 [FIXED]
- [x] (4) Spinner while searching is running https://github.com/owncloud/android/pull/4754#issuecomment-3779546722 [FIXED]
- [x] (5) Lag when typing fast https://github.com/owncloud/android/pull/4754#issuecomment-3783157085 [FIXED]
- [X] (6) Accessibility: no way to reach the `+` for adding members https://github.com/owncloud/android/pull/4754#issuecomment-3783304368 [WONT FIX]
- [X] (7) Clear info after new search https://github.com/owncloud/android/pull/4754#issuecomment-3789875850 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
@@ -71,6 +71,8 @@ class AddMemberFragment: Fragment() {
         collectLatestLifecycleFlow(spaceMembersViewModel.members) { uiState ->
             if (uiState.isLoading) {
                 binding.indeterminateProgressBar.visibility = View.VISIBLE
+                binding.emptyDataParent.root.visibility = View.GONE
+                binding.membersRecyclerView.visibility = View.GONE
             } else {
                 binding.indeterminateProgressBar.visibility = View.GONE
                 val listOfMembersFiltered = uiState.members.filter { member ->


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        collectLatestLifecycleFlow(spaceMembersViewModel.spaceMembers) { event ->
+            event?.let {
+                when (val uiResult = event.peekContent()) {
+                    is UIResult.Success -> {
+                        uiResult.data?.let {
+                            val listOfMembersFiltered = listOfMembers.filter { member -> !it.members.any {
+                                spaceMember -> spaceMember.id == "u:${member.id}" || spaceMember.id == "g:${member.id}" }
+                            }
```
jesmrec: I'd write this block in another way for better readability:

```
val listOfMembersFiltered = listOfMembers.filter { member ->
    !it.members.any { spaceMember ->
        spaceMember.id == "u:${member.id}" || spaceMember.id == "g:${member.id}"
    }
}
```

or 

```
val listOfMembersFiltered = listOfMembers.filter { member ->
    !it.members.any { spaceMember ->
        spaceMember.id == "u:${member.id}" || 
        spaceMember.id == "g:${member.id}"
    }
}
```

jesmrec: OTOH, is it correct to do the calculation about the filtered members in the fragment? shouldn't a `ViewModel` responsibility?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
```diff
@@ -0,0 +1,145 @@
...
+            requestFocus()
+            setOnQueryTextListener(object : androidx.appcompat.widget.SearchView.OnQueryTextListener {
+                override fun onQueryTextSubmit(query: String): Boolean = true
+
+                override fun onQueryTextChange(newText: String): Boolean {
+                    if (newText.length > 2) { spaceMembersViewModel.searchMembers(newText) } else { spaceMembersViewModel.clearSearch() }
+                    return true
+                }
+            })
+        }
```
jesmrec, line 102: Could be used a coroutine in the VW to be able to cancel ongoing requests replaced by others that contain the newest typed chars?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SearchMembersAdapter.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    class SearchMembersViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+        val binding = MemberItemBinding.bind(itemView)
+    }
+
+    companion object {
+        private const val USER_SURNAME = "User"
+        private const val GROUP_SURNAME = "Group"
+    }
+
```
jesmrec: extra blank line

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
```diff
@@ -59,8 +80,17 @@ class SpaceMembersActivity: FileActivity() {
             super.onOptionsItemSelected(item)
         }
 
+    override fun addMember(space: OCSpace) {
+        supportFragmentManager.transaction {
```
jesmrec: the `transaction` method is marked as Deprecated. Is that a known issue?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
```diff
@@ -73,7 +73,9 @@ class SpaceMembersAdapter: RecyclerView.Adapter<SpaceMembersAdapter.SpaceMembers
 
     fun setSpaceMembers(spaceMembers: SpaceMembers, roles: List<OCRole>) {
         this.rolesMap = roles.associate { it.id to it.displayName }
-        this.members = spaceMembers.members.sortedByDescending { member -> roles.indexOfFirst { it.id in member.roles } }
+        this.members = spaceMembers.members.sortedWith(compareByDescending<SpaceMember> {
+                member -> roles.indexOfFirst { it.id in member.roles } }.thenBy { member -> member.displayName }
+        )
```
jesmrec, line 78: It sorts by role and then by displayname, right? this is to avoid the side effect when browsing back to the list of members

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersDiffUtil.kt
```diff
@@ -0,0 +1,47 @@
...
+        val newItem = newList[newItemPosition]
+
+        return oldItem.id == newItem.id
+    }
+
+    override fun areContentsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+        val oldItem = oldList[oldItemPosition]
+        val newItem = newList[newItemPosition]
+
+        return !((oldItem.id != newItem.id) || (oldItem.displayName != newItem.displayName) || (oldItem.surname != newItem.surname))
```
jesmrec: Is there any case in which the ids are different and the others (`displayName` or `surname`) are the same?

OTOH: just curiosity, that is totally correct but... why the expression is written in double negation? `not ( not || not || not)` instead of `yes && yes && yes`  (Boole algebra).

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
```diff
@@ -25,21 +25,42 @@ import android.view.Menu
...
+                val totalQuota = quota.total
+                itemSize.text = when {
+                    usedQuota == null -> getString(R.string.drawer_unavailable_used_storage)
+                    totalQuota == 0L -> DisplayUtils.bytesToHumanReadable(usedQuota, baseContext, true)
+                    else -> getString(
+                        R.string.drawer_quota,
+                        DisplayUtils.bytesToHumanReadable(usedQuota, baseContext, true),
+                        DisplayUtils.bytesToHumanReadable(totalQuota, baseContext, true),
+                        quota.getRelative().toString())
+                }
```
jesmrec, line 62: Moved to the activity to be available everywhere inside, good refactor 💯

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersViewModel.kt
```diff
@@ -67,4 +88,17 @@ class SpaceMembersViewModel(
...
+        useCase = searchMembersUseCase,
+        useCaseParams = SearchMembersUseCase.Params(accountName = accountName, query = query),
+        showLoading = false,
+        requiresConnection = true
+    )
+
+    fun clearSearch() {
+        _members.value = Event(UIResult.Success(emptyList()))
+    }
+
```
jesmrec, line 115: Extra blank line

owncloudApp/src/main/res/layout/member_item.xml
```diff
@@ -70,7 +70,9 @@
                     android:layout_marginBottom="@dimen/standard_half_margin"
                     android:text="@string/placeholder_sentence"
                     android:textSize="13sp"
-                    android:textColor="@color/textColor"/>
+                    android:textColor="@color/textColor"
+                    android:ellipsize="middle"
+                    android:maxLines="1"/>
```
jesmrec, line 75: What is this for? why these changes?

owncloudApp/src/main/res/layout/add_member_fragment.xml
```diff
@@ -0,0 +1,51 @@
...
+            android:id="@+id/members_recycler_view"
+            android:layout_width="match_parent"
+            android:layout_height="0dp"
+            android:layout_weight="1"
+            android:scrollbars="vertical"
+            android:visibility="visible" />
+
+    </LinearLayout>
+
+</ScrollView>
```
jesmrec, line 68: I miss some accessibility inputs like `contentDescription`

owncloudApp/src/main/res/layout/members_activity.xml
```diff
@@ -26,6 +26,49 @@
...
+
+        <TextView
+            android:id="@+id/item_name"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginEnd="@dimen/standard_half_margin"
+            android:layout_marginStart="@dimen/standard_half_margin"
+            android:ellipsize="middle"
+            android:singleLine="true"
+            android:text="@string/placeholder_filename"
```
jesmrec, line 53: Why this placeholder here? i guess the real name is filled out and the placeholder is never displayed, but, is that the best name for the field? Could you ellaborate?

owncloudApp/src/main/res/layout/members_activity.xml
```diff
@@ -26,6 +26,49 @@
...
+            android:textSize="16sp"
+            app:layout_constraintStart_toEndOf="@id/item_icon"
+            app:layout_constraintTop_toTopOf="parent"/>
+
+        <TextView
+            android:id="@+id/item_size"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginStart="@dimen/standard_half_margin"
+            android:text="@string/placeholder_filesize"
```
jesmrec, line 64: same here

owncloudApp/src/main/res/layout/members_activity.xml
```diff
@@ -26,6 +26,49 @@
...
+            android:textColor="@color/list_item_lastmod_and_filesize_text"
+            android:textSize="14sp"
+            app:layout_constraintStart_toEndOf="@id/item_icon"
+            app:layout_constraintTop_toBottomOf="@+id/item_name"/>
```
jesmrec: Why  `+` in the `@+id/item_name` ? is that a typo?

owncloudData/src/main/java/com/owncloud/android/data/members/datasources/implementation/OCRemoteMembersDataSource.kt
```diff
@@ -0,0 +1,56 @@
...
+        private const val GROUP_SURNAME = "Group"
+
+        @VisibleForTesting
+        fun MemberResponse.toModel(isGroup: Boolean): OCMember =
+            OCMember(
+                id = id,
+                displayName = displayName,
+                surname = surname ?: if (isGroup) GROUP_SURNAME else USER_SURNAME
+            )
+
```
jesmrec: extra blank line

owncloudData/src/main/java/com/owncloud/android/data/members/repository/OCMembersRepository.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.owncloud.android.domain.members.model.OCMember
+
+class OCMembersRepository(
+    private val remoteMembersDataSource: RemoteMembersDataSource
+): MembersRepository {
+
+    override fun searchMembers(accountName: String, query: String): List<OCMember> =
+        remoteMembersDataSource.searchUsers(accountName, query) + remoteMembersDataSource.searchGroups(accountName, query)
+
+
```
jesmrec: extra blank lines

owncloudData/src/test/java/com/owncloud/android/data/members/OCRemoteMembersDataSourceTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        } returns searchGroupsResult
+
+        val groups = ocRemoteMembersDataSource.searchGroups(OC_ACCOUNT_NAME, query)
+        assertEquals(listOf(GROUP_MEMBER_RESPONSE).map { it.toModel(isGroup = true) }, groups)
+
+        verify(exactly = 1) {
+            clientManager.getMembersService(OC_ACCOUNT_NAME)
+            ocMembersService.searchGroups(query)
+        }
+
```
jesmrec: blank line

owncloudData/src/test/java/com/owncloud/android/data/members/datasources/implementation/OCRemoteMembersDataSourceTest.kt
```diff
@@ -0,0 +1,90 @@
...
+            ocMembersService.searchGroups(query)
+        } returns searchGroupsResult
+
+        val groups = ocRemoteMembersDataSource.searchGroups(OC_ACCOUNT_NAME, query)
+        assertEquals(listOf(GROUP_MEMBER_RESPONSE).map { it.toModel(isGroup = true) }, groups)
+
+        verify(exactly = 1) {
+            clientManager.getMembersService(OC_ACCOUNT_NAME)
+            ocMembersService.searchGroups(query)
+        }
```
jesmrec, line 68: Does it make sense to verify or assert that there is no users (just groups)?

owncloudData/src/test/java/com/owncloud/android/data/members/OCRemoteMembersDataSourceTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        val users = ocRemoteMembersDataSource.searchUsers(OC_ACCOUNT_NAME, query)
+        assertEquals(listOf(USER_MEMBER_RESPONSE).map { it.toModel(isGroup = false) }, users)
+
+        verify(exactly = 1) {
+            clientManager.getMembersService(OC_ACCOUNT_NAME)
+            ocMembersService.searchUsers(query)
+        }
+    }
+
```
jesmrec: same here: blank line + verify/assert no groups

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SearchMembersAdapter.kt
joragua: Removed!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersViewModel.kt
joragua: Removed 👍🏻

owncloudData/src/main/java/com/owncloud/android/data/members/datasources/implementation/OCRemoteMembersDataSource.kt
joragua: Removed!

owncloudData/src/main/java/com/owncloud/android/data/members/repository/OCMembersRepository.kt
joragua: Removed both blank lines 👍🏻

owncloudData/src/test/java/com/owncloud/android/data/members/OCRemoteMembersDataSourceTest.kt
joragua: Removed! 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
joragua: Yeah, that's right! I added this logic to avoid the "glitch" in the list of space members (the endpoint returns a different order on each request) 🙃

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
joragua: 🙌🏻

owncloudApp/src/main/res/layout/member_item.xml
joragua: This is used to ellipsize the member surname when it's too long, just like the display name

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
joragua: I’ve chosen the first option for better readability. Regarding the other comment, there are two different lists being compared: the space members (`spaceMembers`) and the list of all available members from the server (`listOfMembers`). Both lists are observed in the fragment, and their values are handled there. In the viewModel we only set the data to the corresponding flow variables. Is it clear now? 🤔

owncloudApp/src/main/res/layout/members_activity.xml
joragua: Yes, `+` is not necessary here

owncloudApp/src/main/res/layout/add_member_fragment.xml
joragua: The `SearchView` has its own accessibility properties, and Talkback detects them correctly (unless I'm missing something). Additionally, each item in the `RecyclerView` has its own `contentDescription` which is set in the `SearchMemberAdapter` class (not in the XML)

owncloudApp/src/main/res/layout/members_activity.xml
joragua: The placeholder is never displayed as you commented, but it's useful for the layout preview in Android Studio. That's why I added that string (I didn't find a better placeholder and I think it's not necessary to create a new one just for this) 🤔 The same applies to the comment below

owncloudApp/src/main/res/layout/members_activity.xml
joragua: ⬆️ ⬆️

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
joragua: Refactored! Now, `transaction` is used in a different way (not deprecated) 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersDiffUtil.kt
joragua: I'm not sure if the `id` can change. If it's always the same, that condition is not necessary but... 🤔 

Regarding boolean condition: I've changed to `yes && yes && yes` for better clarity. The double negation was there because other diff utils has this same expression.

owncloudData/src/test/java/com/owncloud/android/data/members/datasources/implementation/OCRemoteMembersDataSourceTest.kt
joragua: Hmm, I'm not sure. Since users and groups are represented by the same object (`OCMember`) and there a no differences in terms of properties, it doesn't matter if the list contains only users, only groups or both... What do you think? 🤔

owncloudData/src/test/java/com/owncloud/android/data/members/OCRemoteMembersDataSourceTest.kt
joragua: Removed blank line! About asserts: the same as the comment above...

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
joragua: I'm working on it, but to avoid blocking the CR, I will move it forward. It's possible that something related to this come up during QA, but we will need to evaluate it carefully...

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
jesmrec: that's ok

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
jesmrec: understood!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersDiffUtil.kt
jesmrec: 💯

owncloudApp/src/main/res/layout/member_item.xml
jesmrec: right, i didn't catch it. Cool!

owncloudApp/src/main/res/layout/members_activity.xml
jesmrec: 👍

owncloudData/src/test/java/com/owncloud/android/data/members/datasources/implementation/OCRemoteMembersDataSourceTest.kt
jesmrec: agreed, wfm

owncloudData/src/test/java/com/owncloud/android/data/members/OCRemoteMembersDataSourceTest.kt
jesmrec: 👍

owncloudApp/src/main/res/layout/members_activity.xml
jesmrec: 👍

owncloudApp/src/main/res/layout/add_member_fragment.xml
jesmrec: Let's take a look in Qa phase then  😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
jesmrec: Let's move this improvement to QA-side and review new code there

jesmrec:
### (1) [FIXED]

Typing characters does not refresh screen with results. App should be unfocused and focused to list:

https://github.com/user-attachments/assets/ea872aaa-bbe2-4a1f-903e-5de2851267ad

NOTE: not in every server... crazy

Galaxy Tab A4, Android 15

`c534e2ba4`

jesmrec:
### (2) [FIXED]

In search view, add some special characters like `?` or `$`

**Current:**

Snackbar `Unknown error` 

I realised the query parameter is `$search:  use$`

**Expected:**

No error, just no results in case there are no matches. Checking iOS, the query parameter is enclosed in quotation marks (if it helps):

`$search:  "User$"`

Galaxy Tab A4, Android 15

`c534e2ba4`

jesmrec:
### (3) [FIXED]

In case the typed chars don't match with any user, it'd be a good idea to show a message like `No matches` or `No user or group matches your search` (iOS approach) `No users or groups found` (web approach)

Galaxy Tab A4, Android 15

`c534e2ba4`

joragua:
I think that **(1)** and **(2)** are fixed now @jesmrec 🙌🏻

jesmrec:
### (4) (improvement) [FIXED]

Given a high number of results, the lag for showing the results is perceptible. Could it be posible to add a spinner, so the view does not seem to be frozen. Ideally (it might not be feasible), the spinner would replace the ✖️ while the operation is executed.

Galaxy Tab A4, Android 15

`ff47f47767`

jesmrec:
(2) fixed

jesmrec:
### (5) [FIXED]

Coming from https://github.com/owncloud/android/pull/4754#discussion_r2708335148

The idea should be prevent effects like this one:

https://github.com/user-attachments/assets/bbb1e8ff-6a1c-4981-97f3-c8fda55cae0f

user types quick and the results appear later. Final status is correct ✅ , but, if only last character is taken in account, lag could be reduced

`ff47f47767`

jesmrec:
### (6) [WONT FIX]

About keyboard navigation: the `+` for adding members in the members list view is not reachable.

EDIT: it can be reached but the focused `+` is not visible:


https://github.com/user-attachments/assets/0b8beedc-6046-4532-b6ef-9135f41fae76

Need to check if the focus color could be changed

`ff47f47767`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/AddMemberFragment.kt
```diff
@@ -102,6 +95,16 @@ class AddMemberFragment: Fragment() {
...
     }
 
+    private fun showOrHideEmptyView(hasMembers: Boolean) {
+        binding.membersRecyclerView.isVisible = hasMembers
+        binding.emptyDataParent.apply {
+            root.isVisible = !hasMembers && binding.searchBar.query.length > 2
+            listEmptyDatasetIcon.setImageResource(R.drawable.ic_share_generic_white)
+            listEmptyDatasetTitle.setText(R.string.members_search_failed)
+            listEmptyDatasetSubTitle.setText(R.string.members_search_empty)
+        }
```
jesmrec, line 115: inside the `apply` block you are setting the `root.isVisible` property, but then, you set the icon, title and subtitle, regardless the `isVisible`? Shouldn't be done only if `isVisible = true` ?

joragua: Right! Changed ✅

jesmrec: 💯

joragua:
**(3)** and **(4)** should be fixed 🤔

jesmrec:
(3) and (4) fixed

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersViewModel.kt
```diff
@@ -67,4 +97,27 @@ class SpaceMembersViewModel(
         requiresConnection = true
     )
 
+    fun searchMembers(query: String) {
+        searchJob?.cancel()
+        searchJob = viewModelScope.launch(coroutineDispatcherProvider.io) {
+            delay(500)
```
jesmrec: This line could be removed since the problem is not triggering request, but rejecting results if other requests were launched, right?

jesmrec:
### (7) [FIXED]

1. Type some chars that don't match with any user -> `No result found`
2. Replace any of the chars in order to get some results

**Current:**

spinner and `No results` together in the view

**Expected:**

if user is waiting for results (spinner), results from previous requests should be cleaned

`a0f13c63f2f`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersViewModel.kt
joragua: Yeah, without `delay(500)`, all requests are triggered but the result is correct. Removed!

jesmrec:
(5) and (7) fixed, good performance in searching 💯

jesmrec:
(1) fixed
(6) will not fix, since there are some kind of dependency on the customization layer. After trying different options, neither of them was satisfactory. The decision was not to block the work because of that, going to technical debt.

jesmrec:
Approved on my side, show must go on!! 🤘 🤘

Issue #4763, opened 2026-01-26
[BUG] Permission to list members


jesmrec:
Comes from https://github.com/owncloud/android/issues/4613

First assumption was that list of members was visible for any member, no matter the platform role or space permission. That's not true, so, the user must have the `libre.graph/driveItem/permissions/read` permission to watch the list.

PR #4765, opened 2026-02-04
[FIX] Permission to list members

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4763

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,9 @@ ownCloud admins and users.
    available).
 
    https://github.com/owncloud/android/issues/4612
+   https://github.com/owncloud/android/issues/4763
    https://github.com/owncloud/android/pull/4728
+   https://github.com/owncloud/android/pull/4765
 
 * Enhancement - Set emoji as space image: [#4707](https://github.com/owncloud/android/issues/4707)
 


owncloudDomain/src/main/java/com/owncloud/android/domain/spaces/usecases/FilterSpaceMenuOptionsUseCase.kt
```diff
@@ -45,7 +45,9 @@ class FilterSpaceMenuOptionsUseCase(
 
         val editImagePermission = hasSpacePermission(spacePermissionsResult, DRIVES_MANAGE_PERMISSION)
 
-        if (spacePermissionsResult is UseCaseResult.Success) {
+        val membersPermission = hasSpacePermission(spacePermissionsResult, DRIVES_READ_PERMISSION)
+
+        if (membersPermission) {
```
jesmrec, line 50: Just a question: before, you checked that the result is `UseCaseResult.Success`, but in the new version the `UseCaseResult.Success` is not checked anymore, just the correct permission. Is that correct?

joragua: `hasSpacePermission` first checks whether the result is `UseCaseResult.Success`, and then verifies if the permissions is included in the permissions list.

jesmrec: understood 👍

jesmrec:
Basic checks performed. 

Approved

Issue #4725, opened 2025-11-20
[FEATURE REQUEST] Remove a space member


jesmrec:
An space member can be removed by a space `manager` unless it is the last `manager` in the space. 

- Associated permission: `libre.graph/driveItem/permissions/delete`

A trashbin icon in the list of members together with a confirmation message (same like sharing and links) could be a good option. Also, an `unshare` option inside the `edit` menu would be an **extra**.

The list will be updated without the removed user.

### REQUEST
`DELETE /graph/v1beta1/drives/<drive-id>/root/permissions/u:<user-id>` for users (`g:<group-id>` for groups )

No body required.

### TASKS

- [x] Research (if needed)
- [x] Create branch feature/remove_a_space_member
- [x] Development tasks
  - [x] Show the trashbin icon only for user with proper permissions
  - [x] Create and show the warning dialog when trashbin is clicked
  - [x] Implement logic to remove a space member
  - [x] Implement unit tests (if needed)
- [x] Code review and apply changes requested
- [x] Design test plan
- [x] QA
- [x] Merge branch feature/remove_a_space_member into master

PR #4768, opened 2026-02-09
[FEATURE REQUEST] Remove a space member

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4725

_____

## QA

Reports: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.8/Spaces%20-%20Remove%20member.md

- [x] (1) Adding new manager does not add trashbin to the existing one https://github.com/owncloud/android/pull/4768#issuecomment-3883823744 [FIXED]
- [x] (2) Improve trashbin accessibility message https://github.com/owncloud/android/pull/4768#issuecomment-3884197634  [FIXED]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ ownCloud admins and users.
 * Enhancement - Show members of a space: [#4612](https://github.com/owncloud/android/issues/4612)
 * Enhancement - Add a member to a space: [#4613](https://github.com/owncloud/android/issues/4613)
 * Enhancement - Set emoji as space image: [#4707](https://github.com/owncloud/android/issues/4707)
+* Enhancement - Remove a space member: [#4725](https://github.com/owncloud/android/issues/4725)
 * Enhancement - Workflow to build APK: [#4751](https://github.com/owncloud/android/pull/4751)
 * Enhancement - Workflow to check Conventional Commits: [#4759](https://github.com/owncloud/android/pull/4759)
 
@@ -104,6 +105,14 @@ ownCloud admins and users.
 
    https://github.com/owncloud/android/issues/4707
    https://github.com/owncloud/android/pull/4708
+
+* Enhancement - Remove a space member: [#4725](https://github.com/owncloud/android/issues/4725)
+
+   A new option to remove members from a space has been added. It will be only
+   visible for users with proper permissions.
+
+   https://github.com/owncloud/android/issues/4725
+   https://github.com/owncloud/android/pull/4768
 
 * Enhancement - Workflow to build APK: [#4751](https://github.com/owncloud/android/pull/4751)
 


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
```diff
@@ -98,15 +105,18 @@ class SpaceMembersFragment : Fragment() {
                     is UIResult.Success -> {
                         uiResult.data?.let {
                             if (roles.isNotEmpty()) {
-                                spaceMembersAdapter.setSpaceMembers(it, roles)
+                                val numberOfManagers = it.members.count { spaceMember->
```
jesmrec: blank missing `spaceMember ->`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
```diff
@@ -153,6 +178,20 @@ class SpaceMembersFragment : Fragment() {
...
+        outState.putBoolean(CAN_REMOVE_MEMBERS, canRemoveMembers)
+    }
+
+    override fun onRemoveMember(spaceMember: SpaceMember) {
+        AlertDialog.Builder(requireContext())
+            .setMessage(getString(R.string.members_remove_dialog_message, spaceMember.displayName))
+            .setPositiveButton(getString(R.string.common_yes)) { _, _ -> spaceMembersViewModel.removeMember(spaceMember.id) }
+            .setNegativeButton(getString(R.string.common_no)) { dialog, _ -> dialog.dismiss() }
+            .show()
+            .avoidScreenshotsIfNeeded()
```
jesmrec, line 192: Why this `avoidScreenshotsIfNeeded()` ?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
```diff
@@ -0,0 +1,70 @@
...
+                appendEncodedPath(memberId)
+            }
+
+            val deleteMethod = DeleteMethod(URL(uriBuilder.build().toString()))
+
+            val status = client.executeHttpMethod(deleteMethod)
+
+            val response = deleteMethod.getResponseBodyAsString()
+
+            if (status == HttpConstants.HTTP_NO_CONTENT || status == HttpConstants.HTTP_INTERNAL_SERVER_ERROR) {
```
jesmrec: `500` intended to be OK when the user has already been removed. But, it could cause problems when the  `500` really means a server problem. In case there is no way to distinguish, we should handle the 500 as error.

My first thought -> issue in the server.  This kind of error should be `404` and not `500`. Any input @jvillafanez ?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
```diff
@@ -72,20 +87,26 @@ class SpaceMembersAdapter: RecyclerView.Adapter<SpaceMembersAdapter.SpaceMembers
...
     override fun getItemCount(): Int = members.size
 
-    fun setSpaceMembers(spaceMembers: SpaceMembers, roles: List<OCRole>) {
+    fun setSpaceMembers(spaceMembers: SpaceMembers, roles: List<OCRole>, canRemoveMembers: Boolean, numberOfManagers: Int) {
+        this.canRemoveMembers = canRemoveMembers
         this.rolesMap = roles.associate { it.id to it.displayName }
         val listOfMembersFiltered = spaceMembers.members.sortedWith(compareByDescending<SpaceMember> {
                 member -> roles.indexOfFirst { it.id in member.roles } }.thenBy { member -> member.displayName })
-        val diffCallback = SpaceMembersDiffUtil(this.members, listOfMembersFiltered)
+        val diffCallback = SpaceMembersDiffUtil(this.members, listOfMembersFiltered, this.numberOfManagers, numberOfManagers)
```
jesmrec, line 96: More parameters to the DiffUtils method will increase the complexity. Is there any way to execute the Diff only if there are differences?

I mean, in dirty code:

```
if ( this.members != listOfMembersFiltered || this.numberOfManagers != numberOfManagers) SpaceMembersDiffUtil(...)
```

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
joragua: Added!

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
joragua: The error code returned when trying to add a member that is already part of the space is `409`. Could it be the same in this case? 🤔  Anyway, `500` is not a successful code  -> Removed `HTTP_INTERNAL_SERVER_ERROR` from condition

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
joragua: It's used to prevent screenshots when the variant is not `DEBUG`. This method is applied to all warning/confirmation dialogs in the app...

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
joragua: I've removed `numberOfManagers` from `SpaceMembersDiffUtil`, and now it is only used when the number of managers is the same. If the number differs from before, the entire list is updated using `notifyItemRangeChanged()`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
jesmrec: If it's the rule to do that in the dialogs, OK. But i don't get the point because there is no sensible information there. Any other reason under it?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
jesmrec: I'd go for  `404 Not Found` but, it's the same handling. As you commented, `500` is never a success code.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
jesmrec: 💯  like it, more performant

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersFragment.kt
joragua: Nop, there is no other reason to include that line and it's true that there is no sensible information there 🤔

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
jvillafanez: I don't see that we're returning a 500 error "willingly". Something must happened in the server that caused the error, so there should be logs.

jesmrec:
moving to QA, since https://github.com/owncloud/android/pull/4768#discussion_r2789326879 is out of CR scope. In case changes are required, will be handled as a QA report.

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
jesmrec: Well, the scenario is also reproduced in `demo.owncloud.com`, not specific from a instance. Beyond the server logs (cumbersome for us), the point is if you agree that such scenario should not return a 5xx, but a 4xx. Then, we will handle the issue somehow (probably, an ocis issue)

jesmrec:
### (1) [FIXED]

All steps to be done in the mobile app:

1. Current user is the only manager in a space -> no trashbin ✅ 
2. Add another user as a manager -> new manager has trashbin ✅  but the existing manager does not ❌ 
3. Add another user as a manager -> all managers have now trashbin ✅ 

Galaxy Tab A8, Android 15
`3d0e7993`

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/members/RemoveRemoteMemberOperation.kt
jvillafanez: It seems something weird is going on... it should return a 4xx (likely 404) error

jesmrec: Thanks!! will move on.

joragua:
**(1)**: This bug is reproducible after applying the commit that comes from this comment on the CR: https://github.com/owncloud/android/pull/4768#discussion_r2789375072

I've checked it, and the best solution here is to revert all the changes and removed that commit. It's more efficient to use `SpaceMembersDiffUtil` instead of mixing both methods (avoiding `notifyItemRangeChanged()`). Using the notify method also causes all animations to be lost, which feels so strange for the UX...

I've removed the commit and it should be fixed now 😄

jesmrec:
### (2) [FIXED]

About accessibility:

the trashbin icon is spelt as `Remove member`. Is there any chance to add the name of the user to delete? so, it'd say `Remove member Alice` (for example)

`6863e3f7f`

joragua:
Regarding **(2)**: With the last commit, `Talkback` should now read `"Remove member <displayName> button"`

jesmrec:
(1) and (2) fixed

jesmrec:
Approved on my side  🚀

PR #4771, opened 2026-02-11
[TECHNICAL] Fix identation of xml string

mmattel:
This PR fixes the identation of the `</resources>` keyword that was unintentionally indended.
Though this identation does no harm, we should keep easy human readability.

Bash command used:
```
cd owncloudApp/src/main/res
grep -rl '  </resources>' | xargs sed -i 's#  </resources>#</resources>#g'
```


mmattel:
Note that a translation sync (manually triggered or the regular nightly sync job) currently reverts this change back to two  trailing white spaces...

PR #4776, opened 2026-02-16
[QA] Add content provider to handle settings in QA variant

jesmrec:
## Summary

Adds a `ContentProvider` to the qa variant that allows to handle the shared preferences with testing purposes.

The `ContentProvider` overrides the `call` function that receives three parameters:

`method` : name of the method that will execute the action. In this case, only one method `set_boolean` since it's intended (for the moment) for boolean preferences
`arg`: the name of the preference
`extras`: value of the preference (in this case, `true` or `false`)

## How it works

1. Build the branch with qa variant to get an apk.
2. Plug on a device/emu and install the apk
3. With an `adb` command, the content provider is called with the proper arguments

F. ex: 

to enable the `show_disabled_spaces` preference:

```
adb shell content call --uri content://com.owncloud.android.test.preferences --method set_boolean --arg show_disabled_spaces --extra value:b:true
```

to disable the `show_hidden_files` preference:

```
adb shell content call --uri content://com.owncloud.android.test.preferences --method set_boolean --arg show_hidden_files --extra value:b:false
```

## Security

In order to assure that this implementation is only available in the qa variant, the following checks have been performed:

**1. Check that `Manifest` is OK**

Firstly, a `Manifest` for every variant is generated:

```
./gradlew clean
./gradlew :owncloudApp:processQaReleaseMainManifest
./gradlew :owncloudApp:processOriginalReleaseMainManifest
```

Then, inspect them:

```
grep -n "TestPreferencesProvider" owncloudApp/build/intermediates/merged_manifest/qaRelease/processQaReleaseMainManifest/AndroidManifest.xml
```

Result: 81: android:name="com.owncloud.android.test.TestPreferencesProvider"

That means, the qa manifest contains the `TestPreferencesProvider` class.

Let's check in the original:

```
grep -n "TestPreferencesProvider" owncloudApp/build/intermediates/merged_manifest/originalRelease/processOriginalReleaseMainManifest/AndroidManifest.xml 
```

That returns no results, so the original manifest does not contain the `TestPreferencesProvider` class

**2. Check if the compilings contain the TestPreferencesProvider class**

Execute the following commands

`./gradlew clean` -> clean up directory
`./gradlew :owncloudApp:compileOriginalReleaseKotlin` -> compile the originalRelease variant
`find owncloudApp/build/tmp/kotlin-classes/originalRelease -iname '*TestPreferencesProvider*'` -> tries to find the `TestPreferencesProvider` inside of it. It shouldn't find anything.

`./gradlew clean` -> clean up again
`./gradlew :owncloudApp:compileQaReleaseKotlin` -> compile the qaRelease variant
`find owncloudApp/build/tmp/kotlin-classes/qaRelease  -iname '*TestPreferencesProvider*'` -> it should return `owncloudApp/build/tmp/kotlin-classes/qaRelease/com/owncloud/android/test/TestPreferencesProvider.class`

**3. Check the presence of TestPreferencesProvider in the artifacts**

Build a signed apk with both `original` and `qa` variant, and execute over both of them:

```
$ANDROID_HOME/cmdline-tools/latest/bin/apkanalyzer dex packages owncloud.apk | grep -i testpreferencesprovider
```

(set the right apk name in the command fot every case)

That command list the code files inside the apk. The `TestPreferencesProvider` class should be listed only in the apk that was  built with the qa variant.

Those checks show that the `originalXXX` variants will never include the `TestPreferencesProvider` and the qa `Manifest`. 

(open to other ideas)

App:

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ ownCloud admins and users.
 * Enhancement - Remove a space member: [#4725](https://github.com/owncloud/android/issues/4725)
 * Enhancement - Workflow to build APK: [#4751](https://github.com/owncloud/android/pull/4751)
 * Enhancement - Workflow to check Conventional Commits: [#4759](https://github.com/owncloud/android/pull/4759)
+* Enhancement - QA Content Provider: [#4776](https://github.com/owncloud/android/pull/4776)
 
 ## Details
 
@@ -138,6 +139,13 @@ ownCloud admins and users.
    Conventional Commits convention.
 
    https://github.com/owncloud/android/pull/4759
+
+* Enhancement - QA Content Provider: [#4776](https://github.com/owncloud/android/pull/4776)
+
+   A new content provider has been added to the QA variant in order to work with
+   the sharing preferences with testing purposes.
+
+   https://github.com/owncloud/android/pull/4776
 
 # Changelog for ownCloud Android Client [4.7.0] (2025-11-17)
 


owncloudApp/src/qa/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
...
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <provider
+            android:name="com.owncloud.android.test.TestPreferencesProvider"
+            android:authorities="com.owncloud.android.test.preferences"
+            android:exported="true" />
+    </application>
+
+</manifest>
```
joragua: No newline at end of file

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+    override fun onCreate(): Boolean = true
+
+    override fun call(
+        method: String,
+        arg: String?,
+        extras: Bundle?
+    ): Bundle {
+
+        val prefs = context!!.getSharedPreferences(prefsName, Context.MODE_PRIVATE)
+        if (method == "set_boolean") {
```
joragua: Would it be possible to create a variable for `set_boolean` (similar to `prefsName`)?

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+                prefs.edit {
+                    putBoolean(arg!!, extras!!.getBoolean("value"))
+                }
+            }
```
joragua, line 57: What would happen if `context`, `arg` or `extras` were null? Would it be possible to use a safe calls (`?.`) for these arguments? 🤔

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+    ): Cursor? = null
+
+    override fun insert(uri: Uri, values: ContentValues?): Uri? = null
+
+    override fun delete(uri: Uri, selection: String?, selectionArgs: Array<out String>?): Int = 0
+
+    override fun update(
+        uri: Uri, values: ContentValues?,
+        selection: String?, selectionArgs: Array<out String>?
+    ): Int = 0
```
joragua: I would write this method on a single line for clarity (just my personal opinion) 🙌🏻

owncloudApp/src/qa/AndroidManifest.xml
jesmrec: 👍  Added

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
jesmrec: sure, no problem 👍

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
jesmrec: yep, since the name of the method is static, i'll add it as a companion string, and same for the prefs_name. What do you think?

joragua: Nice! 💯 Is `prefsName` also a static value? We could move both values to the companion object

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
jesmrec: it shouldn't, or the call to the provider will fail. Anyway, let's do some control over its nullability

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
jesmrec: Both moved!

owncloudApp/src/qa/java/com/owncloud/android/test/TestPreferencesProvider.kt
joragua: I would replace `getContext()?.` by `context?.`, as it was before the CR

jesmrec: Done!!

PR #4784, opened 2026-02-26
[FIX] Refresh expiration date on space members list

joragua:
When a member is removed from a space with an expiration date, the member directly below (if it does not have expiration date) inherits the deleted member expiration date. The adapter is reusing the interface and displaying an incorrect field on the screen.

## Steps to reproduce

1. Invite a new member to a space with `Can view` or `Can edit` permission and expiration date
2. Upgrade the new member to `Can Manage`
3. Remove the member from the space

## Related Issues
App: -

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ ownCloud admins and users.
    https://github.com/owncloud/android/pull/4728
    https://github.com/owncloud/android/pull/4765
    https://github.com/owncloud/android/pull/4779
+   https://github.com/owncloud/android/pull/4784
 
 * Enhancement - Add a member to a space: [#4613](https://github.com/owncloud/android/issues/4613)
 


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersAdapter.kt
```diff
@@ -90,10 +90,11 @@ class SpaceMembersAdapter(
...
 
-            member.expirationDateTime?.let {
-                expirationCalendarIcon.visibility = View.VISIBLE
-                expirationDate.visibility = View.VISIBLE
-                expirationDate.text = DisplayUtils.displayDateToHumanReadable(it)
+            val hasExpirationDate = member.expirationDateTime != null
+            expirationCalendarIcon.isVisible = hasExpirationDate
+            expirationDate.isVisible = hasExpirationDate
+            if (hasExpirationDate) {
+                expirationDate.text = DisplayUtils.displayDateToHumanReadable(member.expirationDateTime)
```
jesmrec, line 97: You are checking nullability of `member.expirationDateTime` to perform some actions. Would it be posible to take advantage of a `?.let` clause somehow?

joragua: `?.let` is more appropriate when you don't need to do anything if the value is `null`, but in this case we have to change the visibility when `expirationDate == null`

It's possible to use that clause but it would look something like this:

```
expirationDateTime?.let {
        expirationDateCalendarIcon.visibility = View.VISIBLE
        expirationDate.visibility = View.VISIBLE
        ...
} ?: run {
        expirationDateCalendarIcon.visibility = View.GONE
        expirationDate.visibility = View.GONE
        ...
}
```

IMO, I prefer to use `.isVisible` with a boolean value, as it feels cleaner and more readable. What do you think? 🤔

jesmrec: makes sense, let's keep your approach.

jesmrec:
Approved

Issue #4760, opened 2026-01-17
[BUG] Sort is not alphabetically correct


esaramago:
### Actual behaviour
- when sorting by name, it is treating A and Á as different letters. That's problematic for languages like Portuguese.

### Expected behaviour
- "Áaa" should appear before "Zzz"
 
### Steps to reproduce
1. Create a folder with the name "Áaa"
2. Create another with the name "Zzz"
3. "Áaa" is being sorted after "Zzz"

Can this problem be reproduced with the official owncloud server?
It doesn't happen on the web version.

### Environment data
Android version: 15

Device model: Redmi N13P

Stock or customized system:

ownCloud app version: 4.7

ownCloud server version:

#### Web server error log
```
Not applicable
```

#### ownCloud log (data/owncloud.log)
```
Not applicable
```

![Image](https://github.com/user-attachments/assets/5c31d475-499e-4b42-85f2-b283c8f4bed9)

jesmrec:
@esaramago thanks for reporting, the bug is reproducible. We'll check for a fix in our side or from the community. Good contribution!

mykh-hailo:
@jesmrec Can I work on this issue please?

joragua:
Hi @mykh-hailo! Of course, go ahead! 🍻 Feel free to open a new PR with your changes. If you have some questions, let us know - we will be happy to help you! As a reminder, please make sure to review the [CONTRIBUTING.md](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md) file before getting started 😄

mykh-hailo:
Hi @joragua , I made my PR for this issue. Can you please have a look and leave some comments if any?
Thank you.

joragua:
Thanks for your contribution, @mykh-hailo!  I will check your pull request as soon as possible and I will ping you when it's ready. Stay tuned! 🙌🏻

joragua:
The fix for this issue will be available in the next release (`v4.8.0`)! Thanks a lot for your engagement @esaramago @mykh-hailo! 🚀

PR #4785, opened 2026-02-26
[FIX] Sort is not alphabetically correct

mykh-hailo:
## Related Issues
App: #4760 

### DIFF
--- a/changelog/unreleased/4785
+++ b/changelog/unreleased/4785
@@ -1,0 +1,6 @@
+Bugfix: Sort is not alphabetically correct
+
+Locale-aware collation has been applied to file sorting by name to handle accented characters correctly.
+
+https://github.com/owncloud/android/issues/4760
+https://github.com/owncloud/android/pull/4785


changelog/unreleased/4760
joragua: 1. Incorrect name of the file. The name must be the PR id, in this case: `4785`
2. No newline at end of file

changelog/unreleased/4760
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix alphabetical ordering when sorting files by name
```
joragua: I would re-write the title like this: `Bugfix: Sort is not alphabetically correct`

changelog/unreleased/4760
joragua, line 1: Regarding changelog file: 

1. Incorrect name of the file. The name must be the PR id: `4785` 😄 
2. No newline at end of file
3. We usually use `chore:` prefix for changelog commit

changelog/unreleased/4760
```diff
@@ -0,0 +1,6 @@
...
+
+Use locale-aware collation for file sorting by name to handle accented characters correctly.
```
joragua: I would use the present perfect passive tense for the changelog description (following [TEMPLATE.md](https://github.com/owncloud/android/blob/master/changelog/TEMPLATE.md))

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
```diff
@@ -34,12 +36,20 @@ class SortFilesUtils {
         }
 
     private fun sortByName(listOfFiles: List<OCFile>, ascending: Boolean): List<OCFile> {
-        val newListOfFiles =
-            if (ascending) listOfFiles.sortedBy { it.fileName.lowercase() }
-            else listOfFiles.sortedByDescending { it.fileName.lowercase() }
+        val collator = Collator.getInstance(Locale("en", "US")).apply {
+            strength = Collator.PRIMARY   // Ignore accents and case
+        }
```
joragua: Would it be possible to use `Locale.getDefault()`? There are many languages with different accents and additional characters that should be taken into account 🤔

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
```diff
@@ -34,12 +36,20 @@ class SortFilesUtils {
         }
 
     private fun sortByName(listOfFiles: List<OCFile>, ascending: Boolean): List<OCFile> {
-        val newListOfFiles =
-            if (ascending) listOfFiles.sortedBy { it.fileName.lowercase() }
-            else listOfFiles.sortedByDescending { it.fileName.lowercase() }
+        val collator = Collator.getInstance(Locale("en", "US")).apply {
+            strength = Collator.PRIMARY   // Ignore accents and case
```
joragua, line 40: Is it necessary to add `strength = Collator.PRIMARY`?. It ignores accents and cases...

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
```diff
@@ -34,12 +36,20 @@ class SortFilesUtils {
...
+        }
+
+        val comparator = compareByDescending<OCFile> { it.isFolder }
+            .thenComparator { a, b ->
+                if (ascending) {
+                    collator.compare(a.fileName, b.fileName)
+                } else {
+                    collator.compare(b.fileName, a.fileName)
+                }
+            }
```
joragua: I would not re-write the code. I'd keep the existing implementation but using the collator for comparisons.

I think it's possible to use the `collator` variable in this block without making a lot of changes. Instead of creating a new comparator, `listOfFiles.sortedWith(compareBy(collator)) { ... }` could be used. Does that make sense? 🤔 

```
val newListOfFiles =
            if (ascending) listOfFiles.sortedBy { it.file.fileName.lowercase() }
            else listOfFiles.sortedByDescending { it.file.fileName.lowercase() }
```

owncloudDomain/src/main/java/com/owncloud/android/domain/files/usecases/SortFilesWithSyncInfoUseCase.kt
joragua, line 1: The comments above also apply for this file 😄

mykh-hailo:
@joragua Thank you for your comments.
I updated the code from your comments can you review it again please?

Thank you

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
mykh-hailo: I think it is necessary because our target is to ignore accent when compare two strings.
Also in your codebase, we compared lowercases(); which means we should ignore cases too.

joragua: Okay, let's check in QA if it works correctly 👍🏻

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
```diff
@@ -34,12 +36,14 @@ class SortFilesUtils {
...
     private fun sortByName(listOfFiles: List<OCFile>, ascending: Boolean): List<OCFile> {
-        val newListOfFiles =
-            if (ascending) listOfFiles.sortedBy { it.fileName.lowercase() }
-            else listOfFiles.sortedByDescending { it.fileName.lowercase() }
+        val collator = Collator.getInstance(Locale.getDefault()).apply {
+            strength = Collator.PRIMARY   // Ignore accents and case
+        }
 
-        // Show first the folders when sorting by name
-        return newListOfFiles.sortedByDescending { it.isFolder }
```
joragua, line 42: Why did you remove this comment?  I think it can be useful to clarify the code (just my opinion) The same for the comment in `SortFilesWithSyncInfoUseCase` 😄

owncloudApp/src/main/java/com/owncloud/android/utils/SortFilesUtils.kt
```diff
@@ -34,12 +36,14 @@ class SortFilesUtils {
...
+        val nameComparator =
+            if (ascending) compareBy<OCFile, String>(collator) { it.fileName }
+            else compareByDescending<OCFile, String>(collator) { it.fileName }
```
joragua: Is it necessary to add `<OCFile, String>` in both conditions? It should be enough with only one declaration 🤔  What do you think?

owncloudDomain/src/main/java/com/owncloud/android/domain/files/usecases/SortFilesWithSyncInfoUseCase.kt
joragua, line 1: Same comments for this file 😄

mykh-hailo:
@joragua Thank you for your comments.
I updated the branch for your comments.
Can you review it please?

jesmrec:
I did several checks and tests with every list of files (list of personal, list of space, copy/move file picker, external share and automatic uploads target folder) including the current fix. Everything was OK ✅ , so this is approved on my side and ready to go!! 🚀 

Thanks a lot for the contribution @mykh-hailo!! your fix will be available in the upcoming 4.8.0 version.

Issue #4718, opened 2025-11-12
[BUG] Option "Copy here" in spaces list after rotating device


jesmrec:
### Steps to reproduce
1.  In oCIS account, select any item from the list of files and `Copy` -> list of spaces displayed
2. Rotate the device

### Actual behaviour

Option `Copy here` in the bottom side of the screen. Such option stays, no matter how many times the device rotates

### Expected behaviour

Option `Copy here` not allowed in the list of spaces. Just allowed to copy inside of an space
 
Xiaomi Redmi Note 13
Android 15
`v4.6.2`

Thanush19:
Hey , i can work on this  @jesmrec @joragua

jesmrec:
@Thanush19 go ahead!

jesmrec:
@Thanush19 it's a good idea to check our [CONTRIBUTING](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md) file beforehand, so that the work you do is aligned with ours. Don't be afraid of asking as many questions as you wish, we are happy to help 😄

For sure, that also applies to https://github.com/owncloud/android/issues/4737 and https://github.com/owncloud/android/issues/4673, where you are also willing to contribute 😉 

Eager to get your contributions!

mykh-hailo:
@joragua I made PR for this issue.
Can you check it and leave comments if any please?

joragua:
Hi @mykh-hailo, thanks for your contribution! I will check your PR as soon as possible! 🍻

PR #4467, opened 2024-09-16
[BUG] List filtering not working after rotating device

AwaisKhan128:
## Related Issues
App:

- [x] Add changelog files for the fixed issues in folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
- [x] Add feature to Release Notes in `ReleaseNotesViewModel.kt` creating a new `ReleaseNote()` with String resources (if required)

_____

Issue #4441, opened 2024-07-18
[BUG] List filtering not working after rotating device


jesmrec:
### Steps to reproduce

1. Open list of files  in portrait orientation
2. Rotate device to landscape
3. Click on lens icon
4. Type something to filter the list

### Actual behaviour

Filter does not work. When rotating the device again, filter does not work either. Required to kill the app an reopening to make it come to life again.

### Expected behaviour
Filter works after  (and before) rotating device

Pixel 2, Android 11
v4.3.0

AwaisKhan128:
Is there someone already working on this bug? or may I start contributing into it?

jesmrec:
@AwaisKhan128 go ahead!!

Take a look to our [CONTRIBUTING](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md) file, where you'll find guidelines about our development process (branch names, commit names...), so that your contribution will be aligned with the product itself.

Let us know any question or help you need!

jesmrec:
@AwaisKhan128 did you take a look? something we can help you?

jesmrec:
> Actually, I did try. First of all, gradle build failed due to its not recognizing the kotlin syntax in the gradle script.
Well, definitely, I will start over again on the weekend because, due to full-time commitment, I am unable to spare some time. @jesmrec

No worries, take your time! just for you to know that we are happy to help.

AwaisKhan128:
@jesmrec Could you tell me which server url used to continue in app?

jesmrec:
> @jesmrec Could you tell me which server url used to continue in app?

You can use the production one: https://ocis.ocis.production.owncloud.works/

or the rolling one (latest master): https://ocis.ocis.master.owncloud.works/

with credentials: katherine/gemini

AwaisKhan128:
> > @jesmrec Could you tell me which server url used to continue in app?
> 
> You can use the production one: https://ocis.ocis.production.owncloud.works/
> 
> or the rolling one (latest master): https://ocis.ocis.master.owncloud.works/
> 
> with credentials: katherine/gemini

Thanks. I do have a look on it

AwaisKhan128:
Issue fixed. Soon creating a PR.

PR #4786, opened 2026-02-27
[FIX] Option "Copy here" in spaces list after rotating device

mykh-hailo:
## Related Issues
App: #4718 

### DIFF
--- a/changelog/unreleased/4786
+++ b/changelog/unreleased/4786
@@ -1,0 +1,6 @@
+Bugfix: Option "Copy here" in spaces list after rotating device
+
+The "Copy here" button has been prevented from being displayed after device rotation in the spaces list.
+
+https://github.com/owncloud/android/issues/4178
+https://github.com/owncloud/android/pull/4786


mykh-hailo:
@joragua Can you check this PR please?

owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -350,6 +351,27 @@ open class FolderPickerActivity : FileActivity(),
...
+            // MainFileListFragment is displayed - show button and update per permissions
+            mainFileListFragment?.getCurrentFile()?.let { currentFolder ->
+                updateButtonsVisibilityAccordingToPermissions(currentFolder)
+            } ?: run {
+                binding.folderPickerBtnChoose.isVisible = true
+                binding.folderPickerNoPermissionsMessage.isVisible = false
+            }
+        }
+    }
+
```
joragua: I think there might be a better way to fix this bug instead of creating new method just to change the visibility of the button. You can take a look to `files_folder_picker.xml` file and play with visibility of the `folder_picker_btn_choose` button.

When the device is rotated the visibility does not change because it is set to `visible` by default. What if we only set it to `visible` when necessary (already done in this [line](https://github.com/owncloud/android/blob/master/owncloudApp/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt#L269))? 🤔

changelog/unreleased/4786
```diff
@@ -0,0 +1,6 @@
+Bugfix: Copy here button visibility updates correctly after rotation in folder picker
```
joragua: I would re-write the title like this:

```suggestion
Bugfix: Option "Copy here" in spaces list after rotating device
```

changelog/unreleased/4786
```diff
@@ -0,0 +1,6 @@
...
+
+The "Copy here" button visibility has been fixed to update correctly after device rotation.
```
joragua: I would also change the changelog description:
```suggestion
The "Copy here" button has been prevented from being displayed after device rotation in the spaces list.
```

mykh-hailo:
Thank you for your comment @joragua 
I will fix for your comments asap.
Thank you.

mykh-hailo:
@joragua I implemented much easier solution for this issue.
Please check and let me know if you have any comments.
Thank you.

owncloudApp/src/main/res/layout/files_folder_picker.xml
```diff
@@ -67,6 +67,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_weight="1"
+            android:visibility="gone"
```
joragua: I would add `tools:visibility="visible"` just for the layout preview. It makes it easier for developers to visualize the UI components while developing. Do you agree?

mykh-hailo: That makes sense.

mykh-hailo:
@joragua I made a quick fix on your comment.
I'd appreciate it if you check this again.
Thank you.

jesmrec:
@mykh-hailo Unfortunately the fix it's no valid. It fixes the problem with the `Copy` but breaks the `Move`.

Check the following steps out:

1. In the list of files, select any item to `Move` -> List of files displayed (not spaces list, since it's not allowed to move items through spaces), with `Move here` and `Cancel` options available.
2. Rotate the device

`Move here` is gone (it shouldn't) and only `Cancel` is available.  

It's verified that the current `master` branch shows the correct behaviour for the `Move`, therefore the defect comes from this PR.

Could you take a look? anything you need, don't hesitate to ask us!

joragua:
After some testing, I've noticed that his bug (https://github.com/owncloud/android/issues/4441) is also reproducible in the folder picker. It’s likely that both issues could be fixed with a single solution. But we will need to check it in QA... 🤔 

Here's the other PR (https://github.com/owncloud/android/pull/4467) related to this problem, if it helps @mykh-hailo 🙌🏻

mykh-hailo:
@joragua as far as I checked the code again based on Jesmerc's comment, I think my first approach to call a function to change visibility of that function onResume would be a solution.
What do you think?

jesmrec:
Then, we have two different approaches:

1) Modifying the `Manifest` file adding a property like `android:configChanges="orientation|screenSize"` , suggested by @joragua

2) Initial solution suggested by @mykh-hailo

First one intends to fix two problems in one ✅ , but, `Manifest` stuff is delicated and will need extra testing, but it is the cleanest solution IMO (if nothing is broken, for sure).

Second one seems to be totally safe ✅ , but more code to achieve the same target.

@mykh-hailo @joragua do you have any extra input? pros and cons of every solution? let's make it constructive 😄

jesmrec:
OTOH, just as a comment, [@mykh-hailo](https://github.com/mykh-hailo) we use to write branch names with underscores (check [GUIDELINES](https://github.com/owncloud/android/blob/master/CONTRIBUTING.md)), so, the repo will be tidy and uniform around ;) . Take it in account for future PRs, thanks!!

joragua:
This is my opinion: 

| Approaches | Pros | Cons |
|---|---|---|
| `android:configChanges="orientation\|screenSize"`  | Could fix two different bugs at the same time related to the folder picker. Cleaner solution in terms of logic, fewer lines of code, and better readability. | Requires more testing to ensure it doesn’t break the folder picker since `Manifest` is delicated as mentioned |
| @mykh-hailo solution | Could fix the bug related to the `"Copy here"` option. | Another fix is still needed for the filter in the folder picker (the same as the other approach? redundant code?). Results in more code and more lines. |

mykh-hailo:
@joragua as far as I checked the code the filtering issue is caused because my branch was not up-to-date.

joragua:
Hmm... 🤔  I don't think so, since I can reproduce the filter bug on the `master` branch. Can you confirm this @jesmrec?

jesmrec:
The problem about filtering that @joragua reported is reproducible in `master` , and also visible in the stable version. That means, it's not a matter of your branch @mykh-hailo.

That also means, if you'd wish you can fix it together with the initial bug (both problems with one fix would be cool), but, as a regression problem, it could be addressed separately. Up 2 you ;)

mykh-hailo:
@jesmrec Let me fix the both issues together.

mykh-hailo:
@jesmrec Can you check my updates again?
one fix for 2 issues.

jesmrec:
i'll check it

jesmrec:
@mykh-hailo fix works fine!!

Just the latest action and we will merge the PR: in the changelog entry, please enter a line break at the end:

Screenshot 2026-03-10 at 11 00 03

the red circle indicates that the line break is missing.

And this will be ready for merging, after rebasing with `master`

jesmrec:
Thanks for applying the changes @mykh-hailo!! everything is correct now, so it has my approval. Merging the PR, fixes will be available in 4.8.0

Issue #4625, opened 2025-07-03
[BUG] Spaces' root folder in landscape has no room for the list of files


jesmrec:
This is the root folder of a custom space in portrait and landscape orientations:

| Portrait | Landscape |
|:-:|:-:|
|![Image](https://github.com/user-attachments/assets/e3b3134d-42d4-46d9-a964-97a5f6f33665)|![Image](https://github.com/user-attachments/assets/8ea6b1b5-eadb-4a93-b315-39ebe7ec2e83)|

Portrait is totally correct, but, in landscape, there is no room for the list of files because the sticky header takes the most part of the screen height.

Possible solutions:

- Not showing the header in landscape
- Reducing the height of the header (saving some room for the description, just in case)
- In landscape, make the header scrollable
- other...

Can this problem be reproduced with the official owncloud server?
(url: https://ocis.ocis.master.owncloud.works, user: einstein, password: relativity)

### Environment data
Android version:

Device model: 

Stock or customized system:

ownCloud app version:

ownCloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### ownCloud log (data/owncloud.log)
```
Insert your ownCloud log here
```

mykh-hailo:
Hi @joragua, @jesmrec 
I have idea about this issue.
On landscape screen, instead of show space information and file list in vertical, we can show them in horizontal.
Here is my mockup for this.
Can you confirm it please? If you like my idea, I will implement my idea to your proejct.

<img width="1192" height="542" alt="Image" src="https://github.com/user-attachments/assets/0cc5449b-3553-4746-b5ed-f18cae751003" />

jesmrec:
It could be a solution, but you have to mind that the space could have a subtitle and i don't know if it'd be visible in that way. I'd better go for a height reduction or the scrollable thing... that's my opinion.

mykh-hailo:
Thank you for your opinion.
Can I work on this issue with that scrollable thing?

jesmrec:
Sure @mykh-hailo !

mykh-hailo:
@jesmrec @joragua 
Can you check my PR #4790 for this issue please?

joragua:
Hi @mykh-hailo! Thanks for your contribution! 💯 I will check the PR as soon as possible

PR #4790, opened 2026-03-03
[FIX] Spaces' root folder in landscape has no room for the list of files

mykh-hailo:
## Related Issues
App: #4625 

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/files/filelist/MainFileListFragment.kt
@@ -330,16 +330,23 @@ class MainFileListFragment : Fragment(),
 
     override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
-        updateSpaceHeaderSize()
+        updateConfigDependentSizes()
     }
 
-    private fun updateSpaceHeaderSize() {
+    private fun updateConfigDependentSizes() {
         val cardSize = resources.getDimensionPixelSize(R.dimen.spaces_header_card_height)
         binding.spaceHeader.spaceHeaderCard.layoutParams.apply {
             width = cardSize
             height = cardSize
         }
         binding.spaceHeader.spaceHeaderCard.requestLayout()
+
+        val iconSize = resources.getDimensionPixelSize(R.dimen.empty_icon_size)
+        binding.emptyDataParent.listEmptyDatasetIcon.layoutParams.apply {
+            width = iconSize
+            height = iconSize
+        }
+        binding.emptyDataParent.listEmptyDatasetIcon.requestLayout()
     }
 
     override fun onResume() {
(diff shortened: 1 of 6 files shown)


mykh-hailo:
@joragua about the branch name, I am sorry for not using underscore.
I will keep that mind for further contributions.

Can you check it please?

joragua (changes requested):
Somme comments! 😄 @mykh-hailo 

1. Missing changelog file 

2. I would go for a different approach. I'd reduce the height of the space card. Since `android:configChanges="orientation|screenSize"` is set in the manifest for `FileDisplayActivity`, the activity is not re-created on orientation change. I'd use `onConfigurationChanged()` instead and update the space header dimensions there.

mykh-hailo:
@joragua @jesmrec  I have one new feature among this issue.
If the space subtitle is empty, the space title should be vertically centered since there is too much empty space below it.
As it's not complicated issue, I can make a fix within this PR too.
image
We can update it to...

image

jesmrec:
> @joragua @jesmrec I have one new feature among this issue.
If the space subtitle is empty, the space title should be vertically centered since there is too much empty space below it.
As it's not complicated issue, I can make a fix within this PR too.

ftm, i'd not go for this approach.

joragua:
As @jesmrec mentioned, I wouldn't  go for that approach at this point

changelog/unreleased/4790
```diff
@@ -0,0 +1,6 @@
+Bugfix: Space header size after rotating device
```
joragua: Just a tiny suggestion 😄 

```suggestion
Bugfix: Reduce space header size after rotating device
```

joragua: @mykh-hailo Be careful with the typo in the calens title. Your version:

`Bugfix: Reduce pace header size after rotating device`

mykh-hailo: Sorry for that.
I will take a deep care of that.

jesmrec:
Problem is fixed, now the space title's room is reduced and the list of files better accesible. I noticed a posible improvement. Take a look:

Screenshot 2026-03-10 at 17 34 14

The placeholder `No files in here` , that before was not visible because of the wide title, now is visible but incomplete. Do you think the size could be adapted as well?

mykh-hailo:
@jesmrec To resolve the issue, should I reduce the icon size or place the icon and title horizontally?

jesmrec:
> @jesmrec To resolve the issue, should I reduce the icon size or place the icon and title horizontally?

I'd say, reducing the icon side. Setting the text aside the icon sounds weird 😅

mykh-hailo:
@jesmrec I have reduces the empty icon size on landscape. I'd appreciate it if you review and share your opinion.
Thank you.

jesmrec:
Approved on my side! 🚀 

Thanks for your patience and engagement @mykh-hailo!! The last step is rebasing the branch against `master` to be ready for merging.

jesmrec:
Merged! thanks for your awesome contribution @mykh-hailo !!!

Issue #4753, opened 2026-01-12
[FEATURE REQUEST] Create a link over a space


jesmrec:
Add a new link is only allowed for `Space managers`, same as `Members`.

iOS example:

<img width="344" height="672" alt="Image" src="https://github.com/user-attachments/assets/6c099ebd-5b96-4b17-a47b-2c1c351b4130" />

- Header with item info
- Form
   - Name (optional): max 255 chars
   - Permissions: `Can View` `Can Edit` `Secret File Drop` (hardcoded?)
   - Password: mandatory if enforced (same as item links, capability-based). Option to copy it to the clipboard.
   - Expiration date (optional)

Request to create the link:

`POST https://<url>/graph/v1beta1/drives/<drive-id>/items/<item-id>/createLink`

with the following payload:

```
{
    "displayName": "Test",
    "expirationDateTime": "2026-01-19T13:14:09.021Z",
    "password": "T/Ec90Vo5/!n##1S%x",
    "type": "edit"
}
```

type has three different values:  `view`, `edit`, `createOnly`

Permission is always mandatory, so, if it is not checked, the `Create` button is disabled.

Response body like:

```
{
    "createdDateTime": "2026-01-14T08:40:49.439847853Z",
    "expirationDateTime": "2026-01-21T08:40:47.921Z",
    "hasPassword": true,
    "id": "DDugZscaZAnhDvr",
    "link": {
        "@libre.graph.displayName": "Name",
        "@libre.graph.quickLink": false,
        "preventsDownload": false,
        "type": "view",
        "webUrl": "https://<url>/s/uwNlynuHnkvMtKA"
    }
}
```

After submitting creation, view is closed and the list of links refreshed with the new link

### TASKS

- [ ] Research (if needed)
- [ ] Create branch feature/feature_name
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
  - [ ] Implement unit tests (if needed)
- [ ] Code review and apply changes requested
- [ ] Design test plan
- [ ] QA
- [ ] Merge branch feature/feature_name into master

PR #4794, opened 2026-03-05
[FEATURE REQUEST] Create a link over a space 

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4753

**EXTRA:** Some content descriptions have been refactored (removed redundant `button`) in order to avoid repetitions in `Talkback`. These strings are related to the `Cancel` and `Create` buttons, displayed in some dialogs:

`Cancel` -> Set password dialog, create space dialog and set icon dialog
`Create` -> Create space dialog

_____

## QA

Test plan execution: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.8/Spaces%20-%20Add%20link.md

Reports:

- [x] (1) Problem creating a second link https://github.com/owncloud/android/pull/4794#issuecomment-4161201482 [FIXED]
- [x] (2) Hide soft keyboard after typing link name https://github.com/owncloud/android/pull/4794#issuecomment-4161779992 [FIXED]
- [x] (3) Empty password means no password https://github.com/owncloud/android/pull/4794#issuecomment-4162119107 [FIXED]
- [x] (4) Undercut labels in tablet https://github.com/owncloud/android/pull/4794#issuecomment-4167958881 [FIXED]


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -0,0 +1,298 @@
...
+                it.selectedPermission?.let { selectedPermission ->
+                    binding.optionsLayout.isVisible = true
+                    binding.passwordLayout.apply {
+                        passwordValue.isVisible = hasPassword
+                        setPasswordButton.isVisible = !hasPassword && isPasswordEnforced
+                        removePasswordButton.isVisible = hasPassword && isPasswordEnforced
+                        setPasswordSwitch.isVisible = !isPasswordEnforced
+                        setPasswordSwitch.isChecked = hasPassword
+                    }
+                    binding.createPublicLinkButton.isEnabled = isPasswordEnforced && hasPassword || !isPasswordEnforced
```
jesmrec: What would happen if `!isPasswordEnforced` and permission is not selected yet? OTOH, shouldn't parentheses be necessary?. Not the same:

`binding.createPublicLinkButton.isEnabled = (isPasswordEnforced && hasPassword) || !isPasswordEnforced`

`binding.createPublicLinkButton.isEnabled = isPasswordEnforced && (hasPassword || !isPasswordEnforced)`

i guess the first one is the correct, the one you wanted... and probably the one that kotlin will take. Just in case.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -0,0 +1,298 @@
...
+
+                    binding.createPublicLinkButton.setOnClickListener {
```
jesmrec: Question: since the "set listeners" (this and the following ones) are inside of the  `collectLatestLifecycleFlow` , are they registered in every state change?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -0,0 +1,298 @@
...
+            secretFileDropPublicLinkRadioButton.isChecked = false
+            selectedRadioButton.isChecked = true
+        }
+        val selectedPermission = selectedRadioButton.tag as OCLinkType
+        checkPasswordEnforced(selectedPermission)
+        spaceLinksViewModel.onPermissionSelected(selectedPermission)
+    }
+
+    private fun checkPasswordEnforced(selectedPermission: OCLinkType) {
+        isPasswordEnforced = when (selectedPermission) {
```
jesmrec: this kind of decision, aren't a `ViewModel` responsibility, instead of the fragment?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -0,0 +1,298 @@
...
+import java.util.Locale
+import java.util.TimeZone
+
+class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordListener {
+    private var _binding: AddPublicLinkFragmentBinding? = null
+    private val binding get() = _binding!!
+
+    private val spaceLinksViewModel: SpaceLinksViewModel by activityViewModel {
+        parametersOf(
+            requireArguments().getString(ARG_ACCOUNT_NAME),
```
jesmrec: I noticed `requireArguments().getString(ARG_ACCOUNT_NAME)` to  be called several times through the class. Couldn't be refactored to be read only once or similar?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -0,0 +1,298 @@
...
+        spaceLinksViewModel.onPasswordSelected(password)
+    }
+
+    private fun selectRadioButton(selectedRadioButton: RadioButton) {
+        binding.publicLinkPermissions.apply {
+            canViewPublicLinkRadioButton.isChecked = false
+            canEditPublicLinkRadioButton.isChecked = false
+            secretFileDropPublicLinkRadioButton.isChecked = false
+            selectedRadioButton.isChecked = true
+        }
```
jesmrec, line 204: About this block: you set all the permissions to `false` and then, the selected one to `true`.  Why that? is it required to reset all the buttons before setting a new value?

EDIT: after checking the correspondent layout, i noticed you use three individual `RadioButton` . Would it be better to use a `RadioGroup` that will make all options exclusive?  https://developer.android.com/reference/android/widget/RadioGroup

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        requireArguments().getString(ARG_PASSWORD)?.let {
+            binding.passwordValue.setText(it)
+        }
+
+        capabilityViewModel.capabilities.observe(viewLifecycleOwner) { event->
```
jesmrec: space missing: `event ->`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
```diff
@@ -0,0 +1,267 @@
...
+                        minUppercaseChars = passwordPolicy.minUppercaseCharacters,
+                        minSpecialChars = passwordPolicy.minSpecialCharacters,
+                    )
+                )
+            }
+        }
+
+        binding.copyPasswordButton.setOnClickListener {
+            val clipboard = requireActivity().getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+            val clip = ClipData.newPlainText("Public link", binding.passwordValue.text.toString())
```
jesmrec: the static "Public link" : should be moved to companion?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
```diff
@@ -0,0 +1,85 @@
...
+import kotlinx.coroutines.flow.update
+
+class SpaceLinksViewModel(
+    private val addLinkUseCase: AddLinkUseCase,
+    private val accountName: String,
+    private val space: OCSpace,
+    private val coroutineDispatcherProvider: CoroutinesDispatcherProvider
+): ViewModel() {
+
+    private val _addPublicLinkUIState = MutableStateFlow<AddPublicLinkUIState?>(null)
```
jesmrec, line 42: here, you initialize `_addPublicLinkUIState` to `(null)`.  Just below, in the `init` block, it's initialized again.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
```diff
@@ -100,6 +101,16 @@ class SpaceMembersActivity: FileActivity(), SpaceMembersFragment.SpaceMemberFrag
         }
     }
 
+    override fun addPublicLink(space: OCSpace) {
+        val addPublicLinkFragment = AddPublicLinkFragment.newInstance(account.name, space)
+        val transaction = supportFragmentManager.beginTransaction()
+        transaction.apply {
+            replace(R.id.members_fragment_container, addPublicLinkFragment, TAG_ADD_PUBLIC_LINK_FRAGMENT)
+            addToBackStack(null)
```
jesmrec, line 109: Why is the parameter `null`?

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
```diff
@@ -0,0 +1,157 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <ScrollView
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toTopOf="@id/create_public_link_button">
```
jesmrec, line 31: I noticed in other ConstraintLayout, you added `constraintEnd_toEndOf` . Is it missing here or not required?

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
```diff
@@ -0,0 +1,157 @@
...
+                    android:gravity="center_vertical"
+                    android:text="@string/global_name"
+                    android:textAllCaps="true"
+                    android:textColor="@color/white"
+                    android:textStyle="bold"
+                    android:accessibilityHeading="true"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent"/>
+
+            </androidx.constraintlayout.widget.ConstraintLayout>
```
jesmrec: This is a `ConstraintLayout` with only one item inside. Is that correct? `ConstraintLayout` sets the UI elements inside, in positions depending on others inside the same layout, if there is only one, it may not be required.

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
```diff
@@ -0,0 +1,157 @@
...
+                    android:gravity="center_vertical"
+                    android:text="@string/members_permissions"
+                    android:textAllCaps="true"
+                    android:textColor="@color/white"
+                    android:textStyle="bold"
+                    android:accessibilityHeading="true"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent"/>
+
+            </androidx.constraintlayout.widget.ConstraintLayout>
```
jesmrec: Same as above

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
```diff
@@ -0,0 +1,157 @@
...
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:focusable="true"
+                    android:background="@color/actionbar_start_color">
+
+                    <TextView
+                        android:id="@+id/options_title"
+                        android:layout_width="wrap_content"
+                        android:layout_height="50dp"
+                        android:layout_centerVertical="true"
```
jesmrec: i think this is not useful in a `ConstraintLayout`

owncloudApp/src/main/res/layout/public_link_permissions.xml
```diff
@@ -0,0 +1,211 @@
...
+    </LinearLayout>
+
+    <View
+        android:layout_width="match_parent"
+        android:layout_height="1dp"
+        android:layout_marginTop="@dimen/standard_quarter_margin"
+        android:background="@android:color/darker_gray"
+        android:alpha="0.5"/>
+
+</LinearLayout>
```
jesmrec, line 211: Check the comment above https://github.com/owncloud/android/pull/4794/changes#r2994418254

owncloudApp/src/main/res/layout/set_password_dialog.xml
```diff
@@ -0,0 +1,267 @@
...
+                android:layout_marginEnd="@dimen/standard_half_margin"
+                android:layout_gravity="center_horizontal"
+                android:orientation="horizontal"
+                android:visibility="gone"
+                tools:visibility="visible">
+
+                <TextView
+                    android:id="@+id/password_policy_min_characters_icon"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"/>
```
jesmrec: Why is this a `TextView` if the content is an icon? same for all below ⬇️

owncloudApp/src/main/res/layout/set_password_dialog.xml
```diff
@@ -0,0 +1,267 @@
...
+                    android:id="@+id/password_policy_min_digits_icon"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"/>
+
+                <TextView
+                    android:id="@+id/password_policy_min_digits_text"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"/>
+
+            </LinearLayout>
```
jesmrec: Seems to be repetition there (one LinearLayout for every policy, just changing ids) . Is there any way to refactor, for example, using `includes` somehow ?  or the different ids in every Layout makes it not posible?

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/LinksService.kt
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.lib.resources.links.services
+
+import com.owncloud.android.lib.common.operations.RemoteOperationResult
+import com.owncloud.android.lib.resources.Service
+
+interface LinksService: Service {
+    fun addLink(spaceId: String, displayName: String, type: String, expirationDate: String?, password: String?): RemoteOperationResult<Unit>
```
jesmrec, line 27: Question: could the `type` be `OCLinkType` instead of `String`?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: That code is wrapped in `selectedPermission?.let{ }` block, so the logic doesn't apply when the permission is not selected. Anyway, I'll add parentheses in the condition to make it clearer. The final version:

```
binding.createPublicLinkButton.isEnabled = (isPasswordEnforced && hasPassword) || !isPasswordEnforced
```

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
joragua: Done! ✅

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
joragua: Removed! 👍🏻

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
joragua: In this case, it's not necessary because we already have constraints for top, bottom and start which should be enough. In other cases, we want a constraint for the end to align the UI element at the end of the screen, but that does not apply here. 😄

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
joragua: We have `addToBackStack` method with `null` value in the whole app. The parameter is an optional name for the back stack state. In order to have consistency, I did not add a name as parameter.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
joragua: There are two different values here:
1) `MutableStateFlow` value, that is set to `null` in that line (no objects to observe at first)
2) `AddPublicLinkUIState` value that is initialized in the `init` block (object with `null` properties)

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
joragua: It should be a string from `strings.xml` file because it must be translated depending on the device language. Also, the string should be `Password` not `Public link` (my bad) since the password is copied to the clipboard (not the link) Changed! 👍🏻

owncloudApp/src/main/res/layout/set_password_dialog.xml
joragua: Yeah, they should be `ImageView` elements... Thanks! Changed 😄

owncloudApp/src/main/res/layout/set_password_dialog.xml
joragua: Refactored! Now there is only one layout for password policy requirements, and different `includes` refer to that layout to avoid duplicated code 😄

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/LinksService.kt
joragua: Since `Services` classes only support primitive types, the value is converted to a `String` before executing the remote operation. Anyway, I've moved the casting, and it's done in the datasource and not in the view model.

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
joragua: Yeah, right! `ConstraintLayout` removed and `TextView` used instead 👍🏻

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
joragua: Check comment above ⬆️

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: Refactored! Now, there is one variable (`accountName`) that is used in the fragment in order to avoid duplicated code ✅

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: Yeah, they are registered on every state change, but they depend on the `addPubicLinkUIState` value. I've re-checked the class and moved the listener for creating a new link out of the collect method. The other listeners cannot be move out because they depend on that properties for displaying the correct UI.

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: I think this is not possible in this case since we have a complex layout with `RadioButton` elements. To apply `RadioGroup` behavior, the buttons would need to be direct children That's why there is a method for selecting the clicked button/layout.

owncloudApp/src/main/res/layout/public_link_permissions.xml
joragua: Check comment about `RadioGroup`

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: Yeees! Moved `checkPasswordEnforced` method from the fragment to the viewModel

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: better readable, cool! 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: perfect!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SetPasswordDialogFragment.kt
jesmrec: Makes sense!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/members/SpaceMembersActivity.kt
jesmrec: 👍

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
jesmrec: good to know, thanks!

owncloudApp/src/main/res/layout/set_password_dialog.xml
jesmrec: Right!

owncloudApp/src/main/res/layout/set_password_dialog.xml
jesmrec: good refactor! 💯

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/LinksService.kt
jesmrec: ok! works for me

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: 💯

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
jesmrec: 👍

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
jesmrec: 👍

owncloudApp/src/main/res/layout/add_public_link_fragment.xml
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: Let's go for it, then

owncloudApp/src/main/res/layout/public_link_permissions.xml
jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
jesmrec: You're right, similar variable names but not the same!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: 👍

jesmrec:
### (1) [FIXED]
 
1. Open `Members` on any space with manager permissions and create a link with any setup -> link created and listed ✅ 
2. Click on `+` to create another link

**Current:**

snackbar with `Public link created correctly` is displayed

**Expected:**

Form to create a new link is displayed

`f26cf140e`

joragua:
**(1)** should be fixed with the commit https://github.com/owncloud/android/pull/4794/commits/7598a6d44245eee1a7922b2e0045fb7959e0dc8b

jesmrec:
### (2) (ux improvement) [FIXED]

Check the following steps out:

1. Click on `+` to create a new link
2. Click on the `Name` section to type a link name -> soft keyboard raises up ✅ 
3. After typing the name, select any permission level

**Current:**

Soft keyboard is still on screen and hides the password and expiration date fields. It could be confusing for the users

**Suggestion:**

When the user clicks on any permission level, hide the soft keyboard (if feasible), so, the new fields will be totally visible

`7598a6d4`

joragua:
**(2)** should be done with the commit https://github.com/owncloud/android/pull/4794/commits/2145d57ff1f03200b6afc4c0fe64d8440827583e

jesmrec:
### (3) [FIXED]

As the password policy, we assume that the password policy could have 0 chars (side-effect), if the capability comes with:

```
                "password_policy": {
                    "max_characters": 72,
                    "min_characters": 0,
                    "min_digits": 0,
                    "min_lowercase_characters": 0,
                    "min_special_characters": 0,
                    "min_uppercase_characters": 0
                },
```

the point is, if the password is set with 0 chars, the platform behaviour is like no password was set.

In our case:

1. Set the server with [mandatory password for writeable links](https://doc.owncloud.com/ocis/next/deployment/services/s-list/frontend.html#password-enforcement-for-writeable-public-links) 
2. Click on `+` to create a new link
3. Select `Can View` (password no mandatory)
4. Switch password on
5. Let password field empty
6. Click on `Set` 

**Current**

Password switch is on, with `****` visible

**Expected**

Since password is empty, there is no password, and link is accessible without security, switch should be off with no  `****` visible

`7598a6d4`

joragua:
**(3)** should be fixed with the commit https://github.com/owncloud/android/pull/4794/commits/57c9a46bd40c4364334da3937d88f79f5424e967

Click on `Set` button with an empty password -> `password = null`. If password is not enforced `setPasswordSwitch` -> off. Nothing to change when password is enforced.

jesmrec:
### (4) [FIXED]

To reproduce with a tablet. The labels `Set` and `Remove` of the password layout are a bit undercut:

Screenshot 2026-04-01 at 08 53 36
Screenshot 2026-04-01 at 08 53 12

Samsung Galaxy Tab A8
`57c9a46bd`

joragua:
**(4)** fixed with the commit https://github.com/owncloud/android/pull/4794/commits/3e42b4ba57631fe71dc963afd571ff30b96d30bb

jesmrec:
Approved on my side. By the way, the following issue was created on server side:

https://github.com/owncloud/ocis/issues/12179

Because an incorrect error handling.

🚀

PR #4795, opened 2026-03-05
[TECHNICAL] Update Transifex config

mmattel:
This PR updates the Transifex config to use the new `app` subdomain instead of `www`.
Additionally the `type` and `reasource_name` is now set explicitly.

The reason is, that `translation-sync` gets back translations to the repo correctly, but does not push new ones to TX.


mmattel:
Hold on, I think I got the real root case (this is so :grimacing: )

mmattel:
Closing, the real cause is now identified, a new PR will be created fixing this.

PR #4759, opened 2026-01-15
[TECHNICAL] Workflow to validate commit names

jesmrec:
Workflow to validate Conventional Commits through all the commits inside a PR. 

Validation rules:

- Commit name starts for one of this prefixes: `docs|fix|feat|chore|refactor|test|style|build|ci`
- Then, a colon `:` 
- Then, a blank space
- Then , the description starting for a lowercase character
- Description must have at least 10 chars, to avoid non-descriptive cases

If the commit name starts by `Merge` or `[tx]` will be ignored, as special cases

Every commit in the PR must match the rule, otherwise the check will not pass till it fits.

NOTE: scoped to the latest 100 commits. There is no way to know in advance (before the `checkout`) how many commits conform the current PR. That means, if a PR contains more that 100 commits (very extreme), only the latest 100 will be checked out.

## Related Issues
App:

- [X] Add changelog files for the fixed issues in folder changelog/unreleased. More info [here](https://github.com/owncloud/android/tree/master/changelog#create-changelog-items)
- [ ] Add feature to Release Notes in `ReleaseNotesViewModel.kt` creating a new `ReleaseNote()` with String resources (if required)

_____

PR #4807, opened 2026-03-16
[FEATURE REQUEST] Add action to check source strings

jesmrec:
Two actions were addressed here:

- In the `strings.xml` file, some tiny changes to normalize the file against the validation script: remove some useless line breaks, and replace `&quot;` for the escaped `\"`

- The action will run a script to check if every string in `strings.xml` contains any forbidden character (non-escaped). The list of forbidden characters is set inside the script (i expect the list to be very static). The util `xmlstarlet` will be used to parse the xml file and provide every entry one by one by using the [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html) operation

For the moment, the check will not be mandatory. Let's give a try, and, if it works fine, we could become mandatory

## Related Issues
App:

_____

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ ownCloud admins and users.
 * Enhancement - Copy permanent link of a space: [#4758](https://github.com/owncloud/android/issues/4758)
 * Enhancement - Workflow to check Conventional Commits: [#4759](https://github.com/owncloud/android/pull/4759)
 * Enhancement - QA Content Provider: [#4776](https://github.com/owncloud/android/pull/4776)
+* Enhancement - Workflow to validate source strings: [#4807](https://github.com/owncloud/android/pull/4807)
 
 ## Details
 
@@ -220,6 +221,13 @@ ownCloud admins and users.
    the sharing preferences with testing purposes.
 
    https://github.com/owncloud/android/pull/4776
+
+* Enhancement - Workflow to validate source strings: [#4807](https://github.com/owncloud/android/pull/4807)
+
+   A new workflow has been added to validate the source strings and prevent errors
+   in translation CI system from the first step
+
+   https://github.com/owncloud/android/pull/4807
 
 # Changelog for ownCloud Android Client [4.7.0] (2025-11-17)
 


changelog/unreleased/4807
```diff
@@ -0,0 +1,5 @@
+Enhancement: GitHub action to validate source strings
+
+A new GitHub action has been added to validate the source strings and prevent errors in translation ci system from the first step
+
+https://github.com/owncloud/android/pull/4807
```
joragua: IMO, I'd write the calens like this (juts to align all entries for workflows) 😄 

A similar example: [#4759](https://github.com/owncloud/android/blob/master/changelog/unreleased/4759)
```suggestion
Enhancement: Workflow to validate source strings

A new workflow has been added to validate the source strings and prevent errors in translation CI system from the first step

https://github.com/owncloud/android/pull/4807
```

owncloudApp/src/main/res/values-es/strings.xml
```diff
@@ -857,7 +857,7 @@
     <string name="create_space_dialog_gb_unit">GB</string>
     <string name="create_space_dialog_empty_error">El nombre del space no puede estar vacío</string>
     <string name="create_space_dialog_length_error">El nombre del space no puede exceder 255 caracteres</string>
-    <string name="create_space_dialog_characters_error">Caracteres prohibidos: / \\ . : \? * &amp;quot; \' &gt; &lt; |</string>
+    <string name="create_space_dialog_characters_error">Caracteres prohibidos: / \\ . : \? * \" \' &gt; &lt; |</string>
```
joragua, line 860: I think there are more string files with `&amp;quot;` and `$quot;` values:

`values-de-rDE/strings.xml`
`values-de/strings.xml`

 They should be replaced, right? 🤔

owncloudApp/src/main/res/values/strings.xml
```diff
@@ -372,8 +372,7 @@
     <string name="auth_expired_basic_auth_toast">Please enter the current password</string>
     <string name="auth_connecting_auth_server">Connecting to authentication server …</string>
     <string name="auth_unsupported_auth_method">The server does not support this authentication method</string>
-    <string name="auth_fail_get_user_name">Your server is not returning a correct user ID. Please contact an administrator.
-	</string>
+    <string name="auth_fail_get_user_name">Your server is not returning a correct user ID. Please contact an administrator.</string>
```
joragua, line 375: The same for indentation. I think there are more string files that need a refactor

changelog/unreleased/4807
jesmrec: Done!

owncloudApp/src/main/res/values-es/strings.xml
jesmrec: Right, they seem to be added after the first task in the PR. Fixing...

owncloudApp/src/main/res/values/strings.xml
jesmrec: In this case, i put the focus in the `strings.xml` because is the only file that the workflow checks. Would it be a good idea to check all the translations file to indent or remove line breaks?

jesmrec: The number of involved strings is very high... it may not woth

owncloudApp/src/main/res/values-es/strings.xml
joragua: Nice! 🙌🏻

owncloudApp/src/main/res/values/strings.xml
joragua: Wontfix, then! 😄

joragua (approved):
LGTM! Nice job 💯

PR #4808, opened 2026-03-16
[FIX] Exclude SBOM from Conventional commits

jesmrec:
SBOM workflow generates commit with title `docs: SBOM updated`

That's not compliant with conventional commits because the first letter of the message is a capital, and must be lower case. That causes CI to fail.

Solution: SBOM commits to be ignored in terms of conventional commits.

About changes in SBOM, seems that `PhotoView` dependency has changed its metadata caught by cycloneDX. We'll put an eye on this, just in case. Dependency is not removed just some of its metadata.

## Related Issues
App:

_____


.github/workflows/conventional-commits.yml
```diff
@@ -50,7 +50,7 @@ jobs:
             echo "$commitmsg"
 
             # Skip merge and [tx] commits
-            if [[ "$commitmsg" =~ ^Merge ]] || [[ "$commitmsg" =~ ^\[tx\] ]]; then
+            if [[ "$commitmsg" =~ ^Merge ]] || [[ "$commitmsg" =~ ^\[tx\] ]] ||  [[ "$commitmsg" =~ SBOM\ update ]]; then
```
joragua: Is `SBOM update` the correct commit message to exclude in the action? The automatic commit for SBOM is `docs: SBOM updated` Or am I missing something? 🤔

joragua (changes requested):
Just a tiny comment 😄

.github/workflows/conventional-commits.yml
jesmrec: You are right, `d` is missing. Fixing...

PR #4817, opened 2026-04-09
[FEATURE REQUEST] Edit link over a space

### LINKED ISSUE #4756: [FEATURE REQUEST] Edit a link over a space
Edit the existing link:

- name (optional)
- level of permission (mandatory)
- password (not always mandatory)
- expiration date

Only for `Space managers`. Pemission: `libre.graph/driveItem/permissions/update`

A pencil icon to be added to the link row, so, the link view is opened and user is able to modify the existing values after clicking on a `Save` button.  A `Cancel` option also available to close the view and keep the values before edition.

Rules are the same as in creation: permission mandatory, password mandatory with policy applied, and name optional with max 255 chars.

Extra: clicking on the cell could open the link to be edited.
Extra II: In the view to edit the link, add an option to copy the link to the clipboard.

**Requests to edit**

1) With no password change (name, permission, expiration)

`PATCH https://<url>/graph/v1beta1/drives/<drive-id>/root/permissions/<link-id>`

with body:
```
{
    "expirationDateTime": "2026-01-31T08:30:00.000Z",
    "link": {
        "@libre.graph.displayName": "NewName",
        "type": "createOnly"
    }
}
```

only required the changed values. 

2) Password change

`POST https://<url>/graph/v1beta1/drives/<drive-id>/root/permissions/<link-id>/setPassword`

with body:

```
{
    "password": "<newPassword>"
}
```

in both cases, the response body like:

```
{
    "createdDateTime": "2025-12-18T13:13:01.272005287Z",
    "expirationDateTime": "2026-01-29T08:30:00Z",
    "hasPassword": true,
    "id": "vOUEFZzkFvEClJL",
    "link": {
        "@libre.graph.displayName": "name",
        "@libre.graph.quickLink": false,
        "preventsDownload": false,
        "type": "view",
        "webUrl": "https://<url>/s/aMhKFSTbWlFIXQb"
    }
}
```

After correct edition, view closed and refreshed list of links/members displayed.

### TASKS

- [ ] Research (if needed)
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Design test plan
- [ ] QA

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4756

_____

## QA

Test plan: https://github.com/owncloud/QA/blob/master/Mobile/Android/Executions/Release_4.8/Spaces%20-%20Edit%20link.md

- [x] (1)(improvement) Add an (x) to remove name with a click https://github.com/owncloud/android/pull/4817#issuecomment-4334292246 [FIXED]
- [x] (2) Year elipsized https://github.com/owncloud/android/pull/4817#issuecomment-4334402358 [FIXED]
- [X] (3) Not posible to edit link because lack of password https://github.com/owncloud/android/pull/4817#issuecomment-4336592600 [FIXED]

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
@@ -31,19 +31,15 @@ import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
 import com.owncloud.android.R
 import com.owncloud.android.databinding.AddPublicLinkFragmentBinding
-import com.owncloud.android.domain.capabilities.model.OCCapability
 import com.owncloud.android.domain.links.model.OCLinkType
 import com.owncloud.android.domain.spaces.model.OCSpace
 import com.owncloud.android.extensions.collectLatestLifecycleFlow
 import com.owncloud.android.extensions.hideSoftKeyboard
 import com.owncloud.android.extensions.showErrorInSnackbar
-import com.owncloud.android.presentation.capabilities.CapabilityViewModel
 import com.owncloud.android.presentation.common.UIResult
 import com.owncloud.android.utils.DisplayUtils
 import org.koin.androidx.viewmodel.ext.android.activityViewModel
-import org.koin.androidx.viewmodel.ext.android.viewModel
 import org.koin.core.parameter.parametersOf
-import timber.log.Timber
 import java.text.SimpleDateFormat
 import java.util.Locale
 import java.util.TimeZone
@@ -60,13 +56,7 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
             requireArguments().getParcelable(ARG_CURRENT_SPACE)
         )
     }
-    private val capabilityViewModel: CapabilityViewModel by viewModel {
-        parametersOf(
-            accountName
-        )
-    }
 
-    private var capabilities: OCCapability? = null
     private var isPasswordEnforced = true
     private var hasPassword = false
 
@@ -126,18 +116,6 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
                         }
                         showPasswordDialog(uiState.selectedPassword)
                     }
-                }
-            }
-        }
-
-        capabilityViewModel.capabilities.observe(viewLifecycleOwner) { event->
-            when (val uiResult = event.peekContent()) {
-                is UIResult.Success -> {
-                    capabilities = uiResult.data
-                }
-                is UIResult.Loading -> { }
-                is UIResult.Error -> {
-                    Timber.e(uiResult.error, "Failed to retrieve server capabilities")
                 }
             }
         }
@@ -203,7 +181,7 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
             selectedRadioButton.isChecked = true
         }
         val selectedPermission = selectedRadioButton.tag as OCLinkType
-        isPasswordEnforced = capabilityViewModel.checkPasswordEnforced(selectedPermission, capabilities)
+        isPasswordEnforced = spaceLinksViewModel.checkPasswordEnforced(selectedPermission)
         spaceLinksViewModel.onPermissionSelected(selectedPermission)
     }
 
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
@@ -168,17 +168,7 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
 
         bindRadioButtonsListeners()
 
-        binding.passwordLayout.apply {
-            setPasswordButton.setOnClickListener {
-                showPasswordDialog()
-            }
-            removePasswordButton.setOnClickListener {
-                removePassword()
-            }
-            setPasswordSwitch.setOnClickListener {
-                if (setPasswordSwitch.isChecked) showPasswordDialog() else removePassword()
-            }
-        }
+        bindPasswordListeners()
 
         binding.createPublicLinkButton.setOnClickListener {
             val displayName = binding.publicLinkNameEditText.text.toString().ifEmpty { getString(R.string.public_link_default_display_name) }
@@ -278,6 +268,20 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
                 if (expirationDate == null) {
                     binding.expirationDateLayout.expirationDateSwitch.isChecked = false
                 }
+            }
+        }
+    }
+
+    private fun bindPasswordListeners() {
+        binding.passwordLayout.apply {
+            setPasswordButton.setOnClickListener {
+                showPasswordDialog()
+            }
+            removePasswordButton.setOnClickListener {
+                removePassword()
+            }
+            setPasswordSwitch.setOnClickListener {
+                if (setPasswordSwitch.isChecked) showPasswordDialog() else removePassword()
             }
         }
     }
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
@@ -274,7 +274,10 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
     private fun bindEditMode() {
         selectedPublicLink?.let {
             binding.publicLinkNameEditText.setText(it.displayName)
-            binding.createPublicLinkButton.setText(R.string.share_confirm_public_link_button)
+            binding.createPublicLinkButton.apply {
+                setText(R.string.share_confirm_public_link_button)
+                contentDescription = getString(R.string.share_confirm_public_link_button)
+            }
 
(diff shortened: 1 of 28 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -263,26 +263,61 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
...
+    private fun bindEditMode() {
+        selectedPublicLink?.let {
+            binding.publicLinkNameEditText.setText(it.displayName)
+            binding.createPublicLinkButton.apply {
+                setText(R.string.share_confirm_public_link_button)
+                contentDescription = getString(R.string.share_confirm_public_link_button)
+            }
+
+            // Do not recreate the edit view after the first iteration
+            if (spaceLinksViewModel.addPublicLinkUIState.value?.selectedPermission != null) return
```
jesmrec, line 313: Should this condition to be the first in the bock to be checked? if no permission to return even before the name and the button are binded

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
```diff
@@ -263,26 +263,61 @@ class AddPublicLinkFragment: Fragment(), SetPasswordDialogFragment.SetPasswordLi
...
+            selectedPublicLink: OCLink?
         ): AddPublicLinkFragment {
             val args = Bundle().apply {
                 putString(ARG_ACCOUNT_NAME, accountName)
                 putParcelable(ARG_CURRENT_SPACE, currentSpace)
+                putBoolean(ARG_EDIT_MODE, editMode)
+                putParcelable(ARG_SELECTED_PUBLIC_LINK, selectedPublicLink)
             }
             return AddPublicLinkFragment().apply {
                 arguments = args
```
jesmrec, line 355: I miss `OnDestroyView` in this fragment (not part of this specific review but...)

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/OCLinksService.kt
```diff
@@ -42,6 +44,28 @@ class OCLinksService(override val client: OwnCloudClient) : LinksService {
...
+    ): RemoteOperationResult<Unit> =
+        EditRemoteLinkOperation(
+            spaceId = spaceId,
+            linkId = linkId,
+            displayName = displayName,
+            type = type,
+            expirationDate = expirationDate
+        ).execute(client)
+
+    override fun editPasswordLink(spaceId: String, linkId: String, password: String?): RemoteOperationResult<Unit> =
```
jesmrec: Should be every parameter in a line break like the `fun` above?

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
```diff
@@ -99,14 +117,53 @@ class SpaceLinksViewModel(
...
+                viewModelScope.launch(coroutineDispatcherProvider.io) {
+                    editPasswordLinkUseCase(EditPasswordLinkUseCase.Params(
+                        accountName = accountName,
+                        spaceId = space.id,
+                        linkId = linkId,
+                        password = _addPublicLinkUIState.value?.selectedPassword
+                    ))
+                }
+            }
+        }
```
jesmrec, line 148: there are two different calls since there are two request, that's right. But, what happens if the first request fails (for any reason)? imo, the second one should be cancelled. Not sure if this make sense or it is somehow controlled.

On the same way, what happens if `EditLinkUseCase` successes and `EditPasswordLinkUseCase` fails?

Even when they are two separate request, should they be handled together as an atomic operation?

jesmrec (changes requested):
Some comments from my side @joragua!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: This condition is used to avoid re-creating the edit view when the device is rotated, but the button still needs to be bound because the default label is `CREATE` (it should be `SAVE` in edit mode). The text binding could be moved below the condition.

New version:
```
binding.createPublicLinkButton.apply { ... }

// Do not recreate the edit view after the first iteration
if (spaceLinksViewModel.addPublicLinkUIState.value?.selectedPermission != null) return

binding.publicLinkNameEditText.setText(it.displayName)
```

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
joragua: Right! Added ✅ I've added `onDestroyView` method in `SetPasswordDialogFragment` too

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/OCLinksService.kt
joragua: Done 👍🏻

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
joragua: Changed ✅

The second request (password update) will only be triggered if the first request (edit link operation) succeeds and the password has changed.

There is one corner case with the new implementation: if the first request succeeds but the second one fails, the user will see a snackbar with the following message: `Public link could not be edited because <error_message>` but the link will be partially updated (name, permission and expiration date) There is no way to roll back the changes when the second request fails. Is that OK? 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: 💯

owncloudComLibrary/src/main/java/com/owncloud/android/lib/resources/links/services/OCLinksService.kt
jesmrec: Perfect

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/AddPublicLinkFragment.kt
jesmrec: Wfm!

owncloudApp/src/main/java/com/owncloud/android/presentation/spaces/links/SpaceLinksViewModel.kt
jesmrec: It's a good solution, since prevents the critical problem: an invisible change in the password. For me it's OK

jesmrec (approved):
CR approved, moving to QA

jesmrec:
### (1) (improvement) [FIXED]

It would be handy to add a (x) in the link name row to remove the name with just one click. What do you think?  (iOS has it)

`2617e171`

jesmrec:
### (2) [FIXED]

This is how a link with `Secret File Drop` permission and expiration date looks like:

Screenshot 2026-04-28 at 12 26 54

The year is elipisized and not visible, but, the timestamp `23:59` is visible. I think that the complete date would be more interesting to show. 

`2617e171`

joragua:
**(1)** should be addressed in commit 1adf344 -> When the edit text for the link name is focused, all the text is selected. In the whole app we don't have an **(X)** icon to clear text so I didn't go with that approach. Let me know if it's OK for you @jesmrec 😄 

**(2)** should be fixed in commit 3e99045 -> The expiration date is now ellipsized at the end instead of in the middle in order to display the full date (not the timestamp)

jesmrec:
### (3) [FIXED]


1. In a server with mandatory password for writeable links, click on the pencil icon of a link with `Can view` permission and password not set
2. Change permission to `Can edit`
3. Click on `Set` and `Generate` to create a new password (or type it yourself)
4. Click on `Set` to submit the edition

**Current:**

Snackbar with `Public link could not be edited because unknown error`. 

Server returns `400` on the `PATCH` request. Error `400` happens because the `PATCH` tries to update the link with password mandatory (step 2) and password is not set yet.

**Expected**

Link updated correctly.

Probably, the requests have to be swapped, as iOS does. 

Not sure if the `400` should be handled as well

`2617e171`

joragua:
**(3)** should be fixed in commit https://github.com/owncloud/android/pull/4817/commits/1ecf016e9123367432bff6bf80c4ee8a0de01159

The current behavior is:

- Scenario 1: The user has changed the password -> Run the `POST` operation to update the password. If the operation fails, a snackbar appears on the screen showing the error. If the operation is OK, the `PATCH` operation is triggered (its result is ignored in order to avoid an inconsistent state - two snackbars with different messages at the same time) and a snackbar is shown on the space members screen.

- Scenario 2: The user hasn't changed the password -> Run the `PATCH` operation. If the operation fails, a snackbar appears on the screen showing the error. If the operation is OK, a different snackbar is shown on the space members screen.

On the other hand, there is also a strange scenario (it applies to servers with mandatory password for writeable links and only in edit mode): 

- It's not possible to change the permission from `Can edit` or `Secret File Drop` to `Can view` and remove the password at the same time. This is because the `POST` operation is triggered first, and it's not possible to remove the password from a link with a mandatory password (since the permission hasn't changed yet). In this case, a snackbar with the message `Public link could not be edited because password is mandatory` will appear on the screen.

jesmrec:
Approved on my side. Moving on...

PR #4821, opened 2026-04-13
[FIX] Remove unused resources

jesmrec:
## Related Issues
App:

_____


jesmrec:
The following script will check all `strings*.xml` files and remove those ones that are not in the list

```
#!/usr/bin/env bash

NAMES_FILE="strings_to_delete.txt"

find . -type f -path "*/res/values*/strings*.xml" | while IFS= read -r file; do
  echo "Processing: $file"

  while IFS= read -r name; do
    gsed -i \
      -e "/<string[^>]*name=\"$name\"[^>]*>.*<\/string>/d" \
      -e "/<string[^>]*name=\"$name\"[^>]*>/,/<\/string>/d" \
      "$file"
  done < "$NAMES_FILE"
done

echo "Done."
```

List:

[strings_to_delete.txt.zip](https://github.com/user-attachments/files/26716928/strings_to_delete.txt.zip)

PR #4853, opened 2026-05-12
[TECHNICAL] Adapt changelog workflow to release branches

jesmrec:
In case the push is done from a release branch, the PR will be created against such branch to avoid rebasing.

## Related Issues
App:

_____

### DIFF
--- a/.github/workflows/calens.yml
+++ b/.github/workflows/calens.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - 'release/**'
 
 permissions:
   contents: read
@@ -11,6 +12,8 @@ permissions:
 jobs:
   changelog:
     uses: owncloud/reusable-workflows/.github/workflows/calens.yml@main
+    with:
+      branch: ${{ github.ref_name }}
     secrets:
       APP_ID: ${{ secrets.CALENS_APP_ID }}
       APP_PRIVATE_KEY: ${{ secrets.CALENS_APP_PRIVATE_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/calens.yml
```diff
@@ -4,13 +4,24 @@ on:
...
 
 permissions:
   contents: read
 
 jobs:
-  changelog:
-    uses: owncloud/reusable-workflows/.github/workflows/calens.yml@main
+  changelog-master:
+    if: github.ref_name == 'master'
+    uses: mio/reusable-workflows/.github/workflows/calens.yml@main
```
joragua: What does `mio` refer to? Shouldn't it be `owncloud/reusable-workflows/.github/workflows/calens.yml@main` instead?

joragua (changes requested):
Just one comment @jesmrec!

.github/workflows/calens.yml
jesmrec: so weird, no idea why is that there... anyway, fixed

joragua: 👍🏻

joragua (approved):
LGTM! 💯

DeepDiver1975 (changes requested):
no need to duplicate the code ..... in case of master the ref_name is master as well ... right?

jesmrec:
> no need to duplicate the code ..... in case of master the ref_name is master as well ... right?

you're right... sending the branch to the reusable is enough, no matter which branch triggered the workflow:

- if `master` : calens PR against `master`
- if `release/**` : calens PR against `release/**`

DeepDiver1975 (approved):

PR #4869, opened 2026-05-20
[FIX] Remove unused resources

jesmrec:
Follow up https://github.com/owncloud/android/pull/4821/

The following script will check all `strings*.xml` files and remove those ones that are not in the list

```
#!/usr/bin/env bash

NAMES_FILE="strings_to_delete.txt"

find . -type f -path "*/res/values*/strings*.xml" | while IFS= read -r file; do
  echo "Processing: $file"

  while IFS= read -r name; do
    gsed -i \
      -e "/<string[^>]*name=\"$name\"[^>]*>.*<\/string>/d" \
      -e "/<string[^>]*name=\"$name\"[^>]*>/,/<\/string>/d" \
      "$file"
  done < "$NAMES_FILE"
done

echo "Done."
```

List:

[strings_to_delete.txt.zip](https://github.com/user-attachments/files/26716928/strings_to_delete.txt.zip)

## Related Issues
App:

_____


owncloudApp/src/main/res/values/styles.xml
```diff
@@ -99,10 +97,6 @@
     </style>
 
     <!-- USED IN SHAREACTIVITY!! !! -->
-    <style name="Theme.ownCloud.AlertDialog.NoTitle">
-        <item name="windowNoTitle">true</item>
-        <item name="android:windowOptOutEdgeToEdgeEnforcement" tools:targetApi="35">true</item>
-    </style>
```
joragua: ^^^^^

joragua (changes requested):
Just one comment @jesmrec! I’d remove all comments related to the removed resources because they’re no longer useful 🤔 I've marked an example just below

owncloudApp/src/main/res/values/styles.xml
jesmrec: Good catch! removed useless comments after the cleanup

joragua (approved):
LGTM now! Good job 💯

owncloudApp/src/main/res/values/styles.xml
joragua: Perfect! 👍🏻

PR #4897, opened 2026-06-18
fix: redact cookie headers to prevent information leaks

joragua:
## Related Issues
Comes from: https://kiteworks.atlassian.net/browse/OC10-95

It depends on the `redact_auth_header_logs` configuration in `setup.xml`. If it is set to true, the values of the cookie headers (`Cookie` and `Set-Cookie`) are replaced with `[redacted]` (the same as the `Authorization` header)

Reference: https://github.com/owncloud/android/issues/4249

_____

### DIFF
--- a/changelog/unreleased/4897
+++ b/changelog/unreleased/4897
@@ -1,0 +1,6 @@
+Security: Redact cookie headers in logs
+
+Cookie headers have been redacted in the logs depending on the app
+configuration, in order to prevent sensitive information leaks.
+
+https://github.com/owncloud/android/pull/4897
--- a/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
+++ b/owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
@@ -1,5 +1,5 @@
 /* ownCloud Android Library is available under MIT license
- *   Copyright (C) 2023 ownCloud GmbH.
+ *   Copyright (C) 2026 ownCloud GmbH.
  *
  *   Permission is hereby granted, free of charge, to any person obtaining a copy
  *   of this software and associated documentation files (the "Software"), to deal
@@ -24,7 +24,9 @@
 package com.owncloud.android.lib.common.http.logging
 
 import com.owncloud.android.lib.common.http.HttpConstants.AUTHORIZATION_HEADER
+import com.owncloud.android.lib.common.http.HttpConstants.COOKIE_HEADER
 import com.owncloud.android.lib.common.http.HttpConstants.OC_X_REQUEST_ID
+import com.owncloud.android.lib.common.http.HttpConstants.SET_COOKIE_HEADER
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import okhttp3.Headers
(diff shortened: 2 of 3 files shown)


owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
```diff
@@ -88,9 +90,13 @@ class LogInterceptor : Interceptor {
...
             val authHeaderList = auxHeaders[AUTHORIZATION_HEADER]!!.split(" ")
             val authType = authHeaderList[0]
-            val authInfo = if (redactAuthHeader) "[redacted]" else authHeaderList[1]
+            val authInfo = if (redactAuthHeader) REDACTED_VALUE else authHeaderList[1]
             auxHeaders[AUTHORIZATION_HEADER] = "$authType $authInfo"
         }
+        if (redactAuthHeader) {
+            if (COOKIE_HEADER in auxHeaders) { auxHeaders[COOKIE_HEADER] = REDACTED_VALUE }
+            if (SET_COOKIE_HEADER in auxHeaders) { auxHeaders[SET_COOKIE_HEADER] = REDACTED_VALUE }
+        }
```
jesmrec: Since `redactAuthHeader` is an static value, would it be posible to evaluate the value only once? There are three `if` statements that will return all the same value.

changelog/unreleased/4897
```diff
@@ -0,0 +1,6 @@
+Security: Redact cookie headers in logs
+
+Cookie headers in oC10 have been redacted in the logs depending on the
+app configuration, in order to prevent sensitive information leaks.
+
```
jesmrec, line 5: My proposal:

```
Cookie headers have been redacted in the logs depending on the app configuration, in order to prevent sensitive information leaks.
```

Solution does not depend on oC10/oCIS, right?

jesmrec (changes requested):
Two changes, @joragua

changelog/unreleased/4897
joragua: Changed! 👍🏻

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
joragua: Refactored! `redactAuthHeader` is now used only once. Anyway, it will be necessary to check in QA if the `Authorization` is still being redacted (because the logic has changed)

changelog/unreleased/4897
jesmrec: 👍

owncloudComLibrary/src/main/java/com/owncloud/android/lib/common/http/logging/LogInterceptor.kt
jesmrec: Will check in QA, but the code is now cleaner ;)

jesmrec (approved):
CR Approved, moving to QA 🚀

jesmrec:
QA checks:

1. `redact_auth_header_logs` as `true` (oC10 & oCIS)

- [x] `Cookie` header -> must be redacted (only oC10)
- [x] `Set-Cookie` header -> must be redacted (only oC10)
- [x] `Authentication` header -> must be redacted

2. `redact_auth_header_logs` as `false`  (oC10 & oCIS)

- [x] `Cookie` header -> not redacted (only oC10)
- [x] `Set-Cookie` header -> not redacted (only oC10)
- [x] `Authentication` header -> not redacted

3. Change value of  `redact_auth_header_logs` 

- [x] from `false` to `true`
- [x] from `true` to `false`

jesmrec:
Approved on my side!

PR #4909, opened 2026-06-29
fix: lint error in strings

### LINKED ISSUE #4908: [BUG] code requires API level 27 (current min is 24)
### Actual behaviour

Image

Adding a baseline makes no sense at all. You should fix it !

### Option 1: Move to values-v27/styles.xml
Split the style: keep the base in values/styles.xml (without the cutout item), add the cutout item in values-v27/styles.xml.
 Pros: 
 Correct Android way — safe on all API levels 
 No suppression needed 
 No runtime conflicts on older devices 
Cons: 
 Requires duplicating/splitting style across two files 
 More maintenance overhead 
 
### Option 2: Raise minSdkVersion to 27
Change minSdkVersion 24 → minSdkVersion 27 in build.gradle.
Pros: 
 Simplest change, no code split needed 
 Eliminates all API 24–26 lint issues 
 Cons: 
 Drops support for Android 7.x/8.0 (~5–8% of devices) 
 May conflict with project policy/requirements 

### Option 3: Add tools:targetApi="27" in XML
Add tools:targetApi="27" to the <item> or parent <style>.
 Pros: 
 Suppresses lint, no file splitting 
 Minimal change 
 Cons: 
 Doesn't prevent runtime crash if attribute conflicts on API 24–26 (manufacturer custom attrs) 
 Just a lint suppression — doesn't actually make the code safe

hannesa2:
close https://github.com/owncloud/android/issues/4908

### This is option 1: Move to values-v27/styles.xml

Split the style: keep the base in values/styles.xml (without the cutout item), add the cutout item in values-v27/styles.xml.
Pros:
Correct Android way — safe on all API levels
No suppression needed
No runtime conflicts on older devices
Cons:
Requires duplicating/splitting style across two files
More maintenance overhead

### DIFF
--- a/owncloudApp/src/main/res/values-v27/styles.xml
+++ b/owncloudApp/src/main/res/values-v27/styles.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <!-- Extends Theme.ownCloud.Video with display cutout support (API 27+) -->
+    <style name="Theme.ownCloud.Video">
+        <item name="android:windowLayoutInDisplayCutoutMode">never</item>
+    </style>
+
+</resources>
--- a/owncloudApp/src/main/res/values/styles.xml
+++ b/owncloudApp/src/main/res/values/styles.xml
@@ -73,7 +73,6 @@
     <!-- Extends Theme.ownCloud for video in full screen mode -->
     <style name="Theme.ownCloud.Video">
         <item name="windowActionBarOverlay">true</item>
-        <item name="android:windowLayoutInDisplayCutoutMode">never</item>
         <item name="actionBarTheme">@style/ownCloud.Appbar.Transparent</item>
         <item name="android:windowBackground">@android:color/background_dark</item>
     </style>


jesmrec:
Thanks again @hannesa2 , we will review as soon as we can 😅

DeepDiver1975 (commented):
Thanks for tackling the API-level lint issue (#4908) — the diagnosis is right: `android:windowLayoutInDisplayCutoutMode` is API 27+, while `minSdkVersion` is 24, so having it in `values/styles.xml` legitimately trips lint. Option 1 (a `-v27` resource bucket) is a sound approach in principle.

However, as written this looks like it introduces a regression on API 27+ devices (i.e. nearly all devices):

- Android resource merging does **not** merge two same-name `<style>` definitions item-by-item across qualifier folders. When `Theme.ownCloud.Video` exists in both `values/` and `values-v27/`, the `values-v27/` definition **fully replaces** the base one for API 27+.
- The new `values-v27/styles.xml` declares only `android:windowLayoutInDisplayCutoutMode`. So on API 27+ the style would lose the other three items it currently carries: `windowActionBarOverlay`, `actionBarTheme` (`@style/ownCloud.Appbar.Transparent`), and `android:windowBackground` (`@android:color/background_dark`). The fullscreen video theme would render differently (background, action-bar overlay/transparency) on essentially every device.

Suggested fixes (either works):
1. Have the v27 variant re-declare all four items (copy the base style's items plus the cutout one), keeping them in sync; or
2. Keep a single style and suppress the lint precisely — e.g. extract the cutout item via `tools:targetApi="27"`, or split into a small base style + a `-v27` override that uses `parent="..."` pointing at a *differently named* base style so inheritance is explicit. (A same-named cross-qualifier override can't use `parent` to pull in its own base.)

Option 1 is the cleaner long-term route as you noted — it just needs the v27 copy to be complete.

Changelog: this repo uses calens fragments (`changelog/unreleased/<id>`) and this is a Bugfix closing an issue, so please add `changelog/unreleased/4908` following `changelog/TEMPLATE.md`, e.g.:

```
Bugfix: Fix lint error for display cutout mode below API 27

The windowLayoutInDisplayCutoutMode attribute (API 27+) has been moved to a
values-v27 resource so it is no longer referenced below the minimum SDK level.

https://github.com/owncloud/android/issues/4908
https://github.com/owncloud/android/pull/4909
```

Mergeability: branch is `MERGEABLE` but `BLOCKED` (only `license/cla` reports; required checks/review pending). Requesting changes on the style-override regression and the missing changelog fragment before this should land.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

joragua:
This is no longer necessary, as the minimum SDK version has already been bumped to API 28 (Android 9)

Thanks for your effort! 🙌🏻

PR #4936, opened 2026-07-22
feat: implement the new oCIS sharing UI

### LINKED ISSUE #4839: [FEATURE REQUEST] Sharing NG - preliminary
As preliminary step to implement [Sharing NG](https://github.com/owncloud/android/issues/4511), the `Share` option in the bottom sheet should open:

- The legacy dialog for oC10, since it is still alive
- The new dialog for oCIS, with Sharing NG implementation

For that aim, the backend type should be recovered from somewhere (typically, capabilities), to choose the correct path.

In this first issue, loading the fragment in an empty web is enough.

Also, it should be enabled for items inside of custom spaces, depending on the permission.

Pay attention to the badges in the list of files. They work thanks to the ocs backwards compatibility, but, in oCIS this should be avoided.

### TASKS

- [ ] Research (if needed)
- [ ] Development tasks
  - [ ] Implement whatever
  - [ ] ...
- [ ] Design test plan
- [ ] QA

joragua:
## Related Issues
App: https://github.com/owncloud/android/issues/4839

_____

### DIFF
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/capabilities/CapabilityViewModel.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/capabilities/CapabilityViewModel.kt
@@ -4,8 +4,9 @@
  * @author David González Verdugo
  * @author Abel García de Prada
  * @author Juan Carlos Garrote Gascón
+ * @author Jorge Aguado Recio
  *
- * Copyright (C) 2024 ownCloud GmbH.
+ * Copyright (C) 2026 ownCloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
@@ -73,6 +74,10 @@ class CapabilityViewModel(
             accountName = accountName
         )
     )
+
+    fun isOcisServer(): Boolean = runBlocking(coroutineDispatcherProvider.io) {
+        getStoredCapabilitiesUseCase(GetStoredCapabilitiesUseCase.Params(accountName))?.isSpacesAllowed() == true
+    }
 
     fun checkMultiPersonal(): Boolean = runBlocking(CoroutinesDispatcherProvider().io) {
         val capabilities = withContext(CoroutinesDispatcherProvider().io) {
--- a/owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
+++ b/owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
@@ -34,8 +34,11 @@ import android.view.KeyEvent
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
+import androidx.core.view.isVisible
 import androidx.fragment.app.transaction
 import com.owncloud.android.R
+import com.owncloud.android.databinding.MembersActivityBinding
+import com.owncloud.android.datamodel.ThumbnailsCacheManager
 import com.owncloud.android.domain.files.model.OCFile
 import com.owncloud.android.domain.sharing.shares.model.OCShare
 import com.owncloud.android.domain.sharing.shares.model.ShareType
@@ -43,6 +46,7 @@ import com.owncloud.android.domain.utils.Event.EventObserver
 import com.owncloud.android.extensions.adaptInfiniteEdges
 import com.owncloud.android.extensions.showErrorInSnackbar
 import com.owncloud.android.lib.resources.shares.RemoteShare
+import com.owncloud.android.presentation.capabilities.CapabilityViewModel
 import com.owncloud.android.presentation.common.UIResult
 import com.owncloud.android.presentation.sharing.sharees.EditPrivateShareFragment
 import com.owncloud.android.presentation.sharing.sharees.SearchShareesFragment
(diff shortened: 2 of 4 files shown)


owncloudApp/src/main/java/com/owncloud/android/presentation/capabilities/CapabilityViewModel.kt
```diff
@@ -74,6 +75,10 @@ class CapabilityViewModel(
         )
     )
 
+    fun isOcisServer(): Boolean = runBlocking(coroutineDispatcherProvider.io) {
+        getStoredCapabilitiesUseCase(GetStoredCapabilitiesUseCase.Params(accountName))?.isSpacesAllowed() == true
+    }
```
jesmrec, line 80: Is this executed in the main thread? it could lead to ANR.

On the other hand, every time you need the capabilities, you are fetching them from DB. Would it be better to read them from DB once and store in a variable or similar?

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
```diff
@@ -64,37 +70,63 @@ class ShareActivity : FileActivity(), ShareFragmentListener {
...
+        adaptInfiniteEdges(root)
+        setupStandardToolbar(title = null, displayHomeAsUpEnabled = true, homeButtonEnabled = true, displayShowTitleEnabled = true)
+        supportActionBar?.setHomeActionContentDescription(R.string.common_back)
+    }
+
+    private fun setupFileHeader(binding: MembersActivityBinding) {
+        binding.itemIcon.setImageResource(MimetypeIconUtil.getFileTypeIconId(file.mimeType, file.fileName))
+        if (file.isImage) {
+            val thumbnail = ThumbnailsCacheManager.getBitmapFromDiskCache(file.remoteId.toString())
+            if (thumbnail != null) { binding.itemIcon.setImageBitmap(thumbnail) }
```
jesmrec, line 118: I miss here an `else`. What happens if `thumbnail` is null?

jesmrec (changes requested):
A couple of changes @joragua

owncloudApp/src/main/java/com/owncloud/android/presentation/capabilities/CapabilityViewModel.kt
joragua: The main thread is blocked until the operation finishes, but in this case, it should not take that long... 

Regarding the other comment, it's true that the capabilities are fetched every time you click on `Share` option. How often does that actually happen? Does the user click this option more than once in few seconds? I think we can keep the current behavior for now and check it in the future if it is necessary (if we need the capabilities for other purposes) What do you think? 🤔

owncloudApp/src/main/java/com/owncloud/android/presentation/sharing/ShareActivity.kt
joragua: The `else` branch is not needed here because if thumbnail is `null` the mime-type is set by default (line 115)

jesmrec: 👍

owncloudApp/src/main/java/com/owncloud/android/presentation/capabilities/CapabilityViewModel.kt
jesmrec: > The main thread is blocked until the operation finishes, but in this case, it should not take that long...

Not sure if it will be noticeable in a budget device. We should adapt to any case. 

In case of `oCIS` probably capabilities will not be checked again in the scenario. Afaik, oC10 needs from them again inside.

joragua: Some conclusions:

1) The risk of having `runBlocking` is theoretical, but not practical (it will no take more than 5 seconds to retrieve the capabilities) Anyway, it should be tested in the QA phase

2) The capabilities operation should not be asynchronous in this case because `setContentView` must be called from `onCreate` to determine which UI should be displayed. With this conclusion, I think using `runBlocking` is justified (similar to the `checkMultiPersonal` method)

Anyway, this behavior may change in the future, so we shouldn't forget about this implementation

jesmrec: Let's give a try!

jesmrec (approved):
CR Approved, moving to QA

jesmrec:
Just a consideration: In this PR, the restriction of sharing for items inside custom spaces has been removed, so it depends on the item permission to show or hide the `Share` option. 

That collides with the idea that the permalink is always available, and it is inside of `Share` menu. A user could not be allowed to share, and it'd lose the access to the permalink.

So, the `Share` option should be always available and other permissions will rule the sharing options.

jesmrec:
> Just a consideration: In this PR, the restriction of sharing for items inside custom spaces has been removed, so it depends on the item permission to show or hide the Share option.
> That collides with the idea that the permalink is always available, and it is inside of Share menu. A user could not be allowed to share, and it'd lose the access to the permalink.
> So, the Share option should be always available and other permissions will rule the sharing options.

This research is deeper as expected, so, it will be addressed to other issue/moment.

Approved on my side

PR #4941, opened 2026-07-24
chore: add CLAUDE.md for guidance

joragua:
## Related Issues
App: -

_____

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,114 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Build & Test Commands
+
+```bash
+# Build
+./gradlew assembleDebug
+./gradlew assembleRelease          # requires signing env vars
+
+# Unit tests (all flavors)
+./gradlew testDebugUnitTest testMdmDebugUnitTest --continue
+
+# Single module unit tests
+./gradlew :owncloudApp:testOriginalDebugUnitTest
+./gradlew :owncloudData:testDebugUnitTest
+
+# Instrumented tests (requires running emulator/device)
+./gradlew :owncloudData:connectedAndroidTest
+
+# Lint & static analysis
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
FYI: claude should already read the agents.md file which was added to the repo with the OSPO PR: https://github.com/owncloud/android/blob/master/agents.md

joragua:
I will update the `agents.md` file with additional information to improve the guidelines for AI agents.

Closing this PR and opening a new one 🚀

jesmrec:
> FYI: claude should already read the agents.md file which was added to the repo with the OSPO PR:

i have asked claude about that, and i got the following:

```
❯ is agents.md as useful  as CLAUDE.md for you?                                                                                                                                                                                                          

  Thought for 6s (ctrl+o to expand)

⏺ No — agents.md is not as useful for me. Claude Code automatically loads CLAUDE.md into every conversation's context (you can see it in the system prompt above as "project instructions, checked into the codebase"). agents.md is just a regular file;
  I won't see it unless I explicitly read it.

  You have a few options:

  1. Rename agents.md back to CLAUDE.md — simplest fix, restores automatic loading.
  2. Keep agents.md and create a minimal CLAUDE.md that points to it — something like @agents.md (Claude Code supports @filename imports in CLAUDE.md). This way other agents can use agents.md directly while I still get it auto-loaded.
  3. Keep both names in sync — but that's maintenance overhead.

  Option 2 is probably the best fit given your intent to have a single shared file for all agents. Want me to restore agents.md as-is and create a minimal CLAUDE.md that imports it?

✻ Brewed for 10s

❯ so you don't load agents.md for getting context                                                                                                                                                                                                        

  Thought for 4s (ctrl+o to expand)

⏺ Correct. I only auto-load CLAUDE.md (and any files it imports via @filename). agents.md is invisible to me unless I explicitly read it with a tool call.
     
```

is there something that i'm missing @DeepDiver1975 ?

PR #4976, opened 2026-09-10
fix: show the full description of space permissions

### LINKED ISSUE #4968: [BUG] Incomplete description in space permissions
### Steps to reproduce
1.  Open `Spaces` list
2.  Select any user to add as a member

### Actual behaviour

The list of permissions is correct, but, a longer description is not correctly shown:

Image

### Expected behaviour

that's how that permission looks like in web:

Image

 We should book more room for the description, or ellipsize the text.

Reproduced with Xiaomi Redmi 13, Android 15

retransmit:
## Related Issues
App: https://github.com/owncloud/android/issues/4968

- [x] Add feature to Release Notes in `ReleaseNotesViewModel.kt` creating a new `ReleaseNote()` with String resources (if required) -> not required for this fix

_____

## Description
The role description in the space member permissions list was capped at two lines with `ellipsize="middle"`. Android only applies a middle ellipsis to single-line text, so the third line of longer descriptions such as "Can manage" was silently hidden.

The description now has room for four lines and uses `ellipsize="end"`, which Android does apply to multi-line text, so anything longer is visibly truncated instead of hidden.

## QA
Before / after renderings of the three roles at 360dp on an Android 15 emulator attached below.

### DIFF
--- a/owncloudApp/src/main/res/layout/role_item.xml
+++ b/owncloudApp/src/main/res/layout/role_item.xml
@@ -74,8 +74,8 @@
                 android:text="@string/placeholder_sentence"
                 android:textSize="13sp"
                 android:textColor="@color/textColor"
-                android:ellipsize="middle"
-                android:maxLines="2"/>
+                android:ellipsize="end"
+                android:maxLines="4"/>
 
         </LinearLayout>
 
--- a/changelog/unreleased/4976
+++ b/changelog/unreleased/4976
@@ -1,0 +1,6 @@
+Bugfix: Show the full description of space permissions
+
+More room has been given to the description of a role in the space member permissions list, so longer descriptions are no longer cut after the second line, and any text that still does not fit is ellipsized.
+
+https://github.com/owncloud/android/issues/4968
+https://github.com/owncloud/android/pull/4976


joragua:
Hi @retransmit! Thanks for opening this PR! 🙌🏻 

I will take of the CR as soon as possible. Stay tuned!

owncloudApp/src/androidTest/java/com/owncloud/android/spaces/members/SpaceRolesAdapterTest.kt
joragua, line 1: We only maintain tests for repositories and data sources in the project, so you can remove this test 😄

owncloudApp/src/main/res/layout/role_item.xml
```diff
@@ -74,8 +74,7 @@
                 android:text="@string/placeholder_sentence"
                 android:textSize="13sp"
                 android:textColor="@color/textColor"
-                android:ellipsize="middle"
-                android:maxLines="2"/>
+                android:ellipsize="end"/>
```
joragua: If you remove the `maxLines` property, what would happen to the `ellipsize` behavior? The text would no longer be ellipsized. Do you agree? 🤔 

My suggestion is: keep the ellipsis at the end, but increase the number of `maxLines`

joragua (changes requested):
Good job @retransmit! 💯 Some comments here about the code:

NOTE: We usually have one commit for the calens entry with the message `chore: add calens file`. In this PR, you've included the changelog info in the fix commit. Could you split it into two separate commits? Thanks in advance!

owncloudApp/src/androidTest/java/com/owncloud/android/spaces/members/SpaceRolesAdapterTest.kt
retransmit: Removed. The layout change is now the only code change in the PR.

owncloudApp/src/main/res/layout/role_item.xml
retransmit: Without a cap the end ellipsis was a no-op. Changed to `maxLines="4"` with `ellipsize="end"`, which Android does apply to multi-line text. Four lines fit the longest current descriptions at common phone widths, with room for larger font scaling and longer translations, and anything beyond that now ends with a visible ellipsis instead of being cut silently. Verified on an Android 15 emulator at 360dp, where the "Can manage" description renders on three lines.

retransmit:
Addressed all three points: history is now a `fix:` commit plus a `chore: add calens file` commit, the instrumented test is removed, and the description is capped at four lines with an end ellipsis.

owncloudApp/src/androidTest/java/com/owncloud/android/spaces/members/SpaceRolesAdapterTest.kt
joragua: 💯

owncloudApp/src/main/res/layout/role_item.xml
joragua: Nice! 👍🏻

joragua (approved):
LGTM! Moving it to QA! @jesmrec will start the QA round when he is available 🚀 

Thanks for your effort!

jesmrec:
Basic checks performed over the solution, everything ✅ 

Thanks for the contribution @retransmit, will be part of the following release!

Approved

